Issue #73, opened 2024-06-05
EWS: Push mail


benbucksch:
Listen to mailbox changes, to get notified immediately about:
* New mails
* Deleted mails
* Flag changes

Reproduction:
* Send email to yourself
* Open same mailbox with another email client, and mark something as read, starred etc., and delete a mail

Expected result:
* Mustang show the changes immediately, within 1 second

Actual result:
* The changes show up only after I click [Get Mail] button

PR #104, opened 2024-06-18
EWS: Push mail (#73)

NeilRashbrook:
- Only handles changes for email, not calendar or contacts
- Doesn't handle folder removal


benbucksch:
> * Only handles changes for email, not calendar or contacts
> * Doesn't handle folder removal

Fair enough

benbucksch (approved):

Issue #105, opened 2024-06-19
IMAP: Delete too slow


benbucksch:
Reproduction:
* Mail, Vertical view (select in bottom right)
* Select 5 messages in the msg list, by clicking on the first msg, holding SHIFT, and clicking on the last msg
* Press Del key

Expected result:
* The messages disappear from the list immediately.
* The messages are deleted from the server
* If there was any problem deleting them on the server, they re-appear.

Actual result:
* The messages disappear from the list immediately.
* About 2 seconds later, they re-appear in the list
* About 5 seconds per message, they disappear one by one
* The whole action takes 20-30 seconds

Dev:
* Please investigate the root cause
* Make a PR with a potential fix
* There are likely multiple ways to fix the problem. Make a PR for each of them.

NeilRashbrook:
I wasn't able to reproduce this, so I'm going off code inspection here. The front end deletes all of the messages in separate IMAP commands. This itself isn't a bug because the user could do the same thing by hitting the Delete key five times in quick succession. However, this does allow for a count notification to arrive before all of the deletions have been able to process. At this point the folder will try to refresh its message list, which will include the messages that are still pending deletion.

At this point it could be that the messages are already marked as deleted pending expunge, and it's up to the folder not to add those messages back to the message list when it refreshes itself, but it's also possible that the messages may simply not have been marked as deleted on the server yet, so it would be better to assume that the latter is the case here.

NeilRashbrook:
Actually it doesn't even need to be a count notification - the flag change notification for the mark as deleted flag might be what's triggering the refresh.

PR #109, opened 2024-06-19
Ignore messages pending deletion (#105)

NeilRashbrook:


benbucksch:
This is a great idea.

How about this implementation: IIRC, the message already has a isDeleted flag, because I tried to fix  similar bug. Could we simply reuse this here?

Thanks for the great analysis that the folder flag notifications cause this. In hindsight, it's obvious and logical.
-- 
Sent from my phone. Please excuse the brevity.

20.06.2024 00:10:35 Neil Rashbrook ***@***.***>:

NeilRashbrook:
What's the point of setting flags on a message that you're removed from the folder's message list?

benbucksch (approved):

benbucksch:
> What's the point of setting flags on a message that you're removed from the folder's message list?

I see what you mean: When the folder lists the messages, it no longer has access to that EMail object, because it's already unlinked from the `.messages` array.

Issue #67, opened 2024-06-04
UI: Localize


benbucksch:
Options:
* [svelte-i18n](https://github.com/kaisermann/svelte-i18n)
  * which uses [Format.JS](https://formatjs.io/)
  * which uses [ICU message format](https://unicode-org.github.io/icu/userguide/format_parse/messages/)
* [Fluent](https://projectfluent.org/)

Considerations:
* Hard to change localization format ("message format") above
* In the UI code, I'd like `tr('Some English words ending with {ending}', {ending})`, i.e. have the English string hardcoded. This is a) more fail-safe, and b) avoids that we have to invent new string IDs and c) allows to delay the localization during development, not touching multiple files all the time.
* Must handle plural forms (both options above do that)
* Would be nice, if it could handle brand names without code in the UI. (Fluent can do that, need to check Format.JS)
* L10n format must be supported by online web tools for translators.

Tasks:
* Choose localization library
* Extract all UI strings from the code (Svelte HTML, Svelte JS, logic JS), and localize it.
* Create a script in TypeScript that extracts the strings from the source code and builds the English locale, in the translation files, as the l10n engine expects them. That locale will then be the basis for the translation to other languages.
* Create a sample translation in 1 other language, and test it.
* Later (other ticket): Set up system to translate to other languages using AI tools. See #108

benbucksch:
@jermy-c Could you please complete this task in the next week?

jermy-c:
### Localization Library
[Lingui](https://github.com/lingui/js-lingui) is the localization library chosen.
https://github.com/HenryLie/svelte-i18n-lingui is the svelte wrapper for the library used for UI Localization.

**Pros**

1. It uses natively uses English strings as IDs
2. It fallsback to the English string if there's no translations, which allows delaying localization during development.
3. It allows interpolation of svelte components in between translation which gives translations more context and lets translators translate without touching the HTML. e.g. 
```svelte
<T msg="Click # for more information">
	<a href="/about">{$t`here`}</a>
</T>
```
4. Has a string extraction tool

**Cons**

1. Very new and the svelte wrapper is even newer

benbucksch:
Interesting choice. Looks very good!

Can you please make a PR with the code changes? Before, please carefully review yourself. Make sure the above criteria are met.

benbucksch:
@jermy-c I've fixed the translation of the ts modules, in the way we discussed yesterday, in b364b946

[...]

PR #111, opened 2024-06-29
Localize UI

jermy-c:
* Translate the entire application.
* Use lingui and svelte-i18n-lingui.
* Use GetText .po files for now. 
* Works with weblate.org and other translation tools
* AI translations for 20 common languages - unchecked

#67

TODO:
* Translation of .ts files (using `gt()`) doesn't load, because the .ts modules load before we load the translations.
* A few fixes, see review
* Change to ICU MessageFormat files (instead of GetText .po files)
* Plural form
* Proof-reading of the AI translations by human translators, using weblate.org


app/tsconfig.json
```diff
@@ -23,7 +23,8 @@
     "frontend/**/*.js",
     "frontend/**/*.svelte",
     "logic/util/Observable.ts",
-    "logic/build.ts"
+    "logic/build.ts", 
+    "lingui.config.ts"
```
benbucksch, line 27: Please avoid changing the last line. Could you please add the new file after `Observable.ts`?

app/lingui.config.ts
```diff
@@ -0,0 +1,18 @@
+import { jstsExtractor, svelteExtractor } from 'svelte-i18n-lingui/extractor';
+
+export default {
+	locales: ['ar','cs','da','de','el','en','es','fi','fr',
```
benbucksch, line 4: Please use 2 spaces for indention, not tabs

app/l10n/l10n.ts
```diff
@@ -0,0 +1,21 @@
...
+export async function setLocale(lang: string) {
+  // TODO: Have a better fallback system that checks whether the locale
+  // is missing or if there's actually an error
+  // Support different variants of the same language i.e. en-US en-AU
+
+  // Extract only the 'en' from 'en-US'
+  lang = lang.match(reLang)[0];
+
+  try {
+    const { messages } = await import(`./locales/${lang}/messages.ts`);
```
benbucksch, line 14: Can we please use the static `import` and not the dynamic `import()` ? The latter causes all kinds of hassles.

It's OK to load all languages.

app/frontend/Calendar/EventLine.svelte
```diff
@@ -11,11 +11,12 @@
   import type { Event } from "../../logic/Calendar/Event";
   import { calendarMustangApp } from "./CalendarMustangApp";
   import { selectedEvent } from "./selected";
+  import { t } from "svelte-i18n-lingui";
 
   export let event: Event;
 
   $: startTime = event.startTime.toLocaleString(navigator.language, { hour: "2-digit", minute: "2-digit" });
-  $: eventAsText = `${startTime} – ${Math.ceil((event.endTime.getTime() - event.startTime.getTime()) / 1000 / 60)} min
+  $: eventAsText = `${startTime} – ${Math.ceil((event.endTime.getTime() - event.startTime.getTime()) / 1000 / 60)} ${$t({message: "min", comment: "Minutes Abbreviation"})}
```
benbucksch, line 19: Please put the number as placeholder inside the string, so that plural forms can work properly.

app/frontend/Calendar/ViewSelector.svelte
```diff
@@ -6,25 +6,25 @@
...
+      return $t`Month`;
     } else if (interval == 28) {
-      return "4 Weeks";
+      return $t`4 Weeks`;
     } else if (interval == 2) {
-      return "Day/Month";
+      return $t`Day/Month`;
     } else {
-      return interval + " days";
+      return $t`${interval} days`;
```
benbucksch, line 27: :+1:

app/frontend/Mail/MailChat/MailMessage.svelte
```diff
@@ -2,7 +2,7 @@
   <svelte:fragment slot="above-center">
     <hbox flex />
     <hbox class="recipients">
-      to: {recipientsLine}
+      {$t`to:`} {recipientsLine}
```
benbucksch, line 5:
```suggestion
      {$t`to:`}
      {recipientsLine}
```

It's correct that the recipientsLine is not part of the string.

Let's add a newline to make clear that these are 2 different elements.

benbucksch (changes requested):

benbucksch:
We need the plural form as part of the l10n engine and translated files, *not* part of the app code.
Lingui supports that https://lingui.dev/guides/plurals

benbucksch:
Is is possible to use ICU MessageFormat? I'd prefer that, because the variables and the plural form are much clearer and easier for the translators.

LinGUI supports that: https://lingui.dev/guides/message-format

benbucksch:
SettingsCategories.ts still needs to be translated.

app/l10n/l10n.ts
benbucksch: This may also fix the translation of the ts files.

app/frontend/Mail/Composer/Attachments/AttachmentEntry.svelte
```diff
@@ -7,7 +7,7 @@
       {attachment.filename}
     </hbox>
     <hbox class="size">
-      {Math.ceil($attachment.size / 1024)} KB
+      {Math.ceil($attachment.size / 1024)} {$t`KB`}
```
benbucksch, line 10: Include number in placeholder

app/frontend/Settings/Mail/Account/IdentityBlock.svelte
```diff
@@ -34,31 +34,31 @@
...
+        {$t`Signature`}
         <vbox class="signature-editor-box">
           <HTMLEditorToolbar {editor} />
           <HTMLEditor bind:html={identity.signatureHTML} bind:editor />
           {#if showSentBy}
             <hbox class="sentBy">
               <div>
                 <!--<Icon data={logo} size="20px" />-->
-                Sent by <a href={siteRoot} target="_blank" style="color: #20AE9E"><strong><em>{appName}</em></strong></a>
+                {$t`Sent by`} <a href={siteRoot} target="_blank" style="color: #20AE9E"><strong><em>{appName}</em></strong></a>
```
benbucksch, line 57: Please put the appname including the `<a>` tag as 1 single placeholder.
Do not use the `<T>` here.

(We will need the same string when sending email.)

benbucksch (changes requested):

benbucksch (approved):
This is great!

benbucksch:
@jermy-c Thanks so much for the changes you did yesterday!

I'm impressed that you managed to use JSON as translation source file which we check in. These JSON files are supported by Weblate (during import and export), I tested that.

I'm also impressed that you managed to have them compiled on the fly from the JSON translation files to the JSON files that the app UI code (using lingui) reads at startup. Well done for finding the right lingui configuration. And vite automatically compiles messages.json to messages-compiled.json , on the fly, and we don't have to commit the latter to the source code repository.

And, the translations are much smaller now. l10n dir is now only 430 KB instead of 1.6 MB originally (or 1.2 MB with the .po files only). This is great.

Just one small glitch: When weblate exports the JSON files, it uses 4 space indention, whereas we have 2 space indention in the JSON files. That causes 100% of the files to be changed. Could you please check whether there's a setting in weblate to use 2 space indention, or to adapt to the source files it found?

I've successfully imported the repo into weblate, checked 1/3 of the German translation strings, fixed a number of them, exported the German JSON file back to the source code, and commited it to this branch. So, it works with the round-trip to and from weblate. Congratulations!

I'll land this as-is.

TODO, each in a new PR:
* Fix the review comments above @jermy-c
* Plural form automatic, in the t/gt functions (instead of all places in the code)
* `{brandName}` and similar global replacement strings in the t/gt functions, instead of all places in the code.
* Fix 2 space indention when exporting from weblate @jermy-c

app/frontend/Calendar/EventLine.svelte
benbucksch: Correction: Please keep `startTime` outside the string, but put the duration number inside the translated "min" string - for plural form.

app/frontend/Settings/Mail/Account/IdentityBlock.svelte
jermy-c: How do I put the appname inside an `<a>` tag and inside `$t``string`` `?

benbucksch: Try:
```suggestion
                {@html $t`Sent by ${'<a href={siteRoot} target="_blank" style="color: #20AE9E"><strong><em>{appName}</em></strong></a>'}`}
```

benbucksch: (We'll append the same string to outgoing HTML emails as signature)

PR #114, opened 2024-06-30
Fix L10n: messages-compiled.json are not compiled automatically by vite

jermy-c:
**Changes**

- Removed extra compile step i.e. `messages-compiled.json`
- Use `@lingui/vite-plugin` added `?lingui` to message.json imports for compilation when running vite
- Added `@lingui/vite-plugin` to `e2/electron.vite.cofig.ts`

### DIFF
--- a/app/l10n/l10n.ts
+++ b/app/l10n/l10n.ts
@@ -3,27 +3,27 @@ import { derived } from 'svelte/store';
 
 import { sourceLocale } from './list';
 
-import { messages as en } from './locales/en/messages-compiled.json';
-import { messages as de } from './locales/de/messages-compiled.json';
-import { messages as fr } from './locales/fr/messages-compiled.json';
-import { messages as it } from './locales/it/messages-compiled.json';
-import { messages as es } from './locales/es/messages-compiled.json';
-import { messages as pt } from './locales/pt/messages-compiled.json';
-import { messages as pl } from './locales/pl/messages-compiled.json';
-import { messages as ja } from './locales/ja/messages-compiled.json';
-import { messages as zh } from './locales/zh/messages-compiled.json';
+import { messages as en } from './locales/en/messages.json?lingui';
+import { messages as de } from './locales/de/messages.json?lingui';
+import { messages as fr } from './locales/fr/messages.json?lingui';
+import { messages as it } from './locales/it/messages.json?lingui';
+import { messages as es } from './locales/es/messages.json?lingui';
+import { messages as pt } from './locales/pt/messages.json?lingui';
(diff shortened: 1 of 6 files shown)


app/l10n/l10n.ts
```diff
@@ -3,27 +3,27 @@ import { derived } from 'svelte/store';
...
-import { messages as da } from './locales/da/messages-compiled.json';
-import { messages as el } from './locales/el/messages-compiled.json';
-import { messages as fi } from './locales/fi/messages-compiled.json';
-import { messages as nl } from './locales/nl/messages-compiled.json';
-import { messages as no } from './locales/no/messages-compiled.json';
-import { messages as ro } from './locales/ro/messages-compiled.json';
-import { messages as ru } from './locales/ru/messages-compiled.json';
-import { messages as sv } from './locales/sv/messages-compiled.json';
-import { messages as uk } from './locales/uk/messages-compiled.json';
+import { messages as ar } from './locales/ar/messages.json?lingui';
```
benbucksch, line 16: Is it possible to keep the 'messages.compiled.json' (preferably with `.` instead of `-`) instead of the `?lingui` query? I think that would be safer, avoiding confusion, and might also work better with the installed version of the app.

benbucksch (changes requested):

e2/electron.vite.config.ts
```diff
@@ -12,6 +13,6 @@ export default defineConfig({
     plugins: [externalizeDepsPlugin()]
   },
   renderer: {
-    plugins: [nodePolyfills({include: ['buffer'], globals: {global: false, process: false}}), svelte()]
+    plugins: [nodePolyfills({include: ['buffer'], globals: {global: false, process: false}}), svelte(), lingui({configPath: "../app/lingui.config.ts"})]
```
benbucksch, line 16: Could you please make newlines after every plugin, similar to app/vite.config.js ?

app/l10n/l10n.ts
jermy-c: It doesn't seem possible, in the source code it searches for `.po` or `?lingui`. And there isn't a way to change that.
https://github.com/lingui/js-lingui/blob/0f6661799acfb62c4d169cacc8f6867278263f0e/packages/vite-plugin/src/index.ts#L11

benbucksch: :laughing: :rofl: :roll_eyes:

benbucksch:
@jermy-c Can this PR be dropped, in favor of #446?

jermy-c:
> @jermy-c Can this PR be dropped, in favor of #446?

Yes.

PR #118, opened 2024-07-02
OWA: Add protocol implementation for mail, calendar and contacts

NeilRashbrook:


e2/src/main/index.ts
```diff
@@ -60,6 +60,107 @@ function createWindow(): void {
       });
     });
 
+    ipcMain.handle('fetchSessionData', async function(_event, partition, url, interactive) {
```
benbucksch, line 63: Can we do that in backend.ts using JPC, please? It's also running in the main process.

We should not use Electron IPC at all. In fact, avoiding Electron IPC was the whole reason why I invented JPC.

e2/tsconfig.json
```diff
@@ -1,14 +1,10 @@
...
   "extends": "@electron-toolkit/tsconfig/tsconfig.json",
   "include": [
-    "../app/**/*",
-    "../backend/**/*",
-    "src/renderer/src/env.d.ts",
-    "src/renderer/src/**/*",
-    "src/renderer/src/**/*.svelte",
-    "../electron/src/logic.ts",
-    "../electron/src/logic/**/*",
-    "src/preload/*.d.ts"
```
benbucksch, line 11: Why these removals?

e2/src/main/index.ts
benbucksch: Also, please keep the code in the backend to the absolute minimum. Anything OWA-specific should be in app/logic/ (and passed as arguments), not here.

NeilRashbrook: I wouldn't be able to make the new window modal to the main window, but apart from that, I didn't realise that those were actually the same process.

app/logic/Mail/OWA/OWAFolder.ts
```diff
@@ -1,9 +1,461 @@
...
+export const kMaxCount = 50;
 
 export class OWAFolder extends Folder {
+  account: OWAAccount;
 
   newEMail(): OWAEMail {
     return new OWAEMail(this);
   }
+
+  fromJSON(json: any) {
```
benbucksch, line 18:
```suggestion
  fromOWA(json: any) {
```

app/logic/Mail/OWA/OWAEMail.ts
```diff
@@ -1,4 +1,147 @@
...
+      },
+    };
+    let result = await this.folder.account.callOWA(request);
+    let mimeBase64 = sanitize.nonemptystring(result.Items[0].MimeContent.Value);
+    this.mime = new Uint8Array(await base64ToArrayBuffer(mimeBase64, "message/rfc822"));
+    await this.parseMIME();
+    //await this.save();
+  }
+
+  fromJSON(json) {
```
benbucksch, line 53:
```suggestion
  fromOWA(json) {
```

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -1,4 +1,176 @@
...
+  calendar: OWACalendar;
+
+  get itemID(): string | null {
+    return this.pID;
+  }
+  set itemID(val: string | null) {
+    this.pID = val;
+  }
+
+  fromJSON(json: any) {
```
benbucksch, line 22:
```suggestion
  fromOWA(json: any) {
```

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -1,4 +1,176 @@
...
+      }
+    }
+    if (json.Start) {
+      this.startTime = sanitize.date(json.Start);
+    }
+    if (json.End) {
+      this.endTime = sanitize.date(json.End);
+    }
+    if (json.DueDate) {
+      this.endTime = sanitize.date(json.DueDate);
```
benbucksch, line 41: Thanks for using the `sanitize.*()` functions

app/logic/Mail/AutoConfig/readConfig.ts
```diff
@@ -69,13 +69,17 @@ function readServer(xml: any, displayName: string, fullXML: any, source: ConfigS
...
+      owaURL = owaURL.replace("365", "");
+    }
     let activeSyncURL = sanitize.url(xml.easURL, null);
     assert(ewsURL || owaURL || activeSyncURL, "Malformed or unknown Exchange config");
     assert(ewsURL || owaURL, "ActiveSync not yet supported");
     // Return only the most preferred config (for this backwards compat)
-    protocol = ewsURL ? "ews" : owaURL ? "owa" : "activesync";
-    xml.url = ewsURL ?? owaURL ?? activeSyncURL;
+    protocol = owaURL ? "owa" : ewsURL ? "ews" : "activesync";
+    xml.url = owaURL ?? ewsURL ?? activeSyncURL;
```
benbucksch, line 80: I'd like to default to EWS for Mustang, if both are available, given that EWS is faster with sync and therefore needs less resources.

benbucksch (changes requested):

e2/tsconfig.json
benbucksch: Did you check whether it still works when the app is built as package/installer? (`yarn run build:linux && ./dist/*

NeilRashbrook: I didn't even mean to commit that change, but I wanted my `npx tsc --noEmit` to mostly work.

app/logic/Mail/OWA/OWAFolder.ts
NeilRashbrook: I didn't see you complain about `fromXML` in the EWS code...

e2/src/main/index.ts
benbucksch: OK. Non-modal is OK for now.

(Just for your information: e2/src/main/index.ts line 14 calls `startupBackend()`, which is in backend/backend.ts and starts JPC.)

app/logic/Mail/AutoConfig/readConfig.ts
NeilRashbrook: I didn't mean to commit this either, this was just so I could actually set up an account for test purposes.

e2/tsconfig.json
benbucksch: Fair enough. I can remove the change before landing.

app/logic/Mail/AutoConfig/readConfig.ts
benbucksch: OK. Good. I wasn't sure whether it was for testing or intentional.

The "remove '365'" change is needed, though, right?

e2/src/main/index.ts
NeilRashbrook: It's all OWA-specific because it has to handle the per-session cookies.

app/logic/Mail/AutoConfig/readConfig.ts
NeilRashbrook: I don't know how you want to handle the difference between the Office 365 OWA and EWS hostnames. If you're not going to default to OWA for Office 365, it doesn't really matter.

e2/src/main/index.ts
benbucksch: Per-session cookies useful for many things, e.g. for OAuth2 dialogs.

benbucksch: The main purpose is to keep all OWA-specific knowledge in one code place. Esp. "canary" and stuff.

NeilRashbrook: I've kept all of the canary specific knowledge here. I didn't think it was a good idea to pass the canary back and forth each time I wanted to make a request...

benbucksch: OK, we can disentangle this later. But please do move it into backend and use JPC.
Make a new code file for this, backend/owa.ts .

Issue #121, opened 2024-07-03
EWS: Throttling


benbucksch:
### Reproduction
* Set up ben@beonex.onm account
* Click on a few emails in Inbox

### Actual result
* Error console shows:
```
Account.ts:32 Error: The server cannot service this request right now. Try again later., This operation exceeds the throttling budget for policy part 'ConcurrentSyncCalls', policy value '5',  Budget type: 'Ews'.  Suggested backoff time 5000 ms.
    at EWSAccount.checkResponse (EWSAccount.ts:176:13)
    at EWSAccount.callEWS (EWSAccount.ts:206:19)
    at async EWSFolder.updateChangedMessages (EWSFolder.ts:86:18)
    at async EWSAccount.processNotification (EWSAccount.ts:361:11)
    at async EWSAccount.ts:293:9
```

### Expected result
* We honor policies that the server pro-actively tells us
* When we receive an error message like this, we parse it, and honor the values given in the error message

### Cause
We did too many things in parallel:
* Download Sent folder after setup
* Downloading Inbox
* Marking messages as read
* Folder notifications cause a sync

### Solution
Throttle:
* Number of concurrent calls
* Number of calls per second

Parse errors:
* Error types
* Numeric values in the error data structure
* Numeric values in the error message

benbucksch:
This caused #150

benbucksch:
Throttle:
* Number of concurrent calls
* Number of calls per second

Parse errors:
* Error types
* Values in the error data structure
* Values in the error message

NeilRashbrook:
How many calls per second do you want? (Exchange doesn't have a specific limit or policy for this.)

benbucksch:
Let's start with 5 per second and dial down as needed?

NeilRashbrook:
Five per second? We're defaulting to 20 at once...

benbucksch:
Then let's make a limit of 50 calls per second by default.

benbucksch:
git commit 49297e5a and 3bb4b237

Issue #95, opened 2024-06-18
EWS: Handle server overloaded errors


benbucksch:
See #94 

We should handle "Too many connections" and similar errors appropriately, by slowing down and trying again later.

PR #122, opened 2024-07-03
Honour suggested backoff time #95 #121

NeilRashbrook:
I couldn't get this to trigger using the STR in #121 but what I was able to do was as follows:
- Delete `~/.mustang/mail.db`
- Set up the account when prompted
- Click on a miscellaneous folder in the account itself
- Click on All Accounts

This seems to trigger a list of every single folder in the account in parallel, thus triggering the server busy error. (This doesn't happen after a restart; you have to do it in the same session.)

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -13,8 +13,8 @@ import { OAuth2URLs } from "../../Auth/OAuth2URLs";
 import { ContentDisposition } from "../Attachment";
 import { ConnectError, LoginError } from "../../Abstract/Account";
 import { appGlobal } from "../../app";
-import { blobToBase64 } from "../../util/util";
-import { assert } from "../../util/util";
+import { assert, sleep, blobToBase64, throttleConnectionsPerSecond } from "../../util/util";
+import { SetColl } from "svelte-collections";
 
 type Json = string | number | boolean | null | Json[] | {[key: string]: Json};
 
@@ -25,6 +25,9 @@ export class EWSAccount extends MailAccount {
   readonly port: number = 443;
   readonly tls = TLSSocketType.TLS;
   readonly folderMap = new Map<string, EWSFolder>;
+  maxConcurrency: number = 20;
+  connections = new SetColl<Promise<any>>;
+  nextConnectionTime = new Array(50).fill(0);
 
   constructor() {
     super();
@@ -198,12 +201,53 @@ export class EWSAccount extends MailAccount {
     return options;
   }
 
+  async checkMaxConcurrency(ex: EWSItemError) {
+    if (ex.type != "ErrorServerBusy") {
+      throw ex;
+    }
+    // Check whether the ConcurrentSyncCalls policy was exceeded.
+    if (/'ConcurrentSyncCalls'.*'(\d+)'.*'Ews'/.test(ex.message) && Number(RegExp.$1) < this.maxConcurrency) {
+      this.maxConcurrency = Number(RegExp.$1);
+      console.log(`Server busy, reduced max concurrency to ${this.maxConcurrency}`);
+    }
+    let milliseconds = Number(ex.error.MessageXml?.Value?.Value);
+    if (!milliseconds) {
+      // Server isn't asking us to back off, so give up.
+      throw ex;
+    }
+    if (milliseconds > 20000) {
+      milliseconds = 20000; // Sensible upper limit
+    }
+    console.log(`Server busy, using ${milliseconds}ms backoff time`);
+    await sleep(milliseconds / 1000);
+  }
+
   async callEWS(aRequest: JsonRequest): Promise<any> {
-    let response = await appGlobal.remoteApp.postHTTP(this.url, this.request2XML(aRequest), "text", this.createRequestOptions());
+    await throttleConnectionsPerSecond(this.nextConnectionTime);
+    while (this.connections.length >= this.maxConcurrency) {
+      console.log(`Throttling because there are ${this.maxConcurrency} pending connections`);
+      try {
+        await Promise.race(this.connections);
+      } catch (ex) {
+      }
+    }
+    let connection = appGlobal.remoteApp.postHTTP(this.url, this.request2XML(aRequest), "text", this.createRequestOptions());
+    this.connections.add(connection);
+    let response;
+    try {
+      response = await connection;
+    } finally {
+      this.connections.remove(connection);
+    }
     response.responseXML = this.parseXML(response.data);
     this.fatalError = null;
     if (response.status == 200) {
-      return this.checkResponse(response, aRequest);
+      try {
+        return this.checkResponse(response, aRequest);
+      } catch (ex) {
+        await this.checkMaxConcurrency(ex);
+        return await this.callEWS(aRequest);
+      }
     }
     if (response.status == 401) {
       if (this.oAuth2) {
@@ -513,7 +557,7 @@ class EWSError extends Error {
 
 class EWSItemError extends Error {
   readonly request: JsonRequest;
-  readonly error: Json;
+  readonly error: any;
   readonly type: string;
   constructor(errorResponseJSON: any, aRequest: JsonRequest) {
     if (Array.isArray(errorResponseJSON.MessageXml?.Value)) {
@@ -523,7 +567,7 @@ class EWSItemError extends Error {
     }
     super(errorResponseJSON.MessageText);
     this.request = aRequest;
-    this.error = errorResponseJSON as Json;
+    this.error = errorResponseJSON;
     this.type = errorResponseJSON.ResponseCode;
   }
 }
--- a/app/logic/util/util.ts
+++ b/app/logic/util/util.ts
@@ -76,6 +76,16 @@ export function fileExtensionForMIMEType(mimetype: string) {
   return ".ext";
 }
 
+export async function throttleConnectionsPerSecond(nextConnectionTime: number[]) {
+  let milliseconds = Date.now();
+  nextConnectionTime.push(milliseconds + 1000);
+  milliseconds = nextConnectionTime.shift() - milliseconds;
+  if (milliseconds > 0) {
+    console.log(`Throttling for ${milliseconds}ms because there were ${nextConnectionTime.length} connections in the last second`);
+    await sleep(milliseconds / 1000);
+  }
+}
+
 /** Abstract class as base class for allowing more specific error classes */
 export class SpecificError extends Error {
   constructor(ex: Error, message: string) {


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -203,7 +203,20 @@ export class EWSAccount extends MailAccount {
...
     response.responseXML = this.parseXML(response.data);
     if (response.status == 200) {
-      return this.checkResponse(response, aRequest);
+      try {
+        return this.checkResponse(response, aRequest);
+      } catch (ex) {
+        if (ex.type != "ErrorServerBusy") {
+          throw ex;
+        }
+        let ms = ex.error.MessageXml?.Value?.Value;
```
benbucksch, line 212:
```suggestion
        let milliseconds = ex.error?.MessageXml?.Value?.Value;
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -203,7 +203,20 @@ export class EWSAccount extends MailAccount {
     let response = await appGlobal.remoteApp.postHTTP(this.url, this.request2XML(aRequest), "text", this.createRequestOptions());
     response.responseXML = this.parseXML(response.data);
     if (response.status == 200) {
-      return this.checkResponse(response, aRequest);
+      try {
+        return this.checkResponse(response, aRequest);
+      } catch (ex) {
+        if (ex.type != "ErrorServerBusy") {
```
benbucksch, line 209: The Error message says 
`'ConcurrentSyncCalls', policy value '5',`

Can you please:
* Consider the generic `ErrorServerBusy`, and do something sensible, even if we cannot parse the values?
* Please also check for 'ConcurrentSyncCalls' before parsing the values?
* Also parse the value 5. Given the context, I assume them to mean that they accept only 5 parallel calls. That should be something built into `callEWS()`, a throtteling to not have too many  concurrent calls. The default could be something like 10, and then if the server complaints with this error message, we honor this value and do only 5 parallel calls.
* Good idea to also honor the backoff time, as you're doing below. But there should be a sensible upper limit. I've seen backoffs from Office365 that are 1 hour or 1 day. Obviously, if we were to honor that, the app would simply hang for the end user. So, the maximum backup that we honor should be something like 20 seconds.

benbucksch (changes requested):

benbucksch:
@NeilRashbrook This is low priority. Please stay on #118 for now.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: I don't know what you mean by "parsing the values". The backoff time is explicitly provided as a separate XML element; I'm not parsing it from the message.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -203,7 +203,24 @@ export class EWSAccount extends MailAccount {
...
+        let milliseconds = ex.error.MessageXml?.Value?.Value;
+        if (!milliseconds) {
+          // Server isn't asking us to back off, so give up.
+          throw ex;
+        }
+        if (milliseconds > 20000) {
+          milliseconds = 20000; // Sensible upper limit
+        }
+        console.log(`Server busy, using ${milliseconds}ms backoff time`);
+        await new Promise(resolve => setTimeout(resolve, milliseconds));
```
benbucksch, line 221:
```suggestion
        `sleep(seconds)`;
```

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: > I don't know what you mean by "parsing the values".

1. Backoff time (as you already do)
2. max number of concurrent calls, in the error message: `policy value '5'`, see my last comment point 3. (we don't check that at all: neither read the value, nor have a throttle of this kind. but we need it.)

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -203,7 +203,24 @@ export class EWSAccount extends MailAccount {
     let response = await appGlobal.remoteApp.postHTTP(this.url, this.request2XML(aRequest), "text", this.createRequestOptions());
     response.responseXML = this.parseXML(response.data);
     if (response.status == 200) {
-      return this.checkResponse(response, aRequest);
+      try {
+        return this.checkResponse(response, aRequest);
+      } catch (ex) {
+        if (ex.type != "ErrorServerBusy") {
```
benbucksch, line 209: Could you please move this whole throttling into 1-2 functions?

benbucksch (changes requested):

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: >     1. Backoff time (as you already do)

No, I don't parse the backoff time.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: You can't sleep in the UI thread.

benbucksch: I was referring to our `sleep()` function in util.js, which does exactly what you do here.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: `error` has a number of child fields:
- ResponseClass (always "Error" by the time we get here)
- MessageText (localised error message)
- ResponseCode (always "ErrorServerBusy" by the time we get here)
- DescriptiveLinkKey (not actually used by Exchange; always 0)
- MessageXml (contains additional information that depends on the error)

In the case of an ErrorServerBusy this last field then contains a Value which contains a Value which is the back off time in milliseconds, but it can contain other information e.g. if you try to write to a read-only field then it will contain the field that you were trying to write.

benbucksch: The full error message is:
`Error: The server cannot service this request right now. Try again later., This operation exceeds the throttling budget for policy part 'ConcurrentSyncCalls', policy value '5',  Budget type: 'Ews'.  Suggested backoff time 5000 ms`

If the message contains `'ConcurrentSyncCalls'`, can we please pick that `policy value '5'` out of there and parse it into an integer and use it as max concurrentSyncCalls?

benbucksch: I understand that this is ugly, but this is Microsoft. It with the microsoft.com domain.

NeilRashbrook:
> I was referring to our sleep() function in util.js, which does exactly what you do here.

(Except for being in seconds, meaning that all of the values get round-tripped from and back to milliseconds.)

Anyway, I've implemented maximum concurrency, and added a limit of 50 requests per second (I don't know an easy way to test the latter though).

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -198,12 +201,58 @@ export class EWSAccount extends MailAccount {
     return options;
   }
 
+  async checkMaxConcurrency(ex: EWSItemError) {
+    if (ex.type != "ErrorServerBusy") {
+      throw ex;
+    }
+    // Check whether the ConcurrentSyncCalls policy was exceeded.
+    if (/'ConcurrentSyncCalls'/.test(ex.message) && /'(\d+)'/.test(ex.message) && Number(RegExp.$1) < this.maxConcurrency) {
```
benbucksch, line 209: Can you make the second regexp more specific, please? To avoid that there's another `'4'` before `ConcurrentSyncCalls`. Like:

```suggestion
    if (/'ConcurrentSyncCalls'.*'(\d+)'/.test(ex.message) && Number(RegExp.$1) < this.maxConcurrency) {
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -198,12 +201,58 @@ export class EWSAccount extends MailAccount {
...
+      this.maxConcurrency = Number(RegExp.$1);
+    }
+    let milliseconds = ex.error.MessageXml?.Value?.Value;
```
benbucksch, line 212: What makes this a number?
```suggestion
    let milliseconds = parseInt(ex.error.MessageXml?.Value?.Value);
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -198,12 +201,58 @@ export class EWSAccount extends MailAccount {
...
+
   async callEWS(aRequest: JsonRequest): Promise<any> {
-    let response = await appGlobal.remoteApp.postHTTP(this.url, this.request2XML(aRequest), "text", this.createRequestOptions());
+    let milliseconds = Date.now();
+    this.nextConnectionTime.push(milliseconds + 1000);
+    milliseconds = this.nextConnectionTime.shift() - milliseconds;
+    if (milliseconds > 0) {
+      console.log(`Throttling for ${milliseconds}ms because there were 50 connections in the last second`);
+      await sleep(milliseconds / 1000);
+    }
```
benbucksch, line 231: Please make this a generic function

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -198,12 +201,58 @@ export class EWSAccount extends MailAccount {
...
+    while (this.connections.length >= this.maxConcurrency) {
+      console.log(`Throttling because there are ${this.maxConcurrency} pending connections`);
+      try {
+        await Promise.race(this.connections);
+      } catch (ex) {
+      }
+    }
```
benbucksch, line 238: Please make this another generic function

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: We want to re-use it for OWA

benbucksch (changes requested):

NeilRashbrook:
I've adapted the regex we use for owl (it's a different policy name but otherwise the same idea).
I've put the rate limiting in its own function in `util.ts`.
I haven't done anything with the concurrency because that would cause a race condition (the `add` and `remove` need to be in the same async function where the length is checked).

benbucksch (approved):

app/logic/util/util.ts
```diff
@@ -76,6 +76,16 @@ export function fileExtensionForMIMEType(mimetype: string) {
...
+export async function throttleConnectionsPerSecond(nextConnectionTime: number[]) {
+  let milliseconds = Date.now();
+  nextConnectionTime.push(milliseconds + 1000);
+  milliseconds = nextConnectionTime.shift() - milliseconds;
+  if (milliseconds > 0) {
+    console.log(`Throttling for ${milliseconds}ms because there were ${nextConnectionTime.length} connections in the last second`);
+    await sleep(milliseconds / 1000);
+  }
+}
+
```
benbucksch, line 88: util.js is only for very generic utility functions. Basically functions that should really be part of the web platform, but are not yet.

app/logic/util/util.ts
```diff
@@ -76,6 +76,16 @@ export function fileExtensionForMIMEType(mimetype: string) {
   return ".ext";
 }
 
+export async function throttleConnectionsPerSecond(nextConnectionTime: number[]) {
```
benbucksch, line 79: This parameter **really** needs documentation. Who populates it? (This function, not the caller, just pass it through) How to init it (array length = number of allowed calls per second, and init all values with 0)

benbucksch (changes requested):

Issue #120, opened 2024-07-03
IMAP: Deleted mails come back


benbucksch:
This is a followup to #105 

* browse through the messages in inbox (10000 msgs)
* clicking through them, which marks them immediately as read
* delete a few of them
* mark others as spam
* speed: generally one email per second

Actual result:
* The emails get deleted or marked as read immediately
* The right emails get deleted
* After 10-20 seconds, some of the deleted emails come back, and some of the read emails are marked unread again

Expected result:
* The actions are visible to the user immediately
* They persist (i.e. are not reverted unless I choose to)
* The right emails get deleted (not another one that I didn't delete)

NeilRashbrook:
I don't have easy access to an IMAP folder with 10000 deletable messages so I tried with 100 or so messages and although I was unable to reproduce the problem I did trip over this error whatever it means:

SqliteError: NOT NULL constraint failed: emailPersonRel.emailID
    at Database.run (/home/neil/github/mustang/e2/out/main/index.js:98176:38)
    at JPCWebSocket.funcListener (/home/neil/github/mustang/e2/out/main/index.js:46416:28)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async WSCall._incomingMessage (/home/neil/github/mustang/e2/out/main/index.js:45981:20)
    at async WebSocket.<anonymous> (/home/neil/github/mustang/e2/out/main/index.js:46075:9) {
  code: 'SQLITE_CONSTRAINT_NOTNULL'
}
    at message.js:153:16
    at new Promise (<anonymous>)
    at WSCall.makeCall (message.js:148:12)
    at JPCWebSocket.callRemote (protocol.js:142:31)
    at RemoteClass.run (obj.js:225:55)
    at SQLEMail.saveRecipient (SQLEMail.ts:128:33)
    at async SQLEMail.saveRecipients (SQLEMail.ts:94:5)
    at async SQLEMail.save (SQLEMail.ts:67:5)
    at async IMAPFolder.messageFlagsChanged (IMAPFolder.ts:222:5)
    at async IMAPAccount.ts:161:9

NeilRashbrook:
[...]

PR #125, opened 2024-07-05
IMAP sequence numbers are unstable #120

NeilRashbrook:
Steps to reproduce problem:
- Don't download all messages in a folder in advance
- Delete a message in a folder
- Read a message with a higher sequence number

Expected results: Message you selected is downloaded

Actual results: Message you selected is overwritten by an unrelated message that now has that sequence number, causing it to become cloned

This is why UIDs were invented, after all...

Well, except when expunging messages; sadly the response only mentions the sequence number of the message, as using or including the UID wouldn't be backwards compatible.

Thunderbird's JavaScript IMAP client works around this by simply refreshing the entire folder if it gets an expunge notification from the server.

Thunderbird's C++ IMAP client keeps an array of the UIDs so that it can always associate a sequence number with a UID even for deleted messages (sequence numbers are always consecutive).

### DIFF
--- a/app/logic/Mail/IMAP/IMAPEMail.ts
+++ b/app/logic/Mail/IMAP/IMAPEMail.ts
@@ -25,14 +25,14 @@ export class IMAPEMail extends EMail {
 
   async download() {
     let msgInfo = await this.folder.runCommand(async (conn) => {
-      // TODO uid failed for me for msgs with uid > folder msg count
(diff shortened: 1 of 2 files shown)


benbucksch (approved):
Thanks for catching this!

There's even a TODO. Silly that I didn't see that.

PR #129, opened 2024-07-10
Extract Strings Github Action

jermy-c:
**Changes**

- Add Github action that extracts strings and pushes the extracted strings back to the repo

**Proposed workflow**

1. Trigger Github Action `Extract Strings`, this leaves extracted strings empty so it's safe to have them in the `master` branch
2. Weblate receives changes and triggers automatic translation using desired translation engine e.g. LibreTranslate, DeepL or Google Cloud Translation Basic
3. Translated strings are labeled `Automatically translated` and marked as `Needs editing`
4. Translators edit translation and the `Automatically translated` label and `Needs editing` are automatically removed
5. Changes are pushed to `translations` branch
6. Check for ICU Message Format syntax errors on Weblate and create a PR
7. Merge changes into `master`

### DIFF
--- a/.github/workflows/extract-strings.yml
+++ b/.github/workflows/extract-strings.yml
@@ -9,7 +9,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - name: Extract Strings
+    - name: Extract UI strings to be translated
       uses: actions/setup-node@v3
       with:
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
@@ -19,7 +19,7 @@ jobs:
     - run: cd app; yarn l10n:extract
     - name: Commit & Push Changes
       run: |
-        git config user.name "github-actions[bot]"
-        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
-        git commit -a -m "Extract Strings"
+        git config user.name "Bot"
+        git config user.email "translate@mustang.im"
+        git commit -a -m "Translation: Extract UI strings from code into the English master translation file"
         git push
--- a/.github/workflows/extract-strings.yml
+++ b/.github/workflows/extract-strings.yml
@@ -1,0 +1,25 @@
+name: Extract Strings
+
+on:
+  workflow_dispatch # manually triggered
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/extract-strings.yml
```diff
@@ -0,0 +1,25 @@
...
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+        node-version: 18.x
+#        cache: 'npm'
+    - run: cd app; yarn
+    - run: cd app; yarn l10n:extract
+    - name: Commit & Push Changes
+      run: |
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git commit -a -m "Extract Strings"
```
benbucksch, line 24:
```suggestion
        git config user.name "Bot"
        git config user.email "translate@mustang.im"
        git commit -a -m "Translation: Extract UI strings from code into the English master translation file"
```

.github/workflows/extract-strings.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  workflow_dispatch # manually triggered
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Extract Strings
```
benbucksch, line 12:
```suggestion
    - name: Extract UI strings to be translated
```

benbucksch (changes requested):

benbucksch:
We cannot always wait for translators for every string addition. We do not have translators for all languages that are constantly standing waiting. We can hire translators only maybe once per year.

Can we instead use the following workflow?
* Extract strings
* Push to automatic translation
* Add automatic translations into translated files, in the right place
* Commit
* Push to Weblate
* Release
* Let translators double-check (typically many weeks months later)
* Release

benbucksch:
More concretely, steps 2 and 3 should have their own commands in packages.json. Then, another packages.json command should do steps 1-4 together. Lastly, a github action invokes the command that does steps 1-4, and then makes the git commit, as done here in this PR.

jermy-c:
Sounds good, will do.

benbucksch:
Included in #445

Issue #117, opened 2024-07-01
Gmail Account doesn't stay logged in after setup


jermy-c:
**Reproduction**

1.⁠ ⁠Setup Gmail
2.⁠ ⁠Close/Quit app after setup
3.⁠ ⁠Open the app again
4.⁠ ⁠Go to Mail
5.⁠ ⁠Open Gmail Inbox Folder

**Expected Result**

1.⁠ ⁠Fetches emails
2.⁠ ⁠Displays emails
3.⁠ ⁠No need to login and doesn't show "Please login" at the top

**Actual Result**

1.⁠ ⁠Shows fetched emails from before
2.⁠ ⁠Displays "Please login" at the top

**Error output**
```
Error: Please login
at _OAuth2.login (http://localhost:5454/logic/Auth/OAuth2.ts:86:13)
at async IMAPAccount.connection (http://localhost:5454/logic/Mail/IMAP/IMAPAccount.ts:64:9)
at async IMAPFolder.runCommand (http://localhost:5454/logic/Mail/IMAP/IMAPFolder.ts:50:18)
at async IMAPFolder.listMessages (http://localhost:5454/logic/Mail/IMAP/IMAPFolder.ts:88:5)
at async loadFolder (http://localhost:5454/frontend/Mail/MailApp.svelte:572:4)
```

jermy-c:
Ok, will do!

jermy-c:
### Cause

The `%EMAILADDRESS%` variable for the refresh token in the local storage is not replaced with the user's email address when doing a fresh login.

**Thunderbird Import refresh token**
|  Key  |  Value  |
| --- | --- |
| oauth2.refreshToken.gmailusername@gmail.com.accounts.google.com | [refresh token] |

**Fresh login refresh token**
|  Key  |  Value  |
| --- | --- |
| oauth2.refreshToken.%EMAILADDRESS%.accounts.google.com | [refresh token] |

jermy-c:
<img width="1370" alt="圖片" src="https://github.com/user-attachments/assets/08569fa1-0af7-4d58-ad37-5b1ca0812e56">

### OAuth2 username gets overwritten with null
My guess is that when you manually login you input the email from the manualConfig email and that get's passed in. And when you import from Thunderbird that is empty and it still tries to assign it to that value.

PR #131, opened 2024-07-12
Check if OAuth2 instance exist before assigning username

jermy-c:
Fixes https://github.com/mustang-im/mustang/issues/117#issuecomment-2224114810


jermy-c:
I'll check if it's better to create an instance and assign the username or just ignore it if there's no instance created. Because the TBImport creates an instance afterwards.

jermy-c:
I think I'll leave it as is since the Mail Account contructor should initialize the OAuth2 instance if needed.

benbucksch (approved):

Issue #128, opened 2024-07-10
Cannot open database because the directory does not exist


benbucksch:
* File: backup/mail-backup.db
* Windows
* Happens immediately after first setup, when opening a few mails. Then goes away.

Possible causes:
* `await SQLSourceEMail.save(this);` in `EMail.save()`
* Speculation: `IMAPFolder.downloadMessages()` saves correctly, but `IMAPEMail.download()` does not
* Speculation: Race between the above 2 functions

benbucksch:
As seen by Elena @raffikastudio :

```
index-ogBq-r3Y.js:232409 Error {
  "id": 174,
  "success": false,
  "message": "Cannot open database because the directory does not exist",
  "backendStack": "TypeError: Cannot open database because the directory does not exist\n    at new Database (C:\\Users\\Elena\\AppData\\Local\\Programs\\parula\\resources\\app.asar\\node_modules\\better-sqlite3\\lib\\database.js:65:9)\n    at new Database (C:\\Users\\Elena\\AppData\\Local\\Programs\\parula\\resources\\app.asar\\out\\main\\index.js:124716:5)\n    at getSQLiteDatabase (C:\\Users\\Elena\\AppData\\Local\\Programs\\parula\\resources\\app.asar\\out\\main\\index.js:136759:10)\n    at JPCWebSocket.callListener (C:\\Users\\Elena\\AppData\\Local\\Programs\\parula\\resources\\app.asar\\out\\main\\index.js:35593:18)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async WSCall._incomingMessage (C:\\Users\\Elena\\AppData\\Local\\Programs\\parula\\resources\\app.asar\\out\\main\\index.js:35178:20)\n    at async WebSocket.<anonymous> (C:\\Users\\Elena\\AppData\\Local\\Programs\\parula\\resources\\app.asar\\out\\main\\index.js:35272:9)",
  "originalCall": {
    "id": 174,
    "path": "call",
    "arg": {
      "obj": "17143693227985522688",
      "args": [
        "backup/mail-backup.db"
      ]
    }
  }
}
```

jermy-c:
On MacOS:

- There's a directory created at `/Users/username/Library/Mustang/backup` with no files in it
- There's a directory created at `/Users/username/Library/Application Support/Mustang` with `calendar.db` and other `.db` files

benbucksch:
Does anybody still see this?

PR #132, opened 2024-07-12
Consistently use config and files dir #128

NeilRashbrook:
- Databases are implicitly saved in `Application Support/Mustang`
- As @jermy-c noticed in #128, `SQLSourceDatabase.ts` creates `backup` as a subfolder of `Mustang` rather than of `Application Support/Mustang` which is where the other databases live
- `SQLEmail.ts` and `RawFilesAttachment.ts` save in `Mustang` but uses the variable name normally used for `Application Support/Mustang`, which could be confusing
- `MailZIP.ts` saves in `Mustang` and uses better variable naming, which I have copied to the two above-mentioned files

### DIFF
--- a/app/logic/Mail/SQL/SQLEMail.ts
+++ b/app/logic/Mail/SQL/SQLEMail.ts
@@ -10,7 +10,7 @@ import { ArrayColl } from "svelte-collections";
 import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
 import sql from "../../../../lib/rs-sqlite";
 
-let configDir: string = null;
+let filesDir: string = null;
 
 export class SQLEMail {
   /**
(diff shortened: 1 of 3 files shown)


jermy-c:
This fixes the error for me. Thanks.

But I get this error message now. It only happens when click on an email the first time after loading a folder not just after setup.
圖片

**Reproduction**

1. Start the app
2. Go to Mail
3. Click a folder and wait for it to load
4. Click on an email for the first time after loading the folder

benbucksch:
Thanks, Jeremy, the report. However, looking at the stack trace, this is 
in MailZIP, a different function, so that's another bug. Could you please 
file a new GitHub issue in this issue? It's valid and should be fixed.

benbucksch (approved):
In the future, where possible, it's better to leave extensive renaming and a small significant bug fix in 2 PRs, because it was hard to see what is the actual behavior-changing code change.

benbucksch:
Thanks for finding and fixing this

PR #134, opened 2024-07-16
OAuth2: Use the username and password from the account. Don't duplicate it.

benbucksch:
Alternative fix for issue #117, instead of PRs #130 and #131


benbucksch:
@jermy-c Could you please test this change, whether
a) It fixes #117 for you
b) doesn't create any new problems for accounts that use OAuth2?

jermy-c:
> @jermy-c Could you please test this change, whether a) It fixes #117 for you b) doesn't create any new problems for accounts that use OAuth2?

It works for me and there's no new problems, thanks.

jermy-c (approved):

PR #135, opened 2024-07-16
Release workflow

jermy-c:
**Changes**

- Add write permission for repo content to the job
- Set releaseType to release

**TODO**

- Release workflow for other OSes
- Should we auto generate the release description/CHANGELOG?


benbucksch:
> Should we auto generate the release description/CHANGELOG?

If you can do so easily, yes, please do. Would be nice to have.

benbucksch (approved):

Issue #124, opened 2024-07-04
EWS: Error "terminated"


benbucksch:
After a while of Mustang just sitting there doing nothing, I see an error "terminated".

The stack trace is:
![image](https://github.com/mustang-im/mustang/assets/1907525/2bb6f24a-03dd-4e09-8feb-ca8626b92e44)

benbucksch:
@NeilRashbrook #118 has higher priority than this

NeilRashbrook:
This seems to happen when your HTTPS socket closes unexpectedly. Since this is happening on the backend, there should be another stack trace on the console, not that it would help (the stack trace here references a line number `206` which has never been inside the `callStream` function).

benbucksch:
I've had Mustang with an EWS account running over night, and I don't see the error anymore.

PR #139, opened 2024-07-18
EWS: Re-open stream when the connection is dropped #124

benbucksch:


benbucksch:
@NeilRashbrook Could you please review the code?

NeilRashbrook (approved):

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -223,41 +223,51 @@ export class EWSAccount extends MailAccount {
...
-            let message = Object.values(XML2JSON(messages))[0];
-            if (message.ResponseClass == "Error") {
-              throw new EWSItemError(message, request);
-            }
-            if (message.ConnectionStatus == "Closed") {
-              this.callStream(request, responseCallback);
+    let lastAttempt = Date.now();
+    while (Date.now() - lastAttempt > 10000) { // quit when last failure < 10 seconds ago. TODO throw? But don't show error to user.
+      try {
+        lastAttempt = Date.now();
```
NeilRashbrook, line 229: Whoops, I failed to notice this at the time, but this completely breaks EWS streaming notifications, because the loop never starts. Maybe this should be a `do...while` loop?

PR #144, opened 2024-07-18
Persist OWA data to database, support push mail

NeilRashbrook:
Plus I noticed that multiple calendars caused the data to get mixed; this was because the code was trying to reuse the same account id for each calendar.

### DIFF
--- a/app/logic/Calendar/OWA/OWACalendar.ts
+++ b/app/logic/Calendar/OWA/OWACalendar.ts
@@ -1,9 +1,11 @@
 import { Calendar } from "../Calendar";
 import { OWAEvent } from "./OWAEvent";
 import type { OWAAccount } from "../../Mail/OWA/OWAAccount";
+import { SQLCalendar } from "../SQL/SQLCalendar";
+import { SQLEvent } from "../SQL/SQLEvent";
 import { kMaxCount } from "../../Mail/OWA/OWAFolder";
 import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
-import type { ArrayColl } from "svelte-collections";
+import { ArrayColl } from "svelte-collections";
 
 export class OWACalendar extends Calendar {
   readonly protocol: string = "calendar-owa";
@@ -19,15 +21,22 @@ export class OWACalendar extends Calendar {
   }
 
   async listEvents() {
-    let events: OWAEvent[] = [];
+    if (!this.dbID) {
+      await SQLCalendar.save(this);
+    }
+
+    let events = new ArrayColl<OWAEvent>;
     await this.listFolder("calendar", events);
     /* Disabling tasks for now.
     await this.listFolder("tasks", events);
     */
+    for (let event of this.events.subtract(events)) {
+      SQLEvent.deleteIt(event);
+    }
     this.events.replaceAll(events);
   }
 
-  async listFolder(folder: string, events: OWAEvent[]) {
+  async listFolder(folder: string, events: ArrayColl<OWAEvent>) {
     let request = {
       __type: "FindItemJsonRequest:#Exchange",
       Header: {
@@ -64,7 +73,7 @@ export class OWACalendar extends Calendar {
     }
   }
 
-  async getEvents(eventIDs: string[], events: OWAEvent[]) {
+  async getEvents(eventIDs: string[], events: ArrayColl<OWAEvent>) {
     if (!eventIDs.length) {
       return;
     }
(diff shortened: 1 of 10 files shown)


NeilRashbrook:
Plus I noticed there wasn't any HTTP error checking for EWS streaming notifications.

backend/owa.ts
```diff
@@ -88,6 +92,27 @@ export async function fetchJSON(partition: string, url: string, action: string,
...
 
+export async function streamJSON(partition: string, url: string) {
+  let result = {
+    ok: false,
+    status: 0,
+    statusText: '',
+    body: null,
+  };
+  let session = Session.fromPartition(partition);
+  let cookies = await session.cookies.get({ name: 'X-OWA-CANARY' });
```
benbucksch, line 103: It seems to be that this cookie getter could be moved into a separate function and the canary added to the URL by the caller.

We want to keep the backend functions absolutely minimal and generic. All protocol implementation details (like canary) should be in app/logic/, not in backend.ts.

By refactoring it this way, you may be able to unite functions `streamJSON()` and `streamHTTP()`.

benbucksch (changes requested):

app/logic/Contacts/OWA/OWAAddressbook.ts
```diff
@@ -71,37 +78,71 @@ export class OWAAddressbook extends Addressbook {
...
-        Body: {
-          __type: "GetPersonaRequest:#Exchange",
-          PersonaId: {
-            __type: "ItemId:#Exchange",
-            Id: person.PersonaId.Id,
+    for (let person of this.persons.contents.filter(person => !persons.some(result => result.PersonaId.Id == person.personaID))) {
+      this.persons.remove(person);
+      await SQLPerson.deleteIt(person);
+    }
+    for (let result of persons) {
```
benbucksch, line 85: In the interest of keeping the functions short and understandable, and hopefully even reusable, could you move the parsing of a person and a group into 2 respective functions, please? Essentially, the part inside the try/catches within the for loops.

benbucksch (changes requested):

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -34,6 +34,7 @@ export class OWAAccount extends MailAccount {
   readonly port: number = 443;
   readonly tls = TLSSocketType.TLS;
   readonly folderMap = new Map<string, OWAFolder>;
+  msgfolderroot: string | void;
```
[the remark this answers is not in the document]
benbucksch: (note camelCase)

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -19,15 +21,22 @@ export class OWACalendar extends Calendar {
   }
 
   async listEvents() {
-    let events: OWAEvent[] = [];
+    if (!this.dbID) {
+      await SQLCalendar.save(this);
+    }
```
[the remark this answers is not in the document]
NeilRashbrook: `app/logic/Calendar/AccountsList/SQL.ts` already does this.

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -19,15 +21,22 @@ export class OWACalendar extends Calendar {
...
+
+    let events = new ArrayColl<OWAEvent>;
     await this.listFolder("calendar", events);
     /* Disabling tasks for now.
     await this.listFolder("tasks", events);
     */
+    for (let event of this.events.subtract(events)) {
```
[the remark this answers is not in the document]
NeilRashbrook: Yes they will, because `events` contains objects found on disk for existing events.

app/logic/Contacts/OWA/OWAAddressbook.ts
```diff
@@ -29,6 +32,10 @@ export class OWAAddressbook extends Addressbook {
     if (!this.name) {
       this.name = contacts.DisplayName;
     }
+    if (!this.dbID) {
+      await SQLAddressbook.save(this);
+    }
```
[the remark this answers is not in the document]
NeilRashbrook: `app/logic/Chat/AccountsList/SQL.ts` already does this.

backend/owa.ts
NeilRashbrook: The caller is JPC.

app/logic/Mail/OWA/OWAAccount.ts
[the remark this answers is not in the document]
NeilRashbrook: (`msgfolderroot` is Exchange's name)

app/logic/Contacts/OWA/OWAAddressbook.ts
NeilRashbrook: For consistency with EWS I actually moved the entire loops into their own functions.

NeilRashbrook:
Also added a commit to switch all of the calls to use a persistent partition. For Office 365 accounts, if you select Stay Logged In, then you do in fact seem to stay logged in, at least over a quick restart; other Exchange servers will log you out if you restart. (The persistent partition does at least remember your login name.)

app/logic/Contacts/OWA/OWAAddressbook.ts
benbucksch: Makes sense

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -34,6 +34,13 @@ export class OWAAccount extends MailAccount {
   readonly port: number = 443;
   readonly tls = TLSSocketType.TLS;
   readonly folderMap = new Map<string, OWAFolder>;
+  /*
```
benbucksch, line 37:
```suggestion
  /**
```

JSDoc needs to start with `/**`

backend/owa.ts
benbucksch: I understand your reason is that you want to avoid a JPC-round-trip to get the cookies via JPC, then calling the HTTP function via JPC as well, for every OWA server call.

benbucksch (approved):

PR #145, opened 2024-07-19
Include .deb and .rpm artifacts in release

jermy-c:
**Changes**

- Remove AppImage param from `build:release:linux` script so it uses the targets in `e2/electron-builder.yml`
- Comment out `snap` target in `e2/electron-builder.yml` because it tries to publish to Snap instead of GitHub and also it needs `snapcraft` to be installed. It causes errors also.

Draft release generated by the action: https://github.com/mustang-im/mustang/releases/tag/untagged-beaee3af94b3a3e82b90

### DIFF
--- a/e2/electron-builder.yml
+++ b/e2/electron-builder.yml
@@ -36,7 +36,7 @@ dmg:
 linux:
   target:
     - AppImage
-    - snap
+    # - snap
     - deb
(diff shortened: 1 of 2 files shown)


benbucksch:
Good catch! Thank you!

PR #146, opened 2024-07-19
Check for updates, notify and install on exit

jermy-c:
- Call `autoUpdater.checkForUpdatesAndNotify();` when app is ready. Check for updates, notify and install on exit.

Note: Doesn't work on Mac OS X, yet.

### DIFF
--- a/e2/src/main/index.ts
+++ b/e2/src/main/index.ts
@@ -1,9 +1,9 @@
 import { setMainWindow, startupBackend } from '../../../backend/backend';
 import { app, shell, BrowserWindow } from 'electron'
 import { join } from 'path'
+import { autoUpdater } from 'electron-updater';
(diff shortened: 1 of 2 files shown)


e2/src/main/index.ts
```diff
@@ -3,6 +3,7 @@ import { app, shell, BrowserWindow } from 'electron'
 import { join } from 'path'
 import { electronApp, optimizer, is } from '@electron-toolkit/utils'
 import icon from '../../resources/icon.png?asset'
+import { autoUpdater } from 'electron-updater';
```
benbucksch, line 6:
```suggestion
import { autoUpdater } from 'electron-updater';
import { electronApp, optimizer, is } from '@electron-toolkit/utils'
import icon from '../../resources/icon.png?asset'
```

benbucksch (approved):

benbucksch:
Thanks!

Issue #149, opened 2024-07-21
EWS: Contacts: Type is `null`


benbucksch:
### Actual result
All email addresses and chat contacts show `null` as contact type in the UI.
![image](https://github.com/user-attachments/assets/619919b0-dc73-409c-8ce9-a8575882cc46)

### Expected result
The type shows "Home", "Business", "Mobile" etc., as saved on the EWS server for that email address, phone number, or other contact.

benbucksch:
Then assume Business, please.

PR #152, opened 2024-07-22
Assume Exchange contact email addresses are Work type #149

NeilRashbrook:


benbucksch (approved):

benbucksch (changes requested):
What about phone numbers? Street addresses etc.?

NeilRashbrook:
> What about phone numbers? Street addresses etc.?

Phone numbers and street addresses already have a type.

PR #154, opened 2024-07-22
Release to Parula repo from Mustang repo

jermy-c:
- `parula-brand.sh` now also changes the git repo url to `parula.git`
- Uses a `PAT` to push release directly to the Parula repo
- No syncing required but the Parula repo will be outdated. Only the releases will be updated.
- Not tested, yet.

### DIFF
--- a/.github/workflows/installer-windows-parula.yml
+++ b/.github/workflows/installer-windows-parula.yml
@@ -25,7 +25,7 @@ jobs:
     - run: cd e2; npm install --legacy-peer-deps
     - run: cd e2; npm run build:release:win
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.PARULA_GH_TOKEN }}
     - name: Upload
       uses: actions/upload-artifact@v4
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/installer-windows-parula.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     - run: cd e2; npm install --legacy-peer-deps
     - run: cd e2; npm run build:release:win
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.PARULA_GH_TOKEN }}
```
benbucksch, line 28: I am not sure how that would work. The `GITHUB_TOKEN` is a special variable that GitHub sets. I wouldn't know where to get it from. If we can generate one, then I would know how to set it in the other repo. There's a page in the settings to set things like `PARULA_GH_TOKEN`.

app/build/parula-brand.sh
```diff
@@ -12,11 +12,13 @@ perl -p -i \
   -e "s|https://mustang.im|https://parula.beonex.com|g;" \
   -e "s|\"name\": \"mustang\"|\"name\": \"parula\"|;" \
   -e "s|\"version\": \".*\"|\"version\": \"$VERSION\"|;" \
+  -e "s|mustang.git|parula.git|g;" \
```
benbucksch, line 15: Elegant!

benbucksch (commented):
This is a good PR. Let's remember that for the future, if we every want to put Parula releases on a separate repository than Mustang.

For fixing the bug, let's go with the other option of using channel names.

Issue #153, opened 2024-07-22
OWA: Throttling


benbucksch:
Caused #150. Similar to #121.

Throttle:
* Number of concurrent calls
* Number of calls per second

Parse errors:
* Error types
* Numeric values in the error data structure
* Numeric values in the error message

benbucksch:
Fixed in git commit 997f5b19

PR #159, opened 2024-07-23
OWA: Throttle connections #153

NeilRashbrook:
(The change to `util.ts` is the same as in #122.)


benbucksch:
Re-implemented using classes `Semaphore` and `Throttle` in git commit 997f5b19

PR #171, opened 2024-08-12
Send all extra MIME headers

NeilRashbrook:
As per comments on 4477457.

(Seriously, is there no way to map over a Map?)

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -103,8 +103,8 @@ export class EWSAccount extends MailAccount {
         }))),
       }, "item:Attachments");
     }
-    for (let [header, value] of email.headers.entries()) {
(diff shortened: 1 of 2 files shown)


benbucksch (approved):

NeilRashbrook:
(to answer my own question, with a regular `Map` I don't need `.entries`, but `MapColl` is a `Collection` so its API differs)

benbucksch:
@NeilRashbrook Thanks.
* It now works in OWA for me.
* No observable change in EWS
* In both cases, the header name is sent in all lower case. OK for now.

NeilRashbrook:
I doubt we can do anything about that unless you're prepared to convert the message to MIME first (which we'll need to do for ActiveSync anyway).

benbucksch:
It's OK, it's Good Enough (TM)

PR #172, opened 2024-08-12
Address review comments from #166

NeilRashbrook:


app/logic/Calendar/RecurrenceRule.ts
```diff
@@ -7,16 +7,11 @@ export enum Frequency {
   Yearly = "YEARLY",
 }
 
+/**
+ * Maps between pretty names (also used by Exchange) and
```
benbucksch, line 11: Thanks for the note that these are also used by Exchange, so that we don't change them.

benbucksch (approved):

Issue #162, opened 2024-08-07
OAuth2: Yahoo


benbucksch:
We finally got the OAuth2 client ID from Yahoo. I've added them to branch `yahoo-oauth2`. Yet, I get "invalid request" error message (in the auth window, inside the webpage, at the very bottom, in grey). I don't know why. Is it because the Client ID has not yet been deployed to Yahoo's production servers? Did we misconfigure something? Does Yahoo not like our request? Are parameters missing? Malformed? Too much (like `login_hint`)?

Could you please find out what's wrong, and fix it?

benbucksch:
We're opening auth URL https://api.login.yahoo.com/oauth2/request_auth?client_id=dj0yJmk9d25qSWNOMzJVeDhlJmQ9WVdrOVV6VTJiMjQzTTFZbWNHbzlNQT09JnM9Y29uc3VtZXJzZWNyZXQmc3Y9MCZ4PTE3&response_type=code&redirect_uri=https%3A%2F%2Fmustang.im%2Floggedin&response_mode=query&scope=mail-w&state=00031&login_hint=ben%40yahoo.de

Auth window error msg says: "Developers: Please specify a valid request and submit again"

jermy-c:
`error_description: invalid code challenge or method`
It might be, because it requires PKCE. Trying to confirm if that's the fix.

https://github.com/thunderbird/thunderbird-android/issues/7159#issuecomment-1731285127

jermy-c:
PKCE seems to be the problem, I got through the first part of the OAuth flow when `code_challenge` and `code_challenge_method` are added. But after logging in and authorization there's some errors:

1. The redirect URI doesn't match the one registered on YDN because the app's one is `https://www.mustang.im/loggedin` and the one registered is `https://mustang.im/loggedin`
2. When I set the condition to ignore that I get an `json?.error?.replace is not a function` error but that's an OAuth2Error.ts problem
3. If I `console.log()` I get an error:
```json
{
    "localizedMessage": "client request is not acceptable or not supported",
    "errorId": "INVALID_INPUT",
    "message": "client request is not acceptable or not supported"
}
```

The docs don't mention anything about about PKCE though.

PR #174, opened 2024-08-13
PKCE

jermy-c:
- Implement PKCE
- fixes https://github.com/mustang-im/mustang/issues/162#issuecomment-2274528544


app/logic/Auth/OAuth2.ts
```diff
@@ -182,16 +184,27 @@ export class OAuth2 extends Observable {
    * @throws OAuth2Error
    */
   protected async getAccessTokenFromParams(params: any, additionalHeaders?: any, tokenURL: string | void = this.tokenURL): Promise<string> {
-    params.scope = this.scope;
     params.client_id = this.clientID;
-    params.client_secret = this.clientSecret || undefined;
+    
+    if (this.clientSecret) {
```
benbucksch, line 189: Why this criteria to turn on/off PKCE?
Can we support a) client secret with PKCE and b) no client secret and no PKCE ?
So far, we use case b), so it seems like case that should be possible.

app/logic/Auth/OAuth2.ts
```diff
@@ -182,16 +184,27 @@ export class OAuth2 extends Observable {
...
+      params.code_verifier = this.codeVerifier;
+    }
+
+    let headers = {
+      'Content-Type': 'application/x-www-form-urlencoded',
+      'Accept': 'application/json',
+      ...additionalHeaders,
+    }
+    if (!this.clientSecret) {
+      headers.Authorization = `Basic ${btoa(this.clientID + ":")}`;
```
benbucksch, line 203: Move `let headers = {` up, then this line can go after `params.code_verifier =`

app/logic/Auth/OAuth2.ts
```diff
@@ -32,6 +33,7 @@ export class OAuth2 extends Observable {
   scope: string;
   clientID = "mail";
   clientSecret: string | null = null;
+  codeVerifier?: string;
```
benbucksch, line 36:
```suggestion
  codeVerifierPKCE?: string;
```

app/logic/Auth/OAuth2.ts
benbucksch: Why do we need an Auth: header when there's no client secret?

This has not been needed so far.

app/logic/Auth/PKCE.ts
```diff
@@ -0,0 +1,83 @@
+/*
```
benbucksch, line 1: Given that this is a third party library, either:
* Move this file to /lib/pkce/PKCE.ts
* Make a fork of the original library, and use the fork (I can publish the fork to npm, if you want)

app/logic/Auth/OAuth2.ts
```diff
@@ -222,25 +235,42 @@ export class OAuth2 extends Observable {
...
     this.authDoneURL = doneURL ?? this.authDoneURL; // needed for getAccessTokenFromAuthCode()
-    return this.authURL + "?" + new URLSearchParams({
+    let params = new URLSearchParams({
       client_id: this.clientID,
       response_type: "code",
       redirect_uri: doneURL ?? this.authDoneURL,
-      response_mode: "query",
       scope: this.scope,
-      state: this.verificationToken,
-      login_hint: this.account.username,
```
benbucksch, line 235: Why do you remove those?
* state is required
* login_hint is required for Microsoft
* response_mode is needed as well, I would presume

Did you test this with Microsoft accounts, that you don't regress anything? I think these changes would break Microsoft login (EWS with Office365 and OAuth2)

benbucksch (changes requested):

benbucksch:
As mentioned on chat, could you please make one PR for PKCE only, and another PR for the changes that are needed for Yahoo?

app/logic/Auth/OAuth2.ts
jermy-c: Yes, we could do that. How would you like to turn on/off PKCE? With a property PKCE?

app/logic/Auth/PKCE.ts
jermy-c: We could use the [pkce-challenge](https://www.npmjs.com/package/pkce-challenge) package as is. The only difference is that it returns an object with `code-verifier` and `code-challenge`together.
```
{
    code_verifier: 'u1ta-MQ0e7TcpHjgz33M2DcBnOQu~aMGxuiZt0QMD1C',
    code_challenge: 'CUZX5qE8Wvye6kS_SasIsa8MMxacJftmWdsIA_iKp3I'
}
```

app/logic/Auth/OAuth2.ts
benbucksch: Yes, it should be a property `doPKCE` on the OAuth2 object. We could then set it from the OAuth2URLs list.

app/logic/Auth/OAuth2.ts
jermy-c: I'm not sure. PKCE is mostly just the addition of `code_challange` and `code_challenge_method` when requesting authorization. And the `code_verifier` when getting the token. The changes are mostly a recreation of a successful request on Postman. I'm still not able to track down the root cause.

benbucksch: > The changes are mostly a recreation of a successful request on Postman. I'm still not able to track down the root cause.

I see.

1. Check whether it works with all your changes here. I.e. you can successfully log in to Yahoo using OAuth2. If so:
2. Remove changes here 1 by 1, and see whether it still works. If so:
3. Remove the change from the PR.

For those changes that you found are necessary, separate them into 2 PRs for PKCE and other changes, as mentioned, please.

app/logic/Auth/OAuth2.ts
```diff
@@ -185,11 +189,16 @@ export class OAuth2 extends Observable {
...
+    if (this.doPKCE) {
+      assert(!!this.codeVerifierPKCE, "Missing code verifier");
+      params.code_verifier = this.codeVerifierPKCE;
+    }
 
     let response = await appGlobal.remoteApp.postHTTP(tokenURL, params, "json", {
       headers: {
         'Content-Type': 'application/x-www-form-urlencoded',
-        'Accept': 'text/json',
+        'Accept': 'application/json',
```
benbucksch, line 201: Can you remove this change?

app/logic/Auth/OAuth2.ts
```diff
@@ -234,6 +243,18 @@ export class OAuth2 extends Observable {
...
+
+    try {
+      if (this.doPKCE) {
+        const pkce = await pkceChallenge();
+        this.codeVerifierPKCE = pkce.code_verifier;
+        params.append("code_challenge", pkce.code_challenge);
+        params.append("code_challenge_method", "S256");
+      }
+    } catch (ex) {
+      console.error(ex);
```
benbucksch, line 255:
```suggestion
      backgroundError(ex);
```

app/logic/Auth/OAuth2.ts
```diff
@@ -334,4 +355,4 @@ export class OAuth2 extends Observable {
     let username = this.account.username.replace(/@.*/, "").replaceAll(".", "-");
     return `oauth2.refreshToken.${username}.${host}`;
   }
-}
+}
```
benbucksch, line 358: remove this change (add return after last char)

benbucksch (approved):
Great (with a few small changes)

app/logic/Auth/OAuth2.ts
```diff
@@ -32,6 +34,8 @@ export class OAuth2 extends Observable {
   scope: string;
   clientID = "mail";
   clientSecret: string | null = null;
+  doPKCE: boolean;
```
benbucksch, line 37:
```suggestion
  doPKCE = false;
```

app/logic/Auth/OAuth2.ts
```diff
@@ -45,7 +49,7 @@ export class OAuth2 extends Observable {
   protected expiryTimout: NodeJS.Timeout;
   refreshErrorCallback = (ex: Error) => console.error(ex);
 
-  constructor(account: Account, tokenURL: string, authURL: string, authDoneURL: string | null | undefined, scope: string, clientID: string, clientSecret?: string | null) {
+  constructor(account: Account, tokenURL: string, authURL: string, authDoneURL: string | null | undefined, scope: string, clientID: string, clientSecret?: string | null, doPKCE?: boolean) {
```
benbucksch, line 52:
```suggestion
  constructor(account: Account, tokenURL: string, authURL: string, authDoneURL: string | null | undefined, scope: string, clientID: string, clientSecret?: string | null, doPKCE = false) {
```

app/logic/Auth/OAuth2.ts
```diff
@@ -58,6 +62,7 @@ export class OAuth2 extends Observable {
     this.scope = scope;
     this.clientID = clientID;
     this.clientSecret = clientSecret ?? null;
+    this.doPKCE = doPKCE ?? false;
```
benbucksch, line 65:
```suggestion
    this.doPKCE = doPKCE;
```

benbucksch (changes requested):

benbucksch (changes requested):
Can you please look through all cases of `new OAuth2` and adapt them all? Particularly the to/fromConfigJSON() is important - otherwise it won't work after a restart of the app (!). Could you please also test that latter case?

PR #176, opened 2024-08-14
Microsoft OAuth2

jermy-c:
- Add client ID
- Remove manual config instructions
- Explicitly list scopes


benbucksch:
@jermy-c Which account types did you test?
* @beonex.onmicrosoft.com 
* @outlook.com / hotmail.com

Which protocols did you test?
* EWS
* OWA
* IMAP + SMTP

Which combinations work with this change?

Issue #177, opened 2024-08-14
EWS and OWA: Tags: Save tags to server and read from there


benbucksch:
We have tags (keywords) implemented. We can add them to messages and save them locally to the DB and read them back from there, so they will persist on the local machine.

We also save the tags to the IMAP server as custom flags, and read them back from there. That means that a Mustang on another machine (or even another supporting email client) can read these flags, and set them, and they will be synced.

Please also save them to the EWS and OWA server. I've added stubs `addTagOnServer(tag)` and `removeTagOnServer(tag)` in `EWSEMail` and `OWAEMail`. Please implement them, to save the flags to the server. Also, please also read those custom flags, e.g. when reading message flags like read and junk state. Obviously, while you implement `ActiveSyncEMail`, please implement it there as well.

benbucksch:
That was fast. Thank you!

PR #178, opened 2024-08-14
Save tags as categories #177

NeilRashbrook:
Also fix a JSON parse error I hit trying to start the app up.


benbucksch (approved):
Thank you!

PR #182, opened 2024-08-16
OAuth2: dont' send code_verifier when getting a refresh token

jermy-c:
Fix:

-  Yahoo account not re-logging in after restart.
- `Error: Missing code verifier`


benbucksch (approved):

benbucksch:
Thanks, Jeremy

PR #183, opened 2024-08-21
Resize webview to content

jermy-c:
- Resize webview to content

### DIFF
--- a/app/frontend/Chat/MessageView/Message.svelte
+++ b/app/frontend/Chat/MessageView/Message.svelte
@@ -33,7 +33,7 @@
       <slot name="inner-top" />
       <div class="text selectable">
         <!-- TODO Security: Jail HTML into untrusted <iframe> for additional protection. -->
-        {@html $message.html || ""}
+        <WebView title={$t`text`} html={message.html} />
         <slot name="bubble" />
       </div>
       <slot name="inner-bottom" />
@@ -53,7 +53,9 @@
   import { ChatMessage, DeliveryStatus } from "../../../logic/Chat/Message";
   import { Person } from "../../../logic/Abstract/Person";
   import PersonPicture from "../../Shared/Person/PersonPicture.svelte";
+  import WebView from "../../Shared/WebView.svelte";
   import { getDateString } from "../../Util/date";
+  import { t } from "../../../l10n/l10n";
 
   export let message: Message;
   export let previousMessage: Message = null;
--- a/app/frontend/Chat/MessageView/Message.svelte
+++ b/app/frontend/Chat/MessageView/Message.svelte
@@ -33,7 +33,7 @@
       <slot name="inner-top" />
       <div class="text selectable">
         <!-- TODO Security: Jail HTML into untrusted <iframe> for additional protection. -->
-        <WebView title={$t`text`} html={message.html} />
+        <WebView title={$t`Text`} html={message.html} />
         <slot name="bubble" />
       </div>
       <slot name="inner-bottom" />
--- a/app/frontend/Shared/WebView.svelte
+++ b/app/frontend/Shared/WebView.svelte
@@ -52,8 +52,24 @@
   function haveWebView () {
     webviewE.addEventListener("dom-ready", () => {
       dispatch("webview", webviewE);
+      webviewE.addEventListener("did-finish-load", resizeWebview());
     }, { once: true });
   }
+
+  const heightBuffer = 30;
+  function resizeWebview () {
+    webviewE.executeJavaScript(`
+      new Promise(resolve => {
+        const observer = new ResizeObserver(entries => {
+          const contentRect = entries[0].contentRect;
+          resolve(contentRect.height);
+        });
+        observer.observe(document.body);
+      });
+    `).then(height => {
+      webviewE.style.height = height + heightBuffer + 'px';
+    });
+  };
 </script>
 
 <style>
--- a/app/frontend/Shared/WebView.svelte
+++ b/app/frontend/Shared/WebView.svelte
@@ -58,24 +58,27 @@
 
   const heightBuffer = 20;
   async function resizeWebview () {
-    const dimensions = await webviewE.executeJavaScript(`
-      document.body.style.minHeight = "0px";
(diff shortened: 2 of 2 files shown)


app/frontend/Chat/MessageView/Message.svelte
```diff
@@ -33,7 +33,7 @@
       <slot name="inner-top" />
       <div class="text selectable">
         <!-- TODO Security: Jail HTML into untrusted <iframe> for additional protection. -->
-        {@html $message.html || ""}
+        <WebView title={$t`Text`} html={message.html} />
```
benbucksch, line 36:
```suggestion
        <WebView title={$t`Text`} html={$message.html || ""} />
```

* The `$` is important to refresh the message content. Esp. if the msg wasn't loaded before and it's loaded on-demand.
* The `|| ""` also handles msgs which have not been downloaded or are empty.

app/frontend/Shared/WebView.svelte
```diff
@@ -52,8 +52,34 @@
...
   }
+
+  const heightBuffer = 20;
+  async function resizeWebview () {
+    try {
+      const dimensions = await webviewE.executeJavaScript(`
+        document.body.style.minHeight = "0px";
+        document.body.style.height = "fit-content";
+        document.body.style.width = "fit-content";
+        new Promise(resolve => {
```
benbucksch, line 66:
```suggestion
        return new Promise(resolve => {
```

Does a `return` work here? Or does it break?
Was it implicit before?

app/frontend/Shared/WebView.svelte
```diff
@@ -52,8 +52,34 @@
...
+            resolve({ 
+              width: contentRect.width,
+              height: contentRect.height 
+            });
+          });
+          observer.observe(document.body);
+        });
+      `);
+      webviewE.style.width = dimensions.width + "px";
+      webviewE.style.height = dimensions.height + heightBuffer + "px";
```
benbucksch, line 78:
```suggestion
      webviewE.style.height = (dimensions.height + heightBuffer) + "px";
```

benbucksch (changes requested):
Can you please also set the min-width inside the doc, *before* rendering?

benbucksch:
Can you please also set the min-width inside the doc, *before* rendering?

Ideally, you would set it not in the script, but at the same time as the HTML content, to avoid an additional "reflow" (re-rendering by the browser).

app/frontend/Shared/WebView.svelte
```diff
@@ -52,8 +52,34 @@
...
+  async function resizeWebview () {
+    try {
+      const dimensions = await webviewE.executeJavaScript(`
+        document.body.style.minHeight = "0px";
+        document.body.style.height = "fit-content";
+        document.body.style.width = "fit-content";
+        new Promise(resolve => {
+          const observer = new ResizeObserver(entries => {
+            const contentRect = entries[0].contentRect;
+            resolve({
```
benbucksch, line 69: There's a trailing space here. Please make sure to remove them before submitting a patch.

app/frontend/Shared/WebView.svelte
```diff
@@ -52,8 +52,34 @@
...
+      const dimensions = await webviewE.executeJavaScript(`
+        document.body.style.minHeight = "0px";
+        document.body.style.height = "fit-content";
+        document.body.style.width = "fit-content";
+        new Promise(resolve => {
+          const observer = new ResizeObserver(entries => {
+            const contentRect = entries[0].contentRect;
+            resolve({ 
+              width: contentRect.width,
+              height: contentRect.height
```
benbucksch, line 71:
```suggestion
              height: contentRect.height,
```

for consistency

benbucksch (changes requested):

app/frontend/Shared/WebView.svelte
```diff
@@ -52,8 +52,34 @@
...
+  const heightBuffer = 20;
+  async function resizeWebview () {
+    try {
+      const dimensions = await webviewE.executeJavaScript(`
+        document.body.style.minHeight = "0px";
+        document.body.style.height = "fit-content";
+        document.body.style.width = "fit-content";
+        new Promise(resolve => {
+          const observer = new ResizeObserver(entries => {
+            const contentRect = entries[0].contentRect;
```
benbucksch, line 68:
```suggestion
            observer.unobserve(document.body);
            const contentRect = entries[0].contentRect;
```

I think you also need the unobserve, given that you ignore further calls.

app/frontend/Shared/WebView.svelte
benbucksch: No, the `return` doesn't work. Your version was correct.

benbucksch:
I'm getting double scrollbars, despite the `heightBuffer`.

Can you please see where this extra space comes from, and which property is correct? You're using `contentRect`. Are you including the margin and border in the calculation?

benbucksch:
@jermy-c We should do the auto-sizing only for this new use-case, not for all cases where we use `<WebView>`. We use it in other places, and this change broke those.

Generally, when you make a change to a component or function, could you please check *all* instances where it is used, and make sure that it works everywhere? Not just in the one place which you try to fix. This procedure avoids regressions.

benbucksch:
It starts at a minimal width, and then doesn't enlarge. This means that long texts will be longer, because they don't use the available width. Screenshot:

image

I think I know why: When we first render it, we start with a very small iframe. Then, we render. Then, we make it larger. But the render was based on the small width (see discussion in the issue). 

Before we start to render, we should set the width to the max width that we have available. Then, we render. Then, we measure the size - width and height. Then we set the width and height of the iframe to the width and height of the content, i.e. let it grow in width and height (but not wider than the max width).

benbucksch:
FWIW, these 2 rendering processes are part of which makes it so expensive/slow.

benbucksch:
We also need to include the content stylesheets as `headHTML`, same as `Mail/Message/HTMLDisplay.svelte`

benbucksch:
I've fixed most of the above issues, and merged it to master.

Some issues remain. @jermy-c Could you please fix them in a follow-up PR?

PR #184, opened 2024-08-28
Chat viewer: New Iframe component with better autosizing and lazyloading

jermy-c:
- Inject CSS and JS while generating the `data:URI`
- Set the iframe to the `width` of the content and use `max-width: 100%` to contain it with in the parent
- Only load the iframe when it is in view because each iframe is essentially using the same amount of resource as a browser tab/window.
- We also have to unload the iframe because when switching between the chats the iframes are still there and it tries to load all of the iframes and its contents at once causing the lag
- Uses `message` events and `postMessage()` instead of `executeJavascript()`, and `postMessage()` doesn't work with `<webview>` because it's in a different process
- Set `sandbox` to restrict the iframe, but "allow-scripts" needs to set for the autoSize. Maybe "allow-scripts" only when we need to get the size would be better but I tried setting it just before the postMessage() call to the iframe content and it doesn't work.
- Set "credentialless" an experimental Chrome feature that: "its content will be loaded in a new, ephemeral context."
- Set "csp" an experimental Chrome feature that sets the "Content Security Policy"

**Known Issues**

- The component loads and unloads continuously when on the margin of the frame which makes it look glitch, I tried setting `rootMargin: "100px 0px"` but it doesn't work even with four explicit values for each side doesn't work
- The scroll feels jumpy because of the loading and unloading

### DIFF
--- a/app/frontend/Chat/MessageView/Message.svelte
+++ b/app/frontend/Chat/MessageView/Message.svelte
@@ -33,7 +33,7 @@
       <slot name="inner-top" />
       <div class="text selectable">
         <!-- TODO Security: Jail HTML into untrusted <iframe> for additional protection. -->
-        <Iframe title={$t`Text`} html={$message.html || ""} {headHTML} autoSize lazyLoad />
+        <Iframe title={$t`Text`} html={$message.html || ""} {headHTML} autoSize lazyLoad lazyUnload/>
         <slot name="bubble" />
       </div>
       <slot name="inner-bottom" />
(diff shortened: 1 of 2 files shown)


benbucksch:
Thanks, Jeremy, for this PR and your work on it.

There are a few conceptual problematic issues with it:
* We *must* disable JavaScript within the iframe, and also enable all the other protections. These protections are the entire reason why we want the iframe in the first place. (`<WebView>.executeJavaScript()` is OK, if it's guaranteed that only our JavaScript will run and never that of the page loaded.)
* The lazy loading is a nice idea. However, I am worried that it will lead to massive shuffling on the page, as we scroll up, e.g. in chat, where the latest messages at at the bottom. Given that we don't know the size before load, we necessarily need to change the size after load, which will shift the entire rest of the page. That will happen for every message loaded. If we're lazy loading, that means the page will constantly move around. That doesn't give an impression of being simple and fast and reliable.
* While `<iframe>` would theoretically be an option, I'd much rather stay with the `<WebView>`. Can you see how much of this PR you can port to the WebView component?

PR #185, opened 2024-08-30
Chat viewer: webview resizable

jermy-c:
- `<webview>` is now resizable but by passing in a maxWidth from the parent
- `<webview>` doesn't respect max-width like an iframe so we need to get it from the parent element
- It gets the size only at the first load of the content. After that, it adjusts the width based on the max-width, this should make it much lighter.
- Inject CSS when generating the data:URI so it doesn't reflow

Known Issues:

- The height is still not accurate, and some times the width.
- The height grows every time you switch a chat/person because reuses the same webview and 'fit-content' property fills space available and when we add the buffer it'll become larger. After that, webview is bigger and adds the buffer again.
- The resizing lags sometimes

### DIFF
--- a/app/frontend/Chat/MessageView/Message.svelte
+++ b/app/frontend/Chat/MessageView/Message.svelte
@@ -4,7 +4,6 @@
   class:outgoing={$message.outgoing}
   class:followup
   deliveryStatus={$message instanceof ChatMessage ? $message.deliveryStatus : DeliveryStatus.Unknown}
-  bind:this={messageE}
   >
   {#if !$message.outgoing && !followup}
     <vbox class="avatar">
@@ -34,7 +33,7 @@
       <slot name="inner-top" />
       <div class="text selectable">
         <!-- TODO Security: Jail HTML into untrusted <iframe> for additional protection. -->
-        <WebView title={$t`Text`} html={$message.html || ""} {headHTML} autoSize {maxWidth} />
+        <WebView title={$t`Text`} html={$message.html || ""} {headHTML} autoSize />
         <slot name="bubble" />
       </div>
       <slot name="inner-bottom" />
@@ -60,7 +59,6 @@
   import WebView from "../../Shared/WebView.svelte";
   import { getDateString } from "../../Util/date";
   import { t } from "../../../l10n/l10n";
-  import { onMount } from "svelte";
 
   export let message: Message;
   export let previousMessage: Message = null;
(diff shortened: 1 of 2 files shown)


app/frontend/Chat/MessageView/Message.svelte
```diff
@@ -74,6 +74,21 @@
...
+  onMount(() => {
+    observeMaxWidth();
+  });
+
+  function observeMaxWidth() {
+    const observer = new ResizeObserver((entries) => {
+      const el = entries[0];
+      maxWidth = el.contentRect.width * 0.75;
+    });
+    observer.observe(messageE.parentElement);
```
benbucksch, line 90: Can you move this into the WebView, please? The WebView can access its `parentElement` (or grandparent) just the same, so there should be no need to put it outside the component.

jermy-c: Ok, will do.

benbucksch: Thanks

benbucksch:
Aside from speed, I see 2 issues:
* There's unnecessary white space at the bottom of each msg.
  * Reproduction: Load test data, open Chat app, click on a person) 
  * image
* The width does not use the allowed width, but tries to be as small as possible, making the messages unnessarily thin and tall.

app/frontend/Chat/MessageView/Message.svelte
```diff
@@ -32,10 +32,8 @@
       {/if}
       <slot name="inner-top" />
       <div class="text selectable">
-        {@html $message.html || ""}
-        <!-- TODO Security: Jail HTML into untrusted <iframe> for additional protection.
-        <WebView title={$t`Text`} html={message.html} {headHTML} autoSize />
-        -->
+        <!-- TODO Security: Jail HTML into untrusted <iframe> for additional protection. -->
```
benbucksch, line 35:
```suggestion
```

No longer applies - that's the whole point of this feature.

app/frontend/Shared/WebView.svelte
```diff
@@ -35,6 +35,22 @@
...
 
+  const autoSizeCSS = `<style>
+  body {
+    min-height: 0px !important;
+    min-width: 100px !important;
+    height: fit-content !important;
+    width: fit-content !important;
+    over-flow: visible !important;
+  }
+  </style>`;
```
benbucksch, line 46: Move this into the function where you use it.

benbucksch (changes requested):

benbucksch (approved):

PR #193, opened 2024-09-05
e2: Output ESM instead of CJS for main and preload scripts

jermy-c:
- update better-sqlite3 to 11.2.1 since older versions don't work with electron@32
- update electron-vite to 2.3.0 to support ESM
- update electron to 32.0.1 to support ESM
- use custom ESM shim plugin since the one in `electron-vite` doesn't seem to work nor the `@rollup/esm-shim` plugin, therefore causing the `__dirname` syntax error
- fix syntax errors to be compatible with ESM
- set the config to output ESM for `main` and `preload` scripts
- in the main script set the `preload` to mjs because it outputs it with the`mjs` file extension for some reason

Where should we point the `esm-shim-plugin` package? It is a package based on https://github.com/rollup/plugins/issues/1709#issuecomment-2112362911. Will it be moved to Mustang?

### DIFF
--- a/e2/electron.vite.config.ts
+++ b/e2/electron.vite.config.ts
@@ -1,11 +1,13 @@
 import { defineConfig, externalizeDepsPlugin } from 'electron-vite';
 import { nodePolyfills } from "vite-plugin-node-polyfills";
 import { svelte } from '@sveltejs/vite-plugin-svelte';
+import esmShim from 'esm-shim-plugin';
 
 export default defineConfig({
   main: {
     plugins: [
       externalizeDepsPlugin({ exclude: ["@radically-straightforward/sqlite"] }),
+      esmShim(),
     ],
     build: {
       rollupOptions: {
@@ -16,7 +18,10 @@ export default defineConfig({
     },
   },
   preload: {
-    plugins: [externalizeDepsPlugin()],
+    plugins: [
+      externalizeDepsPlugin(),
+      esmShim(),
+    ],
     build: {
       rollupOptions: {
         output: {
--- a/e2/electron.vite.config.ts
+++ b/e2/electron.vite.config.ts
@@ -1,13 +1,15 @@
 import { defineConfig, externalizeDepsPlugin } from 'electron-vite';
 import { nodePolyfills } from "vite-plugin-node-polyfills";
 import { svelte } from '@sveltejs/vite-plugin-svelte';
-import esmShim from 'esm-shim-plugin';
+import replace from '@rollup/plugin-replace';
 
 export default defineConfig({
   main: {
     plugins: [
       externalizeDepsPlugin({ exclude: ["@radically-straightforward/sqlite"] }),
-      esmShim(),
+      replace({
+        __dirname: 'import.meta.dirname',
+      }),
     ],
     build: {
       rollupOptions: {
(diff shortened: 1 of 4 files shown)


e2/electron.vite.config.ts
```diff
@@ -1,15 +1,34 @@
-import { defineConfig, externalizeDepsPlugin } from 'electron-vite'
+import { defineConfig, externalizeDepsPlugin } from 'electron-vite';
 import { nodePolyfills } from "vite-plugin-node-polyfills";
-import { svelte } from '@sveltejs/vite-plugin-svelte'
+import { svelte } from '@sveltejs/vite-plugin-svelte';
+import esmShim from 'esm-shim-plugin';
```
benbucksch, line 4: Thanks for figuring that out.

However, could we please avoid that shim? Do I understand it correctly that it's only needed to make `__dirname` work? And it uses node.js functions to get the dirname? If so, could we instead define a function `dirname`, and use that? Avoiding the shim altogether?

e2/electron.vite.config.ts
```diff
@@ -1,15 +1,34 @@
...
   preload: {
-    plugins: [externalizeDepsPlugin()]
+    plugins: [
+      externalizeDepsPlugin(),
+      esmShim(),
+    ],
+    build: {
+      rollupOptions: {
+        output: {
+          format: 'es',
```
benbucksch, line 28: Another alternative option would be to not use ESM for preload scripts, but to keep those as CJS.
Only the main would be ESM.
Only a suggestion, in case that makes things a lot easier.

benbucksch (changes requested):

benbucksch:
Good changes. Thanks a lot for figuring that out!
I'm glad that it's not as difficult as I thought to change compliation to ESM.

If you could avoid the shim, and if it works well (please test carefully!), then it's good to go. The remaining changes look good to me, assuming that everything works.

e2/electron.vite.config.ts
jermy-c: You want the __dirname to be replaced with dirname as a function? Because the problem is that the build generates code with __dirname for imports and some other functions. And we’d have to find way to insert it. Another way that might work if we use a replace plugin to replace the __dirname with dirname as function and insert the dirname function into the output.

jermy-c: I replaced it with `@rollup/plugin-replace`, now it replaces all instances of `__dirname` with `import.meta.dirname` for the output.

jermy-c:
You're welcome. Thanks!

- I `replaced esm-shim` with `@rollup/plugin-replace`, now it replaces all instances of `__dirname` with `import.meta.dirname` for the output. But it requires Node `v20.11.0` or later. Would that be better or do we still do the shim()? **NOTE**: It is only a release candidate for now and not stable feature yet. https://nodejs.org/api/esm.html#importmetadirname
- Removed unnecessary config for ESM electron-vite config
- It is normal for the preload script file extension to change to `.mjs` or `.cjs` when outputing ESM. https://electron-vite.org/guide/dev#enabling-esm

benbucksch:
Great! Thank you. Much simpler = better.

There's a `__dirname` in our src/main/index.ts . Would it make sense to replace that?

Do we still need the replacement in this case? I assume yes, and that's OK.

jermy-c:
Yes, but it'll replaced by the replace plugin also. So, we don't need to replace it manually.

benbucksch:
Could you please replace it?

jermy-c:
> Could you please replace it?

Ok, done.

benbucksch:
Thanks!
review+

PR #202, opened 2024-09-10
Fix scrollbar styles overridden by scrollbar-color

jermy-c:
It seems like Chrome ignores all ::webkit-scrollbar if there's a single `scrollbar-*` css rule. So I just removed a single style in the body selector.

https://syntackle.com/blog/changes-to-scrollbar-styling-in-chrome-121/

### DIFF
--- a/app/frontend/app.css
+++ b/app/frontend/app.css
@@ -108,7 +108,6 @@ body {
   place-items: center;
   min-width: 640px;
   min-height: 100vh;
-  scrollbar-color: #22222325 transparent;
 }
 
 /* <copied from="content-body.css"> */


benbucksch:
See also https://developer.chrome.com/docs/css-ui/scrollbar-styling?hl=de

@jermy-c Did you try to move the rule to `html`? Does that work in Chrome and Firefox, on all OS?

> Chrome now prefers the new scrollbar properties on the html element over the ::-webkit-* pseudo selectors.

jermy-c:
Moving the `scrollbar-color` rule to `html` still has the problem. The scrollbar looks the same on both OS with Chrome. But on Firefox it has the same color, border-radius but it is thinner and and doesn't always show but only while scrolling.

benbucksch (approved):

PR #203, opened 2024-09-11
ActiveSync: EMail

NeilRashbrook:
Supported features:
- Provision ActiveSync device
- Send mail using MIME source
- Push notifications
- List folders
- Create, rename, move, delete folders
- List messages
- Read some basic message properties
- Read message source
- Mark message read or starred
- Tag message
- Move or delete message

Not supported:
- Copy message
- Create draft
- Mark all messages read

### DIFF
--- a/app/logic/Mail/EWS/EWSFolder.ts
+++ b/app/logic/Mail/EWS/EWSFolder.ts
@@ -378,7 +378,7 @@ export class EWSFolder extends Folder {
   }
 
   async moveFolderHere(folder: EWSFolder) {
-    assert(folder.account = this.account, "Cannot move folders between accounts");
+    assert(folder.account == this.account, "Cannot move folders between accounts");
     let request = {
       m$MoveFolder: {
         m$ToFolderId: {
--- a/app/logic/Mail/SQL/SQLEMail.ts
+++ b/app/logic/Mail/SQL/SQLEMail.ts
@@ -232,7 +232,7 @@ export class SQLEMail {
     email.pID = typeof(row.pID) == "number"
       ? sanitize.integer(row.pID, null)
       : sanitize.string(row.pID, null);
-    email.id = sanitize.nonemptystring(row.messageID);
+    email.id = sanitize.nonemptystring(row.messageID, "");
     email.inReplyTo = sanitize.string(row.parentMsgID, null);
     email.size = sanitize.integer(row.size, null);
     email.sent = sanitize.date(row.dateSent * 1000, new Date());
--- a/app/logic/util/util.ts
+++ b/app/logic/util/util.ts
@@ -128,6 +128,12 @@ export class NotImplemented extends Error {
   }
 }
 
+export class NotSupported extends Error {
+  constructor(msg?: string) {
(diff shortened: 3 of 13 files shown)


benbucksch:
:tada: :fireworks: :partying_face:

app/logic/Mail/SQL/SQLEMail.ts
```diff
@@ -232,7 +232,7 @@ export class SQLEMail {
     email.pID = typeof(row.pID) == "number"
       ? sanitize.integer(row.pID, null)
       : sanitize.string(row.pID, null);
-    email.id = sanitize.nonemptystring(row.messageID);
+    email.id = sanitize.nonemptystring(row.messageID, "");
```
benbucksch, line 235: Why?

app/package.json
```diff
@@ -38,6 +38,7 @@
     "@xmpp/client": "^0.13.1",
     "axios": "^1.6.2",
     "dompurify": "^3.1.1",
+    "email-addresses": "^5.0.0",
```
benbucksch, line 41: Good choice of library: No dependencies, small code, and focus on RFC compliance.
I'll try to use that in other places where we need as well.

app/logic/Mail/EWS/EWSFolder.ts
```diff
@@ -378,7 +378,7 @@ export class EWSFolder extends Folder {
   }
 
   async moveFolderHere(folder: EWSFolder) {
-    assert(folder.account = this.account, "Cannot move folders between accounts");
+    assert(folder.account == this.account, "Cannot move folders between accounts");
```
benbucksch, line 381: Whoops. Thanks for finding this.

app/logic/util/util.ts
```diff
@@ -128,6 +128,12 @@ export class NotImplemented extends Error {
   }
 }
 
+export class NotSupported extends Error {
+  constructor(msg?: string) {
+    super(msg ?? "Operation not supported");
```
benbucksch, line 133: User-visible strings must be translated. Same elsewhere.

```suggestion
    super(msg ?? gt`Operation not supported`);
```

benbucksch (approved):

benbucksch:
@NeilRashbrook Thanks so much!!!

app/logic/Mail/EWS/EWSFolder.ts
NeilRashbrook: Looks like IMAP is also affected, presumably because I originally copied the assert from there.

app/logic/Mail/SQL/SQLEMail.ts
NeilRashbrook: ActiveSync doesn't expose an email's message ID, so we'll only have the message ID if we actually download the message, but we still need to be able to save the message, otherwise we'll forget that it exists.

app/logic/util/util.ts
NeilRashbrook: I was just copying `NotImplemented` above...

PR #205, opened 2024-09-11
OWA: Outlook.com: Login success not detected - fixes #198

### LINKED ISSUE #198: OWA: Outlook.com: Login success not detected
### Reproduction
1. Set up @outlook.com email address
2. Choose OWA as protocol
3. -> Login window opens
4. Login in
5. Do not remember login, Do not remember this decition

### Actual result
* The OWA webmail frontend shows in the login window
* Login window stays open
* When closing the window, the login fails
* on `startupdata.ashx` call, there is a `X-OWA-CANARY` session cookie, on the `/owa/0/` path (!), on host `outlook.live.com`.

### Expected result
* Login success is detected
* Window closes
* Setup finishes
* Emails fetch into Mustang

### Cause
* The `/owa/0/` path? We have specific code in Owl login code to handle that for hotmail.

jermy-c:


backend/owa.ts
```diff
@@ -1,8 +1,21 @@
...
+const kCookieName = "X-OWA-CANARY";
+const kHotmailServer = "outlook.live.com";
+const reOWAPath = /^\/owa\/(\d+\/){0,1}/;
+
 export async function fetchSessionData(partition: string, url: string, interactive: boolean) {
   let session = Session.fromPartition(partition);
-  let response = await session.fetch(url + 'sessiondata.ashx', { method: 'POST' });
+  let urlObj = new URL(url);
+  let cookies = await session.cookies.get({ name: 'X-OWA-CANARY' });
+  if (cookies.length) {
```
jermy-c, line 11: @benbucksch Do you think we should check if there are existing cookies and setting the URL based on the existing cookies? Because that's the last issue I have with this. Because

https://github.com/mustang-im/mustang/blob/2dca9083b3d6b0c91d18263404799a821731674e/app/logic/Mail/AutoConfig/saveConfig.ts#L15

Calls it with interactive true and was going to use that to determine whether or not to set the URL based on the domain and path of the existing cookies.

And that causes an issue where when you click `Next` it shows popup again because on the first fetch there's an error and interactive is set to true but it is still saving.

backend/owa.ts
```diff
@@ -1,8 +1,21 @@
...
+  let urlObj = new URL(url);
+  let cookies = await session.cookies.get({ name: kCookieName });
+  if (cookies.length) {
+    urlObj.hostname = cookies[0].domain;
+    let owaCookies = cookies.filter(cookie => cookie.path?.match(reOWAPath));
+    if (owaCookies.length && urlObj.pathname.match(reOWAPath)) {
+      urlObj.pathname = owaCookies[0].path + urlObj.pathname.replace(reOWAPath , "");
+    }
+  }
+  let response = await session.fetch(urlObj.toString() + 'sessiondata.ashx', { method: 'POST' });
```
benbucksch, line 18:
```suggestion
  url = urlObj.toString();
  let response = await session.fetch(url + 'sessiondata.ashx', { method: 'POST' });
```

backend/owa.ts
```diff
@@ -1,8 +1,21 @@
 import { session as Session, BrowserWindow } from "electron";
 
+const kCookieName = "X-OWA-CANARY";
+const kHotmailServer = "outlook.live.com";
+const reOWAPath = /^\/owa\/(\d+\/){0,1}/;
```
benbucksch, line 5: Owl has:
```
          details.cookie.domain == kHotmailServer &&
          this.cookieNames.includes(details.cookie.name) &&
          details.cookie.path.startsWith("/owa/")) {
```
Can you please use `.startsWith("/owa/")` instead of the regexp?

backend/owa.ts
```diff
@@ -33,26 +46,40 @@ export async function fetchSessionData(partition: string, url: string, interacti
...
         }
       }
-      let checkCanary = function(_event, cookie, _cause, removed) {
-        if (!removed && cookie.name == 'X-OWA-CANARY') {
-          checkLoginFinished();
+      let checkCanary = async function(_event, cookie, _cause, removed) {
+        // For Hotmail and personal accounts, check the path to the CANARY cookie.
+        if (cookie.domain == kHotmailServer &&
+          cookie.name == kCookieName &&
+          cookie.path?.match(reOWAPath)) {
```
benbucksch, line 58: See above.

backend/owa.ts
benbucksch: Sorry, I don't understand any part of the question.

benbucksch (changes requested):

backend/owa.ts
```diff
@@ -33,26 +46,40 @@ export async function fetchSessionData(partition: string, url: string, interacti
...
+        // Hotmail also sets cookies for /owa/0/, /mail/0/, /calendar/0/ etc.,
+        // but we can use only the /owa/0/ cookie.
+        // We also need to use that URL for the service request.
+        // This needs to happen before CheckLoginFinished().
+          urlObj.hostname = cookie.domain;
+          urlObj.pathname = cookie.path;
+        }
+        if (!removed &&
```
benbucksch, line 66: If `removed` actually happens, then you should check that at the top of the function and make an early exit. See above.

backend/owa.ts
```diff
@@ -33,26 +46,40 @@ export async function fetchSessionData(partition: string, url: string, interacti
...
-          response = await session.fetch(url + 'sessiondata.ashx', { method: 'POST' });
+          response = await session.fetch(urlObj.toString() + 'sessiondata.ashx', { method: 'POST' });
           finish(await response.json());
         } catch (ex) {
         }
       }
-      let checkCanary = function(_event, cookie, _cause, removed) {
-        if (!removed && cookie.name == 'X-OWA-CANARY') {
-          checkLoginFinished();
+      let checkCanary = async function(_event, cookie, _cause, removed) {
```
benbucksch, line 54:
```suggestion
      let checkCanary = async function(_event, cookie, _cause, removed) {
        if (removed) {
          return;
        }
```

benbucksch (changes requested):

backend/owa.ts
```diff
@@ -1,7 +1,20 @@
...
+const kCookieName = "X-OWA-CANARY";
+const kHotmailServer = "outlook.live.com";
+
 export async function fetchSessionData(partition: string, url: string, interactive: boolean) {
   let session = Session.fromPartition(partition);
+  let urlObj = new URL(url);
+  let cookies = await session.cookies.get({ name: kCookieName });
+  if (cookies.length) {
+    urlObj.hostname = cookies[0].domain;
+    let owaCookies = cookies.filter(cookie => cookie.path == "/owa/");
```
benbucksch, line 12: That won't work. It also needs to be `startsWith("/owa/")`

benbucksch: Your logic is still different from Owl. Why don't you just use the code logic? That's known to work.

```
   onCookie() ...
      if (isHotmail &&
          details.cookie.domain == kHotmailServer &&
          this.cookieNames.includes(details.cookie.name) &&
          details.cookie.path.startsWith("/owa/")) {
        // Hotmail also sets cookies for /owa/0/, /mail/0/, /calendar/0/ etc.,
        // but we can use only the /owa/0/ cookie.
        // We also need to use that URL for the service request.
        // This needs to happen before CheckLoginFinished().
        this.serverURL.pathname = details.cookie.path;
      }
```

Please include the comment as well.

backend/owa.ts
```diff
@@ -33,26 +46,43 @@ export async function fetchSessionData(partition: string, url: string, interacti
...
-          response = await session.fetch(url + 'sessiondata.ashx', { method: 'POST' });
+          response = await session.fetch(urlObj.toString() + 'sessiondata.ashx', { method: 'POST' });
           finish(await response.json());
         } catch (ex) {
         }
       }
-      let checkCanary = function(_event, cookie, _cause, removed) {
-        if (!removed && cookie.name == 'X-OWA-CANARY') {
-          checkLoginFinished();
+      let checkCanary = async function(_event, cookie, _cause, removed) {
```
benbucksch, line 54: Let's rename the function while we're at it (you also need to adapt the caller):

```suggestion
      let onCookie = async function(_event, cookie, _cause, removed) {
```

benbucksch (changes requested):

benbucksch:
BTW: When you make a pull request, please mention the github issue number. #198 in this case.

benbucksch:
I think the problem is that we discover a new server URL using the cookie. Owl sets `this.serverURL.pathname = details.cookie.path` , whereas `this` refers to the account object. This changes where the following REST requests (not login) will go to. However, here in Mustang, the code is in the backend, so it doesn't have direct access to the account object to change the URL. So, you need to send that URL back as part of the function result. The function result is sent in `finish(await response.json());`.

backend/owa.ts
```diff
@@ -25,31 +32,52 @@ export async function fetchSessionData(partition: string, url: string, interacti
           return;
         }
         finished = true;
-        session.cookies.removeListener('changed', checkCanary);
+        session.cookies.removeListener('changed', onCookie);
+        if (isHotmail) {
+          data.owaURL = url;
+        }
```
NeilRashbrook, line 38: `data` will be null if the user closes the window instead of logging in. I also found it confusing that you're using the value of `url` set below in `checkLoginFinished`. Could this be moved into `checkLoginFinished`?

jermy-c: Ok, thanks for the catch.

backend/owa.ts
```diff
@@ -25,31 +32,53 @@ export async function fetchSessionData(partition: string, url: string, interacti
...
+          // but we can use only the /owa/0/ cookie.
+          // We also need to use that URL for the service request.
+          // This needs to happen before CheckLoginFinished().
+          urlObj.hostname = cookie.domain;
+          urlObj.pathname = cookie.path;
+          isHotmail = true;
+        }
+        // If we receive the canary cookie, check whether we're logged in
+        if (cookie.name == kCookieName &&
+            cookie.domain == urlObj.hostname) {
```
benbucksch, line 71: Do we need this domain check condition? It wasn't there before.

backend/owa.ts
```diff
@@ -1,10 +1,17 @@
 import { session as Session, BrowserWindow } from "electron";
 
+const kCookieName = "X-OWA-CANARY";
```
benbucksch, line 3:
```suggestion
const kCanaryName = "X-OWA-CANARY";
```

benbucksch (changes requested):
Could you please also change the other 2 cases to the constant?

```patch
-cookies.get({ name: 'X-OWA-CANARY' });
+cookies.get({ name: kCanaryName });
```

backend/owa.ts
```diff
@@ -25,31 +32,53 @@ export async function fetchSessionData(partition: string, url: string, interacti
...
+          // We also need to use that URL for the service request.
+          // This needs to happen before CheckLoginFinished().
+          urlObj.hostname = cookie.domain;
+          urlObj.pathname = cookie.path;
+          isHotmail = true;
+        }
+        // If we receive the canary cookie, check whether we're logged in
+        if (cookie.name == kCookieName &&
+            cookie.domain == urlObj.hostname) {
+            checkLoginFinished();
```
benbucksch, line 72: Missing `await`! This was already wrong in the previous version.

@jermy-c @NeilRashbrook Please pay attention that *every* call to an async function *always* has an `await` in front of it. Or, if you specifically do not want to wait, at the very least a `.catch()`. A call to an async function without these will throw an "unhandled exception", which might kill the entire app, i.e. that's a crash bug.

benbucksch (approved):

backend/owa.ts
jermy-c: Sorry about that. Thanks for the catch! I'll pay more attention to those.

backend/owa.ts
jermy-c: I think we need it just in case we don't get the correct cookie for personal accounts yet i.e. the cookie should be for `outlook.live.com` instead of `outlook.office.com`.

NeilRashbrook: It could also be because the cookie notifications in a browser extension aren't filtered by partition like they are here. (So in fact the Owl code should start checking the `cookieStoreId` of the cookie as well as or instead of the domain.)

Issue #157, opened 2024-07-23
OWA: Login is not saved


benbucksch:
### Reproduction
1. Set up OWA account ben@bexchange
2. Login
3. Restart Mustang
4. Click on account and [Get mail]
5. Click on [Get mail] again

### Actual result
* After step 3, the account shows as logged out (icon: letter with x)
* After step 4, you see an orange error "Please login"
* After step 5, you see the OWA login window. You have to manually enter the username and password.

### Expected result
* After step 3, the account is logged in
* After step 4, you get new mail.
* If you were logged out before, you see the OWA Login window, not "Please login" error.

NeilRashbrook:
(The correct step 4 is to click on Posteingang.)

benbucksch:
Right, clicking on INBOX = the folder, not the account, yes.

NeilRashbrook:
Note that regular OWA won't let you stay logged in after a restart; this feature is limited to the "Stay Logged In" option of Office 365.

benbucksch:
> regular OWA won't let you stay logged in after a restart

How would OWA know about a restart? Are you saying that these are session cookies? Can we tell the browser to make them persistent cookies?

NeilRashbrook:
Of course they're session cookies. They've always been session cookies. We looked at this in Owl five years ago, but nothing ever came of it, presumably because its auto-login content script was good enough. But there's some [code](https://github.com/electron/electron/issues/9995) on GitHub that might work. (I say might because it assumes there's only one partition and that you have somewhere to control adding the cookie listener.)

benbucksch:
So, in Owl, we:
* Do background login - esp. for on-premise Exchange with standard login form
* For Office365 and similar login forms, we open the window, but fill in the email address and password automatically.
In combination, that saves the user from having to log in manually, in many cases.

benbucksch:
Use the Owl source code to implement this.

jermy-c:
> Do background login, by submitting the login form directly in the background, without UI - esp. for on-premise Exchange with standard login form

This is already implemented by Neil on [`neil/owa-login-in-background`](https://github.com/mustang-im/mustang/compare/neil/owa-login-in-background) but not working as it yet doesn't detect any forms. 

> For Office365 and similar login forms, we open the window, but fill in the email address and password automatically.

[...]

PR #218, opened 2024-09-23
Properly escape variables injected into auto fill script #157

NeilRashbrook:
The preferred way to "uneval" a value to be evaluated through `executeJavaScript` appears to be `JSON.stringify` (obviously assuming that the value has a JSON-safe type in the first place). Note that this includes any quotes (or other necessary metacharacters where relevant) so those aren't needed in the template string.

As for the function, it's clearer simply to wrap the function declaration in parentheses, as this turns it into an expression instead. And as a function expression, we can then pass arguments to it, which makes the value injection even clearer still.

### DIFF
--- a/app/logic/Mail/OWA/OWALoginAutoFill.ts
+++ b/app/logic/Mail/OWA/OWALoginAutoFill.ts
@@ -1,13 +1,11 @@
 
 export function owaAutoFillLoginPage(username = "", password = "") {
-  return (functionAsString(autoFillPageScript)
-    .replaceAll("$USERNAME$", username)
-    .replaceAll("$PASSWORD$", password));
+  return `(${autoFillPageScript})(${JSON.stringify(username)}, ${JSON.stringify(password)})`;
 }
 
 /** This script will be exported as string and executed within the login page.
  * Never run it here in frontend. */
-function autoFillPageScript() {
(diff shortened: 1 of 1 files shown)


app/logic/Mail/OWA/OWALoginAutoFill.ts
```diff
@@ -1,13 +1,11 @@
 
 export function owaAutoFillLoginPage(username = "", password = "") {
-  return (functionAsString(autoFillPageScript)
-    .replaceAll("$USERNAME$", username)
-    .replaceAll("$PASSWORD$", password));
+  return `(${autoFillPageScript})(${JSON.stringify(username)}, ${JSON.stringify(password)})`;
```
benbucksch, line 3: Could we just
```suggestion
  return (functionAsString(autoFillPageScript)
    .replaceAll("$USERNAME$", ${JSON.stringify(username)}
    .replaceAll("$PASSWORD$", ${JSON.stringify(password)}));
```
?

NeilRashbrook: I don't know why you think fragile string replacements are better than constructing a function call.

benbucksch (approved):

PR #221, opened 2024-09-27
ActiveSync: Contacts

NeilRashbrook:
- Supports multiple contact folders as separate address books
- Supports three email addresses
- Supports work, home and other street addresses of five lines
- Supports three chat accounts but can't save the type
- Supports two work telephone numbers, two home telephone numbers and one mobile telephone number
- Supports notes
- Would support work and home fax number and company details if mustang did
- Some missing WBXML tags have been added


app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
```diff
@@ -1,16 +1,129 @@
...
+    }
+    try {
+      this.syncKeyBusy = this.makeSyncRequest(data);
+      return await this.syncKeyBusy;
+    } finally {
+      this.syncKeyBusy = null;
+    }
+  }
+
+  protected async makeSyncRequest(data?: any): Promise<any> {
```
benbucksch, line 53: Can we please document the return value, at least in a JSDoc comment?

Also, it's kind-of funny to have an `async syncReq()` ;-) Better: `makeSynchronizeRequest()`. Ditto for `queued` above.

app/logic/Contacts/ActiveSync/ActiveSyncPerson.ts
```diff
@@ -1,4 +1,123 @@
...
+  }
+  set serverID(val) {
+    this.id = val;
+  }
+
+  fromWBXML(wbxmljs: any) {
+    this.firstName = sanitize.nonemptystring(wbxmljs.FirstName, "");
+    this.lastName = sanitize.nonemptystring(wbxmljs.LastName, "");
+    this.name = this.firstName ? this.lastName ? `${this.firstName} ${this.lastName}` : this.firstName : this.lastName;
+    this.emailAddresses.replaceAll([wbxmljs.Email1Address, wbxmljs.Email2Address, wbxmljs.Email3Address].filter(Boolean).map(address => new ContactEntry((parseOneAddress(address) as ParsedMailbox).address, "work", "mailto")));
```
benbucksch, line 41: Can you please document what format `wbxmljs.Email1Address` is in? I'm assuming there's a reason why you use the `email-addresses` module. Is there a "Bob Sample <bob@example.com>" name in there? Or is just just about the `<>`s?

Could you please wrap the result of `parseOneAddress().address` with `sanitize.emailAddress()` ?

app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
```diff
@@ -1,16 +1,129 @@
...
+    if (response.Collections.Collection.Status == "3") {
+      // Out of sync.
+      this.syncState = null;
+      await SQLAddressbook.save(this);
+    }
+    if (response.Collections.Collection.Status != "1") {
+      throw new EASError("Sync", response.Collections.Collection.Status);
+    }
+    this.syncState = response.Collections.Collection.SyncKey;
+    await SQLAddressbook.save(this);
```
benbucksch, line 75: You're saving the sync state before we had a chance to parse and save the result. If this fails for some reason, we missed these changes or (worse) become incoherent with our local copy.

Instead, we need to:
* Lock (I guess that's what `queued()` does)
* Fetch
* Parse the result
* Set and save the sync key.
* Unlock

benbucksch (changes requested):

app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
```diff
@@ -1,16 +1,129 @@
...
   account: ActiveSyncAccount;
+  syncKeyBusy: Promise<any> | null;
+  readonly folderClass = "Contacts";
+
+  get serverID() {
+    return new URL(this.url).searchParams.get("serverID");
+  }
+
+  async ping() {
+    await this.listContacts();
```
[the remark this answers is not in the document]
NeilRashbrook: `listContacts` only ever pulls in changes (unless the sync state gets corrupt, in which case it has to start over).

app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
NeilRashbrook: The return value depends on the request being made, so no, I can't document it.

And it's called `makeSyncRequest` because the name of the request that's being made is "Sync".

app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
NeilRashbrook: Ugh, I've got the same pattern in `ActiveSyncAccount` for the global sync key, so I'll have to fix that first.

app/logic/Contacts/ActiveSync/ActiveSyncPerson.ts
NeilRashbrook: Yes, it's a display address with `<>`s, which  GitHub is inconveniently hiding for us.

Sure, I can wrap a complex address parser with a simple sanitiser, but you won't get any extra protection.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -70,6 +72,13 @@ export class ActiveSyncAccount extends MailAccount {
     } else {
       await this.verifyLogin();
     }
+
+    for (let addressbook of appGlobal.addressbooks) {
+      if (addressbook.protocol == "addressbook-activesync" && addressbook.url.startsWith(this.url + "?") && addressbook.username == this.emailAddress) {
```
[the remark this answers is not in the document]
NeilRashbrook: Well this is all over EWS and OWA too, so...

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -310,11 +321,31 @@ export class ActiveSyncAccount extends MailAccount {
         break;
       case FolderType.Calendar:
       case FolderType.UserCalendar:
-        // (Re)create calendar account here.
+        let calendar = appGlobal.calendars.find((calendar: ActiveSyncCalendar) => calendar.protocol == "addressbook-activesync" && calendar.url == url.toString() && calendar.username == this.emailAddress) as ActiveSyncCalendar | void;
+        if (calendar) {
+          calendar.name = change.DisplayName;
```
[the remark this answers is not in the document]
NeilRashbrook: This only happens when the user renames the calendar on the server.

app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
benbucksch: Please give it some thought which cases can happen and how we should recover from them.

* If, for example, the processing of the changes throws reproducibly on a given change set, then if we don't save the sync key in case of error, the next attempt would run into the same error again and be stuck forever.
* If the error is temporarily, e.g. because the received data is corrupted on the wire, then we would want to re-try.

benbucksch: Regarding the code, one way would be to wrap the async parsing function, so you could still keep saving the sync key in the same function here. Compare `IMAPFolder.runCommand()` and how it's called.

benbucksch: I've filed #225 about this.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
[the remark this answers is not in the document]
benbucksch: OK, I've filed #226 about it.

app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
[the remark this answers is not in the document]
benbucksch: So, it's actually polling. Shall we rename `ping` to `poll`, then?

app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
benbucksch: > The return value depends on the request being made, so no, I can't document it.

You're returning `response.Collections.Collection`, so there must be something that you can say about the result. "It's an array of ..."

> And it's called `makeSyncRequest` because the name of the request that's being made is "Sync".

Yes, I understood (after a while) what you meant, that's why I made a concrere suggestion. 

I thought at first that you make a non-async request. So, "Sync" is ambiguous.

benbucksch (approved):
We'll address the review comments in a followup commit.

PR #222, opened 2024-10-04
OWA: Don't throw "end of json" error for response.json() - fixes #215

### LINKED ISSUE #215: OWA: Outlook.com: Unexpected end of JSON input
### Reproduction
* Set up an @outlook.com personal account
* Choose protocol OWA
* Click on Inbox
* Login

### Actual result
* Login appears to be working (not sure)
* Error: "Unexpected end of JSON input"
* Stack: 
image

### Expected result
* No error
* Login works

### Scope
* Business accounts (neil@beonex.on...) seem to work, without the error

jermy-c:
- `JSON.parse()` normally throws an error when parsing an empty string or null
- `response.json()` under the hood uses `JSON.parse()` so it'll throw an error when it's empty
- `json.cid` would throw an error if json is null so we just use `json ? json.cid : null` to prevent that

### DIFF
--- a/backend/owa.ts
+++ b/backend/owa.ts
@@ -139,9 +139,8 @@ export async function fetchJSON(partition: string, url: string, action: string, 
   try {
     result.json = await response.json();
   } catch (ex) {
-    if (ex instanceof SyntaxError &&
-      ex.message == "Unexpected end of JSON input") {
-    } else {
+    if (ex !instanceof SyntaxError &&
+      ex.message != "Unexpected end of JSON input") {
       throw ex;
     }
   }
--- a/backend/owa.ts
+++ b/backend/owa.ts
@@ -136,7 +136,15 @@ export async function fetchJSON(partition: string, url: string, action: string, 
   result.statusText = response.statusText;
   result.url = response.url;
   result.contentType = response.headers.get('Content-Type');
-  result.json = await response.json();
+  try {
+    result.json = await response.json();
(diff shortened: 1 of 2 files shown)


backend/owa.ts
```diff
@@ -136,7 +136,14 @@ export async function fetchJSON(partition: string, url: string, action: string,
...
   result.url = response.url;
   result.contentType = response.headers.get('Content-Type');
-  result.json = await response.json();
+  try {
+    result.json = await response.json();
+  } catch (ex) {
+    if (ex !instanceof SyntaxError &&
+      ex.message != "Unexpected end of JSON input") {
+      throw ex;
+    }
```
benbucksch, line 145:
```suggestion
    if (ex instanceof SyntaxError && ex.message == "Unexpected end of JSON input" || ex.message.includes("DOCTYPE")) {
      response.json = null;
    } else {
      throw ex;
    }
```

benbucksch (approved):

benbucksch:
@NeilRashbrook Sorry, our comments crossed. Feel free to make a new PR.

benbucksch:
I tested this PR and the error is gone for me. However, please note https://github.com/mustang-im/mustang/issues/215#issuecomment-2395145584

backend/owa.ts
jermy-c: Ok, understood.

Issue #226, opened 2024-10-05
EWS, OWA, ActiveSync: Don't compare username with emailAddress


benbucksch:
Break-off from #155 and https://github.com/mustang-im/mustang/pull/221/files/0cf3795902c0ddd673444d44ce282720d10d57da#r1786799005

EWS, OWA and ActiveSync implementations of AddressBook and Calendar compare username with email address, e.g. addressbook.username == this.emailAddress, which is making the false assumption that username == email address. Please change those to compare apples with apples and oranges with oranges only, i.e. username with username (or emailAddress with emailAddress), never criss-cross.

@jermy-c Can you please make a PR for that? This will not fix #155 properly, but at least fix one bug.

PR #228, opened 2024-10-07
 EWS, OWA, ActiveSync: Don't compare username with emailAddress #226 

jermy-c:
- EWS, OWA, ActiveSync: Compare username with username

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -83,7 +83,7 @@ export class ActiveSyncAccount extends MailAccount {
     }
 
     for (let addressbook of appGlobal.addressbooks) {
-      if (addressbook.protocol == "addressbook-activesync" && addressbook.url.startsWith(this.url + "?") && addressbook.username == this.emailAddress) {
+      if (addressbook.protocol == "addressbook-activesync" && addressbook.url.startsWith(this.url + "?") && addressbook.username == this.username) {
         (addressbook as ActiveSyncAddressbook).account = this;
         await (addressbook as ActiveSyncAddressbook).listContacts();
       }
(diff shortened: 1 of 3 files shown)


benbucksch:
Looks good and correct and obvious to me. Just to double-check:

@NeilRashbrook Was there a specific reason why you compared with `this.emailAddress` instead of `this.username`? The latter seems to be obvious, so I wonder whether you had a reason to do it differently.

@jermy-c 
1. Is `this.username` correctly populated in all cases, including during startup, before this code runs? (This is a code-inspection verification.)
2. Did you test it? (This is a practical verification.)

NeilRashbrook:
> Was there a specific reason why you compared with `this.emailAddress` instead of `this.username`? The latter seems to be obvious, so I wonder whether you had a reason to do it differently.

It's because in the dim and distant past mail account would only reliably have the email address set, as opposed to other accounts which only have a username. (In fact there is still a `this.username || this.emailAddress` in `EWSAccount.ts` which since got copied to `ActiveSyncAccount.ts`; sorry about that.) For OWA it seems that I actually ignore the username which seems to be a bug (I assume account creation will set it to be the same as the email address unless the user manually configures the account).

NeilRashbrook (approved):

benbucksch:
> It's because in the dim and distant past mail account would only reliably have the email address set

@NeilRashbrook I see, thanks. Is that definitely "past" now?

jermy-c:
> Is this.username correctly populated in all cases, including during startup, before this code runs? (This is a code-inspection verification.)
Yes.

> Did you test it? (This is a practical verification.)
EWS and OWA addressbook and calendar works for business accounts. ActiveSync is buggy because it's not showing any folders so I couldn't test the addressbook and calendar.

I didn't see any contacts or calendar events from personal accounts from any of the protocols.

benbucksch:
@jermy-c Thanks for having checked!

PR #231, opened 2024-10-10
EWS, OWA, ActS: When changes are coming from server, don't push them back to the server

NeilRashbrook:
As mentioned in the aforementioned commit these calls are just to update our local copy of the object from the server.

### DIFF
--- a/app/logic/Calendar/EWS/EWSCalendar.ts
+++ b/app/logic/Calendar/EWS/EWSCalendar.ts
@@ -78,7 +78,6 @@ export class EWSCalendar extends Calendar {
           if (event) {
             this.events.remove(event);
             this.events.removeAll(event.instances);
-            await event.deleteFromServer();
             await event.deleteIt();
           }
         }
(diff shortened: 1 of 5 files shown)


benbucksch:
The problem here is a different understanding of the purpose of the `save()` method. I meant it to do only the local save to the local DB or filesystem. You understood it to also save to the server and implemented it this way.

Could you please use `saveToServer()` for the latter? Then this patch here becomes unnecessary, but you'll have to fox the other callers which intended to save to server.

benbucksch:
@jermy-c Do you want to pick this up?

benbucksch:
@NeilRashbrook I don't really know what to do with this PR.

Could you please give the Pull Request a proper title, in the style as the other commit messages? They generally have the form:
<component or protocol>: <Topic>: <3-5 words describing the purpose of the change>

The summary describes the purpose of the change (e.g. "ensure we save"), not the code ("call foo()").

In the description, you describe the rationale of this change, what exact problem it fixes. If there were multiple alternative solutions, you can also describe which alternatives you considered and rejected, and why.

No commit IDs in the summary, please. You can put them in the longer description, but please no more than 10 digits.

benbucksch:
@NeilRashbrook I've adapted the title here, with a summary that describes the high-level idea of this change. I hope this serves as good example.

Thanks for this change. Obviously, the previous changes made no sense. Thanks for catching and fixing this!

PR #232, opened 2024-10-11
Calendar and AB: setStorage()

jermy-c:
This is a fix for https://github.com/mustang-im/mustang/commit/32c2687493dad9870a00dc7451b734e157bcb511#r147825086

### DIFF
--- a/app/logic/Calendar/Calendar.ts
+++ b/app/logic/Calendar/Calendar.ts
@@ -1,5 +1,6 @@
 import { Account } from "../Abstract/Account";
 import { Event } from "./Event";
+import { setStorage } from "./Store/setStorage";
 import { appGlobal } from "../app";
 import { Collection, ArrayColl } from "svelte-collections";
 
@@ -8,6 +9,11 @@ export class Calendar extends Account {
   readonly events = new ArrayColl<Event>();
   storage: CalendarStorage | null = null;
   syncState: string | null = null;
+
+  constructor() {
+    super();
+    setStorage(this);
+  }
 
   newEvent(parentEvent?: Event): Event {
     return new Event(this, parentEvent);
--- a/app/logic/Calendar/SQL/SQLCalendar.ts
+++ b/app/logic/Calendar/SQL/SQLCalendar.ts
@@ -1,6 +1,7 @@
 import type { Calendar } from "../Calendar";
 import { getDatabase } from "./SQLDatabase";
 import { newCalendarForProtocol } from "../AccountsList/Calendars";
+import { setStorage} from "../Store/setStorage";
 import { appGlobal } from "../../app";
 import { backgroundError } from "../../../frontend/Util/error";
 import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
@@ -40,6 +41,7 @@ export class SQLCalendar {
         WHERE id = ${cal.dbID}
         `);
     }
+    setStorage(cal);
   }
 
   /** Also deletes all persons and groups in this address book */
@@ -70,6 +72,7 @@ export class SQLCalendar {
       ? appGlobal.workspaces.find(w => w.id == sanitize.string(row.workspace, null))
       : null;
     cal.syncState = row.syncState;
+    setStorage(cal);
     return cal;
   }
 
--- a/app/logic/Calendar/SQL/SQLCalendar.ts
+++ b/app/logic/Calendar/SQL/SQLCalendar.ts
@@ -1,8 +1,6 @@
-import type { Calendar, CalendarStorage } from "../Calendar";
-import type { Event } from "../Event";
+import type { Calendar } from "../Calendar";
 import { getDatabase } from "./SQLDatabase";
 import { newCalendarForProtocol } from "../AccountsList/Calendars";
-import { SQLEvent } from "./SQLEvent";
 import { appGlobal } from "../../app";
 import { backgroundError } from "../../../frontend/Util/error";
 import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
@@ -10,7 +8,7 @@ import { assert } from "../../util/util";
 import { ArrayColl } from "svelte-collections";
 import sql from "../../../../lib/rs-sqlite";
 
-export class SQLCalendar implements CalendarStorage {
+export class SQLCalendar {
   static async save(cal: Calendar) {
     if (!cal.dbID) {
       let existing = await (await getDatabase()).get(sql`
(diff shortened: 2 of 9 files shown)


jermy-c:
I'm a bit confused because `setStorage()` only for MailAccounts and can't be called in Addressbook or Contact accounts

NeilRashbrook:
The bug with addressbook and contact accounts is that references to `SQLAddressbook.save();` and `SQLCalendar.save();` were changed to `this.save();` however that tries to save using `storage` which only gets set by the above two calls. My best guess at the fix is to change `Addressbook.save` and `Calendar.save` to call those functions for now.

jermy-c:
When I set the storage in the `Addressbook.save` with `setStorage` function for addressbooks, I get an error.

```ts
import { Addressbook } from '../Addressbook';
import { SQLAddressbook } from '../SQL/SQLAddressbook';

export function setStorage(acc: Addressbook) {
  if (!acc.storage) {
    acc.storage = new SQLAddressbook();
  }
}
```

```
OWAAddressbook.ts:8 Uncaught ReferenceError: Cannot access 'Addressbook' before initialization
    at OWAAddressbook.ts:8:37
```

benbucksch:
setStorage should be called during the init phase of the object (or login), not before every save.
-- 
Sent from my phone. Please excuse the brevity.

NeilRashbrook:
I think the problem is that something imports `Addressbook` imports `setStorage` imports `SQLAddressbook` imports `newAddressbookForProtocol` imports `OWAAddressbook` imports `Addressbook`. But `Addressbook` is still being imported at this point, thus its class hasn't been defined yet, thus the error. Unfortunately I can't see an easy fix for that.

jermy-c:
> setStorage should be called during the init phase of the object (or login), not before every save.

I tried that but I'm still getting the same error.

> I think the problem is that something imports `Addressbook` imports `setStorage` imports `SQLAddressbook` imports `newAddressbookForProtocol` imports `OWAAddressbook` imports `Addressbook`. But `Addressbook` is still being imported at this point, thus its class hasn't been defined yet, thus the error. Unfortunately I can't see an easy fix for that.

Seems like this is the case because I get the error whenever I use either of the storage classes in the main class `Addressbook` or `Calendar` but I don't get the error when using it in subclasses like OWACalendar. I just made the commits based on that but then it there's no way for local calendars or addressbooks use the storage class. 

The mail account doesn't have this problem however, and my guess is because of the methods/interface are all in a seperate class the `SQLStorage` class so it's not directly import thus preventing the circular imports problem.

benbucksch:
Yes, setStorage was originally a fix for exactly that cyclic import error.

MailStorage interface (or similar) is therefore abstract and does not use SQL* directly. Moving the dependency into setStorage and calling it during init or login was sufficient to break the cycle. I would hope the same works for addressbook and calendar.

Jeremy, can you see whether you can use the same method for calendar and addressbook, as done for mail? Make a CalendarStorage and AddressbookStorage interface.

benbucksch:
> Yes, setStorage was originally a fix for exactly that cyclic import error.

FYI, setStorage now also is necessary, because I'm working on a different storage implementation with a noSQL database. For mail only first. The abstract MailStorage and setStorage were also indispensible for that effort, and it was another reason. It was also why I changed SQLEvent.save(this) into this.save() recently. But this only as background info, why I did it this way.

benbucksch (approved):

jermy-c:
> Could you please re-do this PR, in a way that works? For example, you might want to remove `setStorage()` from the constructor and have it called in another place instead, depending on the case.

I'm getting this error whenever `setStorage()` is imported. It doesn't matter where I put it. Unless I put it in the subclasses e.g. `OWAAddressbook`.

benbucksch:
Can you look at the dependency graph and find out why? Where exactly the circle is? This will help to know where to break it best.

jermy-c:
## Imports

Below are the order of files that would most likely have an import cycle

**MailAccount**

1.  main
2.  logic/startup
3.  logic/Mail/SQL/store
4.  logic/Mail/SQL/SQLMailAccount
5.  logic/Mail/AccountsList/MailAccounts
6.  logic/Mail/OWA/OWAAccount
7.  logic/Mail/MailAccount
8.  logic/Mail/Store/setStorage
9.  logic/Mail/SQL/SQLStorage
10. logic/Mail/SQL/SQLMailAccount
11. logic/Mail/AccountsList/MailAccounts
12. Repeats from step 6 based on the protocol of the account

**Calendar**

1.  main
2.  logic/startup
3.  logic/Calendar/SQL/store
4.  logic/Calendar/SQL/SQLCalendar
5.  logic/Calendar/AccountsList/Calendars
6.  logic/Calendar/OWA/OWACalendar
7.  logic/Calendar/Calendar
8.  logic/Calendar/Store/setStorage
9.  logic/Calendar/SQL/SQLStorage
10. logic/Calendar/SQL/SQLCalendar
11. logic/Calendar/AccountsList/Calendars
12. Repeats from step 6 based on the protocol of the account

The problem seems to be around what Neil had mentioned. It seems like `newCalendarForProtocol` is importing `OWACalendar` and `Calendar` but `Calendar` is not defined yet. However, it still seems strange to since `MailAccount` has the same pattern.

Cases that break the import cycle:
1. setStorage for only `Calendar` but in `newCalendarForProtocol` import only `EWSCalendar` and `Calendar`
2. setStorage for both `Calendar` and `Addressbook` but in `newCalendarForProtocol` import only `Calendar` and in `newAddressbookForProtocol` import `EWSAddressbook` and `Addressbook`

In the above cases it has to be `EWS*`, any class for other protocols doesn't work. 

Based on the above cases it seems like it's a problem with the subclasses for the protocols.

benbucksch:
@jermy-c Did you try tools like https://github.com/PSeitz/ts-dependency-graph ?

benbucksch:
> It seems like newCalendarForProtocol is importing OWACalendar and Calendar but Calendar is not defined yet.

If everything is fine, then `Calendar` uses only abstract classes that don't require any SQL or specific protocols. So, as soon as the compiler encounters `Calendar`, it can drag in that file, process it fully, and define `Calendar`, before proceeding with `newCalendarForProtocol()`.

If what you say is true, then `Calendar` requires some specific clases like SQL or a protocol. But I cannot see where that happens. Note that the import can be indirect, e.g. if `Calendar` imports `appGlobal`, and that imports `SQLCalendar`, which imports `Calendar` again, then we're still toast.

benbucksch:
Yeah, it's pretty clear why the circular dependency happens with this PR: You add `setStorage()` to the ctor of the generic `Calendar`. `setStorage()` needs `SQLCalendar`, which needs `Calendar`. That's the cycle. So, we cannot add `setStorage()` to the generic classes.

One way might be to call `setStorage()` in `newCalendarForProtocol()`. Have you tried that?

benbucksch:
Yes, I tried that, and that breaks the cycle.

benbucksch:
See #237 for the implementation.

PR #233, opened 2024-10-11
Activesync: Save sync key later

NeilRashbrook:
Commit 893ad47 is the main part of the PR, it moves processing of received data to a callback which is called before the sync key is saved, so if there is an error processing we can retry.

Commit 520f29b is because I decided that the looping to obtain additional results was awkward. Instead of having a clunky while loop which checks for `MoreAvailable`, I moved the loop into the `makeSyncRequest` functions.

Commit 5a595d3 just fixes a console warning I happened to notice at the time.

Note that you may find it easier to view the commits with whitespace hidden.

### DIFF
--- a/app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
+++ b/app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
@@ -28,9 +28,14 @@ export class ActiveSyncAddressbook extends Addressbook implements ActiveSyncPing
   }
 
   /**
-   * Queues a `Sync` request locally. It waits until the server is available.
-   * @param data 
-   * @param responseFunc 
+   * Queues a `Sync` (synchronization) request locally. It waits until the server is available.
+   * @param data information about the request
+   * @param responseFunc
+   * - It performs actions using the parameter `response.Collections.Collection`.
+   * - It is called when the response is not null.
+   * - It will be repeatedly called while `response.Collections.Collection.MoreAvailable` is 
+   * equal to an empty string.
+   * - It may be called many times.
    */
   async queuedSyncRequest(data: any, responseFunc?: (response: any) => Promise<void>): Promise<any> {
     if (!this.syncState && !this.syncKeyBusy) try {
@@ -54,10 +59,15 @@ export class ActiveSyncAddressbook extends Addressbook implements ActiveSyncPing
   }
 
   /**
-   * Makes a `Sync` request to the server. It is called by `queuedSyncRequest`
-   * and it may be called multiple times.
-   * @param data 
-   * @param responseFunc 
+   * Makes a `Sync` (synchronization) request to the server.
+   * You probably want to call it using the wrapper `queuedSyncRequest()`.
+   * @param data information about the request
+   * @param responseFunc
+   * - It performs actions using the parameter `response.Collections.Collection`.
+   * - It is called when the response is not null.
+   * - It will be repeatedly called while `response.Collections.Collection.MoreAvailable` is 
+   * equal to an empty string.
+   * - It may be called many times.
    */
   protected async makeSyncRequest(data?: any, responseFunc?: (response: any) => Promise<void>): Promise<any> {
     let response;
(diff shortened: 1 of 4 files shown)


app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
```diff
@@ -163,14 +161,18 @@ export class ActiveSyncFolder extends Folder implements ActiveSyncPingable {
           await this.storage.saveMessage(email);
           newMsgs.add(email);
         }
+      } catch (ex) {
+        this.account.errorCallback(ex);
       }
-      for (let item of ensureArray(response.Commands?.Delete)) {
+      for (let item of ensureArray(response.Commands?.Delete)) try {
```
benbucksch, line 167: `for` and `try` should be on separate lines and should each have their own `{}` blocks.

app/logic/Contacts/ActiveSync/ActiveSyncAddressbook.ts
```diff
@@ -41,36 +41,40 @@ export class ActiveSyncAddressbook extends Addressbook implements ActiveSyncPing
...
-      },
-    };
-    let response = await this.account.callEAS("Sync", request);
-    if (!response) {
-      return null;
-    }
-    if (response.Collections.Collection.Status == "3") {
-      // Out of sync.
-      this.syncState = null;
+  protected async makeSyncRequest(data?: any, callback?: (response: any) => Promise<void>): Promise<any> {
```
benbucksch, line 51: rename and doc

app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
```diff
@@ -84,7 +84,7 @@ export class ActiveSyncFolder extends Folder implements ActiveSyncPingable {
    * Sync requests for a given folder must be serialised,
    * because they all use the same per-folder sync key.
    */
-  async queuedSyncRequest(data: any): Promise<any> {
+  async queuedSyncRequest(data: any, callback?: (response: any) => Promise<void>): Promise<any> {
```
benbucksch, line 87: rename and doc

benbucksch (changes requested):

app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
```diff
@@ -98,36 +98,40 @@ export class ActiveSyncFolder extends Folder implements ActiveSyncPingable {
...
-      },
-    };
-    let response = await this.account.callEAS("Sync", request);
-    if (!response) {
-      return null;
-    }
-    if (response.Collections.Collection.Status == "3") {
-      // Out of sync.
-      this.syncState = null;
+  protected async makeSyncRequest(data?: any, callback?: (response: any) => Promise<void>): Promise<any> {
```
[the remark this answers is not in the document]
benbucksch: Please document `responseFunc`: what it does, when it's called, and what is the parameter value.
Important: Also document that it may be called many times, which is very important for the caller and its flow.

benbucksch (changes requested):

app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
```diff
@@ -98,36 +102,46 @@ export class ActiveSyncFolder extends Folder implements ActiveSyncPingable {
...
-    let response = await this.account.callEAS("Sync", request);
-    if (!response) {
-      return null;
-    }
-    if (response.Collections.Collection.Status == "3") {
-      // Out of sync.
-      this.syncState = null;
+  /**
+   * Makes a `Sync` request to the server. It is called by
+   * `queuedSyncRequest()` and it may be called multiple times.
```
benbucksch, line 114: Not `makeSyncRequest` may be called multiple times, but `responseFunc`

benbucksch: same everywhere

benbucksch, line 114:
```suggestion
   * Makes a `Sync` (synchronization) request to the server.
   * You probably want to call it using the wrapper `queuedSyncRequest()`.
```

benbucksch (changes requested):

app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
jermy-c: Sorry my bad, I was reading that wrong.

benbucksch (approved):

Issue #234, opened 2024-10-15
OWA, EWS, ActS: separate save() to DB and saveToServer()


benbucksch:
See https://github.com/mustang-im/mustang/pull/231#issuecomment-2408545123

PR #236, opened 2024-10-15
 OWA, EWS, ActS: separate save() to DB and saveToServer() #234 

jermy-c:
- Separate save() to DB and saveToServer() for OWA, EWS and ActS calendar and addressbook
- Separate delete() to DB and deleteFromServer() for OWA, EWS and ActS calendar and addressbook

### DIFF
--- a/app/logic/Abstract/Group.ts
+++ b/app/logic/Abstract/Group.ts
@@ -8,16 +8,36 @@ export class Group extends ContactBase {
   description = "";
   readonly participants = new SetColl<Person>();
 
+  /**
+   * Saves group locally to the database.
+   */
   async save() {
     await super.save();
     await this.addressbook.storage.saveGroup(this);
   }
 
+  /**
+   * Abstract method to implement saving group to the server.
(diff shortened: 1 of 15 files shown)


benbucksch (approved):
I've removed comments that only state the obvious. Code comments should add new information that's not obvious from the function name and signature.

Otherwise good. Thanks!

NeilRashbrook:
Well, there's only three problems with this:
- The UI only calls `save`, so nothing will ever get saved to the server
- The UI can't call `saveToServer` because it's not implemented for local calendars/addressbooks
- The original bug that I filed #231 for has still not been fixed

benbucksch:
> For comparison, `Folder` uses `deleteIt()` to delete a folder both locally and on the server and separate `deleteItLocally()` and (protected) `deleteItOnServer()` functions. This is why I assumed that `deleteIt` would also delete both locally and on the server for contacts and events.

Yes, I see. That was a reasonable assumption at the time.

PR #238, opened 2024-10-16
Mail: Use AceBase for storage

benbucksch:


benbucksch:
Moving aside for now. Can look at this later again.

PR #240, opened 2024-10-17
IMAP/SMTP: Set OAuth2 when logging in and verifying login

jermy-c:
**Problems**

1. IMAP logs out of OAuth2 in `verifylogin()` thus OAuth2 is `undefined` when verifying for SMTP
2. OAuth2 config is not set in the constructor or login for IMAP login

**Error**

```
SetupMail.svelte:216 Error: Google Mail  SMTP: Need OAuth2 login from IMAP
    at assert (util.ts:4:11)
    at SMTPAccount.getTransportOptions (SMTPAccount.ts:27:7)
    at SMTPAccount.verifyLogin (SMTPAccount.ts:100:61)
    at checkConfig (checkConfig.ts:16:29)
    at async CheckConfig.svelte:19:3
```

I've added the code for setting the config for IMAP but I'm not sure about the fix for SMTP verify login because I understand we need to logout in case SMTP is not successful.

### DIFF
--- a/app/logic/Mail/JSON/JSONMailAccount.ts
+++ b/app/logic/Mail/JSON/JSONMailAccount.ts
@@ -39,7 +39,7 @@ export class JSONMailAccount {
     acc.userRealname = sanitize.label(json.userRealname, appGlobal.me.name);
     acc.name = sanitize.label(json.name, acc.emailAddress);
 
-    acc.fromConfigJSON(json.config ?? {});
+    acc.fromConfigJSON(json.configJSON ?? {});
     acc.workspace = json.workspace
       ? appGlobal.workspaces.find(w => w.id == sanitize.string(json.workspace, null))
       : null;
--- a/app/logic/Mail/JSON/JSONMailAccount.ts
+++ b/app/logic/Mail/JSON/JSONMailAccount.ts
@@ -39,7 +39,7 @@ export class JSONMailAccount {
     acc.userRealname = sanitize.label(json.userRealname, appGlobal.me.name);
     acc.name = sanitize.label(json.name, acc.emailAddress);
 
-    acc.fromConfigJSON(json.configJSON ?? {});
+    acc.fromConfigJSON(json.config ?? {});
     acc.workspace = json.workspace
       ? appGlobal.workspaces.find(w => w.id == sanitize.string(json.workspace, null))
       : null;
--- a/app/logic/Mail/AutoConfig/checkConfig.ts
+++ b/app/logic/Mail/AutoConfig/checkConfig.ts
@@ -15,6 +15,9 @@ export async function checkConfig(config: MailAccount, emailAddress: string, pas
     try {
       await config.outgoing.verifyLogin();
     } catch (ex) {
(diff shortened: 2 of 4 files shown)


benbucksch:
Not all OAuth2 configs are in our hardcoded list. We must not delete nor overwrite an OAuth2 config that is already there, but rather store it in the database.

benbucksch (changes requested):
Can you please find the place where we lose the OAuth2 config, and retain it?

app/logic/Mail/JSON/JSONMailAccount.ts
```diff
@@ -39,7 +39,7 @@ export class JSONMailAccount {
     acc.userRealname = sanitize.label(json.userRealname, appGlobal.me.name);
     acc.name = sanitize.label(json.name, acc.emailAddress);
 
-    acc.fromConfigJSON(json.config ?? {});
+    acc.fromConfigJSON(json.configJSON ?? {});
```
benbucksch, line 42: `config` is correct, within the JSON (makes no sense to have JSON within JSON). Where is `configJSON` being used within JSON specifically?

jermy-c: json.config is null because in SQLMailAccount where the MailAccount is being read the field name is configJSON.

here:
https://github.com/mustang-im/mustang/blob/9e3a022aed801ce17b93e5adbedb1a2fb3d68055/app/logic/Mail/SQL/SQLMailAccount.ts#L86-L87

jermy-c:
d90a6853c2754f5e0882ffc978ebffbe8e547604 fixes the problem for SMTP login. IMAP logs out in verifyLogin() which also clears out the OAuth2 config but we also can't save the config in the database yet until SMTP is verified that the login works.

app/logic/Mail/JSON/JSONMailAccount.ts
jermy-c: I renamed the configJSON to config in SQLMailAccount and JSONMailAccount in 80ead4d1ae5de3beecdbe36466ac75b74d01b558.

benbucksch: * In the SQL database, the field name is `configJSON`. But _only_ there.
* In JSON-based storage, the field name is `config`.
* The functions are called `to/fromConfigJSON()`
* In the classes, each value has its own member.

benbucksch:
> IMAP logs out ... which also clears out the OAuth2 config

Ah, that's a bug, then. Logout should not clear the OAuth config. Instead, it should call `oAuth2.logout()`.

(And yes, if we use OAuth, and we do verify login, then we should not log out before SMTP was tested as well.)

jermy-c:
> Ah, that's a bug, then. Logout should not clear the OAuth config. Instead, it should call `oAuth2.logout()`.

Sorry, my mistake. It doesn't clear out the config just the access and refresh tokens.

benbucksch (approved):

benbucksch:
@jermy-c Thank you! Sorry that I missed this PR.

Issue #210, opened 2024-09-16
Yahoo: Auth window stays open despite login success


benbucksch:
### Reproduction
* Set up a @yahoo.com account in Mustang

### Actual result
* You see "404 Not found"
* The page is https://www.mustang.im/loggedin?code=...
* Mustang window says that the yahoo account is ready to be used -> Mustang recognized that the login succeeded.

### Expected result
* Mustang recognizes the finish URL (that seems to be happening, see above)
* Mustang closes the window (that's missing)
* Setup continues (that happens, see above)

jermy-c:
`window.closed` is true right after `window.close()` which is strange because the window did not close.

https://github.com/electron/electron/issues/37545#issuecomment-1539111486 says:

> The 7th step's check fails on pages of different origins because they wouldn't be ["familiar"](https://html.spec.whatwg.org/multipage/document-sequences.html#familiar-with) with each other

I wonder if it might because of the URL.

benbucksch:
I verified that the App registration at Yahoo has https://www.mustang.im/loggedin , same as our done URL in the code.

jermy-c:
`@aol.com` accounts also have the same issue.

jermy-c:
Just another guess that the problem might be because how Chromium works.

> As of [Chromium 88](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/frame/dom_window.cc;l=347;drc=e1facbdd9b6cfe6f5ec53f8b20e662c934e75045), window.close() succeeds if the new window/tab has an [opener](https://developer.mozilla.org/en-US/docs/Web/API/Window/opener) or if the back/forward stack contains fewer than two entries.

https://textslashplain.com/2021/02/04/window-close-restrictions/

It might be because:

1. it considers yahoo.com and mustang.im two entries because the URLs for other providers are on the same domain

A workaround in the mean time might be to just send an ipcmessage to the main process and close it there. I tested it and it works.

benbucksch:
02.10.2024 23:27:50 Jeremy Chang ***@***.***>:

> window.close() succeeds if the new window/tab has an opener

Do we add and keep an opener? Can you try that?

> A workaround in the mean time might be to just send an ipcmessage to 
> the main process and close it there. I tested it and it works.

Can you make a PR for that, please? Make the code as generic and clean 
and short as possible, please.

You wrote "workaround". What did you have in mind as the other better 
solution? You probably stated it, but I didn't understand, sorry.

[...]

PR #241, opened 2024-10-18
OAuth2Window: Close window from main process #210

jermy-c:
- Sends IPC message to the main process to close the window from there

### DIFF
--- a/app/logic/Auth/OAuth2Window.ts
+++ b/app/logic/Auth/OAuth2Window.ts
@@ -25,13 +25,15 @@ export class OAuth2Window extends OAuth2UI {
             ipcRenderer.removeListener('oauth2-navigate', handleNavigation);
             ipcRenderer.removeListener('oauth2-close', handleClose);
             weClosed = true;
-            popup.close();
+            // Workaround for window.close() not closing in some cases
+            ipcRenderer.send('oauth2-close');
             resolve(this.oAuth2.getAuthCodeFromDoneURL(url));
           }
         } catch (ex) {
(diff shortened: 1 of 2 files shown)


benbucksch:
Why?

jermy-c:
> Why? Can you please describe the specific bug that you are trying to fix here? In which situations does it appear?

This is just a temporary workaround for a window that doesn't close with the `window.close()`. Remember the bug where the OAuth2 window doesn't close for Yahoo and AOL mail? This is a workaround for that in the meantime. It seems like when the redirect url is a different domain then window.close() doesn't work because the new domain is considered a new entry. I couldn't find a way to set the opener otherwise that would be a better solution.

> As of [Chromium 88](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/frame/dom_window.cc;l=347;drc=e1facbdd9b6cfe6f5ec53f8b20e662c934e75045), window.close() succeeds if the new window/tab has an [opener](https://developer.mozilla.org/en-US/docs/Web/API/Window/opener) or if the back/forward stack contains fewer than two entries.

> Why do you think that this is the right fix?

I don't think this is the right fix but it is the only solution I have at the moment.

benbucksch (approved):
@jermy-c What happens when there are 2 windows open at the same time? Which one will it close?

This entire code should be re-written using JPC instead of Electron IPC.

NeilRashbrook:
(The window opening side is my fault since at the time I didn't even know that JPC was running in the IPC main process.)

PR #243, opened 2024-10-21
Actually save events and contacts to the server

NeilRashbrook:
I don't really know how you want this to work, but it's a start.


benbucksch (approved):

benbucksch:
Makes sense

PR #244, opened 2024-10-24
Ensure newly discovered addressbooks and calendars have storage

NeilRashbrook:
Just creating the object isn't enough any more, so I'm using `newXForProtocol` which will set up the storage as well.


benbucksch (approved):
Thanks

PR #245, opened 2024-10-24
Properly check whether OAuth2 account is logged in

NeilRashbrook:
Since there's no `oAuth2` object by default, the account always thinks it's logged in. Switching to checking the `authMethod` fixes this.


benbucksch (approved):

benbucksch:
Thanks, Neil

jermy-c:
When using NTLM the account always shows it's logged even if is not logged in or if the server doesn't support NTLM.

**Reproduction**

1. Setup an EWS hosted by Microsoft with `OAuth2`
2. Go to Settings -> Server
3. Change the Auth method to `NTLM`
4. Go Mail
5. Click [Sync]

PR #248, opened 2024-10-25
Make my tsc find all the logic files again

NeilRashbrook:
For some curious reason, after the removal of the line `import { SQLMailAccount } from "../SQL/SQLMailAccount";` by commit 0056359113c4a0701d4e1c2289cb0f52e7b40008, typescript no longer bothers to check my code, which is unfortunate. I found this addition to be the simplest solution for me.

### DIFF
--- a/app/tsconfig.json
+++ b/app/tsconfig.json
@@ -20,6 +20,7 @@
   "include": [
     "logic/util/Observable.ts",
     "logic/build.ts",
+    "logic/startup.ts",
     "lingui.config.ts",
     "frontend/**/*.d.ts",
     "frontend/**/*.ts",


benbucksch (approved):

NeilRashbrook:
With this change I noticed the following errors which are relatively new:
- In commit 09362ddd1ad53c83b55b177dd177585c7c220afa you didn't fix `AllFolders.listMessages` so currently it doesn't return anything.
- In commit ccd7cdbe80c1336f62aea156c8a7a1bda28d1ab1 you didn't fix `SQLSearchEmail` which still thinks `dbID` is a number.

I had previously tried adding `logic/**/*.ts` which makes TypeScript check more files, generating 55 more errors in total. Some of the errors are in test or unused files and presumably not important, however, I did notice the following errors:
- In PR #201 you didn't update `MailDir.ts` from `email.save();` to `emai.saveCompleteMessage();`.
- In `app/logic/util/Abortable.ts`, you're trying to return an `AbortablePromise<T>` from an async function which isn't allowed (you yourself commented later in `PriorityAbortable.run` that your `cancel` function had gone missing). Async functions always wrap their return in a native Promise, so just removing the `async` will allow `makeAbortable` to return your `AbortablePromise<T>`, although your type casting is still wrong (you need the `as AbortablePromise<T>` on the previous line).

TypeScript also doesn't like `JXON.js`'s use of `toGMTString` which has been deprecated in favour of `toUTCString` for at least 27 years. (ECMAScript 1 even says that it should be the same function object, so `Date.prototype.toGMTString.name` is "toUTCString", although it was a few years before that was fixed in Gecko.)

NeilRashbrook:
Also, `lib/jpc/obj.js` line 20 contains `this._subscribers.remove` but it should be `delete`, not `remove`.

benbucksch:
I'm accepting PRs for that ;-)

PR #249, opened 2024-11-05
Kludge to fix recurrence rule short month bug

NeilRashbrook:
The bug occurs when we fill more occurrences (so they have to be filled twice). This doesn't normally happen because both the front and back end code try to fill a couple of years at once to avoid having to fill more occurrences. However if the last filled occurrence was moved to the last day of a short month then the fill code would erroneously continue from that day.


benbucksch:
Merged #250 instead.

PR #251, opened 2024-11-11
JSONEmail & JSONFolder: Read & write property names don't match

jermy-c:
I changed property names in the `read()` methods for the `JSON*` classes because they are different than the ones in the `save()` methods making them read properties that contain `null` values.

This fixes a bug where all the emails have the date and time of the current time when viewing the emails.

### DIFF
--- a/app/logic/Mail/JSON/JSONEMail.ts
+++ b/app/logic/Mail/JSON/JSONEMail.ts
@@ -77,7 +77,7 @@ export class JSONEMail {
     assert(this.filesDir, "Please call init() first");
     let json: any = [];
     json.filename = a.filename;
-    json.filepath = a.filepathLocal?.replace(this.filesDir + "/", "");
+    json.filepathLocal = a.filepathLocal?.replace(this.filesDir + "/", "");
     json.mimeType = a.mimeType;
     json.size = a.size;
     json.contentID = a.contentID;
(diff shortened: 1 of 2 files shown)


jermy-c:
One question is better to use the property names in the `save()` method or the same names used for columns in the `SQL*` classes?

benbucksch:
> One question is better to use the property names in the `save()` method or the same names used for columns in the `SQL*` classes?

JSON should match the class property names and using JS naming conventions. The SQL column names are adapted to SQL naming conventions and restrictions.

app/logic/Mail/JSON/JSONEMail.ts
```diff
@@ -203,7 +203,7 @@ export class JSONEMail {
       a.mimeType = sanitize.nonemptystring(json.mimeType, "application/octet-stream");
       a.contentID = sanitize.nonemptystring(json.contentID, "" + fallbackID);
       a.filename = sanitize.nonemptystring(json.filename, "attachment-" + fallbackID + "." + fileExtensionForMIMEType(a.mimeType));
-      let filepathLocal = sanitize.string(json.filepathLocal, null)
+      let filepathLocal = sanitize.string(json.filepath, null)
```
benbucksch, line 206: Why not `Local` here?

benbucksch (changes requested):

benbucksch (approved):

app/logic/Mail/JSON/JSONEMail.ts
jermy-c: I just used the property name in the `save()` method. But I see you fixed it thanks!

PR #252, opened 2024-11-12
AceBase: Emails filtered

jermy-c:
All emails are iterated regardless if they belong in the folder. I add a `forEachFiltered()` method to the AceBaseHandle and used that to only iterate through emails that belong in the Folder.


benbucksch (approved):
Nice! Thanks.

PR #254, opened 2024-11-13
JSON: Calendar: Add implementations

jermy-c:
JSON: Calendar: Add implementations
- JSONCalendar
- JSONEvent

### DIFF
--- a/app/logic/Calendar/JSON/JSONEvent.ts
+++ b/app/logic/Calendar/JSON/JSONEvent.ts
@@ -10,13 +10,14 @@ export class JSONEvent {
   static save(event: Event): any {
     assert(event.calendar?.id, "Need calendar ID to save the event");
     let json: any = {};
-    json.pID = event.pID;
+    json.id = event.id;
     json.title = event.title;
     json.descriptionText = event.descriptionText;
     json.descriptionHTML = event.descriptionHTML;
     json.startTime = event.startTime.toISOString();
     json.endTime = event.endTime.toISOString();
     json.calUID = event.calUID;
+    json.pID = event.pID;
     json.calendarID = event.calendar.id;
     json.recurrenceRule = event.recurrenceRule?.getCalString();
     json.recurrenceMasterEventID = event.parentEvent?.pID;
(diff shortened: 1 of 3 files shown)


app/logic/Calendar/JSON/JSONEvent.ts
```diff
@@ -0,0 +1,107 @@
...
+import { appGlobal } from "../../app";
+import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
+import { assert } from "../../util/util";
+import { Collection } from "svelte-collections";
+
+export class JSONEvent {
+  static save(event: Event): any {
+    assert(event.calendar?.id, "Need calendar ID to save the event");
+    let json: any = {};
+    json.pID = event.pID;
```
jermy-c, line 13: What do we use for id here? `Event.id` is readonly so we can't read and set it when reading. But we also can't use pID because then local calendar would not work

benbucksch: You're right. Remove the `readonly` flag in `Event` class.

app/logic/Calendar/JSON/JSONEvent.ts
```diff
@@ -0,0 +1,107 @@
...
+      json.participants.push(this.saveParticipant(personUID));
+    }
+    return json;
+  }
+
+  static saveParticipant(personUID: PersonUID): any {
+    let json: any = {};
+    json.name = personUID.name;
+    json.emailAddress = personUID.emailAddress;
+    json.confirmed = null;
```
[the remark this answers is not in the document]
benbucksch: This is not yet implemented, but we need to be able to save it and read it, so it's needed here.

benbucksch (changes requested):

app/logic/Calendar/JSON/JSONEvent.ts
```diff
@@ -10,13 +10,14 @@ export class JSONEvent {
...
     let json: any = {};
-    json.pID = event.pID;
+    json.id = event.id;
     json.title = event.title;
     json.descriptionText = event.descriptionText;
     json.descriptionHTML = event.descriptionHTML;
     json.startTime = event.startTime.toISOString();
     json.endTime = event.endTime.toISOString();
     json.calUID = event.calUID;
+    json.pID = event.pID;
```
benbucksch, line 20: Good!

benbucksch (approved):

PR #257, opened 2024-11-14
Acebase: Missing awaits

jermy-c:
- The functions are missing awaits.
- We also need to add await to the `forEachFiltered()` function in #252


benbucksch (approved):
Thanks!

PR #260, opened 2024-11-16
AceEmail: Add more query options

jermy-c:
- emails were not ordered correctly by date in descending order
- this adds the option to do that
- it creates an index before doing the query otherwise it would be very slow, because sort is very slow
- it is very slow when there's 16,000 messages


benbucksch (approved):

benbucksch:
Thanks!

benbucksch:
> it is very slow when there's 16,000 messages

Is it faster to sort using the collections classes, using `sortBy()` ?

Try to:
* Replace `EMailCollection` with `ArrayColl` in `Folder`
* Move function `EMailCollection.readFolder()` back into `Folder`, and adapt the caller of `readFolder()`.
* The UI already does `sortBy()`, so you don't need to add it. (`EMailCollection` overrides it.)

PR #262, opened 2024-11-19
JSON: Mail: Tags: Fix reading

jermy-c:
- `emailJSON.tags` is an array of strings, not an array of `{ name: string }` 
- `readTag()` now just reads the param directly instead of access `tag.name`


benbucksch (approved):
Thanks

PR #263, opened 2024-11-19
Fix race condition in makeSyncRequest

NeilRashbrook:
`makeSyncRequest` fails to wait for its callback to complete. In the case of calendar and contacts, this means that it saves the new sync key without checking that the callback was successful. In the case of email, this means that new messages don't get added to the folder correctly.


benbucksch (approved):
outch. Thanks for fixing this.

PR #265, opened 2024-11-20
EWS: Add missing awaits

jermy-c:


benbucksch (approved):

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -360,7 +360,7 @@ export class EWSAccount extends MailAccount {
         m$ConnectionTimeout: 29, // minutes
       },
     };
-    this.callStream(streamRequest, async message => {
+    await this.callStream(streamRequest, async message => {
```
NeilRashbrook, line 363: This was deliberately not an `await`, because it takes 30 minutes to complete.

benbucksch: Ah, thanks for the note. We'll change that to `this.callStream(...).catch(this.errorCallback);`

@NeilRashbrook  Having neither of those is *always* a bug. It leads to uncaught exceptions, and even to crashes. So, every `async` function call (and every promise run) *must* be either with `await` or with `.catch()`.

NeilRashbrook: `callStream` already has its own `try/catch`.

benbucksch: Fixed in git commit 9cd12445

benbucksch: @NeilRashbrook Thanks for alerting about this. Indeed, `await` for 29 minutes would have been bad here.

Issue #268, opened 2024-11-21
Invitation: Acceptance email is parsed as event - causes UI freeze


benbucksch:
### Reproduction
* Check out branch `master` or `neil/invitation-wip`
* Set up neil@beo email account, as EWS
* Open folder `Inbox/testfolder` (no number)
* Open email "Accepted: New Meeting"

### Actual result
* UI freezes
* Error on console:
```
DisplayEvent.svelte:118 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getDay')
    at instance (DisplayEvent.svelte:118:69)
    at init (index.mjs:2165:11)
    at new DisplayEvent (DisplayEvent.svelte:111:23)
```
* The `message.event` object has almost all properties as undefined (that's the bug), and only a `descriptionText` = "Test User <test.temp@tiamo.pro> has accepted your event invitation.\r\n"

### Expected result
* The `message.event` object is either a complete `Event` with all properties, or is empty. There is never an `Event` with no start time etc.
* No UI freeze
* What to display for the email: Not sure. If in doubt, show the normal email body.

NeilRashbrook:
Turns out only ActiveSync provides calendar data for meeting responses as well as requests. (And even then it doesn't actually provide the attendee data, you're supposed to infer that from the message class.) So maybe I should just remove `loadEvent` from all of the subclasses and use the generic code which works for all iTIP messages...

benbucksch:
@NeilRashbrook The problem is that an invitation response isn't an invitation nor an event. It's 2 inferences away: event -> invitation to event -> response to invitation to event.

It should be different functions responsible for handling them, with maybe a shared function called for part of the implementation.

benbucksch:
And please don't use the Event class to represent a response to an invitation, but rather the response should reference the event and the invitee (email address).

benbucksch:
> And please don't use the Event class to represent a response to an invitation, but rather the response should reference the event and the invitation (or invitee email address).

NeilRashbrook:
You're going to run into the same "issue" when sending invitations and responses - they are still all events really.

benbucksch:
I've commited a hotfix in git commit ffbbc67b which stops the UI freeze.

However, the event object is still invalid. All instances of `Event` need to have at least the start time, and a few other important properties.

benbucksch:
@NeilRashbrook Either we will the event object with the correct data, or we remove it.

NeilRashbrook:
[...]

PR #270, opened 2024-11-22
Calendar: Don't try to load iTIP data through Exchange APIs #268

NeilRashbrook:
In the case of EWS and OWA, this only works for invitations.
In the case of ActiveSync, this doesn't provide complete data.

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncEMail.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncEMail.ts
@@ -192,6 +192,7 @@ export class ActiveSyncEMail extends EMail {
     await this.deleteMessageLocally(); // Exchange deletes the message from the inbox
   }
 
+  /* Disabling because ActiveSync does not provide complete event data.
   async loadEvent() {
     assert(this.scheduling, "This is not an invitation or response");
     assert(!this.event, "Event has already been loaded");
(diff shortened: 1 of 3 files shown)


benbucksch:
In this PR, you are simply deleting the entire event fetch implementation, for all of OWA, EWS, and ActiveSync. Are you saying that they are completely useless?

@NeilRashbrook Can you please explain more about the why?

> In the case of EWS and OWA, this only works for invitations.

What other cases other than invitations exist here?

> In the case of ActiveSync, this doesn't provide complete data.

Which data is missing?

app/logic/Mail/ActiveSync/ActiveSyncEMail.ts
```diff
@@ -191,34 +191,6 @@ export class ActiveSyncEMail extends EMail {
     await super.sendInvitationResponse(response); // needs 16.x to do this automatically
     await this.deleteMessageLocally(); // Exchange deletes the message from the inbox
   }
-
-  async loadEvent() {
```
benbucksch, line 195: Are you sure that they will never ever be needed anymore, for nothing?

I'd rather keep the code, disabled, by renaming the function to `loadEvent_disabled()`, and adding a comment why we don't use it, and that this will now implicitly use the implementation from `EMail.loadEvent()`, which uses iTIP.

benbucksch (changes requested):

NeilRashbrook:
> > In the case of EWS and OWA, this only works for invitations.
> 
> What other cases other than invitations exist here?

See #268.

> > In the case of ActiveSync, this doesn't provide complete data.
> 
> Which data is missing?

Attendees are definitely missing.

benbucksch (approved):

PR #286, opened 2024-11-30
Mail: SQL: Use transactions without async

jermy-c:
- journal_mode is set to WAL
- use `executeTransaction()` provided by `@radically-straightforward/sqlite`
- We have to save all persons before saving the email because when we do it while in the transaction the insert doesn't appear in the database until the entire transaction is completed so our query for selecting the personID is always null the first time
- setting `UNIQUE("messageID", "pID", "folderID")` just a little bit faster

### DIFF
--- a/app/logic/Mail/SQL/SQLEMail.ts
+++ b/app/logic/Mail/SQL/SQLEMail.ts
@@ -223,19 +223,47 @@ export class SQLEMail {
   static async saveMultiple(emails: Collection<EMail>) {
     let lock = await this.transactionLock.lock();
     try {
-      await (await getDatabase()).run(sql`BEGIN TRANSACTION`);
-      for (let email of emails) {
-        if (!email.subject) {
-          continue;
+      await (await getDatabase()).executeTransaction(() => {
+        for (let email of emails) {
+          if (!email.subject) {
+            continue;
+          }
+          this.savePerson(email.from);
+          for (let p of email.to) {
+            this.savePerson(p);
+          }
+          for (let cc of email.cc) {
+            this.savePerson(cc);
+          }
+          for (let bcc of email.bcc) {
+            this.savePerson(bcc);
+          }
+          if (email.replyTo) {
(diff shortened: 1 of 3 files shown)


benbucksch:
Can you separate the changes (which can be used independently fromeqch other) in separate PRs, please? See "atomic commits".

benbucksch:
There are still "await"s within the `executeTransaction()`, therefore it won't work.
All the backend calls to the DB are async.

app/logic/Mail/SQL/SQLEMail.ts
```diff
@@ -223,19 +223,89 @@ export class SQLEMail {
...
+      let db = await getDatabase();
+      await db.executeTransaction(() => {
+        for (let email of emails) {
+          if (!email.subject) {
+            continue;
+          }
+          this.saveMainProperties(db, email);
+          this.savePersons(db, email);
         }
-        await this.save(email);
```
benbucksch, line 231: This saves not only the email and persons, but also attachments and tags etc., which are in different tables.

app/logic/Mail/SQL/SQLEMail.ts
```diff
@@ -223,19 +223,89 @@ export class SQLEMail {
...
-      await (await getDatabase()).run(sql`END TRANSACTION`);
+      });
     } finally {
       lock.release();
     }
   }
 
+  protected static saveMainProperties(db: Database, email: EMail) {
+    let contact = email.contact as PersonUID;
+    db.run(sql`
```
benbucksch, line 245: This is an async function that you call without `await`. *All* calls to the database are always going to the backend, and the communication between frontend and backend is happening over WebSockets and therefore always async. There's no way to make sync calls to the DB from the frontend.

benbucksch (changes requested):

benbucksch:
I fear that there is no way to make this work sync, given our architecture.

benbucksch:
Sorry! :-(

PR #287, opened 2024-12-02
Mail: SQL: Pragma settings for better performance

jermy-c:


benbucksch:
@jermy-c Could you please describe exactly why this change should be made? What tests did you do? How did you measure? What were your results?

I tested WAL mode, and the write speed (when fetching 5-10 folders at once) was identical to before.

jermy-c:
- `synchronous = NORMAL` makes the app startup faster
- `journal_mode = WAL` makes transactions faster but transactions needs to implemented properly

I'll test and measure this later.

benbucksch:
> I'll test and measure this later.

I'm still waiting for the numbers. In my testing, this didn't increate speed at all.

jermy-c:
> I'm still waiting for the numbers. In my testing, this didn't increate speed at all.

Sorry, about that. I didn't see any increased speed either.

Issue #278, opened 2024-11-26
OWA: Only 50 emails are downloaded per folder


jermy-c:
**Reproduction**

1. Login to OWA
2. Click folder properties on a folder
3. Switch between folders

**Actual Result**

1. Each folder only has a maximum of 50 emails
2. Even if you click download it still doesn't download the other emails

**Expected Result**

1. Downloads all emails

benbucksch:
I can confirm that, with neil@ account set up as OWA in a fresh profile, and folders "Deleted Items", "Sent Items" and others.

### Actual result
* After clicking on the folder, I see onljy 50 messages
* In the folder properties (hover over folder name, then click "..."), I see that the folder has over 4000 msgs, 50 local messages
* Clicking [Download all messages] has no effect

### Expected result
* The headers of all messages show up
* Messages are downloaded in the background, in batches of 50.

### Cause
* Apparently, the batching code is stopping and not looping anymore.
* I also tried with v0.6.14, before my changed to the locking code, and the bug is still there.
* Could it be that it deadlocks itself, in the OWA locking/queueing code? Just a guess.

benbucksch:
When I do the same with an EWS account, and I can see in folder properties how the "local messages" are counting up in steps of 50. Not with OWA.

PR #288, opened 2024-12-02
Fix usage of boolean types in OWA code #278

NeilRashbrook:
Unlike ActiveSync and EWS, where data is transferred using XML, and therefore is stringly typed, OWA uses JSON, and properties actually have useful types. Unfortunately when porting EWS code I didn't always remember this. As well as the case of #278 where only 50 messages can be downloaded, I found a second instance in calendar code, which would prevent it from reading the alarm status of any event.


benbucksch (approved):

benbucksch:
@NeilRashbrook Use `sanitize.boolean()`

Issue #271, opened 2024-11-23
Calendar: Day/Month view: Should start at 1. 


Unwrap-at:
### Reproduction
* Open Calendar, Day/Month view
* On the month view, change the month, using the < > arrows

### Actual result
* The top row starts with the 18. of the month

### Expected result
* The top row has the 1. of the month
* Depending on which week day is the 1. of the month, we may see the rest of the previous month. E.g. if the 1.11. is a Thursday, the first row starts with the 29.10.
* Optional: The vertical month name does not show at the top left, if it's not the 1. I.e. if we show November, the top left should not show "October". The top left is a special case, only in this case. All other days show the month, as before (at the beginning of the line, and on the 1.)

PR #290, opened 2024-12-02
Improve month view navigtation #271

NeilRashbrook:
- Automatically rewind the month view so that the first day of the month appears in the first week
- Advance the month view by 42 days at a time, otherwise months such as March 2025 don't advance properly
- Rewind the month view by 28 days at a time, otherwise months such as February 2025 don't rewind properly
- Use the middle of the view date range to generate the month label, which is the month the user is looking at


benbucksch:
Doesn't always work, e.g. April 2025, by paging from today.

grafik

All following months are then also off by 1 week. In fact, the problem accumulates, e.g. July 2025 is 2 weeks off:

grafik

NeilRashbrook:
This is Day/Month view, I only fixed Month view.

benbucksch (approved):
Quite hacky. Would be nice to make this less co-dependent between different components.

But improves status quo, so merging.

PR #292, opened 2024-12-03
Setup: Check outgoing acc.id also

jermy-c:
- I think it's reusing the same `acc.id` becuase it doesn't check the `acc.id` of outgoing accounts because it doesn't get added by `getAllAccounts()`

### DIFF
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -116,6 +116,7 @@ function findFreeAccountID(): string {
 function getAllAccounts(): Collection<Account> {
   let allAccounts = new ArrayColl<Account>();
   allAccounts.addAll(appGlobal.emailAccounts);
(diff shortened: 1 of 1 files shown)


app/logic/Abstract/Account.ts
```diff
@@ -116,6 +116,7 @@ function findFreeAccountID(): string {
 function getAllAccounts(): Collection<Account> {
   let allAccounts = new ArrayColl<Account>();
   allAccounts.addAll(appGlobal.emailAccounts);
+  allAccounts.addAll(appGlobal.emailAccounts.map(acc => acc.outgoing));
```
NeilRashbrook, line 119: Note that not all email accounts have outgoing accounts associated with them.

jermy-c: Thanks for the heads up.

benbucksch (approved):
I've reverted the statements, because it's simpler to read for me.
Otherwise good.
Tested and works for me.

PR #300, opened 2024-12-11
Build: Mac Universal build with ARM and x86 binaries

jermy-c:
- Our build for mac was not set to universal
- `target: default` is mandatory
- This was the reason for the slow startup on MacOS
- https://github.com/mustang-im/mustang/actions/runs/12248650829/job/34168650435#step:11:568


benbucksch (commented):
Thanks for fixing this

PR #316, opened 2024-12-16
Translations

jermy-c:


app/l10n/locales/az/messages.json
```diff
@@ -23,7 +23,7 @@
   "J6ypuD": "AXTAR",
   "AiAMXv": "Hesabınız üçün konfiqurasiya {0} mövcud ola bilər:\\n\\n{1}\\n\\n Oxunub:?",
   "uyJsf6": "Haqqında",
-  "uQiTch": "Qeydiyyat",
+  "uQiTch": "Accept call",
```
benbucksch, line 26: Many strings here change from translated back to English. Why?

app/l10n/locales/az/messages.json
```diff
@@ -483,7 +483,7 @@
   "cQxjXv": "Qalereya",
   "vPemUP": "Kateqoriya",
   "G6669w": "{0}",
-  "FIkhWn": "HAQQIMIZDA",
+  "FIkhWn": "KB",
```
benbucksch, line 486: Does that mean Kilobyte? This looks like it should be pluralform in the original.

benbucksch (changes requested):
Most of the changes here in this PR look wrong. They appear to change translated strings back to English.

app/l10n/locales/az/messages.json
jermy-c: Because the same translation was used for multiple strings, it was an issue that we didn't notice before.

app/l10n/locales/az/messages.json
jermy-c: The source string is "KB"

app/l10n/locales/ko/messages.json
```diff
@@ -47,12 +47,12 @@
   "1kLn6M": "1일 전",
   "+bHmdA": "모든 다운로드.",
   "dtwyLS": "모든 메시지",
-  "4xBHUg": "이름 *",
+  "4xBHUg": "Answer",
```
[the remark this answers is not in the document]
jermy-c: This is the issue where the same translation was used for multiple translations.

benbucksch:
@jermy-c wrote in chat:
Many of the strings were wrong or used the same string for many translation, that’s why they were reverted to English.

Issue #317, opened 2024-12-17
Mac: Starting another backend


jermy-c:
**Reproduction**

1. Open Mustang
2. Close Mustang, don't quit the app
3. Open Mustang again

**Actual Result**

1. Displays an error dialog
2. Click "OK"
3. The app starts again

**Expected Result**

1. Displays the app

<img width="372" alt="圖片" src="https://github.com/user-attachments/assets/403659ec-d29d-4357-b215-f9bbe54da594" />

NeilRashbrook:
I created commit 47e4860 which I think might address this to some extent but I'd have to dig out my ancient Mac and try and install Node on it to fully test it so I don't suppose you could try it out?

benbucksch:
@jermy-c Could you please try this proposed fix?

benbucksch:
Branch `neil/restart-backend`

This is stopping the backend when the main window closes, and restarts it when the window opens. I reviewed the code, and it looks right to me - if we follow this logic.

However, that fix means that once the main window is closed, we will not do any mail check anymore (which we currently don't do anyways), and we will not be able to do that anymore.

Additionally, if we were to open a second main window, and close it, the backend would die and leave the first main window stranded. Again, that's not a case that we have right now. But it's something that might potentially come up in the future.

benbucksch:
@NeilRashbrook Can you make that into a PR? The code looks good to me and is a step forward.

jermy-c:
It works I don't see the dialog but IMAP has a problem after closing and opening 2-3 times without quitting.

```
index-uad3MjUS.js:37969 LoginError: Check your login, username, and password.
Too many simultaneous connections. (Failure)
    at IMAPAccount.connection (index-uad3MjUS.js:66563:35)
    at async IMAPAccount.login (index-uad3MjUS.js:66496:5)
    at async loadFolder (index-uad3MjUS.js:386869:9)
```

benbucksch:
a) We need to stop IMAP connections and all other persistent connections. This should likely happen from the frontend in an unload handler.
b) We need to drop all objects in the backend that are no longer needed. Specifically, the callbacks needs to be removed. These are the reason for bug #61.

In other words, we need to shut down from 2 sides: The frontend needs to power down everything that it can think of, and the backend needs to drop everything that remains, because we're likely to have missed something, now or in the far future, and we don't want to leak it.

benbucksch:
[...]

PR #322, opened 2024-12-17
Allow the backend to be restarted on the Mac #317

NeilRashbrook:
IMAP connections won't be closed automatically though, so either they need to be tracked and shut down by the backend or the front end should use an unload handler to close any persistent connections.


benbucksch:
> IMAP connections won't be closed automatically though, so either they need to be tracked and shut down by the backend or the front end should use an unload handler to close any persistent connections.

Can you please file a GitHub issue about that?

benbucksch (approved):

Issue #191, opened 2024-09-02
Msg viewer: Links don't open the system browser


Unwrap-at:
**Reproduction steps:**
Click on a message 
Click on a link

**Expected result:**
The browser should show me the website.

**Actual result:**
When I click on the link, the link becomes red and nothing happens.

benbucksch:
* This depends on your "default browser" settings in your OS. What do you have selected there?
* Also, is that browser already open/running when you click on the link?
* Does the browser open a tab with the page of the link?

Unwrap-at:
mm
1. not sure
2. yes it is
3. no, that's the problem, nothing happens when you click

interesting fact though, this problem doesn't exist in the tread view.
In thread view when I click a links it opens on the browser in a different window.

benbucksch:
The "thread view" (left-most icon, in the view switcher at the top left of the email content) is not in an iframe, whereas the normal msg view is in a `<WebView>`.

benbucksch:
On Linux, links in the `<WebView>` open for me (sometimes, depending on system situation).

benbucksch:
See e2/src/main/index.ts line 40 setWindowOpenHandler()

josiaspb:
setWindowOpenHandler() do not see webview events. When a user clicks a link inside the webview, the main process do not notice the event and do not open the link.

When a user clicks a webview link, I propose to two options: 
-Send a message from the render process to the main process with the URL to open by shell.openExternal(). 
-Open the link from the webview.

benbucksch:
> 
> setWindowOpenHandler() do not see webview events. When a user clicks a link inside the webview, the main process do not notice the event and do not open the link.

This was originally the same for the app UI. As you can see from the comment in the code, we had to add target==_blank.


> 
> 
> When a user clicks a webview link, I propose to two options:
> -Send a message from the render process to the main process with the URL to open by shell.openExternal().
> -Open the link from the webview.

Go ahead and make a patch (or multiple alternatives).
-- 
Sent from my phone. Please excuse the brevity.

benbucksch:
@josiaspb Could you please post progress and findings about this bug here in the github issue?

* In app/frontend/Shared/WebView.svelte, function `haveWebView()`, add an event listener for `will-navigate`. In the event listener, do an `alert("changing page");` or `console.log("changing page");` (the latter would appear in the console of the frontend, not the webview)
[...]

PR #328, opened 2024-12-27
Open links in system browser (#191)

jermy-c:
- Opens links in system browser
- Why did I do it this way? `did-start-navigation`,  `will-frame-navigate`, `will-navigate`, `did-frame-navigate`, `did-navigate` events don't trigger nor does `did-create-window`
- The other way might be to use `<webview>.executeJavaScript(code[, userGesture])` and have it send the URL to the main window.

### DIFF
--- a/app/frontend/Shared/WebView.svelte
+++ b/app/frontend/Shared/WebView.svelte
@@ -123,7 +123,9 @@
       url = eventURL;
     });
     appGlobal.remoteApp.addEventListenerWebContents(id, "input-event", (event) => {
-      if (event.type == "mouseDown" && url) {
+      let modifiers = event.modifiers.map(m => m.toLowerCase());
+      let isLeft = ["left", "leftbuttondown"].some(l => modifiers.some(m => m == l));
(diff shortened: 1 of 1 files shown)


jermy-c:
At the moment it opens the link in the system browser without any popup or dialog asking the user whether would like to open the link externally.

benbucksch:
How do you know that the use actually clicked on the link? And that he 
didn't move the mouse somewhere else after the first event?

jermy-c:
Because:

- `url` would be updated to an empty string as soon as the cursor leaves the link
- if the `url` is an empty string then the click event wouldn't open the link

jermy-c:
It opens link only on left click now but it doesn't open links when the it the link is selected/highlighted and left clicked.

benbucksch:
You mean the text selection includes the URL? In this case, it's fine 
that the click doesn't open the URL. I'd call that a feature, because the 
user tried to select and copy the text, not open the URL.

benbucksch:
Thank you!

Issue #327, opened 2024-12-26
ActiveSync: Throttling


benbucksch:
Caused #281. Similar to #121 and #153.

Throttle:
* Number of concurrent calls
* Number of calls per second

benbucksch:
git commit 2806d714 and 874d67e1

PR #329, opened 2024-12-27
ActiveSync: Throttle connections #327

NeilRashbrook:
Port of PR #159.

I defaulted `maxConcurrency` to 20 as that's the figure we use for EWS and OWA, although according to https://learn.microsoft.com/en-us/powershell/module/exchange/set-throttlingpolicy?view=exchange-ps the default is 5 for OWA and 10 for EWS and ActiveSync...

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -12,9 +12,9 @@ import { OAuth2URLs } from "../../Auth/OAuth2URLs";
 import { request2WBXML, WBXML2JSON } from "./WBXML";
 import { ConnectError, LoginError } from "../../Abstract/Account";
 import { appGlobal } from "../../app";
-import { ensureArray, assert, NotSupported } from "../../util/util";
+import { ensureArray, assert, NotSupported, sleep, throttleConnectionsPerSecond } from "../../util/util";
 import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
-import { ArrayColl } from "svelte-collections";
+import { ArrayColl, SetColl } from "svelte-collections";
 import { gt } from "../../../l10n/l10n";
 
 const kFolderSyncKeyError = "9";
@@ -36,6 +36,9 @@ export class ActiveSyncAccount extends MailAccount {
   heartbeat = 60;
   maxPings = kMaxCount;
   listening = false;
+  maxConcurrency: number = 20;
+  connections = new SetColl<Promise<any>>;
+  nextConnectionTime = new Array(50).fill(0);
   policyKey: Promise<string> | string;
   syncKeyBusy: Promise<any> | null;
   protocolVersion: string;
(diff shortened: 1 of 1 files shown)


app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -245,6 +263,10 @@ export class ActiveSyncAccount extends MailAccount {
       if (!wbxmljs.Status || wbxmljs.Status == "1" || aCommand == "Ping") {
         return wbxmljs;
       }
+      if (wbxmljs.Status == "111") {
```
benbucksch, line 266: This is "Retryable server error", right?
1. Did you see exactly this error, in reality, when we're running into throttling on the server? Are there any other error calls that might appear in this case?
2. In case this status 111 happens for reasons other than throttling, and the server keeps throwing this error for 24 hours. What would prevent us from infinitely re-trying? We need some sort of counter that limits the number of times we try. You could express that as `RetryCounter` class as well.
3. We need to avoid that other calls keep hitting the server at the same time, and run into the same error as well.

The following would address 3, but it's still missing the `RetryCounter`:
```
      if (wbxmljs.Status == "111") { // Retryable server error
        let lock = await this.semaphore.lock();
        await sleep(5);
        lock.release();
        return await this.callEAS(aCommand, aRequest, heartbeat);
      }
```

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -222,7 +233,14 @@ export class ActiveSyncAccount extends MailAccount {
...
     let wbxml = await request2WBXML({ [aCommand]: aRequest });
-    let response = await appGlobal.remoteApp.postHTTP(String(url), wbxml, "arrayBuffer", options);
+    let connection = appGlobal.remoteApp.postHTTP(String(url), wbxml, "arrayBuffer", options);
+    this.connections.add(connection);
+    let response;
+    try {
+      response = await connection;
+    } finally {
+      this.connections.remove(connection);
+    }
```
benbucksch, line 243: This should be in a class `Semaphore`. I've implemented that.
@NeilRashbrook Could you please review my new code on master?

benbucksch (changes requested):

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
benbucksch: Or better yet, the following would address 3, but it's still missing the `RetryCounter`:
```
      if (wbxmljs.Status == "111") { // Retryable server error
        await this.throttle.waitFor(5);
        return await this.callEAS(aCommand, aRequest, heartbeat);
      }
```

benbucksch: Implemented in git commit 874d67e1

benbucksch:
Re-implemented using classes `Semaphore` and `Throttle` in git commit 2806d714.
Handle of throttle error 111 is implemented in git commit 874d67e1.

benbucksch:
> In case this status 111 happens for reasons other than throttling, and the server keeps throwing this error for 24 hours. What would prevent us from infinitely re-trying? We need some sort of counter that limits the number of times we try. You could express that as RetryCounter class as well.

@NeilRashbrook Please look out for such error cases in the future.

> Did you see exactly this error, in reality, when we're running into throttling on the server? Are there any other error calls that might appear in this case?

@NeilRashbrook This answer would still be helpful, depsite the commits.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: > 1. Did you see exactly this error, in reality, when we're running into throttling on the server?

No, I only added the logic because you'd said you'd seen that error. (I didn't think to click on the all folders account myself even though I know that generates a silly amount of parallel requests.)

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: The code is fine, but the comment looks as if it was copied and pasted...

PR #334, opened 2024-12-30
Context menu roles for predefined actions

jermy-c:
- Menu Item constructor takes a string for roles or a function for custom actions
- Copy works now
- Cut and paste actions are implemented but not tested yet


benbucksch (approved):

jermy-c:
I just realized that this change means SvelteUI menu would break since it calls the `menu.action` and it doesn't have access to the predefined actions provided by electron.

PR #335, opened 2024-12-30
Allow JPC to work with Stanza

NeilRashbrook:
There were some issues here:
- Stanza's client object defines an own `subscribe` function which means that we don't send any properties because we assume observing it will achieve this.
- Stanza's client object contains recursive references to itself.

Thus I have created three commits:
- 8c0729a adds support for objects with recursive references, although I'm not entirely sure this works, I should write a test for it and then let you know.
- 6e445e4 requires the `subscribe` function to be on a prototype rather than an own property, although I'm not sure how many Svelte subscribers we actually send over JPC.
- c91bece actually adds support for own method properties on objects.

### DIFF
--- a/lib/jpc/obj.js
+++ b/lib/jpc/obj.js
@@ -173,8 +173,10 @@ export default class BaseProtocol {
       proto = await proto;
     }
     let stub = Object.create(proto);
+    this._recursiveObjects.set(objDescrJSON.idSender, stub);
     stub._jpc_id = objDescrJSON.idSender;
     await this.updateObjectProperties(stub, objDescrJSON.properties);
+    this._recursiveObjects.delete(objDescrJSON.idSender);
     // Timing issue: This needs to happen after the caller sets the
     // Promise it gets from calling us into the remote object map.
     this.addRemoteObject(objDescrJSON.idSender, stub);
@@ -291,7 +293,7 @@ export default class BaseProtocol {
         return new Uint8Array(obj.uint8array);
       }
       if (obj.idSender) {
-        let stub = this.getRemoteObject(obj.idSender);
+        let stub = this._recursiveObjects.get(obj.idSender) || this.getRemoteObject(obj.idSender);
         if (stub) {
           if (stub instanceof Promise) {
             stub = await stub;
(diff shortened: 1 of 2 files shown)


benbucksch:
Thank you. I merged the commit 2 and 3 (with a small change) to master, individually.

benbucksch:
Regarding commit 1:
`getOrCreateIDForLocalObject()` already has an `id = this._localObjectsToIDs.get(obj)`. Am I missing something, or should _localObjectsToIDs already give a hit in all cases where `_recursiveObjects` would give a hit?

(The only difference I see is that `_recursiveObjects` is deleted earlier, but if `_recursiveObjects` doesn't get a hit, then `_localObjectsToIDs` nonetheless would return the cached ID anyways.)

benbucksch:
I tested the change, with real XMPP logins, and I am getting a timeout on `getRoster()` now. I am not sure where it's timing out, whether in JPC or in Stanza XMPP, but I suspect in JPC.

NeilRashbrook:
> Am I missing something, or should _localObjectsToIDs already give a hit in all cases where `_recursiveObjects` would give a hit?

This was what I was thinking about in the commit comment, but I'm not sure it's safe, as the client could garbage collect its stub, and it would need to be recreated if the server needed to send the object again.

NeilRashbrook:
I've added code to break the recursive object deadlock to this PR. (Because everything else has already landed, those commits have been rebased away.)

benbucksch (approved):

benbucksch:
@NeilRashbrook This is still not working for me. The `getRoster()` call is timing out.

benbucksch:
@NeilRashbrook Code review: I don't like the code impact on the normal code path, which was previously only:
```
return this.createObjectDescription(obj, this.getOrCreateIDForLocalObjec
t(obj));
```

Could the `_recursiveObjects.get()/set()` be integrated into the `getOrCreateIDForLocalObject()` function?

NeilRashbrook:
> Could the `_recursiveObjects.get()/set()` be integrated into the `getOrCreateIDForLocalObject()` function?

I on;y put it in `mapOutgoingObjects` because I was editing code there already.

NeilRashbrook:
> This is still not working for me. The `getRoster()` call is timing out.

Commit 239fef9 is not on the `stanza` branch. If I cherry-pick it, the test works for me.

Issue #84, opened 2024-06-14
EWS: NTLM authentication


benbucksch:
Check how difficult it is. If easy (< 4 hours), please implement it.

At the very least, don't show "Password incorrect", but "NTLM not supported".

benbucksch:
Add support for NTLM. First for EWS, then for all other applicable protocols.

* Must work with on-prem Exchange.
* Must work by passing the password, on any operating system (pure HTTP)
* Ideally, should alternatively work by using the OS authentication on Windows 11, without re-entering the password. No idea how that works.

benbucksch:
@NeilRashbrook Could you please implement this?

NeilRashbrook:
Sadly `@ewsjs/ntlm-client` doesn't actually work, despite parsing the WWW-Authenticate header properly, so `httpntlm` (which doesn't, so I have to parse it manually) it is.

benbucksch:
`httpntlm` has a number of unnecessary and unwanted dependencies. Can you extract only the parts that we need and make a new lib from that?

PR #337, opened 2025-01-01
EWS: NTLM authentication #84

NeilRashbrook:
`ntlm.js` is not quite the version from `httpntlm`, since I've changed the API to take separate parameters instead of objects, I've made `workstation` default to `hostname()`, I've parsed the `WWWAuthenticate` header more reliably, I've tweaked the creation of a buffer of 8 random bytes, I've converted it from CommonJS to ESM, and I've also reformatted it because it looked awful.


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -283,16 +293,25 @@ export class EWSAccount extends MailAccount {
...
+        let authorization;
+        if (this.NTLMauthorization) {
+          let response = await appGlobal.remoteApp.streamHTTP(this.url, requestXML, this.createRequestOptions());
+          if (response.status != 401) {
+            console.error("Unexpected NTLM negotiation failure in callStream");
+            return;
+          }
+          authorization = await appGlobal.remoteApp.createType3MessageFromType2Message(response.WWWAuthenticate, this.username, this.password);
+        }
+        let response = await appGlobal.remoteApp.streamHTTP(this.url, requestXML, this.createRequestOptions(authorization));
```
NeilRashbrook, line 312:
```suggestion
        let response = await appGlobal.remoteApp.streamHTTP(this.url, requestXML, this.createRequestOptions());
        if (this.NTLMauthorization) {
          if (response.status != 401) {
            console.error("Unexpected NTLM negotiation failure in callStream");
            return;
          }
          let authorization = await appGlobal.remoteApp.createType3MessageFromType2Message(response.WWWAuthenticate, this.username, this.password);
          response = await appGlobal.remoteApp.streamHTTP(this.url, requestXML, this.createRequestOptions(authorization));
        }
```

backend/ntlm.js
```diff
@@ -0,0 +1,463 @@
...
+  password = password.toUpperCase();
+  var passwordBytes = Buffer.from(password, "ascii");
+
+  var passwordBytesPadded = Buffer.alloc(14);
+  passwordBytesPadded.fill("\0");
+  var sourceEnd = 14;
+  if (passwordBytes.length < 14) {
+    sourceEnd = passwordBytes.length;
+  }
+  passwordBytes.copy(passwordBytesPadded, 0, 0, sourceEnd);
```
NeilRashbrook, line 275:
```suggestion
  passwordBytes.copy(passwordBytesPadded);
```

backend/ntlm.js
```diff
@@ -0,0 +1,463 @@
...
+
+  // split into 2 parts of 7 bytes:
+  var firstPart = passwordBytesPadded.slice(0, 7);
+  var secondPart = passwordBytesPadded.slice(7);
+
+  function encrypt(buf){
```
NeilRashbrook, line 281:
```suggestion
  function encrypt(buf) {
```

backend/ntlm.js
```diff
@@ -0,0 +1,463 @@
...
+  var md4 = createHashMD4();
+  md4.update(buf);
+  return Buffer.from(md4.digest());
+}
+
+function calc_resp(password_hash, server_challenge) {
+  // padding with zeros to make the hash 21 bytes long
+  var passHashPadded = Buffer.alloc(21);
+  passHashPadded.fill("\0");
+  password_hash.copy(passHashPadded, 0, 0, password_hash.length);
```
NeilRashbrook, line 387:
```suggestion
  password_hash.copy(passHashPadded);
```

backend/ntlm.js
```diff
@@ -0,0 +1,463 @@
...
+  var hmac = createHmac('md5', key);
+  hmac.update(data);
+  return hmac.digest();
+}
+
+function ntlm2sr_calc_resp(responseKeyNT, serverChallenge, clientChallenge) {
+  // padding with zeros to make the hash 16 bytes longer
+  var lmChallengeResponse = Buffer.alloc(clientChallenge.length + 16);
+  lmChallengeResponse.fill("\0");
+  clientChallenge.copy(lmChallengeResponse, 0, 0, clientChallenge.length);
```
NeilRashbrook, line 413:
```suggestion
  clientChallenge.copy(lmChallengeResponse);
```

backend/ntlm.js
benbucksch: @NeilRashbrook I've commited those changes. But...
Why do you think it's unnecessary to fill the buffer with 0s?

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -269,9 +274,14 @@ export class EWSAccount extends MailAccount {
         await this.oAuth2.reset();
         await this.oAuth2.login(false); // will throw error, if interactive login is needed
         return await this.callEWS(aRequest); // repeat the call
+      } else if (this.NTLMauthorization && !authorization) {
+        return await this.callEWS(aRequest, await appGlobal.remoteApp.createType3MessageFromType2Message(response.WWWAuthenticate, this.username, this.password));
```
benbucksch, line 278: Could you avoid this new parameter? E.g. could you set `this.NTLMauthorization` instead?

backend/backend.ts
```diff
@@ -9,6 +9,7 @@ import ky from 'ky';
 import { shell, nativeTheme, Notification, Tray, nativeImage, app, BrowserWindow, webContents, Menu, MenuItemConstructorOptions } from "electron";
 import nodemailer from 'nodemailer';
 import MailComposer from 'nodemailer/lib/mail-composer';
+import { createType1Message, decodeType2Message, createType3Message } from "./ntlm";
```
benbucksch, line 12:
```suggestion
import * as NTLMProtocol from "./ntlm";
```

backend/backend.ts
```diff
@@ -56,6 +57,8 @@ async function createSharedAppObject() {
     sendMailNodemailer,
     verifyServerNodemailer,
     getMIMENodemailer,
+    createType1Message,
+    createType3MessageFromType2Message,
```
benbucksch, line 61:
```suggestion
    NTLMProtocol,
```

benbucksch (changes requested):

benbucksch:
@NeilRashbrook I've extracted the `while` change - because it didn't belong here -, commited it on master, and rebased this branch to master.

backend/ntlm.js
NeilRashbrook: `new Buffer(size)` originally gave you uninitialised memory. This was changed in Node 8 to always zero-initialise the buffer. However, it was also deprecated in Node 6; in some versions of Node 7 it would always log a warning to the console, although since Node 10 it only does that for your main code, not in a library. If this code predates Node 8 then they could have converted from `new Buffer` to `Buffer.alloc` and possibly not realised that this gives you zero-initialised memory?

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: The way NTLM works is that you have to make each request twice, the first with a static challenge and the second with a computed response to the server's challenge. I save the static challenge string in `this.NTLMauthorization` and then when I get the server's challenge I can then send the computed response.

backend/backend.ts
```diff
@@ -70,6 +73,10 @@ async function createSharedAppObject() {
   };
 }
 
+function createType3MessageFromType2Message(WWWAuthenticate, username, password) {
```
NeilRashbrook, line 76:
```suggestion
export function createType3MessageFromType2Message(WWWAuthenticate, username, password) {
```
And move this into `ntlm.js` of course. (The reason I had this separate is that the `@ewsjs/ntlm-client` exported the functions like this, but I can't call them separately because the return value from `decodeType2Message` doesn't round-trip through JPC.)

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Yes, I understood that. I meant the way you're passing the challenges and responses within your code.
Instead of the new parameter to `callEWS()`, you could use 1 or 2 object properties, like you're already using NTLMauthorization, you could use either re-use the same property, or make a second one.

Actually, what would be even nicer would be to make a class NTLM, which computes the `Authorization:` HTTP header, in the similar way the OAuth2 class works. The API to the EWS class is only the 1 header value, which the EWS class takes from the NTLM class, and the NTLM class has all the rest of the logic.

backend/backend.ts
benbucksch: > the return value from decodeType2Message doesn't round-trip through JPC

Ah, that's the part I was missing. Makes sense.

Why does it not round-trip? Which value does JPC fail to support?

benbucksch:
I've rebased the branch to master

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: I've tried to implement this in branch `ben/ntlm` = #343 , but it's not finished. You should see the idea, though.

NeilRashbrook: Yes, and it won't work, because for each POST request you first have to send the type 1 message, get a type 2 response, and then repeat the request with a type 3 message which will be unique to that request. Type 3 messages cannot be reused for multiple requests!

benbucksch:
You're saying that very single request has to be repeated and needs 2 HTTPS calls? Seriously? There's no way to avoid that?

Boah, that's awful!

benbucksch:
Doing it both in the backend, like you did, makes sense, though. So, keep it as before.

NeilRashbrook:
> You're saying that very single request has to be repeated and needs 2 HTTPS calls? Seriously? There's no way to avoid that? Boah, that's awful!

Every single POST request needs to be repeated, although technically you don't need to send any data on the first request. (GET requests work differently, as the server will allow multiple requests on the same connection once the first has been authenticated.)

benbucksch:
ben@bexchange account (after restart of app):
```
startup.ts:74 ConnectError: Bexchange: Failed or unsupported authentication protocol(s): Negotiate, NTLM
    at EWSAccount.callEWS (EWSAccount.ts:283:33)
    at async EWSAccount.callEWS (EWSAccount.ts:278:16)
    at async EWSFolder.getNewMessageHeaders (EWSFolder.ts:287:33)
    at async EWSFolder.updateChangedMessages (EWSFolder.ts:102:34)
    at async EWSFolder.listMessages (EWSFolder.ts:44:12)
    at async EWSFolder.getNewMessages (EWSFolder.ts:347:19)
    at async emailAccountLogin (startup.ts:64:5)
EWSAccount.ts:283 Uncaught (in promise) ConnectError: Failed or unsupported authentication protocol(s): Negotiate, NTLM
    at EWSAccount.callEWS (EWSAccount.ts:283:33)
    at async EWSAccount.callEWS (EWSAccount.ts:278:16)
    at async EWSEMail.download (EWSEMail.ts:55:18)
    at async EWSEMail.loadMIME (EMail.ts:337:5)
```

NeilRashbrook:
> Uncaught (in promise)

That's very wrong; I filed PR #347.

benbucksch:
Thanks for fixing the "uncaught error". I've merged that to master and re-based your branch here.
I re-tested, and login with NTLM now works. It might be that it failed, because NTLM is multi-step and the missing `await` meant that we failed to wait for the previous step.

benbucksch:
But actually, the missing `await` was not in the middle of an NTLM auth, so it should not have caused by error above.

Isn't there a race condition here? Multiple EWS calls may happen at the same time. Can the second EWS call "steal" the first auth step of the first EWS call, making the second one succeed and the first one fail?

We cannot simply extend the scope of the lock, because it would deadlock immediately.

Instead, how about - if we use NTLM - making an empty EWS request first without POST body, as the first step of the NTLM exchange, and then do the second EWS request as the actual EWS call? That means we don't need to "repeat", which means we avoid the recursive call, which makes locking easier and is also safer.

NeilRashbrook:
How carefully do you want me to lock? I've created branch `neil/ews-ntlm-lock` which completely locks the entire NTLM sequence (challenge, response, authenticate), unlocking before waiting for the actual data, which will be at least somewhat less inefficient then setting the semaphore to one concurrent connection.

benbucksch:
Are you sure sure that we need the 2 server requests for *every* server request with NTLM ? It's so inefficient, it's hard for me to believe. Where you do get the info from? I'm double-check, because that question will have a major impact on the code structure, and on the performance.

NeilRashbrook:
I got that information from the following page:

https://davenport.sourceforge.net/ntlm.html#ntlmHttpAuthentication

> This scheme differs from most "normal" HTTP authentication mechanisms, in that subsequent requests over the authenticated connection are not themselves authenticated; NTLM is connection-oriented, rather than request-oriented. So a second request for "/index.html" would not carry any authentication information, and the server would request none. If the server detects that the connection to the client has been dropped, a request for "/index.html" would result in the server reinitiating the NTLM handshake.
>
> A notable exception to the above is the client's behavior when submitting a POST request (typically employed when the client is sending form data to the server). If the client determines that the server is not the local host, the client will initiate reauthentication for POST requests over the active connection. The client will first submit an empty POST request with a Type 1 message in the "Authorization" header; the server responds with the Type 2 message (in the "WWW-Authenticate" header as shown above). The client then resubmits the POST with the Type 3 message, sending the form data with the request.

However, trying this manually, this does appear to be unnecessary. Instead, for NTLM, we could authenticate without a `WWW-Authenticate` header, and then if that returns a 401 then we need to perform the other two requests. I've come up with two potential approaches, one in branch `neil/ews-ntlm-headerless` and one in branch `neil/ews-ntlm-3stage`; the difference is that the the latter one performs all of the requests under the semaphore. Both branches appear to function normally in my testing.

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Are you sure that we cannot reuse the `Authentication` header? From how I read the protocol, we can.

NeilRashbrook: I already replied to your other comment.

benbucksch (changes requested):
If we really need to generate the auth header for every call, then we cannot save it in a class property, because we have multiple parallel calls happening. This can lead to race conditions.

NeilRashbrook:
> If we really need to generate the auth header for every call, then we cannot save it in a class property, because we have multiple parallel calls happening. This can lead to race conditions.

The only thing I save in a class property is the type 1 message, which is the same for every call.

Issue #110, opened 2024-06-29
Add context menu to browser views


benbucksch:
### Reproduction 1
* Open an email
* Select some text
* Right-click on the selected text

### Expected result 1
* A context menu with a "Copy" menu item

### Actual result
* Nothing

### Reproduction 2
* Open an email with a URL
* Right-click on the URL

### Expected result 2
* A context menu with "Open link" and "Copy link" menu items

### Actual result
* Nothing

### Reproduction 3
* Open an email with an inline image
* Right-click on the image

### Expected result 3
* A context menu with a "Copy image" menu item

### Actual result
* Nothing

benbucksch:
* Same in the webapps (button "Apps")
* Technically, this is a [`<webview>`](https://www.electronjs.org/docs/latest/api/webview-tag)

benbucksch:
See #192 for implementation attempts and ideas.

benbucksch:
I've implemented the ideas from #192 in the last commit.
DONE:
* Build menu items, based on context
* Implement most of the functions behind the menu items
* Show the menu as Electron menu
* Show the menu as SvelteUI menu (commented-out)

TODO:
* [Electron menu](https://www.electronjs.org/docs/latest/api/menu) (only): Clicking on a menu item throws an unhandled exception on the backend.
* [SvelteUI menu](https://svelteui.dev/core/menu) (only)
  * Menu doesn't appear in the right place, but at the bottom. SvelteUI Menu [doesn't offer any way to pass](https://github.com/svelteuidev/svelteui/blob/main/packages/svelteui-core/src/components/Popper/Popper.svelte) in the x,y coordinates. :-(
  * Close by clicking into the content (mail) doesn't work, because the click event stays in the content document. Same problem with all our other menus. See #216
* Functions not working
  * download()
  * saveAs dialog
  * Copy to clipboard
* Get the `webContents` for `win`
  * Clipboard etc.

@jermy-c Do you want to take over and see which of these you can fix?

jermy-c:
Ok, sure. I'll see how much I can fix.

jermy-c:
PR #334 is a regression since SvelteUI menu doesn't have access to "electron menu roles". We would have revert it to how it was before where each menu action is a function and not have any roles.

benbucksch:
@jermy-c This is fixed now, isn't it?

benbucksch:
All reproductions in the initial description now work for me.
Thanks, Jeremy

PR #338, opened 2025-01-02
Revert "WebView: Context menu roles for predefined actions (#334)"(#110)

jermy-c:
This reverts commit db21fea03d1045a4137835511ab184c055521ac3.

PR https://github.com/mustang-im/mustang/pull/334 is a regression since SvelteUI menu doesn't have access to "electron menu roles". We would have revert it to how it was before where each menu action is a function and not have any roles.


benbucksch (approved):

benbucksch:
@jermy-c Thanks that you did the revert as a separate PR. This is makes the git history clearer. Well done.

PR #340, opened 2025-01-02
JPC: Fix TypeScript errors

NeilRashbrook:
There are currently 70 TypeScript errors, including `lib/jpc/obj.js(20,34): error TS2339: Property 'remove' does not exist on type 'Set<any>'.`, which is fixed in the first commit, while the remainder are fixed in the second commit, although you may not like all of the changes TypeScript demands.

### DIFF
--- a/lib/jpc-ws/WSCall.js
+++ b/lib/jpc-ws/WSCall.js
@@ -3,7 +3,7 @@ import { assert } from "../jpc/util.js";
 
 export default class WSCall extends MessageCall {
   /**
-   * @param webSocket {WebSocket from ws}
+   * @param webSocket {import("ws").WebSocket}
    */
   constructor(webSocket) {
     assert(typeof (webSocket.on) == "function");
--- a/lib/jpc-ws/protocol.js
+++ b/lib/jpc-ws/protocol.js
@@ -3,6 +3,9 @@ import JPCProtocol from "../jpc/protocol.js";
 import WebSocketNode, { WebSocketServer } from "ws";
 import { assert } from "../jpc/util.js";
 
+/** @typedef {number} Integer */
+/** @typedef {object} JSON */
+
 /**
  * Wire protocol API
  */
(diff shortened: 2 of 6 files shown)


benbucksch:
Interesting that TypeScript reads the JSDoc. I'm surprised that you need only so few changes, actually. Seems like we've been diligent in the JSDoc typing.

Only 2 things that I don't like: `JSON` (vs. `object`) and `Integer` (vs. `number`) carries actual information for the developer and is important to keep. Luckily, there's a solution: Typedefs. We already define `type URLString = string;` in app/logic/util/util . We could also do:
`type Integer = number;` and `type JSONO = object`  or `any` or `Record<string, JSONO>` or similar.
This way, we can continue to use `Integer` as type and the TypeScript compiler would no longer complain. You could put these in JPC directly for now. We might later put them in `lib/util/util` or elsewhere, because we will eventually want to use them in app/logic/ as well.

Could you use those 2 typedefs, please? The rest is fine for me. Not sure how you define typedefs in .js files, though. @jermy-c , do you know?

benbucksch:
@NeilRashbrook You called this PR "Fix JPC unsubscribing bug", but the patch contains only JSDoc changes, to it cannot possibly fix any bugs. I also don't see any other PR for that purpose. Did you mean to also submit a separate PR with the unsubscribe bug fix?

NeilRashbrook:
> @NeilRashbrook You called this PR "Fix JPC unsubscribing bug", but the patch contains only JSDoc changes

I specifically put the JPC unsubscribing bug in its own commit...

NeilRashbrook:
... but I messed up and backed it out in the other commit... sorry about that...

NeilRashbrook:
I've reverted all of the `Integer -> number` and `JSON -> object` changes and added `@typedef`s to the relevant files (I don't know a good way of adding typedefs globally and importing them is a pain).

benbucksch:
Hey Neil, could you make a separate PR for the actual code fix, please?

benbucksch (approved):

PR #342, opened 2025-01-02
EWS, OWA, ActS: Fix import of ResponseType

NeilRashbrook:
`OWAEvent.ts` was using `import { type ResponseType }`, although its values need to be accessible. This doesn't cause a run-time error because the global `ResponseType` object gets used instead, so instead the code just fails silently.

Additionally the import filename was overly long (caused by a move of the file) and this trims the import and groups it with the other imports from the same directory.

The overly long import filename also happens in two other files but in this case that's the only change as they weren't using `import { type ResponseType }`.


benbucksch:
Hey Neil, thank you for this perfect PR. You've described the reason for this change really well. It was very clear and easy for me to understand. Thank you.

I'm sure this was awful to debug, when you import A and it actually uses a completely different A at runtime. You probably found it with the debugger.

It's also great that you made this change a separate PR. Thank you.

PR #343, opened 2025-01-03
EWS: NTLM authentication 

### LINKED ISSUE #84: EWS: NTLM authentication

benbucksch:
* Fixes #84. This is the same as #337, but as class.
* NTLM state machine in `NTLM` class. API similar to `OAuth2` class, but not identical.
* First makes an empty call
* Then reuses the token for all subsequent calls


benbucksch:
This is looping. I have a logic error somewhere.

benbucksch:
@NeilRashbrook Can you adopt this patch, find my mistakes, and make it work, based on this overall code structure?

This code is assuming that we can re-use a login `Authentication:` header (because it increases speed by factor 2, and makes the code cleaner). Please test this assumption.

NeilRashbrook:
> @NeilRashbrook Can you adopt this patch, find my mistakes, and make it work, based on this overall code structure?

No; at the very least I have to remove your per-account NTLM state, because NTLM doesn't work like that.

A potential problem with trying to detect supported methods in setup is that it would fail on our old Rackspace account where you could get some servers supporting Basic authentication and others supporting NTLM. As it is, you didn't actually provide UI for changing an existing account to use NTLM authentication.

I've pushed the changes made to get it to work to branch `neil/ben/ntlm`.

> This code is assuming that we can re-use a login `Authentication:` header (because it increases speed by factor 2, and makes the code cleaner). Please test this assumption.

Your assumption is false; the only "re-use" we can make is the initial `Authorization` challenge string.

app/frontend/Settings/Mail/Manual/ManualConfigURL.svelte
```diff
@@ -29,6 +29,9 @@
     <hbox class="authMethod" class:error={authError}>
       <select bind:value={config.authMethod} required>
         <option value={AuthMethod.Password}>{$t`Password`}</option>
+        {#if $config.protocol == "ews"}
```
NeilRashbrook, line 32: Any reason to use `$config` here and regular `config` two lines above? (Rest of the PR seems fine.)

Issue #140, opened 2024-07-18
Mail viewer: Allow external images, on user request


benbucksch:
As a user, if I receive an email with inline images, where the image is hosted on the open web using a http URL, I want to have the option to see the images, on explicit request.

By default, such requests MUST be blocked, to prevent the sender from tracking the user, when he reads the email, and where he is (IP address). However, in some cases, or for some senders, I want to see the images. The same applies not only to images, but all remote content, be it external CSS style sheets or anything else with a HTTP URL.

To be safe, we use 2 layers to stop the load: Using the HTML sanitizer and using `Content-Security-Policy`. Code: convertHTML.ts and WebView.svelte.

The latter already has a property that should toggle between external content or not. But it needs to be implemented, and made to work, in both cases. We cannot weaken our default defenses. But we need to specifically allow external resource loads on request, without disabling the other HTML sanitization.

PR #345, opened 2025-01-03
 Mail viewer: Allow external images and css, on user request (#140)

jermy-c:
- CSP scope `'unsafe-inline'` is used in all cases because without it inline CSS would not load but `unsafe-inline` is not recommended
- `default-src 'none'` is used in all cases because it's also the fallback directive for scripts, workers and iframes
- It only allows all sources for `img-src` and `style-src` on user request
- For sanitizeHTML I just added an extra condition to add `src` or `href` attribute if it's an image or external css

### DIFF
--- a/app/logic/Abstract/Message.ts
+++ b/app/logic/Abstract/Message.ts
@@ -1,7 +1,7 @@
 import type { Contact } from "./Contact";
 import type { Person } from "./Person";
 import type { PersonUID } from "../Abstract/PersonUID";
-import { convertHTMLToText, convertTextToHTML, sanitizeHTML } from "../util/convertHTML";
+import { convertHTMLToText, convertTextToHTML, sanitizeHTML, sanitizeHTMLExternal } from "../util/convertHTML";
 import { Observable, notifyChangedProperty } from "../util/Observable";
 import { backgroundError } from "../../frontend/Util/error";
 import { MapColl } from "svelte-collections";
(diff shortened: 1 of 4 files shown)


app/logic/Abstract/Message.ts
```diff
@@ -85,6 +90,24 @@ export class Message extends Observable {
     }
     return null;
   }
+
+  get htmlWithExternal(): string {
```
benbucksch, line 94: This doesn't work, because `EMail.html` overwrites `Message.html` and also downloads the msg.

Also, we don't copy complex functions like this.

Rather, I'd suggest to make a flag `withExternalContent = false` on the msg, and then check in it the `html` getter.

app/logic/util/convertHTML.ts
```diff
@@ -133,6 +150,12 @@ DOMPurify.addHook('afterSanitizeAttributes', node => {
           node.setAttribute(attribute, "");
           node.setAttribute("title", ex.message ?? ex + "");
         }
+      } else if ((node.tagName.toLocaleLowerCase() == "img" && attribute == "src") ||
+          (node.tagName.toLocaleLowerCase() == "link" && node.getAttribute("rel") == "stylesheet" && attribute == "href")
+          && includeExternal) {
+        let url = node.getAttribute(attribute);
+        let newURL = externalSrc(url);
```
benbucksch, line 157: Given that this no longer allows `data:` and `cid:` URLs, it now breaks inline images.

benbucksch (changes requested):

benbucksch:
> CSP scope `'unsafe-inline'` is used in all cases because without it inline CSS would not load but unsafe-inline is not recommended

* Could we please avoid that?
* Without this PR, are we blocking normal inline CSS in the HTML?
* How common is CSS loaded from https: in email?

benbucksch:
> * Could we please avoid `'unsafe-inline'`?
> * Without this PR, are we blocking normal inline CSS in the HTML?

* Without this PR, we are loading inline CSS.
* With this PR, we are loading inline CSS
* With this PR, but without `'unsafe-inline'`, we do not load inline CSS, even from within the msg, which messes up the msg layout and style.

Issue #282, opened 2024-11-28
ActiveSync: Calendar: Error Sync status 3


benbucksch:
### Reproduction
* Randomly clicking around on folders in the ActiveSync account

### Actual result
* Error: ActiveSync Sync status 3
* Stack:
![image](https://github.com/user-attachments/assets/ff4dc00c-3b6b-4663-bb79-8154e9e648a3)
* Note that the stack shows the calendar

benbucksch:
@NeilRashbrook Can you please:
* Handle the errors that we can recover from? E.g. this one should automatically trigger the re-sync, as requested, and not throw an error.
* Show human-readable, translated error messages for all other sync codes that we cannot handle?

NeilRashbrook:
What do you want to do in the case of an edit operation? Doing a re-sync will throw away all our data and refresh it from the server.

benbucksch:
Keep the edit in a cache, resync, then re-apply the chnage on top of the newly synced data.

NeilRashbrook:
If you're getting a lot of sync status 3, it probably means you have duplicate entries in your calendar database. I noticed I had two sets of ActiveSync calendars and when I deleted one set the sync status 3 errors went away.

NeilRashbrook:
One of my calendars duplicated itself again; I'm not sure how that would happen, the code checks `appGlobal.calendars` for a matching calendar first. Strange also that it was only one calendar that got duplicated.

NeilRashbrook:
> Keep the edit in a cache, resync, then re-apply the chnage on top of the newly synced data.

I've just realised that this doesn't make sense because mustang doesn't actually keep track of what the change was. Anyway, I've created branch `neil/activesync-sync-status-3` which checks each `Sync` command for a status code of `3`, and if so, first resyncs, then checks whether the item being edited still exists, and if so, retries the operation. (I special-case a delete operation so that it does not fail if the item has already been deleted on the server.)

benbucksch:
Yes, this is pretty much what I had in mind.

I see what you do with the makeSyncRequest in ActiveSyncEMail class. Here's abother idea with less changes:
* Do the catch error 3 and resync in folder.
* Add a "retryOnlyIf" function param to makeSyncRequest
* Let callers in ActiveSyncEMail pass that for deleted or not deleted checks

Other point: You do addPingable() on every call to sync(). Do the pingables accumulate or does the callee check for already existing ones?

NeilRashbrook:
[...]

PR #348, opened 2025-01-06
ActiveSync: Stop creating duplicate calendars

NeilRashbrook:
This might be the cause of #282.


benbucksch (approved):
Ooops

benbucksch:
@NeilRashbrook Thanks for finding and fixing this.

PR #353, opened 2025-01-07
Webview: Make necessary only WebContents API available in the frontend (#110)

jermy-c:
- We need access to some WebContents in the frontend for some of the context menu actions e.g. `copyImageAt()`
- The change wraps the original WebContents so only the necessary methods are exposed but this is only possible for some of the methods e.g. `win.session.addWordToSpellCheckerDictionary()` is not possible since we'll have to expose the entire `win.session`
- Exposing `win.session` maybe dangerous since there's a methods called `ses.setPreloads(preloads)` which runs node-level scripts

### DIFF
--- a/app/frontend/Shared/ContextMenu.ts
+++ b/app/frontend/Shared/ContextMenu.ts
@@ -29,11 +29,9 @@ import { appGlobal } from "../../logic/app";
  * });
  * ```
  */
-export async function buildContextMenu(context: ContextInfo, webview: any): Promise<ArrayColl<MenuItem>> {
+export async function buildContextMenu(context: ContextInfo, win: any): Promise<ArrayColl<MenuItem>> {
   context.isText = context.selectionText?.length > 0;
   context.isLink = !!context.linkURL;
-
-  const win = await appGlobal.remoteApp.getWebContents(webview.getWebContentsId());
 
   let menuItems = new ArrayColl<MenuItem>();
   function add(id: string, label, icon: string, action: Function, disabled: boolean = false) {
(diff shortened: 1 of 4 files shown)


benbucksch:
Functionality:
* Which functions worked before this PR?
* Which functions (exposed in the UI) work only after this PR?

Code review:
* You will need `await` for all accesses of `win` now.
* "wraps the original WebContents so only the necessary methods are exposed": Is this only for security, or is this necessary to make it work? If security: Which functions do you not want to be exposed?

jermy-c:
> Which functions worked before this PR?

Tested
- copy
- cut
- paste
- copyLink

Not tested but from reading the code
- searchWeb
- copyImageURL
- copyVideoURL

> Which functions (exposed in the UI) work only after this PR?

- copyImage

> "wraps the original WebContents so only the necessary methods are exposed": Is this only for security, or is this necessary to make it work? If security: Which functions do you not want to be exposed?

This is only for security, I haven't checked the entire API but `win.session.setPreloads()` maybe dangerous. It sets preload scripts.  And also `win.session.loadExtensions()`.

https://www.electronjs.org/docs/latest/api/session#sessetpreloadspreloads
https://www.electronjs.org/docs/latest/api/session#sesloadextensionpath-options

benbucksch (approved):

benbucksch:
> > Which functions (exposed in the UI) work only after this PR?
> 
>     * copyImage (not in the UI yet)

Could you please add this menu item for it, in this PR? Then it's clearer what the purpose is.

> This is only for security, I haven't checked the entire API but `win.session.setPreloads()` maybe dangerous. It sets preload scripts. And also `win.session.loadExtensions()`.
> 
> https://www.electronjs.org/docs/latest/api/session#sessetpreloadspreloads https://www.electronjs.org/docs/latest/api/session#sesloadextensionpath-options

Great, thanks for thinking about this. Could you please add this as code comment above your delegator object? Please start with: "Delegator object", and then your comment above.

As a rule of thumb,  when you do something for security,  and it's not immediately obvious,  make a comment about what exactly you're protecting against.

benbucksch:
Actually, could you please use this backend object only for the one function where you need it? And let the rest use `webview`?

benbucksch:
Also, please put the delegator object impl in its own file. It's large enough to warrant that.

benbucksch (approved):

PR #355, opened 2025-01-07
Task description should be able to scroll horizontally

NeilRashbrook:
Current behaviour is that it pushes the right pane out of the window.

Remove the min-width if you want it to force wrap instead.

### DIFF
--- a/app/frontend/Calendar/EditEvent/EditEvent.svelte
+++ b/app/frontend/Calendar/EditEvent/EditEvent.svelte
@@ -350,6 +350,7 @@
   }
   .left-pane {
     flex: 2 0 0;
+    width: 0px;
   }
   .right-pane {
     flex: 1 0 0;
@@ -393,6 +394,7 @@
   .editor {
     font-family: unset;
     padding: 8px 12px;
+    min-width: fit-content;
   }
   grid.time {
     grid-template-columns: max-content max-content;


benbucksch (approved):

app/frontend/Calendar/EditEvent/EditEvent.svelte
```diff
@@ -350,6 +350,7 @@
   }
   .left-pane {
     flex: 2 0 0;
+    width: 0px;
```
benbucksch, line 353: What is this?

NeilRashbrook: This tells layout not to take the width of the content into account when determining the width of the left pane. By default this will cause the editor text to wrap, so I added the min-width on the editor if you'd prefer it to scroll.

benbucksch:
The dialog was re-written.
The bug was still there.
I fixed it with `word-break: break-all;` in git commit 1ddfa1ad and git commit 3227a76. This fixes long words.
I've also tested large images, they were not a problem.

The real problem here is that the editor is within a `<Scroll>`, but still pushes out the window. `<Scroll>` should *never* under any circumstance grow larger than the size attributed to it by the parent element, completely independent of its content. @NeilRashbrook Do you know how to achieve that?

NeilRashbrook:
It's not the `<Scroll>`'s fault; its ancestor has already decided to grow to show all of the text; this is because the default `min-width` of a flex item is `auto`, meaning "feel free take up as much width as you need in order to show all of your content"*. Your `word-break` approach allows its ancestor not to grow in width because it reduces the automatic minimum width of the content. (I also think it looks better for very long links.) By comparison my approach above was to set `width: 0px;` on the specific flex element causing the problem, as this allows its ancestors to ignore the width of its content (the relevant element is now `.column2 > .section.outer > .inner`). The other approach is simply to set `min-width: 0px;` on all flexible children (at line 174 of app.css) to override the default flexbox behaviour. (In this case the children that need it are `.main-window > hbox > vbox`, `.column2` and `.column2 > .section.outer > .inner`.) However in neither case does it help to set any CSS on the `<Scroll>` itself.

*After discovering this I found this site that might explain it more clearly: https://defensivecss.dev/tip/flexbox-min-content-size/

PR #365, opened 2025-01-09
Don't always switch back to dual view

NeilRashbrook:
Currently the calendar always switches to dual view after viewing or editing an event or switching to another app and back. This seems unintentional, particularly after viewing an event. This makes the selected view persists throughout the session.


benbucksch:
r+

PR #366, opened 2025-01-09
Week views should not scroll to clicked date

NeilRashbrook:
When clicking on a day in the week or four week view, the view will scroll, sometimes unhelpfully actually past the clicked date. This fixes this by separating the date viewed from the date clicked. (This ends up simplifying the dual view slightly.)

### DIFF
--- a/app/frontend/Calendar/selected.ts
+++ b/app/frontend/Calendar/selected.ts
@@ -5,9 +5,8 @@ import { writable, type Writable } from "svelte/store";
 /** Which event is highlighted right now. */
 export let selectedEvent: Writable<Event> = writable(null);
 
-// TODO Move into observable prefs?
-export let startDate = writable(new Date());
 export let selectedDate = writable(new Date());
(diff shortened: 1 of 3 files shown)


app/frontend/Calendar/selected.ts
```diff
@@ -6,6 +6,7 @@ import { writable, type Writable } from "svelte/store";
 export let selectedEvent: Writable<Event> = writable(null);
 
 // TODO Move into observable prefs?
+export let startDate = writable(new Date());
 export let selectedDate = writable(new Date());
 export type DateInterval = 0 | 1 | 2 | 7 | 31 | 28;
 export let selectedDateInterval = writable(2 as DateInterval);
```
benbucksch, line 12:
```suggestion
export let selectedDate = writable(new Date());
export type DateInterval = 0 | 1 | 2 | 7 | 31 | 28;
export let selectedDateInterval = writable(2 as DateInterval);
export let startDate = writable(new Date());
```

benbucksch (approved):

PR #370, opened 2025-01-10
Make ICS processor run

NeilRashbrook:
Unfortunately the old code is valid TypeScript; all it does is declare that `runOn` will always have the value `1`, without actually setting it...


benbucksch:
Yeah... not obvious. Rule of thumb: Everything after ":" is only types and striped out during compilation.

Issue #357, opened 2025-01-08
HTML: "Copy link address" fails with "Document is not focused"


benbucksch:
### Reproduction
1. Open an HTML email
2. Right-click on a link / URL
3. Select "Copy link address" menu item
4. Paste in another app, e.g. a text editor

### Actual result
* Nothing is copied
* No error msg is visible to end user
* Console shows error "NotAllowedError: Failed to execute 'writeText' on 'Clipboard': Document is not focused."
* Stack trace
```
  at copyToClipboard (../../../../app/frontend/Shared/ContextMenu.ts:264:25)
  at copyLink (../../../../app/frontend/Shared/ContextMenu.ts:226:3)
  at menuItem (../../../../app/frontend/Shared/ContextMenu.ts:38:56)
  at catchErrors (../../../../app/frontend/Util/error.ts:60:11)
  at <object>.click (../../../../app/frontend/Shared/WebView.svelte:145:20)
...
```

### Expected result
* The URL is pasted
* On Linux, the secondary clipboard is used (the clipboard where I explicitly copy and explicitly paste something, not the clipboard where every selection is copied and a mouse middle click pastes it)

### Additional info
* "Copy" on a selected text works

benbucksch:
@jermy-c Have you tried focusing the right document? Given that the `writeText()` is baked by the frontend, it should probably be the frontend and not the webview that is focused. A simple `document.documentElement.focus()` might be sufficient.

jermy-c:
I have tried:
- `document.documentElement.focus()`
- `<webview>.focus()`
- `window.focus()`
- `webcontents.focus()`

It be because the while the context menu is in the backend and while the function is executing it will focus on the window it is called.

> **Chrome**
> 
>> Document is not focused
> 
> This error happens in Chrome when there is delay in writing to clipboard and user is not focusing on the window.

https://medium.com/nerd-for-tech/navigator-clipboard-api-b96be187df5d

benbucksch:
https://github.com/mustang-im/mustang/pull/372

PR #372, opened 2025-01-11
Webview: Context Menu: Copy link and *URLs (#110)

jermy-c:
- There's this `Failed to execute 'writeText' on 'Clipboard': Document is not focused` error when using the `navigator.clipboard` so I decided to use the one provided by electron
- The `read()` functions aren't exposed in the renderer process because it poses a security threat where the some code could be used to check what is on the users system clipboard
- So far I haven't found any way to fix `Failed to execute 'writeText' on 'Clipboard': Document is not focused` error to avoid using this method


benbucksch:
https://github.com/mustang-im/mustang/issues/357#issuecomment-2585238963

benbucksch (approved):

Issue #368, opened 2025-01-10
Mail: Persons tabs creates huge number of server requests


benbucksch:
### Reproduction
1. Go to Mail | Vertical view
2. Select the Persons tab
3. Select a person
4. Unselect it
5. Select one or multiple other persons

### Actual result
* App creates thousands of server requests to the mail server. Happens in IMAP (seen by Nick) and Exchange protocols (seen by Ben many months ago), so it's protocol independent.
* Server freaks out and blocks the user. Might cause #350 
* No other user observable problem, other than the app being very slow and erratic

### Expected result
* Only very few network requests
* No looping
* No insane (unnecessarily much) processing

benbucksch:
To add logging:
* For IMAP, you can add a console.log() in `IMAPFolder.runCommand()`. Even `IMAPEMail.download()` etc goes through that.
* For HTTP requests (e.g. Exchange protocols), you can use backend.ts line 150:
```// console.log("Calling server", "input", input, "options", options, "defaults", defaultOptions);` ``

PR #373, opened 2025-01-13
Mail: Person search: Limit reactivity

benbucksch:
* Avoid that the `$:` reactivity statements trigger on the value that they change. Put the action in a function.
* Avoid re-running the search twice for the same person
* Handle that the user clicked elsewhere before the search completed.

Helps with #368, but not a complete fix


nix-universe:
With this change, I the person list is not rendered in the pane at all. Without the change it is rendered.

nix-universe (approved):
Locally tested: this resolves the 500,000 mail db request issue.

PR #382, opened 2025-01-13
Mustang/Parula branding

jermy-c:
- replace `app.setName()` with a build setting so it changes according to the `productName` when building
- There's a bug with the `productName` building setting so that's why we use `extraMetadata.productName` also. See: https://github.com/electron-userland/electron-builder/issues/6714#issuecomment-2150777424
- When right clicking the app icon in the bottom menu bar for Mac OS the `appName` doesn't change even with the build setting because it's similar to a tab and it uses the `<title>` tag, so I used the `<svelte:head>` tag for that.
- I wasn't sure where to put the `<svelte:head>` so it's in the `MainWindow` component now.


e2/electron-builder.yml
```diff
@@ -56,3 +56,6 @@ publish:
     channel: mustang
 #  - provider generic
 #    url: https://mustang.im/auto-updates
+extraMetadata:
+  # <https://github.com/electron-userland/electron-builder/issues/6714#issuecomment-2150777424>
+  productName: Mustang
```
benbucksch, line 61: This will be replaced by the build script with "Parula".

benbucksch, line 61: You were missing a newline at the end of the file. At the end of files, the cursor should be at the left position, directly _below_ the last line.

git diff will warn you about this, and also git PR here makes an (-) icon to warn you.

benbucksch (approved):

e2/electron-builder.yml
benbucksch: I fixed this

Issue #376, opened 2025-01-13
OWA: Upload MIME message


benbucksch:
Implement function `OWAFolder.addMessage(message: EMail)`.
You can assume that `message.mime` is populated.

Use cases:
* Move or copy a message between accounts (or protocols)
* Save Draft

PR #388, opened 2025-01-14
EWS/OWA: Upload MIME message #375 #376

NeilRashbrook:
Also some minimal (! - I couldn't work out how to show more than one account at once) UI for actually copying messages between accounts...

OWA's `moveOrCopyMessagesOnServer` was actually broken, which is a shame, as i needed a copy of the code for uploading (OWA can only upload messages to the Drafts folder), so I had to fix that first.

I also improved some of the parameter type definitions.


benbucksch (approved):
Amazing!!! Thanks so much

Issue #384, opened 2025-01-14
Branding: Logo at OS level


benbucksch:
### Environment
* Parula 0.7.10 and Mustang git master
* Linux, Ubuntu, with Mate desktop environment

### Reproduction
1. Install Parula on Linux (in my case, as .deb)
2. Start it
3. Look at the app menu entry
4. Look at the system-level task bar

### Actual result
* Task bar icon is generic window
* ![image](https://github.com/user-attachments/assets/38d1f110-9ff3-4987-8e9b-79eb0a45351e)
* App menu starter entry is generic app
* ![image](https://github.com/user-attachments/assets/8ce7d7d8-7c4f-4346-b1a1-db208d2894ae)

### Expected result
* Task bar icon is Parula logo
* App menu starter entry is Parula logo

jermy-c:
The issue for Linux logo not set might be because it's automatcially generated from the MacOS icon but there should be a way to set it manually.

https://github.com/electron-userland/electron-builder/issues/6905

jermy-c:
Icon files are not included by default, they need to be included explicitly.

https://www.electron.build/app-builder-lib.interface.metadatadirectories#buildresources

jermy-c:
Fixed by #463

benbucksch:
I've made a relese 0.8.5.

Unfortunately, it's not working on Linux with the .deb. Both in the applications start menu, and in the task bar after the app launched, I still see the generic icon, not the logo.

benbucksch:
It also fails the Parula Mac build, because the PNG icon is too small:
"image /Users/runner/work/mustang/mustang/e2/build/icon.png must be at least 512x512"
https://github.com/mustang-im/mustang/actions/runs/13701744790/job/38317518713

benbucksch:
![Image](https://github.com/user-attachments/assets/585dde41-08c9-42d8-be8f-867c6f21a8e0)
![Image](https://github.com/user-attachments/assets/fb749f22-7f7a-4bf5-8edc-c000367a444c)

If this icon is too small or not good, please ask @elenaux for the right one.

PR #389, opened 2025-01-14
 Branding: Logo at OS level #384 

jermy-c:
- Explicitly set the icon for Linux


benbucksch:
@jermy-c Can you please explain a little more why you make this change? Is this just a guess, or do you have good reason to believe that this will fix the bug?

Re: e2/build/icon.png

Where does the Parula version of this logo come from? I don't see anything in app/build/parula-brand.sh

benbucksch:
@jermy-c I tested it with your branch here, using `cd e2; yarn run build:linux` and then installing the .deb, but the problem remains as before.

Issue #391, opened 2025-01-15
MailZIP: ZIP files are empty


benbucksch:
### Reproduction
1. Set up a mail account
2. Open a few folders
3. Click on a few messages in the folders (only clicking on the msgs will download them entirely, and we save only entire mails to the ZIP file)
4. Wait a bit (we write to disk only once there are no further writes within 3 seconds)

### Actual result
* You see ZIP files in your profile dir in backup/email/$account$/$foldername$.zip
* They all have 22 bytes.
* They are all empty. (You can open them with a ZIP app to verify that they ZIP file is correct, but empty)

### Expected result
* The ZIP files contain files with numbers as filenames
* The files contain the MIME source of the emails

### Other reproduction
1. `yarn test test/logic/Mail/Store/MailZIP.test.ts`
2. The second test checks for this bug and currently fails

benbucksch:
Neil found the bus and submitted a PR.

I tested with the app, and the ZIP file now contain content. Yay! Big progress. Now we're ready for POP3 😉

The test is failing for me. Bug in the test. I get an Uint8Array (test fails), while Neil gets a Buffer object (test succeeds). We're both on Linux. Strange.

nix-universe:
> Neil found the bus and submitted a PR.
> 
> I tested with the app, and the ZIP file now contain content. Yay! Big progress. Now we're ready for POP3 😉

🥳 🎸 🚀

PR #396, opened 2025-01-16
Fix empty MailZip files #391

NeilRashbrook:
Sorry if I'm stepping on someone else's toes here.

The actual problem was a missing `await`; see line 25 of `MailZIP.ts`.

However, this isn't enough to fix the test, as it has other problems:
- It's using the wrong filename to open the ZIP file (so the one it does open is still always empty)
- It's trying to compare a Buffer to a Uint8Array

Investigating the filename issue further, I noticed that `writeZip`'s `filename` is just for locking purposes... but `save` passes the wrong filename, so the locking doesn't actually work. I'm not convinced that locking on the filename is ideal; the other alternative is keying your lock and zip maps on the folder, which would at least avoid an extra call to `getFolderZIPFilePath`, but you might want to use `WeakMap`s if you do that.

While debugging I also confused myself at one point as the test was opening the file saved from the previous test. I've made the test clean up after itself, assuming it succeeds...

On a mostly unrelated note, is it worth using the `beforeAll` and `afterAll` functions to start and stop the backend automatically?

### DIFF
--- a/app/logic/Mail/Store/MailZIP.ts
+++ b/app/logic/Mail/Store/MailZIP.ts
@@ -22,7 +22,7 @@ export class MailZIP implements MailContentStorage {
     assert(email.mime, "Need MIME source to save the email as mail.zip");
     let zip = await this.getFolderZIP(email.folder);
     let filename = this.getEMailFilename(email);
-    if (zip.getEntry(filename)) {
+    if (await zip.getEntry(filename)) {
       return;
     }
     // https://github.com/cthackers/adm-zip/wiki/ADM-ZIP-Introduction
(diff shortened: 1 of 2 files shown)


benbucksch (approved):

benbucksch:
Thanks a lot!

> is it worth using the beforeAll and afterAll functions to start and stop the backend automatically?

No, we need to build a proper testing infra. Ideally, these functions would run exactly once, no matter whether you run the entire testsuite or a just a single test/file. I hope that there's an API in jest/vitest for that, but I didn't check.

NeilRashbrook:
Apparently you can configure a setup file which can run code before and after all of the tests.

nix-universe:
> Sorry if I'm stepping on someone else's toes here.
> 
They only get stronger this way... Or it motivates to buy a pair of S3 shoes.
Your help is invaluable Neil!

> The actual problem was a missing `await`; see line 25 of `MailZIP.ts`.
💥  🥳 👍

nix-universe:
> Thanks a lot!
> 
> > is it worth using the beforeAll and afterAll functions to start and stop the backend automatically?
> 
> No, we need to build a proper testing infra. Ideally, these functions would run exactly once, no matter whether you run the entire testsuite or a just a single test/file. 

TESTINFRA mentioned! Let's go!!!! 🚀

nix-universe (commented):
Well done!

NeilRashbrook:
The problem with setup files is that they run for every test, even those that don't use JPC. But if you want to go that way, I created branch `neil/setup-backend` as proof of concept.

As it happens, all of the relevent tests are already using `backend.test.ts` to set up JPC, so we can just start and shut down the back end there. I created branch `neil/backend-test` as proof of concept.

Issue #407, opened 2025-01-17
Save draft


benbucksch:

benbucksch:
See EMail.ts line 610.

What happens is that JPC `observe()` function tries to serialize the entire object tree, starting with the `email.folder` and `email.identity`, and walking `folder.messages`, and serializing all messages. Obviously, that doesn't make sense.

benbucksch:
@NeilRashbrook Could you help me out here, please? I'm stuck on this.

NeilRashbrook:
You shouldn't be sending an `email` object across IPC. To generate MIME for a message, use `await SMTPAccount.getMIME(email);`. See https://github.com/mustang-im/mustang/blob/20a2fd5d8e518f82d55b02905c1fceae97ad6dIb particuarl, 38/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts#L122

`observe()` is a red herring; this would also happen with an object that doesn't support Svelte directly in `createObjectDescription`.

When sending an object across JPC, in order to accurately represent the object, all of its own properties also get sent, unless they begin with `_`, as those are assumed to be backing values for getters. Examples of own properties on the `EMail` object are `to`, `cc`, `bcc`, `attachments`, `tags`, `headers`, `folder`, `storageLock`, `identity`.

NeilRashbrook:
(I don't know what happened to that link, it looked fine in preview.) https://github.com/mustang-im/mustang/blob/20a2fd5d8e518f82d55b02905c1fceae97ad6d38/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts#L122

NeilRashbrook:
We haven't run into this before because none of our front end calls pass classes as parameters and most of them also return primitive values or plain objects. One obvious exception is our `WebContents` wrapper but that uses `_win` for its reference which avoids being serialised. I don't know about the other Node imports which we serialise from the backend to the front end (HTTPServer, ImapFlow, Database, Zip, shell, Tray, Notification, fs).

benbucksch:
@NeilRashbrook Thanks, Neil, for the good patch.

We still have one problem: If I drag a file (e.g. a PDF file) into the mail composer as attachment, then save as draft, it saves apparently correctly. I can open the file while viewing the draft as message in the viewer (click on the attachment icon to open it in a local app). I can even re-open the draft, using the [Edit Draft] button, and the attachment shows up correctly in the mail composer again. However, I can then neither send nor save as draft again. I get error:
[...]

PR #411, opened 2025-01-17
Generate MIME that can be saved as draft #407

NeilRashbrook:
Unfortunately we can't import `SMTPAccount` from `EMail` as that produces a circular import and the app fails to start, so instead I moved all of the static functions from `SMTPAccount` into an `NMUtils` object exported from `MailAccount` which all users already import.

Or I guess it could go into a new file, but I only just thought of that.


benbucksch (approved):

benbucksch:
I've ended up with almost the same commit as your PR here. Thank you for making me aware of the circular imports. These are nasty, indeed.

I've also separated out `EMailActions`.

PR #412, opened 2025-01-17
Simplify Scheduling values

NeilRashbrook:
I originally provided five values because Exchange exposed them all but they're not needed.

The values are renamed so that they can be readily mapped from iMIP methods.

I used `6` for `REPLY` because in practice we only compare against `REQUEST` and `CANCEL`, but these can be changed if you must.

### DIFF
--- a/app/logic/Calendar/Invitation.ts
+++ b/app/logic/Calendar/Invitation.ts
@@ -1,15 +1,14 @@
 
 /**
  * For an inbox item that represents a scheduling message, the type of message:
- * Accepted/Tentative/Declined responses, invitations, or cancellations.
+ * invitations, responses or cancellations.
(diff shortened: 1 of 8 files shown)


app/logic/Calendar/Invitation.ts
```diff
@@ -1,15 +1,14 @@
...
-  None = 0,
-  Accepted = 1,
-  Tentative = 2,
-  Declined = 3,
-  Request = 4,
-  Cancellation = 5,
+  NONE = 0,
+  REQUEST = 4,
+  REPLY = 6,
+  CANCEL = 5,
```
benbucksch, line 11: Could you please keep the old values? They are much more logical.
(And they should be Capitalized, not UPPERCASE.)

benbucksch (changes requested):

Issue #420, opened 2025-01-27
EWS, OWA, ActS: Delete old folders


benbucksch:
### Reproduction
1. Set up account A in Mustang and in another client (or as a second account in Mustang)
2. With the other client, delete a folder
3. Restart Mustang
4. Click on the deleted folder

### Actual result
* The folder is still here
* You get an error that the ID could not be found

### Expected result
* Folders that were deleted on the server will be deleted on the client. Either immediately and at the latest during the next login.

### Code
* Compare IMAP and JMAP code for `listFolders()`
* Should probably move the code into generic `MailAccount` and use it in all protocol implementations.

benbucksch:
ActiveSync: "Folder hierarchy changed" error is #293

PR #421, opened 2025-01-27
EWS, OWA: Synchronise folder deletions from the server during login #420

NeilRashbrook:
ActiveSync is trickier; it normally doesn't have this problem, but it would do if we ever get a status 9.

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -512,6 +512,7 @@ export class EWSAccount extends MailAccount {
       },
     };
     let result = await this.callEWS(query);
+    this.folderMap.clear();
     for (let folder of result.RootFolder.Folders.Folder) {
       if (!folder.FolderClass || folder.FolderClass == "IPF.Note" || folder.FolderClass.startsWith("IPF.Note.")) {
         let parent = this.folderMap.get(folder.ParentFolderId.Id);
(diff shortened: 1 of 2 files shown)


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -526,6 +527,12 @@ export class EWSAccount extends MailAccount {
         this.folderMap.set(folder.FolderId.Id, ewsFolder);
       }
     }
+    // Iterate from deepest to shallowest
+    for (let folder of this.getAllFolders().reverse()) {
+      if (this.folderMap.get(folder.id) != folder) {
```
benbucksch, line 532:
```suggestion
      if (!this.folderMap.has(folder.id)) {
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -257,6 +258,12 @@ export class OWAAccount extends MailAccount {
         this.folderMap.set(folder.FolderId.Id, owaFolder);
       }
     }
+    // Iterate from deepest to shallowest
+    for (let folder of this.getAllFolders().reverse()) {
+      if (this.folderMap.get(folder.id) != folder) {
```
benbucksch, line 263:
```suggestion
      if (!this.folderMap.has(folder.id)) {
```

benbucksch (approved):

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -526,6 +527,12 @@ export class EWSAccount extends MailAccount {
         this.folderMap.set(folder.FolderId.Id, ewsFolder);
       }
     }
+    // Iterate from deepest to shallowest
+    for (let folder of this.getAllFolders().reverse()) {
+      if (!this.folderMap.has(folder.id)) {
```
benbucksch, line 532: Don't delete, if we accidentally have 2 folder objects for the same folder id.
**Should** not happen, but if it does, we shouldn't destroy the folder. Always opt for safe.
I already commited the change.

NeilRashbrook: Should not happen? Actually it always happens when you move a folder... at least the folder still "works" in its previous location I guess... I'll see whether I can detect this case and move the folder.

PR #422, opened 2025-01-27
ActiveSync: Remove stale folders after a full resync #420

NeilRashbrook:
Normally we automatically sync folder deletions from the server but if a full resync is triggered for some reason then we need to clean up any stale folders.

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -366,7 +366,7 @@ export class ActiveSyncAccount extends MailAccount {
     }
   }
 
-  protected async makeRequest(command: string, request: any, callback?: (response: any, resync?: true) => Promise<void>): Promise<any> {
+  protected async makeRequest(command: string, request: any, callback?: (response: any) => Promise<void>): Promise<any> {
     try {
       // The SyncKey must be the first element of the request.
       request = Object.assign({ SyncKey: this.getStorageItem("sync_key") || "0" }, request);
@@ -378,7 +378,8 @@ export class ActiveSyncAccount extends MailAccount {
       if (callback && ex.code == kFolderSyncKeyError) {
         // Try to resync from start.
         let response = await this.callEAS(command, { SyncKey: "0" });
-        await callback(response, true);
+        response.errorCode = ex.code;
+        await callback(response);
         this.setStorageItem("sync_key", response.SyncKey);
         return response;
       }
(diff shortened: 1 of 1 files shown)


app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -366,7 +366,7 @@ export class ActiveSyncAccount extends MailAccount {
     }
   }
 
-  protected async makeRequest(command: string, request: any, callback?: (response: any) => Promise<void>): Promise<any> {
+  protected async makeRequest(command: string, request: any, callback?: (response: any, resync?: true) => Promise<void>): Promise<any> {
```
benbucksch, line 369: Please do not add edge case parameters to very generic functions.
See a suggestion below on how to fix it.

benbucksch, line 369:
```suggestion
  protected async makeRequest(command: string, request: any, callback?: (response: any) => Promise<void>): Promise<any> {
```

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -378,7 +378,7 @@ export class ActiveSyncAccount extends MailAccount {
       if (callback && ex.code == kFolderSyncKeyError) {
         // Try to resync from start.
         let response = await this.callEAS(command, { SyncKey: "0" });
-        await callback(response);
+        await callback(response, true);
```
benbucksch, line 381:
```suggestion
        response.errorCode = ex.code;
        await callback(response);
```

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -392,7 +392,11 @@ export class ActiveSyncAccount extends MailAccount {
       this.setStorageItem("sync_key", "0");
     }
 
-    await this.queuedRequest("FolderSync", {}, async response => {
+    let missingFolders = new ArrayColl<Folder>();
+    await this.queuedRequest("FolderSync", {}, async (response, resync) => {
+      if (resync) {
```
benbucksch, line 397:
```suggestion
    await this.queuedRequest("FolderSync", {}, async (response) => {
      if (response.errorCode == kFolderSyncKeyError) {
```

benbucksch (changes requested):
Please do not add edge case parameters to very generic functions.

benbucksch:
@NeilRashbrook Could you please check whether my code suggestion works?

benbucksch:
Does this also fix the "Folder hierarchy is out of sync" error when I add or remove a folder in our UI?

NeilRashbrook:
Your version does work, although you hadn't completely backed out the parameter changes, so I added a commit for that.

I don't know whether it's the error that you're referring to, but I happened to notice that a hierarchy change can interrupt the Ping command although apparently only if you don't have the debugger open... This is useful to detect simultaneous hierarchy changes from another client, so I'll open a new PR on that.

PR #427, opened 2025-01-30
EWS, OWA: Custom sent folder

NeilRashbrook:
This turned out to be harder to test than necessary, because if you have multiple accounts, and try to send messages between them, Mustang will always set the `From` email address to the first account, and changing the identity doesn't change the `From` address...

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -1,5 +1,6 @@
 import { AuthMethod, MailAccount, TLSSocketType } from "../MailAccount";
 import type { EMail } from "../EMail";
+import { SpecialFolder } from "../Folder";
 import { EWSFolder } from "./EWSFolder";
 import EWSCreateItemRequest from "./EWSCreateItemRequest";
 import type EWSDeleteItemRequest from "./EWSDeleteItemRequest";
@@ -91,7 +92,10 @@ export class EWSAccount extends MailAccount {
   }
 
   async send(email: EMail): Promise<void> {
-    assert(email.folder.id, "Bad Sent folder");
+    if (email.folder?.account != this) {
+      // Bad sent folder. Ideally this would be an assertion...
+      email.folder = this.getSpecialFolder(SpecialFolder.Sent);
+    }
     let request = new EWSCreateItemRequest({ m$SavedItemFolderId: { t$FolderId: { Id: email.folder.id } }, MessageDisposition: "SendAndSaveCopy" });
     request.addField("Message", "ItemClass", "IPM.Note", "item:ItemClass");
     request.addField("Message", "Subject", email.subject, "item:Subject");
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -92,9 +92,7 @@ export class EWSAccount extends MailAccount {
   }
 
   async send(email: EMail): Promise<void> {
-    if (email.folder?.account != this) {
-      email.folder = this.getSpecialFolder(SpecialFolder.Sent);
-    }
+    assert(email.folder?.id, "Need folder to save the sent email in");
     let request = new EWSCreateItemRequest({ m$SavedItemFolderId: { t$FolderId: { Id: email.folder.id } }, MessageDisposition: "SendAndSaveCopy" });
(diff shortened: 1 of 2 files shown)


benbucksch:
As specced in the issue, could you please make the code work when there's no folder set, or if the folder is the sent folder? In these cases, send the request as before. You can easily implement this by assembling this part of the request as a var in a line above, and set the var to undefined, if you don't want a special sent folder.

NeilRashbrook:
If there's no folder set then that's a type error...

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -91,7 +92,11 @@ export class EWSAccount extends MailAccount {
   }
 
   async send(email: EMail): Promise<void> {
-    let request = new EWSCreateItemRequest({MessageDisposition: "SendAndSaveCopy"});
+    if (!email.folder) { // This is a type error but Ben is paranoid so...
+      email.folder = this.getSpecialFolder(SpecialFolder.Sent);
+    }
+    assert(email.folder.id, "Bad Sent folder");
+    let request = new EWSCreateItemRequest({ m$SavedItemFolderId: { t$FolderId: { Id: email.folder.id } }, MessageDisposition: "SendAndSaveCopy" });
```
benbucksch, line 99:
```suggestion
    let request = new EWSCreateItemRequest({ m$SavedItemFolderId: email.folder?.id ? { t$FolderId: { Id: email.folder.id } } : undefined, MessageDisposition: "SendAndSaveCopy" });
```

NeilRashbrook: I've improved the bad folder test.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -91,7 +92,11 @@ export class EWSAccount extends MailAccount {
   }
 
   async send(email: EMail): Promise<void> {
-    let request = new EWSCreateItemRequest({MessageDisposition: "SendAndSaveCopy"});
+    if (email.folder?.account != this) {
+      // Bad sent folder. Ideally this would be an assertion...
```
benbucksch, line 96:
```suggestion
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -127,7 +128,11 @@ export class OWAAccount extends MailAccount {
   }
 
   async send(email: EMail): Promise<void> {
-    let request = new OWACreateItemRequest({MessageDisposition: "SendAndSaveCopy"});
+    if (email.folder?.account != this) {
+      // Bad sent folder. Ideally this would be an assertion...
```
benbucksch, line 132:
```suggestion
```

benbucksch (changes requested):

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -91,7 +92,10 @@ export class EWSAccount extends MailAccount {
   }
 
   async send(email: EMail): Promise<void> {
-    let request = new EWSCreateItemRequest({MessageDisposition: "SendAndSaveCopy"});
+    if (email.folder?.account != this) {
+      email.folder = this.getSpecialFolder(SpecialFolder.Sent);
+    }
```
benbucksch, line 97:
```suggestion
    assert(email.folder?.id, "Need folder to save the sent email in");
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -127,7 +128,10 @@ export class OWAAccount extends MailAccount {
   }
 
   async send(email: EMail): Promise<void> {
-    let request = new OWACreateItemRequest({MessageDisposition: "SendAndSaveCopy"});
+    if (email.folder?.account != this) {
+      email.folder = this.getSpecialFolder(SpecialFolder.Sent);
+    }
```
benbucksch, line 133:
```suggestion
    assert(email.folder?.id, "Need folder to save the sent email in");
```

benbucksch (changes requested):

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: (This change makes the import of `SpecialFolder` unnecessary.)

Issue #415, opened 2025-01-21
Translation: Placeholders do not work in release builds with untranslated messages


benbucksch:
### Reproduction
1. Use a release build
2. Change language to German (or another language, presumably)
3. Go to folder properties, by clicking [...] on the row of a folder
4. Look at the messages counts in the folder – `FolderGeneral.svelte`
5. Alternatively, look at the message counts at the bottom of the msg list – `FolderFooter.svelte`

### Actual result
* "0 neu, 8068 ungelesen, 19.270 insgesamt. {0} downloaded messages."
* "0 neu, 0 ungelesen, 1.140 insgesamt. {0} displayed. {0} downloaded messages." – `FolderGeneral.svelte`
* "{0} new, {1} unread, {2} total" – `FolderFooter.svelte`
* `{0}` are supposed to be replaced with the messages counts. But it's actually the number of the placeholder.
* Messages that have been translated to German seem to work correctly, and have their placeholders replaced with the message count.
* Messages that have not yet been translated to German, and might not even have been extracted yet, but which are using the English string hardcoded in the source code, are not working.

### Expected result
* In all cases, the placeholder should be replaced with the number.

PR #443, opened 2025-02-17
Translation: Extract strings and generate placeholder before build (#415)

jermy-c:
- The placeholders for the untranslated strings were not displaying properly because they were not extracted and compiled
- It fallsback to the source string which is just a plain string since it is not compile
- Now there's a new npm command `l10n` which extracts, generates placeholders/template and compiles the strings before building a release


app/package.json
```diff
@@ -4,13 +4,15 @@
   "private": true,
   "version": "0.8.2-dev",
   "scripts": {
-    "dev": "npm run l10n:compile && vite",
-    "build": "npm run l10n:compile && vite build",
+    "dev": "npm run l10n && vite",
```
benbucksch, line 7: Why removing the `:compile`?

Does this make the dev compile slower?

app/package.json
```diff
@@ -71,6 +73,7 @@
     "adm-zip": "^0.5.12",
     "happy-dom": "^16.0.0",
     "html-formatter": "^0.1.9",
+    "json-autotranslate": "^1.14.5",
```
benbucksch, line 76: Can you tell me more about this project?

How does it compare to crowdin etc.?

app/package.json
jermy-c: >  Why removing the :compile?
Because I've combined `l10n:extract`, `l10n:placeholder` and `l10n:compile` into a single `l10n`. But you still do the `l10n:*` separate if you want.

> Does this make the dev compile slower?
Just a little.

benbucksch: Can you quantify, please? Before, after, in seconds?

app/package.json
jermy-c: > Can you tell me more about this project?

It compares the strings and only translate the ones that are still in English. And it replaces ICU Mesageformat placeholders and plurals with plain strings so the translation service doesn't get confused.

> How does it compare to crowdin etc.?

This script allows you to autotranslate locally or on any machine. With Crowdin and other services you'd need to upload it to their services and with each uploaded string you're charged for it if you're over the free limit. Then Crowdin would send parse, compare the strings, send it to the translation service, received the translation back again and apply them(this is what `json-autotranslate` does). Then you'd have to use the Crowdin CLI again and download and apply the translated strings to the repo.

With the script it is to parse, compare and send to translation service. And then receive from translation service and apply the translations again. With Crowdin it is to we have to upload to Crowdin to parse, compare and send from their servers to the translation service. Then Crowndin receives the translation and applies them on their servers. We then need to download and apply those translation from Crowdin servers to our repo.

app/package.json
jermy-c: Before: 3.28s
After: 6.45s

That's more than it felt.

app/package.json
benbucksch: Thanks for the clear explanation!

Yes, it seems like json-autotranslate is what I wanted since the start. Thanks for finding that!

benbucksch (approved):

Issue #108, opened 2024-06-19
Translate app UI using AI tools


benbucksch:
Once #67 is complete, and we have an English locale generated, we need to translate the app into at least 12 to 20 languages.

### Requirements

Find an online service dedicated to translating UI strings that:
* Can import the locale string format that we use in #67, and export back to the same format. Nice to have (not a requirement): For bonus points, it can directly import from and export to GitHub repos.
* Take the English locale for strings to be translated. Then skips the existing strings that are already translated and mark only the missing strings (which are in the English locale, but not yet in the target locale) as to be translated.
* Offers an online web tool to allow human translators to provide the missing translations for the strings needed. The human translators are per language.
* Offers automatic translations using AI tools (similar to deepl.com and Google Translate) to translate the UI strings. Ideally, the tool would be optimized for translating UI strings. It should offer at least the 20 most common languages.
* Ideally, each string has a state, like "not translated", "draft" and "approved", or similar. The tool should know about software versions. Each software version should show statistics of how many strings are not translated, draft, or approved.
* It would be nice, if the tool was Open-Source. Paid tools are acceptable, as long as the prices are reasonable.

### Plan
1. We will run the translation tool over the strings and generate a first translation for each locale using AI.
2. We will hire human translators that will check and correct the strings that are not translated well. The translator can approve individual strings, or correct them.
3. Export from the tool back into our app repo, and release.

jermy-c:
### Weblate troubleshooting

**Conflict Resolution**

1.  Download Translation Memory for backup
2. Commit changes to weblate repo
3. `git remote add weblate https://hosted.weblate.org/git/project/component/` in the GitHub source code repo
4. `git remote update weblate`
5. `git merge weblate/main`, resolve any conflicts
6. `git rebase origin/main`, resolve any conflicts
7. `git push`
8. Reset repo in the Weblate Repository Maintenance section
9.  Weblate pull

**Can't push to translation branch**

1.  Delete `translation` branch
2.  Push again

jermy-c:
### TODO

- [ ] Have Weblate parse ICU plurals as plurals so it's easier for translators to translate and the AI translation will be more accurate. The docs say it's not supported JSON files. https://docs.weblate.org/en/latest/user/translating.html#plurals https://docs.weblate.org/en/latest/formats.html#translation-types-capabilities

benbucksch:
[...]

PR #445, opened 2025-02-19
Auto Translations Script and Workflow(#108)

jermy-c:
- Auto Translations workflow finally works even for Placeholders and Plurals!🎉🎉🎉
- Works with DeepL only


benbucksch (approved):
Yay!!! Thank you!

benbucksch:
I've added the DeelP API key as secret in github, and tried the workflow, and it works! Well done.

Unfortunately, it's over-zealous, see bug #448

PR #446, opened 2025-02-19
Translations: Use FormatJS runtime (#108)

jermy-c:
- Lingui requires extraction and compilation for it to properly function even for fallback
- FormatJS is the standard library for ICU Message format so I picked that and it also works for as a replacement runtime
- FormatJS as a runtime doesn't require extraction or compilation
- You can now edit and save the translation and hot reload would reflect it immediately without restarting the dev server
- No more `*-compiled.json` is needed
- Better solution than #114

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -4,8 +4,8 @@
   "private": true,
   "version": "0.8.3-dev",
   "scripts": {
-    "dev": "npm run l10n && vite",
-    "build": "npm run l10n && vite build",
+    "dev": "vite",
+    "build": "vite build",
     "preview": "vite preview",
     "test": "vitest",
     "check": "svelte-check --tsconfig ./tsconfig.json",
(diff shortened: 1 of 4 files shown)


benbucksch:
This is great! This is what we needed.

Unfortunately, there's only 1 big problem: The new library has 1 MB (!) of new dependencies. Can you please check the dependencies, how much of that is actually JS code?

We don't want the compiled JS code bundle size to go up. We're already at 13 MB, and that's a serious factor for startup time, and even more so for webmail...

app/package.json
```diff
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@faker-js/faker": "^9.0.0",
+    "@formatjs/intl": "^3.1.4",
```
benbucksch, line 25: *Huge* new indirect dependencies

benbucksch (approved):
Code is great.
Just the dependency size.

benbucksch:
Actual size of minified JS source code for webmail increased from 13.532.651 bytes to 13.605.925 bytes, an increase by 73274 bytes = 73 KB = 0.5%. Not great, but acceptable.

Issue #431, opened 2025-02-03
ActiveSync: Save sent copy in specific folder


benbucksch:
### Reproduction
* Use an ActiveSync email account
* Reply to a message in a normal folder (not Inbox or any other special folder)
* Send it

### Actual result
* Message is saved in the Sent folder

### Expected
* Message is saved in the folder where the replied-to message is

### Code
* `MailAccount.send()` should send the message and save it in the folder `email.folder`, if set, otherwise in the Sent folder.

### Implementation
* Wait until msg is saved in server Sent folder (save it after listing folders initially)
* List/update sent folder
* Find the sent msg
* Move the message to our defined folder

PR #450, opened 2025-02-20
ActiveSync: Custom sent folder #431

NeilRashbrook:
Note that the `SendMail` API doesn't actually send the mail, it only queues it to the Outbox, so we have to rely on a subsequent `Ping` to tell us that the mail got sent.

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -36,7 +36,6 @@ export class ActiveSyncAccount extends MailAccount {
   readonly port: number = 443;
   readonly tls = TLSSocketType.TLS;
   readonly pingsMRU = new ArrayColl<ActiveSyncPingable>();
-  readonly customSentFolders: Folder[] = [];
   heartbeat = 60;
   maxPings = kMaxCount;
   listening = false;
@@ -119,11 +118,19 @@ export class ActiveSyncAccount extends MailAccount {
       throw new NotSupported("ActiveSync does not support BCC");
     }
     assert(email.folder?.id, "Need folder to save the sent email in");
+    let defaultSentFolder = this.getSpecialFolder(SpecialFolder.Sent) as ActiveSyncFolder;
     if (email.folder.specialFolder != SpecialFolder.Sent) {
-      let defaultSentFolder = this.getSpecialFolder(SpecialFolder.Sent);
       await defaultSentFolder.listMessages();
     }
-    this.customSentFolders.push(email.folder);
+    defaultSentFolder.newMessageCallbacks.add(message => {
+      if (message.subject != email.subject) {
+        return false;
+      }
+      if (email.folder.specialFolder != SpecialFolder.Sent) {
+        email.folder.moveMessageHere(message).catch(this.errorCallback);
+      }
+      return true;
+    });
     let request = {
       ClientId: await this.nextClientID(),
       SaveInSentItems: {},
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -1,6 +1,6 @@
 import { AuthMethod, MailAccount, TLSSocketType } from "../MailAccount";
 import type { EMail } from "../EMail";
-import type { Folder } from "../Folder";
+import { type Folder, SpecialFolder } from "../Folder";
 import { kMaxCount, ActiveSyncFolder, FolderType } from "./ActiveSyncFolder";
 import { ActiveSyncError } from "./ActiveSyncError";
 import { CreateMIME } from "../SMTP/CreateMIME";
@@ -36,6 +36,7 @@ export class ActiveSyncAccount extends MailAccount {
   readonly port: number = 443;
   readonly tls = TLSSocketType.TLS;
   readonly pingsMRU = new ArrayColl<ActiveSyncPingable>();
+  readonly customSentFolders: Folder[] = [];
   heartbeat = 60;
   maxPings = kMaxCount;
   listening = false;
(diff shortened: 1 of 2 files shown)


app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
```diff
@@ -174,6 +174,17 @@ export class ActiveSyncFolder extends Folder implements ActiveSyncPingable {
     });
     this.messages.addAll(newMsgs);
     this.account.addPingable(this);
+    if (this.specialFolder == SpecialFolder.Sent) {
+      for (let email of newMsgs) {
+        let customSentFolder = this.account.customSentFolders.shift();
```
benbucksch, line 179: * Save the sent emails in an array, instead of the `customSentFolders`. The email contains the property `.folder`.
* When `newMsgs` come in, you need to match up the new mail that comes in, with the email we sent. Either based message ID, if possible, or based on subject.

app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
```diff
@@ -174,6 +174,17 @@ export class ActiveSyncFolder extends Folder implements ActiveSyncPingable {
     });
     this.messages.addAll(newMsgs);
     this.account.addPingable(this);
+    if (this.specialFolder == SpecialFolder.Sent) {
```
benbucksch, line 177: Here, add a possibility to add an observer function.
The sending code then adds an observer and does what you do here.

benbucksch (changes requested):

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -117,6 +117,20 @@ export class ActiveSyncAccount extends MailAccount {
     if (email.bcc.hasItems) {
       throw new NotSupported("ActiveSync does not support BCC");
     }
+    assert(email.folder?.id, "Need folder to save the sent email in");
+    let defaultSentFolder = this.getSpecialFolder(SpecialFolder.Sent) as ActiveSyncFolder;
+    if (email.folder.specialFolder != SpecialFolder.Sent) {
+      await defaultSentFolder.listMessages();
+    }
+    defaultSentFolder.newMessageCallbacks.add(message => {
```
benbucksch, line 125: Can we do that only when the sent folder is not the default ActiveSync sent folder? I.e. within the `if` block above?

NeilRashbrook: The idea is that we track all sent messages so that we can explicitly ignore the ones that shouldn't be moved.

benbucksch: What's the advantage? Why not avoid adding the listener at all in this case? We already know at the time before we add the listener.

I want to avoid running this code altogether in the normal case, and do this special stuff only when the user explicitly sets the sent folder to a non-default folder.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -117,6 +117,20 @@ export class ActiveSyncAccount extends MailAccount {
     if (email.bcc.hasItems) {
       throw new NotSupported("ActiveSync does not support BCC");
     }
+    assert(email.folder?.id, "Need folder to save the sent email in");
+    let defaultSentFolder = this.getSpecialFolder(SpecialFolder.Sent) as ActiveSyncFolder;
```
benbucksch, line 121: Note that the user can change the Sent folder in settings. In this case, the Sent folder flag does not match the server's idea of what the Sent folder is.

benbucksch: Well, this very PR would allow us to remove that limitation.

No assumptions, remember?

benbucksch: TODO: When listing folders, remember (in a variable) the sent folder that the server told us, and use that sent folder here. That allows the user to set a custom sent folder.

app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
```diff
@@ -174,6 +175,14 @@ export class ActiveSyncFolder extends Folder implements ActiveSyncPingable {
     });
     this.messages.addAll(newMsgs);
     this.account.addPingable(this);
+    for (let email of newMsgs) {
+      for (let callback of this.newMessageCallbacks) {
+        if (callback(email)) {
```
benbucksch, line 180:
```suggestion
        let done = callback(email);
        if (done) {
```

app/logic/Mail/ActiveSync/ActiveSyncFolder.ts
```diff
@@ -31,6 +31,7 @@ export enum FolderType {
 export class ActiveSyncFolder extends Folder implements ActiveSyncPingable {
   account: ActiveSyncAccount;
   messages: EMailCollection<ActiveSyncEMail>;
+  readonly newMessageCallbacks = new ArrayColl<{(message: ActiveSyncEMail): boolean}>();
```
benbucksch, line 34:
```suggestion
  /** Callbacks that are called when new messages arrive in this folder.
   * If the callback returns true, it will be removed and not be called anymore. */
  readonly newMessageCallbacks = new ArrayColl<{(message: ActiveSyncEMail): boolean}>();
```

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
benbucksch: Sorry, I don't understand. If we do *not* need to move a message, i.e. there's nothing to do, why would we add a callback?

benbucksch (changes requested):

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: It'll need to be more than a variable, because we don't get reminded each restart.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: Here are some sent messages. Some of them need to be moved. I'm not going to tell you which ones; you seem to know what you're doing.

PR #451, opened 2025-02-20
Always set the from address when changing identity

NeilRashbrook:
Particularly relevant for Exchange accounts which typically don't let you fake the from address.

I'm also updating the name as well.


benbucksch:
Do you have a reproduction for the bug that you are fixing here?

I've seen cases where the identity selection didn't apply, but couldn't see a reproducible pattern. If you fixed this here, that would be great!

NeilRashbrook:
These are the steps where I was hitting the problem, although there may be simpler ones
- Set up different Exchange accounts
- Copy a message from one account to the other
- Reply to the copied message
- Mustang tries to set the identity to the original account
- Change to the other account
- Try to send the reply

Although changing the identity changed the account that was asked to send the message, the send would fail because the from address was still that of the original account.

PR #452, opened 2025-02-20
Fix CORS request header filtering

NeilRashbrook:
I was able to make `getAccessTokenFromParams` use a regular `fetch` to get its data (and of course the fetch also shows up on the developer tools).


benbucksch:
Great! Thanks for this fix!

Code Review: I'd like to avoid regexps wherever possible.

NeilRashbrook:
I only used regex because I was worried about the sec-fetch headers but it seems to work without them for now so I've switched to a faster check.

benbucksch (approved):
I deleted the lower case, not the correct casing. OOOHHHH STUPID ME!!!

THANKS!!!

PR #453, opened 2025-02-22
Translation: Use lightweight L10n lib

jermy-c:
- Use `svelte-icu-l10n` package because it is lighter


app/l10n/l10n.ts
```diff
@@ -139,7 +138,7 @@ function initLocale() {
 
 /** Used in Svelte files, e.g.
  * $t`Hello World!` or $t`Hello ${username}!` */
-export const t = derived(locale, () => gt);
+export const t = _t;
```
benbucksch, line 141: There's no need to replace this function.

app/vite.config.ts
```diff
@@ -15,6 +15,9 @@ export default defineConfig({
   },
 
   plugins: [
+    replace({
+      'svelte-i18n-lingui': 'svelte-icu-l10n',
```
benbucksch, line 19: You don't need this replacement.
* You can remove `T` component and replace it with simple `{$t...}`
* Where `plural()` is used, import it from our l10n.ts.

app/package.json
```diff
@@ -61,15 +58,16 @@
     "pretty-bytes": "^6.1.1",
     "psl": "https://github.com/lupomontero/psl",
     "svelte-collections": "^1.2.4",
-    "svelte-i18n-lingui": "^0.1.7",
     "svelte-icon": "^2.0.0",
+    "svelte-icu-l10n": "https://github.com/mustang-im/svelte-icu-l10n",
```
benbucksch, line 62: As mentioned in chat, please put the functions directly into our l10n.ts.

Please simplify the functions as well. Use as much of our existing functions as possible, then let the new functions do the minimal amount of work to do the rest.

Please document the new functions, to say *why* they are needed. (Don't describe *what* they do, but *why* it's needed.)

benbucksch (changes requested):

benbucksch (approved):

benbucksch:
Thanks, Jeremy!

Both dev and release builds show the placeholders correctly. Plural form seems to work in Calendar Event Repeat.

I've fixed up the license a bit, and cleaned up the code.

PR #454, opened 2025-02-24
Create Participant subclass of PersonUID

NeilRashbrook:
This includes the display of the response status of participants.

Or it would, if I could figure out how to view meeting participants in the new UI... they seem to have all disappeared, along with the close button...

Originally I thought creating the subclass was the way to go, but in the end it might have been be better to just add the property to PersonUID in the first place...

### DIFF
--- a/app/logic/Abstract/PersonUID.ts
+++ b/app/logic/Abstract/PersonUID.ts
@@ -12,6 +12,10 @@ export class PersonUID extends Observable {
   emailAddress: string;
   @notifyChangedProperty
   person?: Person;
+  // Used so that autocompleted meeting participants have a default value.
+  // They should really be Participant objects, but I'm cheating for now.
+  @notifyChangedProperty
+  protected response = 0;
 
   constructor(emailAddress?: string, name?: string) {
     super();
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -1,4 +1,3 @@
-import { PersonUID, findPerson } from "../Abstract/PersonUID";
 import type { Calendar } from "./Calendar";
 import type { RecurrenceRule } from "./RecurrenceRule";
 import { ResponseType, type Responses } from "./Invitation";
@@ -6,18 +5,6 @@ import { Observable, notifyChangedAccessor, notifyChangedProperty } from "../uti
 import { Lock } from "../util/Lock";
 import { assert, randomID, AbstractFunction } from "../util/util";
 import { ArrayColl } from "svelte-collections";
-
-export class Participant extends PersonUID {
-  @notifyChangedProperty
-  response: ResponseType;
-
-  constructor(emailAddress: string, name: string, response: ResponseType) {
-    let person = findPerson(emailAddress);
-    super(emailAddress, name ?? person?.name ?? emailAddress);
-    this.person = person;
-    this.response = response;
-  }
-}
 
 export class Event extends Observable {
   id: string;
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -1,4 +1,4 @@
-import type { PersonUID } from "../Abstract/PersonUID";
+import { PersonUID, findPerson } from "../Abstract/PersonUID";
 import type { Calendar } from "./Calendar";
 import type { RecurrenceRule } from "./RecurrenceRule";
 import { ResponseType, type Responses } from "./Invitation";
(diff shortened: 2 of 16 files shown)


benbucksch:
> Originally I thought creating the subclass was the way to go, but in the end it might have been be better to just add the property to PersonUID in the first place...

It shouldn't be on the generic PersonUID. Your choice was reasonable. Another choice would have been to do Participant class with a person property, but I like your solution better. It makes sense.

app/logic/Calendar/SQL/createDatabase.ts
```diff
@@ -132,11 +132,13 @@ export const calendarDatabaseSchema = sql`
...
-    -- 3 = tentative accepted
-    -- null = not responded
-    "confirmed" INTEGER default null,
+    -- 0 = unknown
+    -- 1 = organizer
+    -- 2 = tentative
+    -- 3 = accepted
+    -- 4 = rejected
+    -- 5 = not responded
+    "response" INTEGER default 0,
```
benbucksch: We don't have a database migration for this, so this would trash all existing profiles and throw errors for all testers and users. Even those who have shortly tried Mustang in the past and will try it again. So, please avoid database changes. If you do, please start a migration.

benbucksch: I'm changing this column name back.
I understand that the values changed and will now be wrong for existing databases, but nobody other than you tested this feature, because it didn't exist.
In contrast, a DB schema change will break the entire app for 100% everybody.

benbucksch (changes requested):

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -307,9 +307,9 @@ export class OWAEvent extends Event {
   }
 }
 
-function addParticipants(attendees, participants: PersonUID[]) {
+function addParticipants(attendees, participants: Participant[]) {
   for (let attendee of attendees) {
-    participants.push(findOrCreatePersonUID(sanitize.emailAddress(attendee.Mailbox.EmailAddress), sanitize.nonemptystring(attendee.Mailbox.Name, null)));
+    participants.push(new Participant(sanitize.emailAddress(attendee.Mailbox.EmailAddress), sanitize.nonemptystring(attendee.Mailbox.Name, null), sanitize.integer(ResponseType[attendee.ResponseType], ResponseType.Unknown)));
```
benbucksch, line 312: Use `sanitize.enum()` instead of `.integer()`

app/logic/Abstract/PersonUID.ts
```diff
@@ -12,6 +12,10 @@ export class PersonUID extends Observable {
   emailAddress: string;
   @notifyChangedProperty
   person?: Person;
+  // Used so that autocompleted meeting participants have a default value.
+  // They should really be Participant objects, but I'm cheating for now.
```
benbucksch, line 16: Yes, the types are wrong. We need to fix that properly.

app/logic/Calendar/SQL/createDatabase.ts
NeilRashbrook: When did you start requiring database migrations? Some warning would have been nice.

app/logic/Calendar/Event.ts
```diff
@@ -1,4 +1,3 @@
-import { PersonUID, findPerson } from "../Abstract/PersonUID";
```
NeilRashbrook: I can't do this as a suggestion, but unsurprisingly `Event` uses the `Participant` type.
```typescript
import type { Participant } from "./Participant";
```

app/logic/Calendar/SQL/createDatabase.ts
benbucksch: Since always. I already told you when you added the current fields (`respondToOrganizer`). That caused tons of drama for our testers.

benbucksch: My plan is to remove those fields for emails, and add a generic JSON column to all tables. This is particularly important for email, where we have lots of other structured data potentially in the emails that we want to save, and we cannot make columns for each.

The thing stopping me is exactly that migration necessary.

app/logic/Calendar/Event.ts
benbucksch: Fixed in git commit ddde84ce

app/logic/Calendar/SQL/createDatabase.ts
NeilRashbrook: I'll have to take your word for it, as I can't find any record of it.

PR #455, opened 2025-02-24
OWA: Teams Meetings

NeilRashbrook:
I see you restored `isOnline`!

Creating an event in the correct calendar is still awkward.
- Open Calendar view
- Click `+`
- Select the correct calendar from the picker
- Click `X`
- Click `+`
- Now create the event.

Note that if you don't add a description to the event before turning it into a Teams meeting, you won't see the Teams meeting description, because it's in text, not HTML.

### DIFF
--- a/app/logic/Calendar/OWA/OWAEvent.ts
+++ b/app/logic/Calendar/OWA/OWAEvent.ts
@@ -186,10 +186,13 @@ export class OWAEvent extends Event {
     }
     let response = await this.calendar.account.callOWA(request);
     this.itemID = sanitize.nonemptystring(response.Items[0].ItemId.Id);
+
+    // The server will set the online meeting URL and append the description.
+    // Get the new values back from the server.
     if (this.calendar.account.isOffice365() && this.isOnline && !this.onlineMeetingURL) {
       // Sadly we can't get all of the changes in one API call
-      await this.updateOnlineMeeting();
-      await this.updateOnlineMeetingURL();
+      await this.getOnlineMeetingDescription();
+      await this.getOnlineMeetingURL();
     }
     if (!this.calUID) {
       // Need an extra server roundtrip to get the UID
@@ -197,7 +200,7 @@ export class OWAEvent extends Event {
     }
   }
 
-  protected async updateOnlineMeeting() {
+  protected async getOnlineMeetingDescription() {
     let request = {
       __type: "GetItemJsonRequest:#Exchange",
       Header: {
@@ -243,7 +246,7 @@ export class OWAEvent extends Event {
     }
   }
 
-  protected async updateOnlineMeetingURL() {
+  protected async getOnlineMeetingURL() {
     let request = {
       __type: "GetCalendarEventJsonRequest:#Exchange",
       Header: {
@@ -263,7 +266,7 @@ export class OWAEvent extends Event {
       },
     };
     let response = await this.calendar.account.callOWA(request);
-    this.onlineMeetingURL = sanitize.nonemptystring(response.Items[0].OnlineMeetingJoinUrl, "");
+    this.onlineMeetingURL = sanitize.url(response.Items[0].OnlineMeetingJoinUrl, null);
   }
 
   protected async updateUID() {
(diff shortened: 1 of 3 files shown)


benbucksch:
> Click X Click +

That means the account picker doesn't work in this dialog as it should. Feel free to file a bug report with reproduction in these cases, as new issue. I'll take a look at it.

> if you don't add a description to the event before turning it into a Teams meeting, you won't see the Teams meeting description, because it's in text, not HTML.

That means that the Event.descriptionText/HTML doesn't work as it should. They should automatically sychronize. (Unless the server messes us up, but even then, we still need to deal with it.)

benbucksch (changes requested):
Can you please make sure that you always use the most specific sanitize.*() function available? E.g. sanitize.url() instead of sanitize.string() etc.

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -133,14 +133,27 @@ export class OWAEvent extends Event {
...
+      new OWAUpdateOffice365OccurrenceRequest(this) :
+      // Office 365 requires an explicit saved item folder id
+      new OWACreateOffice365EventRequest({ SavedItemFolderId: { __type: "TargetFolderId:#Exchange", BaseFolderId: { __type: "DistinguishedFolderId:#Exchange", Id: "calendar" } } });
+  }
+
+  async saveCalendarItem() {
+    let request = this.calendar.account.isOffice365() ? this.getOffice365Request() : this.getItemRequest();
     request.addField("CalendarItem", "Subject", this.title, "item:Subject");
-    request.addField("CalendarItem", "Body", this.descriptionHTML ? { __type: "BodyContentType:#Exchange", BodyType: "HTML", Value: this.descriptionHTML } : this.descriptionText ? { __type: "BodyContentType:#Exchange", BodyType: "Text", Value: this.descriptionText } : null, "item:Body");
+    request.addField("CalendarItem", "Body", this.descriptionHTML ? { __type: "BodyContentType:#Exchange", BodyType: "HTML", Value: this.descriptionHTML } : { __type: "BodyContentType:#Exchange", BodyType: "Text", Value: this.descriptionText || "" }, "item:Body");
```
benbucksch: Defaults should use `??` instead of `||`, unless you want to specifically include `""` (which is not the case here, because `""` itself is the fallback).

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -133,14 +133,27 @@ export class OWAEvent extends Event {
     */
   }
 
-  async saveCalendarItem() {
-    let request: any = this.itemID ?
+  getItemRequest() {
```
benbucksch: I've changed this to protected, to make it clearer that this is just a helper function.

benbucksch (changes requested):

benbucksch:
@NeilRashbrook Could you please file a bug about descriptionHTML and fix it, as mentioned in https://github.com/mustang-im/mustang/pull/455#issuecomment-2680147828 ?

NeilRashbrook:
> That means that the Event.descriptionText/HTML doesn't work as it should. They should automatically sychronize. (Unless the server messes us up, but even then, we still need to deal with it.)

If the server only has text, then it only sends us text. It's up to us to convert it to HTML in that case via #449.

NeilRashbrook:
> ```diff
> +    this.onlineMeetingURL = sanitize.nonemptystring(json.OnlineMeetingJoinUrl, "");
> +    this.isOnline = !!json.IsOnlineMeeting;
> ```

Did you mean to switch this to `sanitize.url`? It got deleted for some reason.

app/logic/Calendar/OWA/OWAEvent.ts
NeilRashbrook: (#449 wants to change this anyway...)

NeilRashbrook:
> That means the account picker doesn't work in this dialog as it should. Feel free to file a bug report with reproduction in these cases, as new issue.

I don't know what you were expecting it to do; it updates `selectedCalendar`, but nothing in `EditEvent` actually reads `selectedCalendar`, it only gets read in `CalendarApp`, `MonthView.EventsLineCell` and `DayView.EventsBlockCell`.

PR #457, opened 2025-02-26
Translations: Generate unique hashID for each single word string instance - Not Ready Yet

jermy-c:
- The hashIDs were the same for all instances of the same single word string. E.g. "Read" in "SearchCriteria.svelte" which means the "past tense" and "Read" in "SettingsCategories.ts" which means "the action". The hashID would be the same for both which meant that there would only be one translation
- If the string is single worded then a unique HashID would be generated based on `string + fileName`
- It only works for plain string without placeholders or plurals so `{0} user` would not generate an unique HashID nor `{num, plural, one {# user} other {# users}}`
- There's a new formatter outputting the below format for json-autotranslate
```json
{
  "id": {
    "message": "[string]",
    "source": "[fileName]"
  }
}
```
- I still need to test if the hashIDs always match


jermy-c:
#459 Implements the fixes which this PR was to fix

PR #459, opened 2025-02-26
Translations: Strings extraction using vite

jermy-c:
- Lingui Extractor was had many parse errors because it did use Vite preprocessors for parsing
- The parse errors caused many of the strings to be deleted
- We need the following format for the autotranslate to group the strings together
```json
{
  "Options": {
    "abcd": "A user-readable string",
  },
}
- The new extractor deletes any strings not found in the source code
- It still sorts the strings based on the value/message
```
- Remaining issues <T> component extraction is still not working


benbucksch:
Review:
* Please do not copy foreign libraries into our code. Use a normal dependency.
* Why do we need another vite.config.ts in l10n/? Normally, the one int he parent dir should be sufficient. If you really, absolutely need that file there, you can import the vite.config.ts from the parent dir, because it's a normal typescript file.
* getCallerFile(): Calling this at runtime is going to be:
  1. Slow (creating exception, getting stack trace etc)
  2. Fragile (if the algo goes wrong for some file types)
  Is there a way to add something to the specific string, e.g. $t(`read`, { extraData: "This is the past tense verb" }) `´?

jermy-c:
>  Please do not copy foreign libraries into our code. Use a normal dependency. 

Sorry, I had didn't know how SvelteJS was compiling/parsing the code without errors so I just copied it and modified it. But I eventually found out how to parse the code without errors so we can extract all the strings. We were just missing the Svelte Precessor. Now it is just a plain script.


>  Why do we need another vite.config.ts in l10n/? Normally, the one int he parent dir should be sufficient. If you really, absolutely need that file there, you can import the vite.config.ts from the parent dir, because it's a normal typescript file.

Because I couldn't figure out how to do the same processing as the config and so I just copied the vite.config and modify with just the precessors need to extract the strings without errors. But now I just placed everything into `l10n/config.ts` and the only plugin we need is the `conditionalCompile`.

>  Instead, is there a way to add something to the specific string, e.g. $t(`read`, { extraData: "This is the past tense verb" }) `´?

I implemented `$t'read *=> This is the past tense verb'`. Where ` *=> ` with space before and after is the commentSymbol anything after the symbol is considered a comment.

jermy-c:
<T> component extraction works now for simple cases where it is just a string msg="string" but there might still a problem with msg={"message with ${interpolation}"} since that can only be processed at runtime. So the hashID might be different.

benbucksch:
Don't worry about the extraction from `<T>` component. It's also fine to just say msg={$t`Text`} instead of `msg="text"`. And even if we have to do extract the strings from T manually, that's still fine. We have only 3 instances.

benbucksch:
> Now it is just a plain script.

Thanks!

benbucksch (approved):
Did you write the files in l10n/extractor/ (other than utils.ts) ?

jermy-c:
> Did you write the files in l10n/extractor/ (other than utils.ts) ?

- `tsExtract.ts` and `svelteExtract` are based on https://github.com/HenryLie/svelte-i18n-lingui/blob/main/src/lib/extractor.js#L130
- But `svelteExtract` is different in our one because we do the Svelte preprocessing in our one which wasn't done it that one.
- `extractor.ts` is completely new since no other lib does Vite preprocessing before parsing and extracting for avoiding errors.

PR #460, opened 2025-02-27
Sanitise calendar UIDs using alphanumdash

NeilRashbrook:


benbucksch:
WONTFIX
As you said, other senders can use non-alphanumdash UIDs.

PR #463, opened 2025-03-04
Logo: Generate for Linux and Replace for Parula

jermy-c:
- Linux logo must be generated from an `.icns` file
- There are only two ways to get it: 1. running `yarn build:mac --dir`, 2. using an package to generate it
- Running `yarn build:mac --dir` is slow and won't work on Github Actions since, a Linux machine cannot build for MacOS
- Added a script to `parula-brand.sh` in Perl because `cp <sourceFile> <destinationFile>` doesn't work on Windows, their command is `copy`
- Only recommended icon sizes are generated for Linux: 16,32,48,64,128,256,512

https://www.electron.build/icons#linux


jermy-c:
The 512x512 png is needed for Parula, right now the resolution is 256x256

Issue #410, opened 2025-01-17
Mac: Window titlebar buttons (minimize etc) on left after overlapping with Mustang logo (MC)


benbucksch:
* Make Mac window buttons appear permanently
* Move logo to the left
* Move app name to the left
* On Mac only
* Verified that the buttons are always on the top left, no way to customize.

PR #464, opened 2025-03-04
Mac: Native style title bar (#410)

jermy-c:
圖片

- Add styles for MacOS only
  - Move app logo to the right of the traffic light
  - Move title closer to the logo
  - Remove Mustang Minimize and Close buttons
- Only enable if not Webmail and the OS name is "macintosh"

Haven't tested if it affects other OS builds yet.


benbucksch (approved):
Nice patch!

Issue #414, opened 2025-01-18
Mail Composer: Size images to window


benbucksch:
Should be simple:
```
  max-width: 90%;
  height: auto;
  margin: 10px;
```

but tiptap's own content model is getting into the way again.

Whatever we do, it needs to work both in the mail composer, as well as when the recipient views the message in an HTML-capable mail client.

benbucksch:
Workaround in git commit 161ef622

PR #465, opened 2025-03-05
Mail Composer: Set image size to window and add the ability to resize #414

jermy-c:
- Set image size to window and add the ability to resize
- You can also align the Image
- The sent result is the same


jermy-c:
Bug: The alignment icons don't show up.

benbucksch:
> The alignment icons don't show up.

Maybe the icon pack is missing from the build?

benbucksch:
nice!

And the dependency is small (11KB) with 0 dependencies. Great!

jermy-c:
> Maybe the icon pack is missing from the build?

Yeah, it's because the I'm in dev mode it's missing but it shows up in the build.

圖片

PR #468, opened 2025-03-07
Logo: Linux logo fix for release build and Mac icon logo fix for Parula build

jermy-c:
- Fixes Linux logo for release build because I forgot add the `npm run icon:icns` command to the `build:release:linux` npm script
- Fixes the Mac OS logo for Parula build


benbucksch (changes requested):
* You forgot the build and build:release scripts
* Please test it
* Is there a way to integrate it into the build system, e.g. vite-config?

jermy-c:
> You forgot the build and build:release scripts

I didn't set it for the others because they don't require or don't use `*.icns`

> Please test it

Ok, will do.

> Is there a way to integrate it into the build system, e.g. vite-config?

I was searching the something similar but I haven't found any yet.

benbucksch (approved):
OK, I see. Only Linux needs it. Gotcha.

PR #470, opened 2025-03-08
Save main window size and position

jermy-c:
- Saves main window size and position in local storage
- Uses browser APIs to save and set window size and position
- Delays showing a bit to prevent initial flash when resizing and repositioning

### DIFF
--- a/app/frontend/MainWindow/MainWindow.svelte
+++ b/app/frontend/MainWindow/MainWindow.svelte
@@ -2,10 +2,8 @@
   <title>{ appName }</title>
 </svelte:head>
 <svelte:window
-  bind:outerWidth={windowWidth}
-  bind:outerHeight={windowHeight}
-  on:resize={saveWindowSize}
-  on:load={loadWindowSize}
+  on:visibilitychange={saveWindowSettings}
+  on:load={loadWindowSettings}
 />
 <vbox flex class="main-window" dir={rtl}>
   <WindowHeader selectedApp={$selectedApp} />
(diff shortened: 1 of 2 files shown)


jermy-c:
Not tested on multiple monitors yet

jermy-c:
Minimum size not implemented yet

app/frontend/MainWindow/MainWindow.svelte
```diff
@@ -1,6 +1,10 @@
 <svelte:head>
   <title>{ appName }</title>
 </svelte:head>
+<svelte:window
+  on:visibilitychange={saveWindowSettings}
+  on:load={loadWindowSettings}
```
benbucksch, line 6: try on domready or similar event that comes earlier

benbucksch: Also, you need to `catchErrors()`

app/frontend/MainWindow/MainWindow.svelte
```diff
@@ -97,6 +103,24 @@
     appGlobal.remoteApp.setTheme(theme);
   }
 
+  let windowSize = getLocalStorage("window.size", [ window.outerWidth, window.outerHeight ]);
+  let windowPosition = getLocalStorage("window.position", [ window.screenX, window.screenY ]);
```
benbucksch, line 107: Use a default value to avoid an error on the very first load after install

app/frontend/MainWindow/MainWindow.svelte
```diff
@@ -97,6 +103,24 @@
...
+  function loadWindowSettings() {
+    assert($windowSize.value.length == 2 && $windowSize.value.every((e) => typeof(e) == "number"),
```
benbucksch, line 109: Thanks for checking the values.

Use `sanitize.integer()` everywhere

benbucksch (changes requested):
Nice that you found a way to do it within the frontend code!

See whether you can use an earlier event, and hopefully we don't need to show the window later. We want to show the window as early as possible.

benbucksch (approved):

benbucksch:
Thanks, Jeremy, for this nice idea!

My suggestion with the "domready" (actually "DOMContentLoaded") event most likely wouldn't have worked, because Svelte needs to load, and the event has probably fired already when MainWindow.svelte has loaded, so that won't work. That's why I moved the code into main.ts.

I've removed the load delay. With this new code, I didn't see any flashing, at least on Linux.

Could you please test whether it flashes for you on Mac?

jermy-c:
> Could you please test whether it flashes for you on Mac?

No, I don't see any flashes. Thanks for improving the code!

app/frontend/main.ts
```diff
@@ -16,3 +21,13 @@ const app = new MainWindow({
 });
 
 export default app;
+
+function loadWindowSettings() {
+  let windowSize = getLocalStorage("window.size", [window.outerWidth, window.outerHeight]).value;
+  let windowPosition = getLocalStorage("window.position", [window.screenX, window.screenY]).value;
+  assert(windowSize.length == 2 && windowSize.every(i => sanitize.integer(i)), gt`Bad window size` + windowSize);
+  assert(windowPosition.length == 2 && windowPosition.every(i => sanitize.integer(i)),  gt`Bad window position` + windowPosition);
```
NeilRashbrook, line 29: This complains if your window is positioned at `[0, 0]`; perhaps `typeof i == "number"` would work better?

Issue #476, opened 2025-03-10
Build: Open-Source licenses file is not in production build


benbucksch:
### Environment
* Official release build, e.g. of Parula

### Reproduction
* Go to Settings | About | Open Source
* See the main content of the window

### Actual result
* It's empty
* `<iframe src="public/attribution.txt.gz" title="Licenses" sandbox="" class="s-Zf6EAjcWadop"></iframe>`
* public/attribution.txt.gz doesn't exist

### Expected result
* Scrolling text window with a long list of licenses appears

### Other constallations
* It works in the dev builds
* Icons loaded from public/ seem to work

benbucksch:
https://github.com/mustang-im/mustang/pull/480#issuecomment-2719428359

jermy-c:
`Uncaught (in promise) TypeError: Failed to fetch` only happens in production but in dev I get `Instead of /public/attribution.txt.gz?url, use /attribution.txt.gz?url.`

benbucksch:
Works in production and dev

PR #480, opened 2025-03-11
Build: Open-Source licenses file is not in production build #476

jermy-c:
There were 3 problems that caused this bug:

1. The public assets were never copied to the dist folder because the default is '<root>/public' and in the case of `e2` it is `e2/public` which has no assets in there. I've set it to `../../../app/public`
2. The file was using the `file:///` url instead of HTTP which doesn't tell the browser it is a gzipped file and it needs to unzip it. I've implemented a `fetchGzip()` function for that and it fetches and unzips it. In development these static assets are served through HTTP so that's why there's no issue.
3. The URL is wrong in production when you set `public/attribution.txt.gz` it turn into `<root>/public/attribution.txt.gz` which is not the correct location in production since the static assets are copied to root and there's no public directory.

Attempts tried:

1. fetching the resource with HTTP and adding `Content-Encoding: gzip, [...other compressions]` and `Accept-Encoding: gzip, [...other compressions]`.
2. search a plugin to unzip or process zipped assets
3. search for a setting in Vite for zipped files
4. search for a setting for iframes
5. tried `config.base`  with `./`, `''`, `http://localhost:5454`, it seems like it doesn't change the URL for No.3
6. Also did `new URL(resource, import.meta.url).href` but that doesn't look nice

### DIFF
--- a/app/frontend/Settings/About/Licenses.svelte
+++ b/app/frontend/Settings/About/Licenses.svelte
@@ -13,7 +13,7 @@
 </div>
 
 <iframe 
-  src="" 
+  src={attributionURL}
   title="Licenses" sandbox=""
   bind:this={iframeE}
 />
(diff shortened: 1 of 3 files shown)


benbucksch (approved):

benbucksch:
Nice, thanks a lot, Jeremy!

benbucksch:
@jermy-c ?

jermy-c:
> Is it even right in dev? I see the following message on the console:
> 
> ```
> Instead of /public/attribution.txt.gz?url, use /attribution.txt.gz?url.
> ```

No, it's not. Thanks for checking that. There's an `Uncaught (in promise) TypeError: Failed to fetch` error, which I don't know why.

Issue #471, opened 2025-03-08
Branding: Linux: Icon on taskbar is generic


benbucksch:
### Environment
* Parula 0.8.6
* Linux, Ubuntu, with Mate desktop environment

### Reproduction
1. Install Parula on Linux (in my case, as .deb)
2. Start it
3. Look at the system-level task bar

### Actual result
* Task bar says "Parula"
* Task bar icon is generic
* ![Image](https://github.com/user-attachments/assets/adb253a3-3b88-4428-bc0e-a1774bd0ba9b)

### Expected result
* Task bar says "Parula"
* Task bar icon is Parula logo

### Additional info
* App start menu has the correct icon now (since #468 and #463): ![Image](https://github.com/user-attachments/assets/ebaf9992-53e9-4ae6-957d-9aa6c29b3b01)

PR #481, opened 2025-03-12
Branding: Linux: Use correct icon (#471)

jermy-c:
- The imported icon was never changed with the correct icon file
- The resources folder indeed is for the `main` and `preload` folder but to continue using that folder we would have to copy the icon to that folder so I just pointed to the file in build folder.


benbucksch:
worksforme! Yay, finally! Thank you.

benbucksch:
Yay @elenaux @jermy-c 

grafik

Issue #168, opened 2024-08-12
Calendar: Recurrence: First event appears twice


benbucksch:
Reproduction:
* Create a recurring event

Actual result:
* The first instance appears twice
* It seems that both the "master" and the generated "recurrance" appears.

![image](https://github.com/user-attachments/assets/e92f28bd-4ec3-483f-986e-40c3f4ddaa92)

Expected result:
* The first instance appears only once

NeilRashbrook:
They are two different event objects; currently you don't have any other UI for choosing which one you want to edit.

benbucksch:
Showing the same appointment twice is obviously wrong.
For the UI, I've filed #169.

benbucksch:
@NeilRashbrook Could you please fix this?

benbucksch:
This is fixed now

PR #483, opened 2025-03-12
Stop displaying recurring master events #168

NeilRashbrook:


app/frontend/Calendar/CalendarApp.svelte
```diff
@@ -20,7 +20,7 @@
   import { mergeColls } from "svelte-collections";
   import { t } from "../../l10n/l10n";
 
-  $: events = mergeColls(appGlobal.calendars.map(cal => cal.fillRecurrences(new Date(Date.now() + 1e11)))).sortBy(ev => ev.startTime);
+  $: events = mergeColls(appGlobal.calendars.map(cal => cal.fillRecurrences(new Date(Date.now() + 1e11)))).filter(ev => !ev.recurrenceRule).sortBy(ev => ev.startTime);
```
benbucksch, line 23: (Not about your current change, but about the `fillRecurrences()` etc.)
Are you aware that this runs every time we open the calendar view?
Could we cache this somehow?

benbucksch (approved):

app/frontend/Calendar/CalendarApp.svelte
NeilRashbrook: We already cache the results; we only run this to ensure that the cache is up-to-date.

Issue #469, opened 2025-03-07
Mac: Window titlebar buttons overlap for RTL languages


jermy-c:
<img width="1440" alt="Image" src="https://github.com/user-attachments/assets/18e5aa0f-2001-4320-b3db-30d0263e19d7" />

- The window titlebar buttons overlap for RTL languages

benbucksch:
Did you check what MacOS does, if the system setting is RTL? Does it put the window buttons on the right side?

jermy-c:
Use https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language to get the system locale.

There's also one other case now thinking about, if the system setting is LTR.

So 4 cases in total:
- System LTR, App LTR
- System LTR, App RTL
- System RTL, App RTL
- System RTL, App LTR

PR #491, opened 2025-03-13
Mac: WindowHeader: Styling to fix button overlap in RTL languages (#469)

jermy-c:
- Styling to fix button over lap in RTL languages on Mac

圖片


benbucksch:
Did you check what MacOS does, if the system setting is RTL? Does it put the window buttons on the right side?

benbucksch (approved):

PR #494, opened 2025-03-13
Comment on some differences between Exchange protocols

NeilRashbrook:
Is this the sort of thing you were looking for in your comment on a48b133?

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -57,8 +57,10 @@ export class ActiveSyncAccount extends MailAccount {
     return new ActiveSyncFolder(this);
   }
 
-  // Currently storing the folder sync key, protocol version and policy key
-  // (if any) in local storage. Should this migrate to configJSON?
+  /**
+   * Currently storing the folder sync key, protocol version and policy key
+   * (if any) in local storage. Should this migrate to configJSON?
+   */
   getStorageItem(key) {
     return localStorage.getItem(`mail.${this.id}.${key}`);
   }
@@ -97,7 +99,7 @@ export class ActiveSyncAccount extends MailAccount {
 
     await this.listFolders();
 
-    // `listFolders` will subscribe to new user address books,
+    // `listFolders` will subscribe to new user-added address books,
     // but until #155 is fixed we need to link existing ones manually.
     for (let addressbook of appGlobal.addressbooks) {
       if (addressbook.protocol == "addressbook-activesync" && addressbook.url.startsWith(this.url + "?") && addressbook.username == this.username) {
@@ -106,7 +108,7 @@ export class ActiveSyncAccount extends MailAccount {
       }
     }
 
-    // `listFolders` will subscribe to new user calendars,
+    // `listFolders` will subscribe to new user-added calendars,
     // but until #155 is fixed we need to link existing ones manually.
     for (let calendar of appGlobal.calendars) {
       if (calendar.protocol == "calendar-activesync" && calendar.url.startsWith(this.url + "?") && calendar.username == this.username) {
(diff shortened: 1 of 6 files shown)


app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -95,19 +97,28 @@ export class ActiveSyncAccount extends MailAccount {
...
+    // but until #155 is fixed we need to link existing ones manually.
     for (let calendar of appGlobal.calendars) {
       if (calendar.protocol == "calendar-activesync" && calendar.url.startsWith(this.url + "?") && calendar.username == this.username) {
         (calendar as ActiveSyncCalendar).account = this;
         await (calendar as ActiveSyncCalendar).listEvents();
       }
     }
+
+    // ActiveSync doesn't have streaming notifications, instead it
+    // can tell us when a pingable needs to resync again. There's no
```
benbucksch, line 119: "it can tell us"

Can you please briefly (in a few words, with helpful pointers) describe *how* it tells us?

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -95,19 +97,28 @@ export class ActiveSyncAccount extends MailAccount {
...
+    // point asking about pingables that we know aren't in sync, so
```
benbucksch, line 120: "There's no point asking about pingables that we know aren't in sync, so ..."

-> "Pingables start from a sync point, so..."

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -58,6 +58,8 @@ export class EWSAccount extends MailAccount {
     }
     await this.listFolders();
 
+    // Link (until #155) or create the default address book.
+    // TODO: Support user address books.
```
benbucksch, line 62:
```suggestion
    // TODO: Support additional address books that the user manually added.
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -89,6 +93,8 @@ export class EWSAccount extends MailAccount {
     await this.oAuth2?.logout();
   }
 
+  // This uses Exchange to construct the message rather than
+  // building the MIME ourselves.
```
benbucksch, line 97: When the comment pertains to an entire function, like this one here, and several other comments below, can you please put it in JSDoc format? That allows the IDE to pick up the comment and show it even when hovering over the caller (!).

Like so:
```suggestion
  /** This uses Exchange to construct the message rather than
   * building the MIME ourselves. */
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -53,17 +53,23 @@ export class OWAAccount extends MailAccount {
     return new OWAFolder(this);
   }
 
+  // The cookie jar used for OWA requests. This allows you to have
+  // multiple OWA accounts for the same host.
```
benbucksch, line 57: Great comment! I know that, but other devs would not.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -53,17 +53,23 @@ export class OWAAccount extends MailAccount {
...
   get partition(): string {
     return 'persist:login:' + this.id;
   }
 
+  // OWA full page login resembles OAuth2, so we label it as such,
+  // although it's actually Office 365 itself doing its own OAuth2.
```
benbucksch, line 63: Great comment, but belongs to `login()` function.

benbucksch (changes requested):
Hey Neil, yes, these comments are great! Thank you for adding them. Would be nice to keep sprinkling such informative comments, even to other existing code.

I requested a few small, trivial improvements.

benbucksch:
@NeilRashbrook ping

benbucksch:
Ah. A message here in the PR will do :) Thank you

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -95,19 +99,29 @@ export class ActiveSyncAccount extends MailAccount {
...
         (calendar as ActiveSyncCalendar).account = this;
         await (calendar as ActiveSyncCalendar).listEvents();
       }
     }
+
+    // ActiveSync doesn't have streaming notifications, instead it
+    // provides the Ping operation which will tell us when a pingable
+    // has gone out of sync. This only makes sense once the pingable
+    // is in sync, so each pingable registers with the account when
+    // it's ready to be specified in the Ping operation.
```
benbucksch, line 124: New comment is great. Thanks!

benbucksch (approved):

PR #495, opened 2025-03-13
Mac: WindowHeader: Styling for system RTL (#469)

jermy-c:
- Fixes window header styling for when the system and app text directions are different

### Cases

1. System LTR, App LTR
圖片

2. System LTR, App RTL
圖片

3. System RTL, App RTL
圖片

4. System RTL, App LTR
圖片

### DIFF
--- a/app/frontend/MainWindow/WindowHeader.svelte
+++ b/app/frontend/MainWindow/WindowHeader.svelte
@@ -1,4 +1,4 @@
-<hbox class="window-header {osRTL ? "os-rtl" : ""}" class:mac
+<hbox class="window-header" class:os-rtl={osRTL} class:mac
   style="--workspace-color: {$selectedWorkspace?.color ?? "var(--windowheader-bg)"}">
   <vbox class="app-logo">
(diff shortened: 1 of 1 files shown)


benbucksch (changes requested):
Thanks for the patch.

To avoid confusion between browser RTL and OS RTL, can you please make the class and variable name unambiguous? I've made a proposal, but the rest of the code needs to be adapted to that.

Thanks for the screenshots!

jermy-c:
Ok, great, thanks for the proposal!

app/frontend/MainWindow/WindowHeader.svelte
```diff
@@ -1,4 +1,4 @@
-<hbox class="window-header" class:mac class:rtl
+<hbox class="window-header {osRTL ? "os-rtl" : ""}" class:mac
```
benbucksch, line 1:
```suggestion
<hbox class="window-header" class:os-rtl={osRTL} class:mac
```

benbucksch (approved):

PR #496, opened 2025-03-13
Polyfill global for frontend

NeilRashbrook:
`window.global = window;` doesn't work when the file gets included by a test.


benbucksch (approved):
Nice! Thanks, Neil.

Issue #498, opened 2025-03-14
Mail: Some inline images are rendered only partially


benbucksch:
### Reproduction
* Click on an inline image

### Actual result
* You see only the top part of the image
* The entire space of the image is reserved. The bottom part is blank.

### Expected result
* The entire image renders

### Conditions
* This doesn't happen for the first rendering of the email. Only the second open, or maybe you even need to restart Mustang to see the problem.
* Doesn't happen for all mails with inline images, only for some.

benbucksch:
Sent example email to Jeremy by email.

jermy-c:
I had a guess was that the generated `data:` was half way because there are some sync and async functions. And `.html` was returning the html before the `data:` was properly generated. So I was trying to see if it was possible to run the `addCID()` in the background and update once it's done. But that was just a guess.

benbucksch:
I have the same effect even when I re-open the image. Even when I shut down the app and restart it. And, I have 3 different (!) emails with the same image - created one after the other, not replies/copied of each other -, and the image stop at the same point in all 3 images. So, it's looks like it's not a race-condition (your guess) or timing issue (80% certainty on that.)

I have the same effect with another image, so the image file (e.g. original corrupted) is not the problem. The other image stops at a different point (50% vs. 10% shown).

benbucksch:
Image 1, email 1
(only top 10% visible)
![Image](https://github.com/user-attachments/assets/c0e72262-dd92-4dff-b544-c306a3f7cfe5)

Image 1, email 2
(ditto)
![Image](https://github.com/user-attachments/assets/5f3b5dad-e05b-450c-a789-b80275177f7c)

Image 1, email 3
(ditto)
![Image](https://github.com/user-attachments/assets/208a0ed9-c814-45ed-b35a-8e7364db9ee5)

The above 3 emails look the same on my 2 IMAP accounts, and in Sent folder (sent copy) and Inbox (email to myself), I have the 3 emails above in 4 copies each (12 total), and they all look the same.


Image 2, email 4
(only top 50% visible)
![Image](https://github.com/user-attachments/assets/88a30ba2-8f1c-4c02-be15-5d4ef03508c2)

Image 2, email 4 (same email), but in a different IMAP account:
(this image renders completely)
![Image](https://github.com/user-attachments/assets/ffc066de-08fb-4e03-a6be-b99717038f8f)

benbucksch:
@NeilRashbrook Any ideas, guesses, where this problem could come from?

[...]

PR #504, opened 2025-03-14
Mail: Load inline messages fix (#498)

jermy-c:
- Changed `FileHandle.read()` to `FileHandle.readFile()` because the buffer limit makes the image partially rendered
> buffer [<Buffer>](https://nodejs.org/api/buffer.html#class-buffer) | [<TypedArray>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray) | [<DataView>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) A buffer that will be filled with the file data read. Default: Buffer.alloc(16384)

https://nodejs.org/api/fs.html#filehandlereadoptions
https://nodejs.org/api/fs.html#filehandlereadfileoptions

- Add a new `loadExternalImages(val: boolean)` to the one in `Messages.ts` sets `sanitizedHTML` to null and doesn't addCID back

https://github.com/mustang-im/mustang/blob/1d5fe19b4603f947285db3354250bbf01ce41a2a/app/logic/Abstract/Message.ts#L121-L127

### DIFF
--- a/app/logic/Mail/EMail.ts
+++ b/app/logic/Mail/EMail.ts
@@ -416,6 +416,8 @@ export class EMail extends Message {
     if (this._loadExternalImages == val) {
       return;
     }
+    // Load body again instead of setting _sanitizedHTML to null
+    // For adding CID specifically
     this.loadBody();
     this._loadExternalImages = val; // notifyChangedProperty triggers update
   }
--- a/app/logic/Mail/EMail.ts
+++ b/app/logic/Mail/EMail.ts
@@ -411,6 +411,14 @@ export class EMail extends Message {
     super.html = val;
     this.needToLoadBody = true;
   }
+  
+  set loadExternalImages(val: boolean) {
+    if (this._loadExternalImages == val) {
+      return;
+    }
+    this.loadBody();
+    this._loadExternalImages = val; // notifyChangedProperty triggers update
+  }
 
   async download() {
     throw new AbstractFunction();
--- a/backend/backend.ts
+++ b/backend/backend.ts
@@ -82,9 +82,9 @@ function createType3MessageFromType2Message(WWWAuthenticate, username, password)
   return createType3Message(decodeType2Message(WWWAuthenticate), username, password);
 }
 
-async function readFile(path: string): Promise<Uint8Array> {
+async function readFile(path: string): Promise<ArrayBufferLike> {
   let fileHandle = await fsPromises.open(path, "r");
-  let { buffer } = await fileHandle.read();
+  let { buffer } = await fileHandle.readFile();
   await fileHandle.close();
   return buffer;
 }


app/logic/Mail/EMail.ts
```diff
@@ -411,6 +411,16 @@ export class EMail extends Message {
...
+  
+  set loadExternalImages(val: boolean) {
+    if (this._loadExternalImages == val) {
+      return;
+    }
+    // Load body again instead of setting _sanitizedHTML to null
+    // For adding CID specifically
+    this.loadBody();
+    this._loadExternalImages = val; // notifyChangedProperty triggers update
+  }
```
benbucksch, line 423: Why is this part necessary?

Note: "external images" != "inline images"

backend/backend.ts
```diff
@@ -82,9 +82,9 @@ function createType3MessageFromType2Message(WWWAuthenticate, username, password)
   return createType3Message(decodeType2Message(WWWAuthenticate), username, password);
 }
 
-async function readFile(path: string): Promise<Uint8Array> {
+async function readFile(path: string): Promise<ArrayBufferLike> {
   let fileHandle = await fsPromises.open(path, "r");
-  let { buffer } = await fileHandle.read();
+  let { buffer } = await fileHandle.readFile();
```
benbucksch, line 87: Interesting. I have no idea why `read()` doesn't work. Do you?
But indeed, this fixes the bug. Thank you!

benbucksch: I've commited this change independently, in git commit c3dd0255

benbucksch (changes requested):

benbucksch:
>   buffer | | A buffer that will be filled with the file data read. Default: Buffer.alloc(16384)

> https://nodejs.org/api/fs.html#filehandlereadoptions
> https://nodejs.org/api/fs.html#filehandlereadfileoptions

Ah, makes sense!

app/logic/Mail/EMail.ts
jermy-c: Because when you go to external images view it's empty and that's because CID is not added to the HTML. The loadExternalImages method sets the sanitizedHTML with the CID added to null and then generates and another one without the CID added.
https://github.com/mustang-im/mustang/blob/1d5fe19b4603f947285db3354250bbf01ce41a2a/app/logic/Abstract/Message.ts#L121-L127

https://github.com/mustang-im/mustang/blob/c3dd02559f19661e35a98771248e0a2911a60f6e/app/frontend/Mail/Message/MessageBody.svelte#L48

jermy-c: And `loadBody()` is not called again after the sanitizedHTML is set to `null`.

PR #509, opened 2025-03-19
Fix some bugs converting from parsed iCal to mustang Event object

NeilRashbrook:
These were found while writing tests based on libical's test data.
- 0fa5ef0 Detect the organiser case-insensitively
- 76625b5 Don't try to detect the organiser twice
- 6ae3265 Recurrence rule should be parsed case-insensitively
- 92b32cf Recurrence rule needs to output recurrence end in allDay format
- c68d52f Correctly handle all-day events


benbucksch:
Thank you!

Issue #35, opened 2024-02-23
Mail Composer: Newline vs. paragraph


benbucksch:
* In HTML, a paragraph has vertical whitespace (= an empty line = bottom margin) between paragraphs.
* In editors, when users press Enter key, they expect a simple newline, not 2 newlines.
* In TipTap, when you press Enter key, it creates a new paragraph, including the unexpected empty newline.
* I am currently working around that by styling a `<p>` to not have a bottom margin. This works only when the recipient also renders CSS.

### Reproduction
* In mail composer
* Type 1
* Hit Enter
* Type 2
* Hit Enter
* Hit Enter
* Type 3
* Send

### Actual result
(without the CSS workaround)

1. Every Enter creates a `<p>`.

HTML:
```html
<p>1</p>
<p>2</p>
<p></p>
<p>3</p>
```

Display:
```
1

2


3
```

### Expected result

1. The first Enter creates a `<br>`.
2. Two Enter successively (= in a row) create a `<p>`.
3. Third Enter creates `<br>` within the `<p>`,
4. the forth Enter creates a second `<p>`, and so on.

HTML:
```html
<p>1<br>
2</p>
<p>3</p>
```

Display:
```
1
2

3
```

### Process
* You need to first remove my CSS workaround, in file `content.css`.
* If `<br>` is not yet allowed, you need to whitelist this HTML element.
* Create a TipTap extension which implements the above behavior.
* It has to harmonize with #32 (in both modes), #30 and other extensions we have.
* Test it with existing documents (e.g. pasting them from websites)

benbucksch:
@jermy-c Can you please pick this up? This is a release blocker now.

benbucksch:
Basically, if the cursor is right before or after a `<br/>`, and the user presses ENTER, then remove the `<br/>` and replace it with `</p><p>`.

benbucksch:
FWIW (just in case it's relevant here), Links `<a href>` should stop once you press enter.

benbucksch:
Check what Thunderbird does and sends out as HTML.
Make sure that <p> is rendered with newline in other email clients.

benbucksch:
### Reproduction
* In the mail composer, in HTML, enter text, and send it to yourself
```
1 line
2 line
3 line

4 line new p
5 line
6 line

7 after new p


8 after 2 new p
```

### Actual result
1. Plaintext: Linebreaks are doubled
```
1 line

2 line

3 line



4 line new p

5 line

6 line



[...]

PR #510, opened 2025-03-21
Mail Composer: Newline vs. paragraph (#35)

jermy-c:
- Adds a `<br>` when there's no `<br>`
- Splits paragraph into 2 paragraphs when there's already a `<br>`
- Only happens in first level `<p>` so `<li><p></li>` is not affected and still has the default behavior

### DIFF
--- a/app/frontend/Shared/Editor/HTMLEditor.svelte
+++ b/app/frontend/Shared/Editor/HTMLEditor.svelte
@@ -12,6 +12,7 @@
   import { SplitBlockquote } from './SplitBlockquote';
   import { Footer } from './Footer';
   import { BoldStar, ItalicSlash } from './StdConventions';
+  import { ParagraphNewLine } from './ParagraphNewLine';
   // import CodeBlockLowlightFeature from '@tiptap/extension-code-block-lowlight';
   // import { common as lowlightCommon, createLowlight } from 'lowlight'
   import { onMount, onDestroy } from 'svelte';
@@ -49,6 +50,7 @@
         }),
         BoldStar,
         ItalicSlash,
+        ParagraphNewLine,
         // CodeBlockLowlightFeature.configure({
         //  lowlight: createLowlight(lowlightCommon),
         // }),
--- a/app/frontend/Shared/Editor/HTMLEditor.svelte
+++ b/app/frontend/Shared/Editor/HTMLEditor.svelte
@@ -6,7 +6,6 @@
   import { Editor } from '@tiptap/core';
   import StarterKit from '@tiptap/starter-kit';
   import LinkFeature from '@tiptap/extension-link';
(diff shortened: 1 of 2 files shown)


jermy-c:
Can someone please test if the current behavior is what's normal?

jermy-c:
One possible issue if when you have the html:

```html
<p>line<br>
line<br>
line</p>
```

And then you make it a list

```html
<ul>
<li><p>line<br>
line<br>
line</p></li>
</ul>
```

The entire thing just becomes one list item and this is because it wraps each `<p>` in a `<li>` instead of using newlines to split

benbucksch (approved):
Looks great!

benbucksch:
### Test case
I've tested it and it works really well. What works:
* First ENTER creates `<br>` :+1: 
* Second consequtive ENTER removes `<br>` and creates `<p>` :+1:
* Bulletet list works :+1:
* ENTER in bulleted list creates new list item  :+1:  
* SHIFT-ENTER in bulleted list creates `<br>`  :+1: 
* Second SHIFT-ENTER in bulleted list creates second `<br>` :+1: 
* ENTER twice in bulleted list ends the bulleted list and creates a normal `<p>`  :+1: 

### Visual results
The result
* looks correct in our editor (after my stylesheet fixes, please check)
* looks correct in our mail viewer
* looks correct in our plaintext mail viewer (top right [T] button)
* looks correct in Thunderbird's mail viewer
* looks correct in Thunderbird's plaintext mail viewer (menu | View | Message body as | plain text)
* looks correct when replying with our editor
* looks correct when replying with Thunderbird's HTML editor
* looks correct when replying with Thunderbird's plaintext editor (SHIFT-click on Reply button in Thunderbird), apart from bulleted list items that have excessive newlines, see below.

### Mail source results
The mail source code: (message menu [...] | View Source)
* is correct in HTML. The HTML elements are exactly as specced above. There is no stylesheet needed. :+1: 
* is correct in plaintext. 

### Problems
The only problems I could find were with bulleted lists:
1. List items contain `<p>`, and visible margins between list items, but they should not. I get `<ul><li><p>Text</p></li></ul>`. Expected: `<ul><li>Text</li></ul>`.
2. Consequence: That is also the reason why you get excessive newlines when replying in the plaintext editor in Thunderbird. Reason: Thunderbird must be down-converting the HTML before going into the plaintext editor, instead of replyintg to the plaintext straight.
3. Text ENTER `* List item` should create a bulleted list, but it does not. Text ENTER ENTER `* List item` works as expected. That means bulleted lists are not created after a `<br>`, only after a `<p>`.

Can you please fix those 2 problems in a new PR?

benbucksch:
Fixes #35

benbucksch:
This is dramatically better than before. Thank you!

PR #511, opened 2025-03-21
EWS, OWA: Fix save draft

NeilRashbrook:
Port of ebdb226.

(Why do we have `AbstractFunction` when we can just use TypeScript's abstract classes?)


benbucksch (approved):
What about ActiveSync?

benbucksch:
> (Why do we have AbstractFunction when we can just use TypeScript's abstract classes?)

Because I didn't know about it? :)

NeilRashbrook:
(ActiveSync didn't support Save Draft until #513 so there was nothing to fix.)

PR #512, opened 2025-03-21
Provide useful message if ActiveSync server sends Retry-After header

NeilRashbrook:
I had to increase the timeout otherwise `ky` will just abort the request. (I tried 20s but that was too low.)

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -233,7 +233,7 @@ export class ActiveSyncAccount extends MailAccount {
         "MS-ASProtocolVersion": this.protocolVersion,
         Cookie: `DefaultAnchorMailbox=${encodeURI(this.emailAddress)}`, // required for 14.0
       },
-      timeout: heartbeat * 1000 + 10000, // extra timeout for Ping commands
+      timeout: heartbeat * 1000 + 25000, // extra timeout for Ping commands
     };
     if (this.oAuth2) {
       options.headers.Authorization = this.oAuth2.authorizationHeader;
@@ -300,6 +300,9 @@ export class ActiveSyncAccount extends MailAccount {
       }
     }
     this.throttle.waitForSecond(1);
+    if (response.RetryAfter) {
+      throw new Error(`The server is overloaded. Please try again after ${response.RetryAfter} seconds.`);
+    }
     throw new Error(`HTTP ${response.status} ${response.statusText}`);
   }
 
--- a/backend/backend.ts
+++ b/backend/backend.ts
@@ -239,6 +239,7 @@ async function postHTTP(url: string, data: any, responseType: string, config: an
     status: response.status,
     statusText: response.statusText,
     data: await response[responseType](),
+    RetryAfter: response.headers.get("Retry-After"),
     WWWAuthenticate: response.headers.get("WWW-Authenticate"),
   };
 }


app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -233,7 +233,7 @@ export class ActiveSyncAccount extends MailAccount {
         "MS-ASProtocolVersion": this.protocolVersion,
         Cookie: `DefaultAnchorMailbox=${encodeURI(this.emailAddress)}`, // required for 14.0
       },
-      timeout: heartbeat * 1000 + 10000, // extra timeout for Ping commands
+      timeout: heartbeat * 1000 + 25000, // extra timeout for Ping commands
```
benbucksch, line 236: We don't want all functions to have a timeout of 26 seconds, but 5 seconds.
This needs to be fixed for merging.

Of course heartbeat functions can be longer.

benbucksch (changes requested):

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: ... the whole point is that we won't get the 503 if we don't wait 25 seconds for it...

benbucksch: I guess I don't understand what you're trying to do here, and how, and what the server does when. Could you please explain more verbatim? The PR description here and your answer is very terse, and I cannot decipher what you try to do and why.

Normal RPC server calls MUST finish within 5 seconds, or fail. For normal function calls, we cannot wait 30 seconds, because that will regress a number of other situations. We will enforce that. But that's for normal calls.

I'm find with having a long timeout for some specific calls like Pings or calls that specifically check for this error situation.

NeilRashbrook: I thought you would be interested in knowing when the server wants you to retry after, but the only way to find that out is to wait the 25 seconds. I guess this would still work for the Ping command, as that waits for 65 seconds by default, but that only helps you once you already have one running; I don't know whether that's enough for you.

benbucksch: Can you please explain more of the circumstances?
* Is this a generic server-side policy setting, which applies always? If so, one call is enough to find out the value, and then we store it
* Is that only in cases where the server is overloaded, i.e. situation-based? If so, what happens when we don't wait 25 seconds? The call will fail, yes, but how? And how will we react to that failure? Retry automatically? In which way? Or fail and leave it to user to try again?
* Is that throttling, which is specific to us? We sent too many calls in a given time, and the server is specifically limiting us and telling us to back off?
* Does that happen only for specific calls that we make, and is based on the parameters? I.e. we request too much data at once? And a changed parameter would fix it?

Depending on the case, we need different strategies to fix it. As it is right now, I am not clear which case you're trying to fix.

NeilRashbrook: Yes, this is dynamic throttling, so the response will depend on the time left before the server-side throttle times out. If we don't wait 25 seconds, `ky` will abort the request; at best this just results in an error banner that the operation timed out. I don't know if there's a specific call, since I usually only notice it at login, after a previous session has tripped the throttling for some reason.

benbucksch: Can you give a reproduction (at least for the case that *you* saw, even if not reproducible for me), from user perspective and which functions are called, so that I understand more precisely?

benbucksch (approved):

PR #513, opened 2025-03-21
ActiveSync: Save draft if 16.1 available

NeilRashbrook:
The alternative would be to directly support 16.1 as a protocol version, which would mean having to check all of the APIs we currently use to see if they work differently in 16.1

[Sync Command Request](https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-asemail/a8602ea5-a3f3-4426-83b5-a8d5315a953d)

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -163,7 +163,7 @@ export class ActiveSyncAccount extends MailAccount {
     if (response.ok) {
       let versions = (response.MSASProtocolVersions || "").split(",");
       if (versions.includes("14.1")) {
-        this.protocolVersion = "14.1";
+        this.protocolVersion = versions.includes("16.1") ? "16.1" : "14.1";
         this.setStorageItem("protocolVersion", this.protocolVersion);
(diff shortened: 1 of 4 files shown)


app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -163,7 +163,7 @@ export class ActiveSyncAccount extends MailAccount {
     if (response.ok) {
       let versions = (response.MSASProtocolVersions || "").split(",");
       if (versions.includes("14.1")) {
-        this.protocolVersion = "14.1";
+        this.protocolVersion = versions.includes("16.1") ? "16.1" : "14.1";
```
benbucksch, line 166: That's a little strange, given the `if` above, but alright...

benbucksch (approved):

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: I need to be sure that the server supports both versions. I wrote that in my PR comment but I guess I should have put it in as a code comment as well...

Issue #508, opened 2025-03-19
Context Menu: Save image does not work


jermy-c:
### Reproduction

1. Open an email with an image
2. Right click to open the context menu
3. Click `Save image`

### Actual Result

1. Does nothing
2. No errors in console

### Expected Result

1. Save image to system downloads folder

### Alternative Cases

1. `Save as` with a pop up dialog
2. External images where the `src` in an URL
3. CID images where the images are `data:`
4. Web Mail

jermy-c:
I think we might have to do it in the `<webview>` or `webContents` with `executeJavascript()` and  the html download attribute.

PR #515, opened 2025-03-22
Context Menu: Save image does not work (#508)

jermy-c:
- Saves images directly for `Save Image`
- Creates a dialog and saves image in selected location for `Save as`
- Right now it uses `<a download="filename">` but the modern way is with `showSaveFilePicker()` but it is not available in all browsers and also has a security warning
- CID images are not saved with original filename, we might need to get that by matching the data URI with the matching attachment object which has the original filename
- If there's an existing file with the same name it gives an error also, instead of the default behavior of `image(n).png` where n is incremented

### DIFF
--- a/app/frontend/Shared/ContextMenu.ts
+++ b/app/frontend/Shared/ContextMenu.ts
@@ -269,7 +269,7 @@ async function openBrowser(url: URLString) {
   await appGlobal.remoteApp.openExternalURL(url);
 }
 
-async function download(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
+async function download(url: URLString, window: any, howSaveAsDialog: boolean, filename?: string) {
   let urlObj = new URL(url);
   if (urlObj.protocol == "http:" || urlObj.protocol == "https:") {
     let blob = await (await fetch(url)).blob();
--- a/app/frontend/Shared/ContextMenu.ts
+++ b/app/frontend/Shared/ContextMenu.ts
@@ -274,12 +274,10 @@ export async function openBrowser(url: URLString) {
 }
 
 export function download(url: URLString, window: any, howSaveAsDialog: boolean, filename?: string) {
-  let a = document.createElement("a");
+  let a: HTMLAnchorElement = document.createElement("a");
   a.href = url;
-  a.setAttribute("target", "_blank"); // for non-same-origin images
-  if (filename) {
-    a.setAttribute("download", filename); // TODO sanitize filename
-  }
+  a.target = "_blank" // for non-same-origin images
+  a.download = "image.png";
   a.click();
   a.href = "";
 }
--- a/app/frontend/Shared/ContextMenu.ts
+++ b/app/frontend/Shared/ContextMenu.ts
@@ -276,6 +276,17 @@ export async function openBrowser(url: URLString) {
 }
 
 export async function download(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
+  let a = document.createElement("a");
+  a.href = url;
+  a.setAttribute("target", "_blank"); // for non-same-origin images
+  if (filename) {
+    a.setAttribute("download", filename); // TODO sanitize filename
+  }
+  a.click();
+  a.href = "";
+}
+
+export async function downloadWebView(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
   let urlObj = new URL(url);
   let blob: Blob;
   if (urlObj.protocol == "http:" || urlObj.protocol == "https:") {
@@ -290,7 +301,12 @@ export async function download(url: URLString, win: any, howSaveAsDialog: boolea
   let ext = fileExtensionForMIMEType(blob.type);
   filename = sanitize.filename(filename, `${filename ?? "untitled"}.${ext}`);
   if (howSaveAsDialog) {
-    saveBlobAsFile(blob, filename);
+    let a = document.createElement("a");
+    a.href = url;
+    a.setAttribute("target", "_blank"); // for non-same-origin images
+    a.setAttribute("download", filename);
+    a.click();
+    a.href = "";
   } else {
     let filePath = await appGlobal.remoteApp.path.join(await appGlobal.remoteApp.getHomeDir(), "Downloads", filename);
     await appGlobal.remoteApp.writeFile(filePath, 0o666, await blob.arrayBuffer());
--- a/app/frontend/Shared/ContextMenu.ts
+++ b/app/frontend/Shared/ContextMenu.ts
@@ -275,6 +275,8 @@ export async function openBrowser(url: URLString) {
   await appGlobal.remoteApp.openExternalURL(url);
 }
 
+// #if [WEBMAIL]
+// @ts-ignore ts2300
 export async function download(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
   let a = document.createElement("a");
   a.href = url;
@@ -285,8 +287,9 @@ export async function download(url: URLString, win: any, howSaveAsDialog: boolea
   a.click();
   a.href = "";
 }
-
-export async function downloadWebView(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
+// #else
+// @ts-ignore ts2300
+export async function download(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
   let urlObj = new URL(url);
   let blob: Blob;
   if (urlObj.protocol == "http:" || urlObj.protocol == "https:") {
@@ -312,3 +315,4 @@ export async function downloadWebView(url: URLString, win: any, howSaveAsDialog:
     await appGlobal.remoteApp.writeFile(filePath, 0o666, await blob.arrayBuffer());
   }
 }
+// #endif
--- a/app/frontend/Shared/ContextMenu.ts
+++ b/app/frontend/Shared/ContextMenu.ts
@@ -292,6 +292,7 @@ export async function download(url: URLString, win: any, howSaveAsDialog: boolea
   if (howSaveAsDialog) {
     saveBlobAsFile(blob, filename);
   } else {
-    saveBlobAsFile(blob, filename);
+    let filePath = await appGlobal.remoteApp.path.join(await appGlobal.remoteApp.getHomeDir(), "Downloads", filename);
+    await appGlobal.remoteApp.writeFile(filePath, 0o666, await blob.arrayBuffer());
   }
 }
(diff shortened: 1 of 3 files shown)


benbucksch:
Can you please not change the generic `download()` function, and instead add a function dedicated to `WebView`?

This also has to work in webmail, where we have no Electron functions and no backend.

benbucksch:
@jermy-c ping

jermy-c:
I don't think it's possible to have a `Save Image` action for web mail because Browser API require explicit user actions. I've tried already.

benbucksch:
@jermy-c On the web, the web browser hooks up the context menu and offers this feature. (In fact, we're merely trying to re-create this useful web browser feature here.) So, on web mail, we should not open a context menu for images, but let the browser handle it.

That's what I meant with "has to work in webmail". Sorry for being vague.

benbucksch:
I also meant that we use the `download()` function in webmail for other purposes, and we cannot break these. The `download()` function before your change should work in webmail, and this should not regress.

app/frontend/Shared/ContextMenu.ts
```diff
@@ -283,3 +287,32 @@ export function download(url: URLString, window: any, howSaveAsDialog: boolean,
...
+  if (howSaveAsDialog) {
+    let a = document.createElement("a");
+    a.href = url;
+    a.setAttribute("target", "_blank"); // for non-same-origin images
+    a.setAttribute("download", filename);
+    a.click();
+    a.href = "";
+  } else {
+    let filePath = await appGlobal.remoteApp.path.join(await appGlobal.remoteApp.getHomeDir(), "Downloads", filename);
+    await appGlobal.remoteApp.writeFile(filePath, 0o666, await blob.arrayBuffer());
```
benbucksch, line 315:
```suggestion
    await appGlobal.remoteApp.writeFile(filePath, 0o644, await blob.arrayBuffer());
```

You made the files world-writable

backend/backend.ts
```diff
@@ -402,6 +403,14 @@ function directory(type: string): string {
   return app.getPath(type as any);
 }
 
+
+/**
+ * Get the user's home directory.
+ */
+function getHomeDir() {
```
benbucksch, line 410: Can you use `directory("home")` instead? (It's right above)

app/frontend/Shared/ContextMenu.ts
```diff
@@ -283,3 +287,32 @@ export function download(url: URLString, window: any, howSaveAsDialog: boolean,
...
+  let ext = fileExtensionForMIMEType(blob.type);
+  filename = sanitize.filename(filename, `${filename ?? "untitled"}.${ext}`);
+  if (howSaveAsDialog) {
+    let a = document.createElement("a");
+    a.href = url;
+    a.setAttribute("target", "_blank"); // for non-same-origin images
+    a.setAttribute("download", filename);
+    a.click();
+    a.href = "";
+  } else {
```
benbucksch, line 313: Do we need this? Why do we need both "Save image" and "Save image as..."? Isn't one sufficient? I'd keep "Save image as..." (1), but rename it to "Save image…" (with the … ellipsis)

(1) given that it's simpler to implement, and more helpful to select the file location

app/frontend/Shared/ContextMenu.ts
```diff
@@ -273,7 +275,9 @@ export async function openBrowser(url: URLString) {
   await appGlobal.remoteApp.openExternalURL(url);
 }
 
-export function download(url: URLString, window: any, howSaveAsDialog: boolean, filename?: string) {
```
benbucksch, line 276: Why is this `export` ? Nobody else uses this. I thought this was the generic function, but it's actually used only in this file.

Please remove the `export` from all functions in this file that are not used by other files.

app/frontend/Shared/ContextMenu.ts
benbucksch: If you keep only "save image as..." (= `howSaveAsDialog = true`), then do we still need 2 different download() function implementations? Do you still need to convert to blob?

Actually, you are making a lot of effort to convert to blob, and then you don't use it at all in the `howSaveAsDialog = true` case.

benbucksch (changes requested):

app/frontend/Shared/ContextMenu.ts
```diff
@@ -273,13 +270,23 @@ export async function openBrowser(url: URLString) {
...
+export async function download(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
+  let urlObj = new URL(url);
+  let blob: Blob;
+  if (urlObj.protocol == "http:" || urlObj.protocol == "https:") {
+    let res = await fetch(urlObj.href);
+    filename = urlObj.pathname.split("/").pop();
+    blob = await res.blob();
+  } else if (urlObj.protocol == "data:") {
+    blob = await dataURLToBlob(url);
+  } else {
```
benbucksch, line 282: While you're at it, can you please also handle `blob:` URLs? We have them, too. They should obviously be trivial in this case, given that you need a `Blob` object.

app/frontend/Shared/ContextMenu.ts
```diff
@@ -66,8 +66,7 @@ export async function buildContextMenu(context: ContextInfo, win: any): Promise<
   if (context.mediaType == "image") {
     add("copyImage", gt`Copy image`, null, copyImage);
     add("copyImageURL", gt`Copy image address`, null, copyImageURL);
-    add("saveImage", gt`Save image`, null, saveImage);
-    add("saveImageAs", gt`Save image as…`, null, saveImageAs);
+    add("saveImage", gt`Save image...`, null, saveImage);
```
benbucksch, line 69: With the reverted `download()` (and `saveBlobAsFile()`) function, do we get a file picker? Is it dependent on settings? (in Firefox, it is.)

benbucksch (changes requested):

app/frontend/Shared/ContextMenu.ts
jermy-c: It is dependent on the Browser settings.

app/frontend/Shared/ContextMenu.ts
```diff
@@ -233,52 +229,62 @@ export async function copyLink(context: ContextInfo, win: any) {
...
+  } else {
+    a.href = url;
+  }
   a.setAttribute("target", "_blank"); // for non-same-origin images
   if (filename) {
+    sanitize.filename(filename);
     a.setAttribute("download", filename); // TODO sanitize filename
+  } else {
+    a.setAttribute("download", "");
   }
```
benbucksch, line 288:
```suggestion
  a.setAttribute("download", sanitize.filename(filename, ""));
```

Last parameter to sanitize.foo() is a default value, in case the requirement is not met. You can use this to avoid the `if` here entirely.

benbucksch (changes requested):

app/frontend/Shared/ContextMenu.ts
```diff
@@ -233,53 +229,58 @@ export async function copyLink(context: ContextInfo, win: any) {
...
   let a = document.createElement("a");
-  a.href = url;
-  a.setAttribute("target", "_blank"); // for non-same-origin images
-  if (filename) {
-    a.setAttribute("download", filename); // TODO sanitize filename
+  let urlObj = new URL(url);
+  if (urlObj.protocol == "http:" || urlObj.protocol == "https:") {
+    let blob = await (await fetch(url)).blob();
+    a.href = await blobToDataURL(blob);
+    filename = filename ?? urlObj.pathname.split("/").pop();
```
benbucksch, line 278: I trust you that this special handling of `http:` URLs was necessary, but can you please explain to me why, so that I understand?

benbucksch (approved):

app/frontend/Shared/ContextMenu.ts
```diff
@@ -233,53 +229,58 @@ export async function copyLink(context: ContextInfo, win: any) {
...
 }
 
-export async function openBrowser(url: URLString) {
+async function openBrowser(url: URLString) {
   await appGlobal.remoteApp.openExternalURL(url);
 }
 
-export function download(url: URLString, window: any, howSaveAsDialog: boolean, filename?: string) {
+async function download(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
   let a = document.createElement("a");
```
benbucksch, line 273:
```suggestion
```

app/frontend/Shared/ContextMenu.ts
```diff
@@ -233,53 +229,58 @@ export async function copyLink(context: ContextInfo, win: any) {
...
-  if (filename) {
-    a.setAttribute("download", filename); // TODO sanitize filename
+  let urlObj = new URL(url);
+  if (urlObj.protocol == "http:" || urlObj.protocol == "https:") {
+    let blob = await (await fetch(url)).blob();
+    a.href = await blobToDataURL(blob);
+    filename = filename ?? urlObj.pathname.split("/").pop();
+  } else {
+    a.href = url;
   }
```
benbucksch, line 281:
```suggestion
    url = await blobToDataURL(blob);
    filename = filename ?? urlObj.pathname.split("/").pop();
  }

  let a = document.createElement("a");
  a.href = url;
```

benbucksch (changes requested):

app/frontend/Shared/ContextMenu.ts
```diff
@@ -233,53 +229,58 @@ export async function copyLink(context: ContextInfo, win: any) {
...
 }
 
-export async function openBrowser(url: URLString) {
+async function openBrowser(url: URLString) {
   await appGlobal.remoteApp.openExternalURL(url);
 }
 
-export function download(url: URLString, window: any, howSaveAsDialog: boolean, filename?: string) {
+async function download(url: URLString, win: any, howSaveAsDialog: boolean, filename?: string) {
```
benbucksch, line 272:
```suggestion
async function download(url: URLString, window: any, howSaveAsDialog: boolean, filename?: string) {
```

benbucksch (approved):

app/frontend/Shared/ContextMenu.ts
jermy-c: Because without it, it opens a new tab in the browser instead of downloading the file. And also it gets the original filename instead of having `untitled.png`

benbucksch: In this particular case where the browser did not initiate a download, did we set the `download` attribute to a filename? Not an empty string? Can you please try that?

Normally, if the `download` attribute is set, the browser should start a download, instead of opening a tab. I can see code paths, both in the old and the new code, where `download` attribute would be empty or non-existing, which would cause this. That's why I'm asking whether that might be the reason. If so, we could avoid the special handing for http -> blob.

benbucksch: ah! "download only works for same-origin URLs" - That's probably why you see this behavior.

PR #517, opened 2025-03-25
Add some more timezone support

NeilRashbrook:
ActiveSync, EWS, OWA: Read timezone from server events

Invitations: Read timezone from .ics attachment and handle invitations given in local time

### DIFF
--- a/app/logic/Calendar/OWA/OWAEvent.ts
+++ b/app/logic/Calendar/OWA/OWAEvent.ts
@@ -70,6 +70,7 @@ export class OWAEvent extends Event {
     if (json.DueDate) {
       this.endTime = sanitize.date(json.DueDate);
     }
+    this.timezone = fromWindowsZone(json.StartTimeZoneId);
     this.allDay = sanitize.boolean(json.IsAllDayEvent);
     if (json.Recurrence) {
       this.repeat = true;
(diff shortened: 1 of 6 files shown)


app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -330,3 +331,12 @@ function addParticipants(attendees, participants: Participant[]) {
 function extractWeekdays(daysOfWeek: string): Weekday[] | null {
   return daysOfWeek ? daysOfWeek.split(" ").map(day => sanitize.integer(Weekday[day])) : null;
 }
+
+function fromWindowsZone(windowsZone): string | null {
+  for (let iana in WindowsTimezones) {
```
benbucksch, line 336: Does OWA use only Windows timezones? Not IANA timezones at all?

Would be nice to also support IANA timezones. In the off chance that Microsoft decides one day to use standards. Or so...

benbucksch: Same for the other protocols.

benbucksch (approved):
Nice, thank you!

app/logic/Calendar/OWA/OWAEvent.ts
NeilRashbrook: I can't prove it for sure but the list of timezones in the OWA UI does seem to match the list of Windows zones. Neither the EWS nor the ActiveSync documentation is very helpful.

Issue #518, opened 2025-03-26
L10n: Extract: quasis is null in generateMessageFromTemplate()


benbucksch:
### Reproduction

`yarn run l10n:extract`

### Actual result
Error:
```
 Cannot read properties of undefined (reading 'map')
    at generateMessageFromTemplate (/home/runner/work/mustang/mustang/app/l10n/extractor/utils.ts:132:33)
    at extractComponent (/home/runner/work/mustang/mustang/app/l10n/extractor/utils.ts:116:40)
```
https://github.com/mustang-im/mustang/actions/runs/14076830744/job/39421213991

Source code:
```
function generateMessageFromTemplate(node): string {
  const rawQuasis = node.quasis.map((q) => q.value.raw);
```

That means `node.quasis` is null

### Impact

Strangely, the translation still seems to run through.

PR #524, opened 2025-03-27
L10n: Use <T> component only for interpolation (#518)

jermy-c:
- There were many problems with the <T> component when extracting string from it since it doesn't know what values are there at compile time.
- I replaced all instances where <T> is used for translation with $t.


benbucksch (approved):
Good change!

PR #525, opened 2025-03-28
L10n: Auto-Translate: Include context for messages: Adapt Build script

jermy-c:
- Npm script to use https://github.com/mustang-im/json-autotranslate/pull/2
- Fix Git Ignores for L10n files

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -29,4 +29,4 @@ public/olm.wasm
 
 # autotranslate
 .json-autotranslate-cache
-*.template.json
+l10n/locales/en/messages.template.json
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -12,8 +12,6 @@ dist
 dist-ssr
 public/olm.wasm
 *.local
-# Temp workaround for bug #112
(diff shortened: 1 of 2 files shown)


app/.gitignore
```diff
@@ -29,5 +27,6 @@ l10n/locales/*/messages-compiled.json
 # Sentry Config File
 .env.sentry-build-plugin
 
-# json-autotranslate
-.json-autotranslate-cache
\ No newline at end of file
+# autotranslate
+.json-autotranslate-cache
+*.template.json
```
benbucksch, line 32: Make them specific to the directory, please, e.g.:
```suggestion
l10n/locales/*/*.json-autotranslate-cache
l10n/locales/en/messages.template.json
```
or similar

benbucksch (changes requested):

app/.gitignore
jermy-c: `.json-autotranslate-cache` is a generated directory in `app/`

benbucksch (approved):
Thanks

Issue #489, opened 2025-03-13
Mail composer: Move to CC freezes the app


benbucksch:
### Reproduction
* Write email or reply, with multiple people in To
* Do *not* have the CC line open
* Click on one of the persons
* Click [CC] to move him/her to CC.

### Actual result
* App freezes
* Error: `(null).openPopup()`

### Expected result
* CC line opens
* Person is removed from To line
* Person is added to CC line
* When sending, person is in CC:, not To: list

benbucksch:
I don't know why the `<PersonsAutocomplete>` does not properly observe changes in `mail.cc`.

I've disabled the button for now, to avoid this bug, in git commit 093ed2d4.

jermy-c:
I get a `could not parse command` as soon as opening the composer.
<img width="909" alt="Image" src="https://github.com/user-attachments/assets/043cf0cd-1206-42e1-b9c4-980102345eec" />

benbucksch:
That's a different bug, unrelated.

NeilRashbrook:
Based on the state of the tree at commit 8af516b, I was able to get the buttons working (including automatically showing the Cc line) by renaming the variable from `personUID` to `person` in the following places:
- `MailAutocomplete.svelte:9`
- `MailComposer.svelte:39-41,63-65,72-74`

I don't know why that works, but I suspect it's because `PersonsAutocomplete.svelte:10` has `person` and not `personUID` as the variable in the `#each` loop.

(If you want to use `personUID` throughout, don't forget to change the various references in Calendar.)

jermy-c:
@NeilRashbrook  was right, it was because of naming of the variable because:

1. In `MailComposer.svelte` the svelte:fragment is passing a `personUID` variable:

https://github.com/mustang-im/mustang/blob/22aa830f86d410fc1f7911f19c2761e8ff13b43d/app/frontend/Mail/Composer/MailComposer.svelte#L39-L42

2. In `MailAutoComplate.svelte` the slot takes a personUID variable:

https://github.com/mustang-im/mustang/blob/22aa830f86d410fc1f7911f19c2761e8ff13b43d/app/frontend/Mail/Composer/MailAutocomplete.svelte#L9

3. But in `PersonsAutocomplete.svelte` the slot is taking a person variable, just changing `{person}` to `personUID` fixes the error for me:

[...]

PR #530, opened 2025-03-31
Mail composer: Use props properly in slots (#489)

jermy-c:
- There were 2 issues I found that caused the UI to freeze, the `person` prop was access before mount is always `undefined` and in the inner components the slots took a `person` prop instead of a `personUID` but `personUID` was the prop passed only in MailComposer, in Display even it uses a `person` also

### DIFF
--- a/app/frontend/Mail/Composer/MailComposer.svelte
+++ b/app/frontend/Mail/Composer/MailComposer.svelte
@@ -36,9 +36,9 @@
       <hbox flex>
         <hbox flex>
           <MailAutocomplete addresses={mail.to} placeholder={$t`Add recipient`} tabindex={1} autofocus={mail.to.isEmpty}>
-            <svelte:fragment slot="person-popup-buttons" let:person={personUID}>
-              <Button plain label={$t`CC`} onClick={() => onMoveToCC(personUID)} />
-              <Button plain label={$t`BCC`} onClick={() => onMoveToBCC(personUID)} />
+            <svelte:fragment slot="person-popup-buttons" let:person>
+              <Button plain label={$t`CC`} onClick={() => onMoveToCC(person)} />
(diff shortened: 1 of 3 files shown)


app/frontend/Mail/Composer/MailComposer.svelte
```diff
@@ -36,10 +36,10 @@
       <hbox flex>
         <hbox flex>
           <MailAutocomplete addresses={mail.to} placeholder={$t`Add recipient`} tabindex={1} autofocus={mail.to.isEmpty}>
-            <!--<svelte:fragment slot="person-popup-buttons" let:personUID>
-              <Button plain label={$t`CC`} onClick={() => onMoveToCC(personUID)} />
-              <Button plain label={$t`BCC`} onClick={() => onMoveToBCC(personUID)} />
-            </svelte:fragment>-->
+            <svelte:fragment slot="person-popup-buttons" let:person>
+              <Button plain label={$t`CC`} onClick={() => onMoveToCC(person)} />
+              <Button plain label={$t`BCC`} onClick={() => onMoveToBCC(person)} />
```
jermy-c, line 41: Because the slot takes a `person` prop in `PersonsAutoComplete`. It should have the same prop name also.

app/frontend/Shared/PersonAutocomplete/PersonEntry.svelte
```diff
@@ -28,7 +28,7 @@
   import PersonPopup from "./PersonPopup.svelte";
   import PersonPicture from "../Person/PersonPicture.svelte";
   import Popup from "../Popup.svelte";
-  import { createEventDispatcher } from 'svelte';
+  import { createEventDispatcher, onMount } from 'svelte';
   const dispatchEvent = createEventDispatcher<{ focusNext: void }>();
 
   export let person: PersonUID;
```
jermy-c, line 34: Here is the proper prop name.

Issue #521, opened 2025-03-27
Invitations: Time doesn't show


benbucksch:
### Environment
* Mustang master 839e2e44
* Neil@beonex
* EWS

### Reproduction
* Open email "Owl Activesync 128 from 115" in the inbox. It contains the invitations cited below.

### Actual result
* Invitation shows, with title and participants, and confirmation buttons
* Start and end time is missing

![Image](https://github.com/user-attachments/assets/0d4ccd21-4d2f-43a7-9b86-38da8fcd0730)

### Expected result
* Invitation shows, with title and participants, and confirmation buttons
* Start time and duration (1 hour) shows

### Data
```
BEGIN:VCALENDAR
METHOD:REQUEST
PRODID:Microsoft Exchange Server 2010
VERSION:2.0
BEGIN:VTIMEZONE
TZID:W. Europe Standard Time
BEGIN:STANDARD
DTSTART:16010101T030000
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=-1SU;BYMONTH=10
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:16010101T020000
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=-1SU;BYMONTH=3
END:DAYLIGHT
END:VTIMEZONE
BEGIN:VEVENT
ORGANIZER;CN=Ben Bucksch:mailto:ben@beonex.foo
ATTENDEE;ROLE=REQ-PARTICIPANT;PARTSTAT=NEEDS-ACTION;RSVP=TRUE;CN=Neil Rashb
 rook:mailto:neil@beonex.foo
DESCRIPTION;LANGUAGE=en-GB:  *   ssgdsgdsg\n
UID:040000008200E00074C5B7101A82E00800000000FB63F76F978EDB01000000000000000
 010000000C6F10602BDD1AA4C9719CA7E15D8742F
SUMMARY;LANGUAGE=en-GB:Owl ActiveSync 128 from 115
DTSTART;TZID=W. Europe Standard Time:20250305T104500
DTEND;TZID=W. Europe Standard Time:20250305T114500
CLASS:PUBLIC
PRIORITY:5
DTSTAMP:20250306T125820Z
TRANSP:OPAQUE
STATUS:CONFIRMED
SEQUENCE:0
[...]

PR #531, opened 2025-03-31
Parse Windows zones in iMIP messages #521

NeilRashbrook:
I also created separate mapping files to simplify the logic.

I had to change the country code I was using to detect the UTC offset. I had used one that works in Node but it doesn't work in Electron. Sigh...


benbucksch:
> I also created separate mapping files to simplify the logic.

Great!

Could you please rename it from FooZone to FooTimezone? File names and variable names.

benbucksch (changes requested):
Looks all good, apart from the 'Zone ;-)

NeilRashbrook:
Renamed as requested.

benbucksch (approved):

PR #536, opened 2025-04-01
Don't try to unset the recurrence of an instance

NeilRashbrook:
Exchange Online EWS doesn't like it any more.

Obviously I didn't want to break any of the back end for handling recurring events, but I basically had to rewrite parts of the UI in order to get it into a position where I could test the change to make sure that, were the UI to work, the back end would actually do the right thing.

I was also disappointed to find that the UI to set the end date of a recurring event was never restored.

### DIFF
--- a/app/logic/Calendar/EWS/EWSEvent.ts
+++ b/app/logic/Calendar/EWS/EWSEvent.ts
@@ -154,7 +154,9 @@ export class EWSEvent extends Event {
     request.addField("CalendarItem", "Body", this.descriptionHTML ? { BodyType: "HTML", _TextContent_: this.descriptionHTML } : this.descriptionText ? { BodyType: "Text", _TextContent_: this.descriptionText } : "", "item:Body");
     request.addField("CalendarItem", "ReminderIsSet", this.alarm != null, "item:ReminderIsSet");
     request.addField("CalendarItem", "ReminderMinutesBeforeStart", this.alarmMinutesBeforeStart(), "item:ReminderMinutesBeforeStart");
-    request.addField("CalendarItem", "Recurrence", this.recurrenceRule ? this.saveRule(this.recurrenceRule) : null, "calendar:Recurrence");
+    if (!this.parentEvent) { // Exchange Online now requires this for EWS
+      request.addField("CalendarItem", "Recurrence", this.recurrenceRule ? this.saveRule(this.recurrenceRule) : null, "calendar:Recurrence");
+    }
     if (this.calUID && !this.itemID && !this.parentEvent) {
       // This probably only makes sense when creating an event.
       // (And it's not even needed then as Exchange will auto-generate one.)
--- a/app/logic/Calendar/EWS/EWSEvent.ts
+++ b/app/logic/Calendar/EWS/EWSEvent.ts
@@ -154,7 +154,7 @@ export class EWSEvent extends Event {
     request.addField("CalendarItem", "Body", this.descriptionHTML ? { BodyType: "HTML", _TextContent_: this.descriptionHTML } : this.descriptionText ? { BodyType: "Text", _TextContent_: this.descriptionText } : "", "item:Body");
     request.addField("CalendarItem", "ReminderIsSet", this.alarm != null, "item:ReminderIsSet");
     request.addField("CalendarItem", "ReminderMinutesBeforeStart", this.alarmMinutesBeforeStart(), "item:ReminderMinutesBeforeStart");
-    if (!this.parentEvent) { // Exchange Online now requires this for EWS
+    if (!this.parentEvent) { // Exchange Online requires not to write the `Recurrence` prop for recurrence instances
       request.addField("CalendarItem", "Recurrence", this.recurrenceRule ? this.saveRule(this.recurrenceRule) : null, "calendar:Recurrence");
     }
     if (this.calUID && !this.itemID && !this.parentEvent) {
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -34,7 +34,17 @@ export class Event extends Observable {
   /** Only used by the editing UI. */
   @notifyChangedProperty
   repeat = false; // TODO remove
-  /** Only used by recurring masters, describes the recurrence pattern. */
+  /** Only used by recurring masters, describes the recurrence pattern.
+   *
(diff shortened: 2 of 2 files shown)


benbucksch:
I don't doubt the fix, but I don't understand it. Could you please help me understand?

without `Event.parentEvent`, without `Event.recurrenceRule`: Normal event
without `Event.parentEvent`, with `Event.recurrenceRule`: Recurrence master
with `Event.parentEvent`, with `Event.recurrenceRule`: Recurrence instance
with `Event.parentEvent`, without `Event.recurrenceRule`: Recurrenc exception

Is the above correct? (If not, please correct me.) If so, could you please
a) document that in the `Event` API, above either one of the property definitions?
b) explain for all 4 cases what is expected to be done here in this function, any why?

benbucksch:
Is this a recent server-side change at Office365? Do we need the same change for Owl and ExQuilla?

benbucksch:
> I was also disappointed to find that the UI to set the end date of a recurring event was never restored.

Yes, sorry, that was a deliberate decision to not include that.

NeilRashbrook:
> with `Event.parentEvent`, with `Event.recurrenceRule`: Recurrence instance
> with `Event.parentEvent`, without `Event.recurrenceRule`: Recurrence exception

These are incorrect; both instances and exceptions have a `parentEvent` but no `recurrenceRule` (it shouldn't be possible to have both).

> Is this a recent server-side change at Office365? Do we need the same change for Owl and ExQuilla?

I assume it's recent because the code used to work. ExQuilla doesn't support calendar. Owl does of course, but fortunately Lightning saves the before and after state so Owl only sends the fields that change (and in the case of an instance, that never includes the recurrence rule of course).

> Yes, sorry, that was a deliberate decision to not include that.

So what's the new UI, create it with no end, then the user has to find the first one they don't want and say "Delete this and all future occurences"?

benbucksch:
> create it with no end, then the user has to find the first one they don't want and say "Delete this and all future occurrences"? (Which only helps in one use case.)

I filed #542 about the "end date" discussion.

benbucksch:
I still have troubles mapping the state with the expected server API, and with the code. Could you please help me understand?

* without `Event.parentEvent`, without `Event.recurrenceRule`: Normal event
* without `Event.parentEvent`, with `Event.recurrenceRule`: Recurrence master
* with `Event.parentEvent`, with `Event.recurrenceRule`: Invalid state
* with `Event.parentEvent`, without `Event.recurrenceRule`: Recurrenc instance or exception

Is the above correct? Could you please
a) document that in the `Event` API, above either one of the property definitions?
b) explain for all 4 cases: 1. the EWS API, 2. what is expected to be done here in this function, any 3. why?

benbucksch:
I still have troubles mapping the state with the expected server API, and with the code. Could you please help me understand?

| `Event.parentEvent` | `Event.recurrenceRule` | Event type
| - | - | -
| without | without | Normal event
| without | with | Recurrence master
| with | with | Invalid state
| with | without | Recurrenc instance or exception

Is the above correct? Could you please
a) document that in the `Event` API, above either one of the property definitions?
b) explain for all 4 cases: 1. the EWS API, 2. what is expected to be done here in this function, any 3. why?

NeilRashbrook:
> (table)
> 
> Is the above correct?

Yes, it is correct.

> Could you please a) document that in the `Event` API, above either one of the property definitions?

`parentEvent` is already documented to link back to the recurring master from an instance, but I can add some more text for the properties.

> b) explain for all 4 cases:

Well, those aren't the cases. The cases are as follows:
1. Editing a single event
2. Changing a single event into a recurring master
3. Changing a recurring master into a single event
4. Editing a recurring master
5. Editing an (occurrence into an) exception

The back end can't distinguish between 2. and 4. - all it knows is that the event is now recurring. So in these cases it has to set the recurrence field to the (new) recurrence rule.

It also can't distinguish between 1. and 3. What we used to do, and still do in OWA and ActiveSync, is to clear the recurrence field. This also works for case 5 and EWS on Exchange 2019 and earlier.

The problem is that Office 365 now doesn't like us clearing the empty recurrence field of an exception, so now we need to special-case case 5.

app/logic/Calendar/EWS/EWSEvent.ts
```diff
@@ -154,7 +154,9 @@ export class EWSEvent extends Event {
     request.addField("CalendarItem", "Body", this.descriptionHTML ? { BodyType: "HTML", _TextContent_: this.descriptionHTML } : this.descriptionText ? { BodyType: "Text", _TextContent_: this.descriptionText } : "", "item:Body");
     request.addField("CalendarItem", "ReminderIsSet", this.alarm != null, "item:ReminderIsSet");
     request.addField("CalendarItem", "ReminderMinutesBeforeStart", this.alarmMinutesBeforeStart(), "item:ReminderMinutesBeforeStart");
-    request.addField("CalendarItem", "Recurrence", this.recurrenceRule ? this.saveRule(this.recurrenceRule) : null, "calendar:Recurrence");
+    if (!this.parentEvent) { // Exchange Online now requires this for EWS
```
benbucksch, line 157:
```suggestion
    if (!this.parentEvent) { // Exchange Online requires not to write the `Recurrence` prop for recurrence instances
```

app/logic/Calendar/Event.ts
```diff
@@ -31,12 +31,13 @@ export class Event extends Observable {
   /** IANA timezone name, e.g. "Europe/Berlin" */
   @notifyChangedProperty
   timezone: string | null = null;
+  /** Only used by the editing UI. */
   @notifyChangedProperty
   repeat = false; // TODO remove
-  /** If `repeat` is set, should be the rule which describes the pattern. */
+  /** Only used by recurring masters, describes the recurrence pattern. */
```
benbucksch, line 37:
```suggestion
  /** Only used by recurring masters, describes the recurrence pattern.
   *
   * | `Event.parentEvent` | `Event.recurrenceRule` | Event type
   * | ------------------- | ---------------------- | -----
   * | without             | without                | Normal event
   * | without             | with                   | Recurrence master
   * | with                | with                   | Invalid state
   * | with                | without                | Recurrence instance or exception
   */
```

benbucksch (approved):

benbucksch (approved):

Issue #486, opened 2025-03-13
Invitation: Move respondToInvitation() from EMail to Calendar/


benbucksch:
This is about the refactoring from the last commit on PR #461, now git commit abc66c72 on branch `neil/refac`.
This ticket is about fixing this branch.

One known bug is:

### Env
* neil@beonex account
* EWS

### Reproduction
1. Start Mustang, go to Email
2. Click on email "Owl ActiveSync 128 from 115" or any of the other emails with an open (not responded) invitation
3. Click on [Accept]

### Actual result
After step 2 (view only):
* You see an invitation
* You see error "Must have event to find calendar"
* You see the error message twice
* ![grafik](https://github.com/user-attachments/assets/134bf503-df62-4888-807c-4aded0bc607f)
* No error on console (!)

After step 3:
* `selectedCalendar` is undefined
```
error.ts:6 TypeError: Cannot read properties of undefined (reading 'respondToInvitation')
    at respond (Invitation.svelte:59:28)
    at onAccept (Invitation.svelte:62:11)
    at HTMLButtonElement.myOnClick (Button.svelte:79:13)
```
* This is on line: `await selectedCalendar.respondToInvitation(message, response);` . That means `selectedCalendar` is undefined.

### Expected result
* All errors are on the local error console, and are also logged to server
* No error during display
* No error during [Accept]
* Response is processed.

_Originally posted by @benbucksch in https://github.com/mustang-im/mustang/issues/461#issuecomment-2719719466_

benbucksch:
The entire code to load calendars in Invitation.svelte is iffy.

NeilRashbrook:
The error message is displayed on the page because of the block:
```
    {:catch ex}
      {ex?.message}
      {ex}
    {/await}
 ```

`message.loadEvent()` is supposed to, well, load the `event`, so I'm not sure offhand why the event is missing at this point.

As for the selected calendar, it's not getting set because of the previous error.

NeilRashbrook:
I've pushed commit a32cd92 to branch `neil/refac` just to see what effect it has for you.

benbucksch:
Is there a particular reason why you load the calendars **after** loading the message event?

NeilRashbrook:
It's future-proofing for processing replies and cancellations.

benbucksch:
Ah, OK, so we don't need the calendars for displaying the events, but for replying? That makes sense. That is a good thing to put as code comment, above the `loadCalendars();` call.

[...]

PR #543, opened 2025-04-07
Send invitations when creating meetings in the local calendar

NeilRashbrook:
Also for when creating meetings in the ActiveSync calendar.

Still to do:
- [X] Send invitations
  - [X] IMAP*
  - [X] ActiveSync*
  - [X] EWS
  - [X] OWA
- [X] Recognise invitation email
- [X] Display invitation data
- [X] Accept or decline invitation
- [X] Recognise invitation event
- [X] Change invitation acceptance
- [X] Recognise acceptance email
- [ ] Update meeting participation (update my calendar event with the response from the participants)
  - [ ] IMAP
  - [X] ActiveSync
  - [X] EWS
  - [X] OWA

Note:
\* Mustang has no way of cancelling individiual attendees


NeilRashbrook (commented):
This PR is all about outgoing invitations which is why there is very little code relevant to incoming invitations and I don't think it makes sense to scope creep it to include #486 in it.

Anyway I've added a commit which addresses some of your review comments.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -111,12 +111,23 @@ export class ActiveSyncEvent extends Event {
...
+    let organizer;
+    if (this.participants.length) {
+      organizer = this.participants.find(participant => participant.emailAddress == this.calendar.account.emailAddress);
+      if (organizer) {
+        organizer.response = ResponseType.Organizer;
+      } else {
+        organizer = new Participant(this.calendar.account.emailAddress, null, ResponseType.Organizer);
+        this.participants.add(organizer);
+      }
+    }
```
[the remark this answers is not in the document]
benbucksch: We need to save that information, on disk.
If you make OutgoingInvitation a class, and store a reference to it in `Event`, and store the data of the invitation in the CalendarEvent DB table, then you should not have to re-find the organizer.

benbucksch:
> Could we please clearly separate IncomingInvitation and OutgoingInvitation, as discussed?
> Here in this PR, I would expect a new class OutgoingInvitation, which is an optional property of the Event object and exists IFF (if and only if) our user is the organizer.

And the OutgoingInvitation data is saved in DB together with the event, as part of the JSON column.

Can you please adapt the PR to use that design? That makes the distinction between event, incoming and outgoing invitation much clearer and the code easier to ready.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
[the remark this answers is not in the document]
NeilRashbrook: We do save that information, but that's unhelpful, because this is an event that needs to be saved, so even if it was an existing event that we saved previously, that saved information will be out of date anyway.

app/logic/Calendar/OutgoingActions.ts
```diff
@@ -0,0 +1,45 @@
+import type { Event } from "./Event";
+import type { Participant } from "./Participant";
+import { ResponseType, type iCalMethod } from "./Invitation";
+import type { MailAccount } from "../Mail/MailAccount";
+import type { PersonUID } from "../Abstract/PersonUID";
+
+export default class OutgoingActions {
```
benbucksch, line 7: Can you please rename this to `OutgoingInvitation` and move it to app/logic/Calendar/Invitation/OutgoingInvitation.ts ?

Please also separate the `ResponseType` enum into `IncomingInvitationResponse` and `OutgoingInvitationResponse`.

benbucksch (changes requested):

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -111,12 +111,25 @@ export class ActiveSyncEvent extends Event {
   }
 
   async saveToServer(): Promise<void> {
+    let organizer;
+    if (this.participants.length && this.response <= ResponseType.Organizer) {
+      organizer = this.participants.find(participant => participant.emailAddress == this.calendar.account.emailAddress);
```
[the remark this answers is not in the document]
NeilRashbrook: Line 115 checks that you organised the meeting; Line 116 wants wants to ensure that you appear as an organiser participant on your own meeting.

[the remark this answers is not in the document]
benbucksch: Line 115: The `<=` is confusing me. It should be `==`
Line 116: What if the email address is not the one you expect / is on the account? We have multiple identities, etc. pp.. Also, I want to remove `account.emailAddress` entirely.

app/logic/Calendar/OutgoingActions.ts
benbucksch: Wherever possible, separate the status into incoming and outgoing, and put them into IncomingInvitation.ts and `OutgoingInvitation.ts`

For the rest, move app/logic/Calendar/Invitation.ts to app/logic/Calendar/Invitation/InvitationStatus.ts

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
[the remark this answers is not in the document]
NeilRashbrook: Ah yes, back when this was written, `<=` was correct, but as of commit 0328b31, I think it might be possible to assume that `==` will work.
Exchange doesn't support organising meetings with alternate identities, except via shared calendering.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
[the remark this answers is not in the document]
NeilRashbrook: Which of these do you want to support?
- Changing an appointment to a meeting
- Changing a meeting to an appointment
- Changing the attendee list of a meeting
- Handling the user doing any of the above using another client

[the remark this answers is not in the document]
benbucksch: I understand where you're aiming at.

However, as far as possible in Exchange, I would like to use an explicit flag - independent from the participants list - that a meeting is organized by me (our user). Does Exchange give that?

Whatever that flag is, once set, I do not think it's legitimate to remove it.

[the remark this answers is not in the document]
NeilRashbrook: I've only ever been talking about meetings organised by the user, so I'm failing to see the relevance of your last question. But to answer your previous question, yes, I don't see how you can guarantee that a particular participant can never be removed from an event, so you have to assume that you have to check for it each time.

NeilRashbrook:
(I had to rebase because the file move conflicted with #570)

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -115,12 +115,25 @@ export class ActiveSyncEvent extends Event {
   }
 
   async saveToServer(): Promise<void> {
+    let organizer;
```
benbucksch, line 118:
```suggestion
    let organizer: Participant;
```

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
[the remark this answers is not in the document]
benbucksch: I still don't understand this code. Here's what's confusing me:

1. `this.response == InvitationResponse.Organizer`

How is me being the organizer a **`response`**? If I'm the organizer, I am sending out. "response" is the wrong direction, logically. Very confusing.

2. Why is this re-adding me as participant? Are you trying to make up for our UI that fails to add myself as participant to meetings I organize? If so, shouldn't we rather fix the UI?
Or are you making up for other clients that fail to do so? Or clients that remove it?

4. What about meetings that I organize, but I am not myself attending the meeting?

app/logic/Calendar/Event.ts
```diff
@@ -194,6 +198,29 @@ export class Event extends Observable {
...
+    return new OutgoingInvitation(this);
+  }
+
+  protected participantMe(): { identity: MailIdentity, participant: Participant, person: PersonUID } {
+    let results = [];
+    for (let account of appGlobal.emailAccounts) {
+      if (!account.canSendInvitations) {
+        continue;
+      }
+      for (let identity of account.identities) {
```
benbucksch, line 211: Use `findIdentityForEMailAddress()`

app/logic/Calendar/Event.ts
```diff
@@ -194,6 +198,29 @@ export class Event extends Observable {
...
+        for (let participant of this.participants) {
+          if (identity.isEMailAddress(participant.emailAddress)) {
+            let person = new PersonUID(participant.emailAddress, identity.realname);
+            results.push({ identity, participant, person });
```
benbucksch, line 215: Why not directly `return {}` here? Given that you return the first only anyway?

benbucksch (changes requested):
Almost there.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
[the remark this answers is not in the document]
NeilRashbrook: 1. `this.response` is just a shorthand for `this.participantMe().response`.
2. I didn't know it was the job of the UI to do that
3. How is that represented in Mustang, in particular, how would I know which of your identities is the organiser?

app/logic/Calendar/Event.ts
NeilRashbrook: That searches accounts which I can't use.

app/logic/Calendar/Event.ts
NeilRashbrook: Only if it's unique.

benbucksch: I thought you just wanted fail, if there are none found.
You're actually also failing, if there are several matching.
I am not sure what the caller is supposed to do with that error.

app/logic/Calendar/Event.ts
benbucksch: Here's what I would do:
* Given that you need it only for outgoing invitations, move the function to the `OutgoingInvitation` class
* Then you can search for the participant which is the organizer. And you can rename it to `organizerParticipant()`
* Validate this result by check that the email address of this participant belongs to our user. You can use `findIdentityForEMailAddress()` for that. It's only a validation, a double-check, to ensure that we won't take over a meeting organized by somebody else. The check could also be elsewhere.

benbucksch:
TODO: We need to add our user as participant with Organizer flag to the meeting. The user won't add himself.
We should do that in the logic, in the `Event` class (on save, if this is a new meeting and it has participants) or even better in the `OutgoingInvitation` class.

app/logic/Calendar/Event.ts
```diff
@@ -194,6 +198,29 @@ export class Event extends Observable {
     this.response = original.response;
   }
 
+  get outgoingInvitation() {
+    return new OutgoingInvitation(this);
```
benbucksch, line 202: You cannot have a getter create new objects on every call.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -176,6 +189,15 @@ export class ActiveSyncEvent extends Event {
         throw new ActiveSyncError("Sync", response.Responses.Delete.Status, this.calendar);
       }
     }
+    if (this.response == InvitationResponse.Organizer) {
+      await this.outgoingInvitation.sendCancellations(this.calendar.account);
+    } else if (this.response) {
+      for (let participant of this.participants) {
+        if (participant.response == InvitationResponse.Organizer) {
+          await this.sendInvitationResponse(InvitationResponse.Decline, this.calendar.account);
+        }
```
benbucksch, line 198: These lines should be on a class `IncomingInvitation`, but class is created in #550, correct? Can you please do the change there in that PR, once this PR here lands?

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -176,6 +189,15 @@ export class ActiveSyncEvent extends Event {
...
+      }
```
benbucksch, line 199: These lines should be on a class `IncomingInvitation`, but class is created in #550, correct? Can you please do the change there in that PR, once this PR here lands?

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -120,25 +120,18 @@ export class ActiveSyncEvent extends Event {
...
-    }
     // Not supporting tasks for now.
     if (this.parentEvent) {
       this.parentEvent.saveFields(this.parentEvent.toFields(this.toFields()));
     } else {
       await this.saveFields(this.toFields());
+      if (!this.calUID) {
+        // If we haven't set a UID yet, Exchange will auto-generate one,
+        // so sync up to find out what it is.
+        await this.calendar.listEvents();
```
benbucksch, line 131: How do we match this event with the one that's coming from the server, given that we don't have the UID to match it?

NeilRashbrook:
This seems to still work. I haven't "optimises" the `outgoingInvitation` getter, since I don't have a good example of what you want here. I also didn't implement `organizerParticipant()` since I didn't notice it until now; I can add it if you really want.

(This would be correct code in C++, since there the compiler will simply optimise out a temporary instance.)

app/logic/Calendar/Event.ts
```diff
@@ -230,6 +236,12 @@ export class Event extends Observable {
     this.unedited = null;
   }
 
+  createMeeting() {
+    assert(this.participants.isEmpty, "This is already a meeting");
+    assert(this.myParticipation <= InvitationResponse.Organizer, "Incoming invitation should already have participants");
```
benbucksch, line 241:
```suggestion
    assert(this.myParticipation == InvitationResponse.Organizer, "Incoming invitation should already have participants");
```

Please do not use `<=` et al at all for enums.

Use `||` or `!` or `IncomingStates.includes(this.myParticipation)` or similar as necessary.

app/logic/Calendar/Event.ts
```diff
@@ -248,7 +260,13 @@ export class Event extends Observable {
   }
 
   async saveToServer(): Promise<void> {
-    // nothing to do for local events
+    if (!this.participants.length || this.myParticipation > InvitationResponse.Organizer) {
```
benbucksch, line 263: ditto

app/logic/Calendar/Event.ts
```diff
@@ -276,7 +294,18 @@ export class Event extends Observable {
...
 
   async deleteFromServer(): Promise<void> {
-    // nothing to do for local events
+    if (!this.participants.length) {
+      return;
+    }
+    if (this.myParticipation == InvitationResponse.Organizer) {
+      await this.outgoingInvitation.sendCancellations();
+    } else if (this.myParticipation) {
+      for (let participant of this.participants) {
```
benbucksch, line 303:
```suggestion
      // TODO Move code to `IncomingInvitation` class
      for (let participant of this.participants) {
```

app/logic/Calendar/Event.ts
```diff
@@ -302,13 +331,29 @@ export class Event extends Observable {
   }
 
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
-    assert(this.response > InvitationResponse.Organizer, "Only invitations can be responded to");
+    assert(this.myParticipation > InvitationResponse.Organizer, "Only invitations can be responded to");
```
benbucksch, line 334: ditto

app/logic/Calendar/Event.ts
```diff
@@ -302,13 +331,29 @@ export class Event extends Observable {
...
     let accounts = appGlobal.emailAccounts.contents.filter(account => account.canSendInvitations && this.participants.some(participant => participant.response != InvitationResponse.Organizer && participant.emailAddress == account.emailAddress));
     assert(accounts.length == 1, "Failed to find matching account for invitation");
     let participant = this.participants.find(participant => participant.emailAddress == accounts[0].emailAddress);
     participant.response = response;
     await this.save();
-    await accounts[0].sendInvitationResponse(this, response);
+    await this.sendInvitationResponse(response, accounts[0]);
+  }
+
+  async sendInvitationResponse(response: InvitationResponseInMessage, account: MailAccount) {
```
benbucksch, line 343:
```suggestion
  // TODO Move code to `IncomingInvitation` class
  async sendInvitationResponse(response: InvitationResponseInMessage, account: MailAccount) {
```

app/logic/Calendar/Event.ts
```diff
@@ -302,13 +331,29 @@ export class Event extends Observable {
   }
 
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
```
benbucksch, line 333:
```suggestion
  // TODO Move code to `IncomingInvitation` class
  async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
```

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  createOrganizer() {
+    let identity = this.identity;
+    this.event.participants.add(new Participant(identity.emailAddress, identity.realname, InvitationResponse.Organizer));
+    this.event.myParticipation = InvitationResponse.Organizer;
+  }
+
+  get identity(): MailIdentity {
+    let account = appGlobal.emailAccounts.find(account => account.canSendInvitations);
+    assert(account, gt`No suitable identity to organise this meeting`);
```
benbucksch, line 25:
```suggestion
    assert(account, gt`No suitable email identity to organise this meeting`);
```

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -322,7 +327,7 @@ export class OWAEvent extends Event {
   }
 
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
-    assert(this.response > InvitationResponse.Organizer, "Only invitations can be responded to");
+    assert(this.myParticipation > InvitationResponse.Organizer, "Only invitations can be responded to");
```
benbucksch, line 330: ditto

app/logic/Calendar/ActiveSync/ActiveSyncOutgoingInvitation.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { EMail } from "../../Mail/EMail";
+
+export default class ActiveSyncOutgoingInvitation extends OutgoingInvitation {
+  event: ActiveSyncEvent;
+
+  get identity(): MailIdentity {
+    return this.event.calendar.account.identities.first;
+  }
+
+  protected newEmailFrom(): EMail {
```
benbucksch, line 13: Why does only ActiveSync override this function, but not OWA and EWS?

benbucksch (changes requested):

app/logic/Calendar/ActiveSync/ActiveSyncOutgoingInvitation.ts
benbucksch: Neil says: Because for the others, we are not sending emails

benbucksch (approved):

benbucksch:
@NeilRashbrook Can you please fix the enum comparisons in another PR?

NeilRashbrook:
> @NeilRashbrook Can you please fix the enum comparisons in another PR?

This was done in #586.

Issue #478, opened 2025-03-10
Calendar: Recurrence: Create exceptions: UI buttons


benbucksch:
### Reproduction
1. Create a repeated event, once per week
2. The second week should be changed to a different time, or have one participant removed, or be different in some other way
3. All other instances stay as they are
4. All events of the series **from now on** should change to another day, or have one more participant
5. All events of the series (including past events) should change to another day, or have one more participant

### Actual result
* Steps 1-3: OK
* Steps 4 and 5: UI doesn't offer a way to do that

### Expected result
* Ask Elena
* Options:
  * Make a radio box somewhere in the Repeat box that shows whether to change only this event, or all future events, or all events.
  * Create multiple [Save] buttons
  * Elena's ideas

benbucksch:
@elenaux Need UX here
Could you please give some ideas, next week?

elenaux:
Here my idea for now

![Image](https://github.com/user-attachments/assets/0298b4f6-7836-473d-9939-f6b689f41c66)

![Image](https://github.com/user-attachments/assets/047c554f-6396-42dd-8849-ec461dad30ef)

benbucksch:
Breaking the series:

Neil wrote:
> Implementation note: This is actually achieved by truncating the old recurrence and creating a new recurring event.

Yes, makes sense. Make it so! :-)

Elena wrote:
> I would't care too much about previous events (n.5 of your list)

Agreed that it doesn't make sense to change the past. No point in asking the user about it, either.

benbucksch:
About exceptions of the old series that are in the future:

We do not delete them, because they might still be valid.
I see 3 optoins:
* We keep them on the old series. Problem: In the specs, if there are exceptions that are after the end of a series, are they still valid? Would other calendar apps render them?
* We could turn them into one-time events, by deleting their reference to the master. It means that I have 2 events on that week: The old exception, and the new series. The user will hopefully notice the duplicate when it comes to that week, and delete the unneeded event. This is likely the safest option.
* We make them exceptions of the new series. However, if the exception is indeed no longer valid, then the user might miss the meeting at the new time. So, I don't think that's a good idea.

benbucksch:
@elenaux The [Delete] button also needs a Chevron and choice between this event only and the whole series. I wouldn't put a label on this, to avoid repeating the same label twice, but just the Trash icon and the Chevron. The Dropdown options then say "Delete only this event" and "Delete the whole series", both with Trash icon.

[...]

PR #544, opened 2025-04-09
Allow future events of a series to be changed #478

NeilRashbrook:
- After deleting all other instances of a series, deleting the last instance deletes the series
- If editing the first instance of a series, provides button to delete entire series
- If editing an instance in the middle of a series, provides button to delete remainder of series
- If editing the first instance of a series, allows recurring master to be replaced with current event
- If editing an instance in the middle of a series, allows a new series to be started from this point (truncates old series)

Notes:
- When replacing part or all of a series, you must specify the recurrence again, as the start date may have changed, invalidating the previous recurrence
- For some reason, after converting a series back to a single event, the UI gets confused and needs a restart to be able to convert the event back to a series

### DIFF
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -152,7 +152,7 @@
       event.startTime.getTime() <= event.endTime.getTime() && $event.hasChanged();
   $: seriesStatus = event.seriesStatus;
 
-  function confirmAndChangeRule(): boolean {
+  function confirmAndChangeRecurrenceRule(): boolean {
     let master = event.parentEvent || event;
     if (!repeatBox) {
       if (!master.recurrenceRule) {
@@ -207,11 +207,11 @@
   }
 
   async function onChangeAll() {
-    if (!confirmAndChangeRule()) {
+    if (!confirmAndChangeRecurrenceRule()) {
       return;
     }
     let master = event.parentEvent;
-    master.copyFrom(event);
+    master.copyEditableFieldsFrom(event);
     await master.saveToServer();
     await master.save();
     onClose();
@@ -219,7 +219,7 @@
 
   async function onChangeForward() {
     let master = event.calendar.newEvent();
-    master.copyFrom(event);
+    master.copyEditableFieldsFrom(event);
     master.recurrenceRule = repeatBox.newRecurrenceRule();
     master.recurrenceCase = RecurrenceCase.Master;
     master.fillRecurrences(new Date(Date.now() + 1e11));
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -23,6 +23,25 @@
             border={false}
             iconSize="16px"
             />
+          {#if seriesStatus == "first"}
+            <RoundButton
+             label={$t`Delete entire series`}
+             icon={DeleteIcon}
+             onClick={onDeleteAll}
+             classes="plain delete"
+             border={false}
+             iconSize="16px"
+             />
+          {:else if seriesStatus == "middle"}
+            <RoundButton
+             label={$t`Delete remainder of series`}
+             icon={DeleteIcon}
+             onClick={onDeleteForward}
+             classes="plain delete"
+             border={false}
+             iconSize="16px"
+             />
+          {/if}
         {/if}
       </hbox>
       <hbox class="account-icon">
@@ -49,6 +68,27 @@
             />
         {/if}
         {#if canSave}
+          {#if seriesStatus == "first"}
+            <RoundButton
+             label={$t`Change entire series`}
+             icon={SaveIcon}
+             onClick={onChangeAll}
+             classes="plain save-or-close"
+             filled={true}
+             iconSize="16px"
+             />
+          {:else if seriesStatus == "middle"}
+            <RoundButton
+             label={$t`Change remainder of series`}
+             icon={SaveIcon}
+             onClick={onChangeForward}
+             classes="plain save-or-close"
+             filled={true}
+             iconSize="16px"
+             />
+          {/if}
+        {/if}
+        {#if canSave && !(event.parentEvent && repeatBox)}
           <RoundButton
             label={$t`Revert`}
             icon={RevertIcon}
@@ -79,7 +119,7 @@
 </vbox>
 
 <script lang="ts">
-  import type { Event } from "../../../logic/Calendar/Event";
+  import { type Event, RecurrenceCase } from "../../../logic/Calendar/Event";
   import { Calendar } from "../../../logic/Calendar/Calendar";
   import { Account } from "../../../logic/Abstract/Account";
   import { EventEditMustangApp, calendarMustangApp } from "../CalendarMustangApp";
@@ -111,6 +151,50 @@
   $: canSave = event && $event.title && $event.startTime && $event.endTime &&
       event.startTime.getTime() <= event.endTime.getTime() && $event.hasChanged();
   $: oldTitle = event?.title || $t`Event`;
+  $: seriesStatus = isInstance(event);
+
+  function isInstance(event) {
+    let master = event.parentEvent;
+    if (!master?.recurrenceRule) {
+      return "none";
+    }
+    let pos = master.instances.indexOf(event);
+    let isFirst = master.instances.getIndexRange(0, pos).every(instance => instance === null);
+    let rule = master.recurrenceRule;
+    let isLast = (rule.count != Infinity || rule.endDate) && master.instances.contents.slice(pos + 1).every(instance => instance === null || instance?.dbID) && !rule.getOccurrenceByIndex(master.instances.length + 1);
+    return isLast ? isFirst ? "only" : "last" : isFirst ? "first" : "middle";
+  }
+
+  function confirmAndChangeRule(): boolean {
+    let master = event.parentEvent || event;
+    if (!repeatBox) {
+      if (!master.recurrenceRule) {
+        // Event had never been a recurring event.
+        return true;
+      }
+      if (!confirm($t`Are you sure you want to remove this unfortunate series of events?`)) {
+        return false;
+      }
+      master.recurrenceRule = null;
+      master.recurrenceCase = RecurrenceCase.Normal;
+    } else {
+      let rule = repeatBox.newRecurrenceRule();
+      if (master.recurrenceRule) {
+        if (event.startTime.getTime() == master.recurrenceRule.startDate.getTime() &&
+            rule.getCalString() == master.recurrenceRule.getCalString()) {
+          // Rule hasn't actually changed.
+          return true;
+        }
+        if (!confirm($t`This change will reset all of your series to default values.`)) {
+          return false;
+        }
+      }
+      master.recurrenceRule = rule;
+      master.recurrenceCase = RecurrenceCase.Master;
+    }
+    master.clearExceptions();
+    return true;
+  }
 
   function onCancel() {
     assert(event.unedited, "need unedited state");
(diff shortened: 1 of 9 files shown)


app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -102,10 +142,54 @@
   $: canSave = event && $event.title && $event.startTime && $event.endTime &&
       event.startTime.getTime() <= event.endTime.getTime();
   $: oldTitle = event?.title || $t`Event`;
+  $: seriesStatus = isInstance(event);
+
+  function isInstance(event) {
```
benbucksch, line 147: This function should be on the `Event` class in logic, not in the UI

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -102,10 +142,54 @@
...
+      }
+      if (!confirm($t`Are you sure you want to remove this unfortunate series of events?`)) {
+        return false;
+      }
+      master.recurrenceRule = null;
+    } else {
+      let rule = repeatBox.newRecurrenceRule();
+      if (master.recurrenceRule) {
+        if (event.startTime.getTime() == master.recurrenceRule.startDate.getTime() &&
+            rule.getCalString() == master.recurrenceRule.getCalString()) {
```
benbucksch, line 174: This check should be in the `Event` class.

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -102,10 +142,54 @@
...
+    let rule = master.recurrenceRule;
+    let isLast = (rule.count != Infinity || rule.endDate) && master.instances.contents.slice(pos + 1).every(instance => instance === null || instance?.dbID) && !rule.getOccurrenceByIndex(master.instances.length + 1);
+    return isLast ? isFirst ? "only" : "last" : isFirst ? "first" : "middle";
+  }
+
+  function confirmAndChangeRule(): boolean {
+    let master = event.parentEvent || event;
+    if (!event.repeat) {
+      if (!master.recurrenceRule) {
+        // Never a recurring event.
```
benbucksch, line 163: I don't understand. 'Never what? a recurring event.'

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -118,14 +202,74 @@
     onClose();
   }
 
+  async function onChangeAll() {
+    if (!confirmAndChangeRule()) {
+      return;
+    }
+    let master = event.parentEvent;
```
benbucksch, line 209: Please move the non-UI part of this function into logic/ .

----

In general:
* JS in the UI (frontend, Svelte) should be only be about 1-2 lines. Any function more complex than that should be in logic/ .
* However, logic MUST NOT contain UI access.

Please try to factor the functions so to make both rules true.

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -118,14 +202,74 @@
...
+    if (!confirm($t`Are you sure you want to remove this unfortunate series of events?`)) {
+      return;
+    }
+    let master = event.parentEvent;
+    await master.deleteFromServer();
+    await master.deleteIt();
+    onClose();
+  }
+
+  async function onDeleteForward() {
```
benbucksch, line 253: ditto. Move to logic/

app/frontend/Calendar/EditEvent/RepeatBox.svelte
```diff
@@ -186,35 +186,6 @@
     }
     return new RecurrenceRule(init);
   }
-
-  // TODO Call from save()
-  export function confirmAndChangeRule(): boolean {
```
benbucksch, line 191: Please try to keep the UI functions regarding recurrence in RepeatBox.svelte, as far as possible.

benbucksch (changes requested):

NeilRashbrook (commented):
I've moved the code as you requested. I also found the cause of the UI issue: I was using `copyFrom` but it copies stuff that it doesn't make sense to copy. The existing caller doesn't want to copy those fields either, so I just removed them.

app/logic/Calendar/RecurrenceRule.ts
```diff
@@ -182,6 +182,17 @@ export class RecurrenceRule implements Readonly<RecurrenceInit> {
     return 'RRULE:' + Object.entries(rule).map(entry => entry.join("=")).join(";");
   }
 
+  /**
+   * Checks whether this rule will invalidate existing occurrences.
+   * We don't check the series length though, as the UI never sets one.
```
benbucksch, line 187: The logic operates independently from the UI. Please assume that there will be 5 entirely different UIs, all working with this same logic. The UI can and will change at will.

If you need to check the length, please do so.

app/logic/Calendar/Event.ts
```diff
@@ -250,6 +249,20 @@ export class Event extends Observable {
     return !this.dbID;
   }
 
+  get seriesStatus() {
```
benbucksch, line 252: Please define the return type, and document all values.
Please add a description comment of what this tells us.

app/logic/Calendar/Event.ts
```diff
@@ -162,17 +162,16 @@ export class Event extends Observable {
     this.duration = days * 86400;
   }
 
-  /** Create a new instance of the same event.
-   * Copy all data of the `original` event into this new Event object */
+  /**
+   * Create a new instance of the same event.
+   * Copy most data of the `original` event into this new Event object.
```
benbucksch, line 167: It must copy *all* data.
If you need a variation, you can pass in optional parameters. Please make sure to document when you would use which variant.

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -102,7 +142,36 @@
   $: event.startEditing(); // not `$event`
   $: canSave = event && $event.title && $event.startTime && $event.endTime &&
       event.startTime.getTime() <= event.endTime.getTime() && $event.hasChanged();
-  $: oldTitle = event?.title || $t`Event`;
+  $: seriesStatus = event.seriesStatus;
+
+  function confirmAndChangeRule(): boolean {
```
benbucksch, line 147:
```suggestion
  function confirmAndChangeRecurrenceRule(): boolean {
```

app/frontend/Calendar/EditEvent/RepeatBox.svelte
benbucksch: I see. Too bad.

benbucksch (changes requested):
Getting better. Thanks for the changes.
A few polish changes still to do.

benbucksch:
ping

NeilRashbrook:
pong

NeilRashbrook:
I've been testing on Exchange and found some bugs, including some fallout from the rebase; I've pushed a fix for all of them except for one, which is that change remainder of series doesn't work on ActiveSync, it's failing to add the new series because it conflicts. (No, I don't understand how you can get a conflict when there isn't an event on the server to conflict with.) I'll have to debug it in more detail next week...

NeilRashbrook:
One conflict was caused by trying to create a new series with the same UID, which isn't allowed. I did run into a second conflict which I now have to debug...

NeilRashbrook:
I figured out the conflict issue; I had forgotten to get the server's auto-generated UID in ActiveSync accounts.

app/frontend/Calendar/EditEvent/EditEvent.svelte
```diff
@@ -93,7 +93,7 @@
 
   export let event: Event;
 
-  $: showRepeat = $event.recurrenceCase != RecurrenceCase.Normal;
```
benbucksch, line 96: Why is this condition no longer correct?
(Could it be modified slightly to remain here?)

app/frontend/Calendar/EditEvent/EditEvent.svelte
```diff
@@ -103,7 +103,6 @@
   let repeatBox: RepeatBox;
 
   function expandRepeat(): void {
```
benbucksch, line 105:
```suggestion
  function expandRepeat(): void {
    // `showRepeat = true` set by `<ExpanderButton>` click handler
```

benbucksch (approved):
Some small polish work still to do please in a followup PR, but let's land this.

app/logic/Calendar/RecurrenceRule.ts
```diff
@@ -182,6 +182,17 @@ export class RecurrenceRule implements Readonly<RecurrenceInit> {
     return 'RRULE:' + Object.entries(rule).map(entry => entry.join("=")).join(";");
   }
 
+  /**
+   * Checks whether this rule will invalidate existing occurrences.
```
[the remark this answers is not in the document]
NeilRashbrook: Note to self:
```typescript
/**
 * When you change the entire series, that could include changing when it recurs.
 * The UI uses this function to tell whether the recurrence was edited,
 * as this will cause all exceptions and exclusions to be reset.
 * But we can't check the series length though, as the UI never sets one.
 */
```

app/logic/Calendar/RecurrenceRule.ts
NeilRashbrook: I don't know what you're trying to ask here.

The UI always creates an unending recurrence. But it still wants to know whether any of the other fields changed. So when we're comparing, we can only compare the other fields.

app/logic/Calendar/Event.ts
```diff
@@ -277,6 +290,33 @@ export class Event extends Observable {
...
+   * "only" - event is the only instance or exception remaining
+   * "first" - event is the earliest instance or exception remaining
+   * "last" - event is the last instance, but there may be more exceptions
+   * "middle" - event is an instance, but none of the above apply
+   * In particular:
+   * "only" - don't delete the event, delete the master instead
+   * "first" - offer to edit the whole series
+   * "middle" - offer to edit the remainder of the series
+   */
+  get seriesStatus(): "none" | "only" | "first" | "last" | "middle" {
```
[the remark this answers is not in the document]
benbucksch: OK, good enough for now.

app/logic/Calendar/RecurrenceRule.ts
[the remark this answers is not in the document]
benbucksch: Thanks for the new comment. This is easier to understand, yes.

> But we can't check the series length though, as the UI never sets one.

Please write the code to support series length and work in case that an series end is set.
Even if the UI never uses it.

app/frontend/Calendar/EditEvent/EditEvent.svelte
benbucksch: > I think I could restore it without the $ by making some other changes.

With the code comment that I added, it's OK for now.

Change it if (and only if) your change makes the code easier to understand or working more reliably in case of future code changes.

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -134,14 +206,53 @@
...
+      return;
+    }
+    let master = event.parentEvent;
+    master.copyEditableFieldsFrom(event);
+    await master.saveToServer();
+    await master.save();
+    onClose();
+  }
+
+  async function onChangeForward() {
```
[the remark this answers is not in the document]
benbucksch: > I obviously can't move the whole button handler itself to logic, plus there is a call to repeatBox.newRecurrenceRule, so it's unclear what parts you want moved.

Ah, I missed the reference to `repeatBox`. All the rest was merely using `master` and `event`, so it could have been moved to logic.

Nevermind.

app/logic/Calendar/RecurrenceRule.ts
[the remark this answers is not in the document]
NeilRashbrook: The old code already did this in the sense that it simply checked whether both rules had the same serialisation, which appears to be what the OWA UI does. This is no longer sufficient for the new UI, which is why it needed a dedicated method.

app/frontend/Calendar/EditEvent/EditEvent.svelte
NeilRashbrook: This does simplify matters, thanks; I should be able to implement that to fix issues #591 and #592. (This will also result in a further change to this condition.)

benbucksch: > This does simplify matters, thanks

Great!

> (This will also result in a further change to this condition.)

That's OK, as long as the changes are inherent. I just want them to make sense independent from the UI.

PR #545, opened 2025-04-11
Calendar: Free/Busy: EWS, OWA, ActiveSync

NeilRashbrook:
Works from the console at least.

Note that EWS and OWA return one availability entry for each event within the time frame, while ActiveSync returns half hour availability entries for the entire time frame which are marked free only if no busy event overlaps the given half hour.


benbucksch (approved):
Nice

Issue #534, opened 2025-04-01
Composer: Delete last line before blockquote


benbucksch:
### Reproduction
1. Mail composer
2. Blockquote, empty line, Foo, i.e. `<blockquote><p>Quote</p></blockquote><p>Foo</p>`

![Image](https://github.com/user-attachments/assets/9d8b26b8-a702-4be1-ae49-8e1e48e8a51c)

3. Put cursor at the empty line
4. Press BACKSPACE to delete the empty line, i.e. between the blockquote and Foo

### Actual result
* Foo becomes part of the blockquote
* The empty line stays
* It acutally has become 2 (!) empty lines now, whereas only the second empty line is selectable and the first empty line behaves like a margin of the 

![Image](https://github.com/user-attachments/assets/9d57dca0-fd3a-46cd-81f5-2f787623ca20)

### Expected result
* The empty line is removed
* Foo is still outside the blockquote
* Quote still remains the only line inside the blockquote

### Other situations
* It works, if you press DELETE instead of BACKSPACE. -- Because: Delete removes the following linebreak, whereas BACKSPACE removes the previous linebreak.

jermy-c:
I think this is cause buy the automatic `joinForward()` and `joinBackward()` commands on or after a transaction.

jermy-c:
This is possible to fix but break something else if not careful. I would need to write a custom command for backspace and the key to that new command. And there's also the `<p>` issue since it's not the default behavior.

benbucksch:
This is low priority

jermy-c:
So, when in a list the shortcuts from the [`listKeyMap`](https://tiptap.dev/docs/editor/extensions/functionality/listkeymap) Extension applies and it has a shortcut for `backspace` and `delete`. Maybe we'll need to extend that extension.

PR #550, opened 2025-04-14
Invitation: Display responses and allow the local calendar to be updated

NeilRashbrook:
Still to do:
- [ ] Send invitations
  - [ ] IMAP
  - [ ] ActiveSync
  - [X] EWS
  - [X] OWA
- [X] Recognise invitation email
- [X] Display invitation data
- [X] Accept or decline invitation
- [X] Recognise invitation event
- [X] Change invitation acceptance
- [X] Recognise acceptance email
- [X] Update meeting participation

### DIFF
--- a/app/frontend/Calendar/DisplayEvent/Invitation.svelte
+++ b/app/frontend/Calendar/DisplayEvent/Invitation.svelte
@@ -1,12 +1,12 @@
 <vbox class="invitation">
   {#if $message.event}
-    <InvitationDisplay event={message.event} {calendars} {selectCalendar} bind:selectedCalendar />
-  {:else if message.invitationMessage}
+    <InvitationDisplay event={message.event} {calendars} bind:selectedCalendar on:select={selectCalendar} />
+  {:else if $message.invitationMessage}
     {#await loadEvent()}
       {$t`Loading event...`}
     {:then}
-      {#if message.event}
-        <InvitationDisplay event={message.event} {calendars} {selectCalendar} bind:selectedCalendar />
+      {#if $message.event}
+        <InvitationDisplay event={$message.event} {calendars} bind:selectedCalendar on:select={selectCalendar} />
       {:else}
         No event found
       {/if}
@@ -15,7 +15,7 @@
     {/await}
   {/if}
   <hbox class="buttons">
-    {#if message.invitationMessage == InvitationMessage.Invitation}
+    {#if $message.invitationMessage == InvitationMessage.Invitation}
       {#if myParticipation == InvitationResponse.Accept}
         <Button
           label={$t`Confirmed *=> A meeting request has been confirmed by you`}
@@ -73,7 +73,7 @@
             classes="maybe font-normal" />
         </ButtonMenu>
       {/if}
-    {:else if message.invitationMessage}
+    {:else if $message.invitationMessage}
       <Button label={$t`Update calendar`} disabled={updateDisabled} onClick={onUpdate} classes="font-normal" />
     {/if}
   </hbox>
@@ -86,14 +86,14 @@
   import type { IncomingInvitation } from "../../../logic/Calendar/Invitation/IncomingInvitation";
   import { InvitationMessage, InvitationResponse, type InvitationResponseInMessage } from "../../../logic/Calendar/Invitation/InvitationStatus";
   import InvitationDisplay from "./InvitationDisplay.svelte";
+  import ButtonMenu from "../../Shared/Menu/ButtonMenu.svelte";
+  import MenuItem from "../../Shared/Menu/MenuItem.svelte";
   import Button from "../../Shared/Button.svelte";
   import AcceptIcon from "lucide-svelte/icons/check-check";
   import DeclineIcon from "lucide-svelte/icons/x";
   import MaybeIcon from "lucide-svelte/icons/circle-help";
   import ChevronDownIcon from "lucide-svelte/icons/chevron-down";
   import { gt, t } from "../../../l10n/l10n";
-  import ButtonMenu from "../../Shared/Menu/ButtonMenu.svelte";
-  import MenuItem from "../../Shared/Menu/MenuItem.svelte";
   import type { Collection } from "svelte-collections";
 
   export let message: EMail;
@@ -132,9 +132,7 @@
   async function respond(response: InvitationResponseInMessage) {
     await incomingInvitation.respondToInvitation(response);
     myParticipation = response;
-    if (!event) {
-      event = selectedCalendar.events.find(event => event.calUID == message.event.calUID);
-    }
+    event ??= selectedCalendar.events.find(event => event.calUID == message.event.calUID);
   }
   async function onAccept() {
     await respond(InvitationResponse.Accept);
--- a/app/frontend/Calendar/DisplayEvent/Invitation.svelte
+++ b/app/frontend/Calendar/DisplayEvent/Invitation.svelte
@@ -1,12 +1,12 @@
 <vbox class="invitation">
   {#if $message.event}
-    <InvitationDisplay event={message.event} />
+    <InvitationDisplay event={message.event} {calendars} {selectCalendar} bind:selectedCalendar />
   {:else if message.invitationMessage}
     {#await loadEvent()}
       {$t`Loading event...`}
     {:then}
       {#if message.event}
-        <InvitationDisplay event={message.event} />
+        <InvitationDisplay event={message.event} {calendars} {selectCalendar} bind:selectedCalendar />
       {:else}
         No event found
       {/if}
@@ -15,9 +15,6 @@
     {/await}
   {/if}
   <hbox class="buttons">
-    {#if selectedCalendar}
-      <AccountDropDown selectedAccount={selectedCalendar} accounts={calendars} withIcon={true} withLabel={true} disabled={calendars.length < 2} on:Select={selectCalendar} />
-    {/if}
     {#if message.invitationMessage == InvitationMessage.Invitation}
       {#if myParticipation == InvitationResponse.Accept}
         <Button
@@ -85,10 +82,10 @@
 <script lang="ts">
   import type { EMail } from "../../../logic/Mail/EMail";
   import type { Event } from "../../../logic/Calendar/Event";
+  import type { Calendar } from "../../../logic/Calendar/Calendar";
   import type { IncomingInvitation } from "../../../logic/Calendar/Invitation/IncomingInvitation";
   import { InvitationMessage, InvitationResponse, type InvitationResponseInMessage } from "../../../logic/Calendar/Invitation/InvitationStatus";
   import InvitationDisplay from "./InvitationDisplay.svelte";
-  import AccountDropDown from "../../Shared/AccountDropDown.svelte";
   import Button from "../../Shared/Button.svelte";
   import AcceptIcon from "lucide-svelte/icons/check-check";
   import DeclineIcon from "lucide-svelte/icons/x";
@@ -97,6 +94,7 @@
   import { gt, t } from "../../../l10n/l10n";
   import ButtonMenu from "../../Shared/Menu/ButtonMenu.svelte";
   import MenuItem from "../../Shared/Menu/MenuItem.svelte";
+  import type { Collection } from "svelte-collections";
 
   export let message: EMail;
   let calendars: Collection<Calendar>;
--- a/app/frontend/Calendar/DisplayEvent/Invitation.svelte
+++ b/app/frontend/Calendar/DisplayEvent/Invitation.svelte
@@ -16,72 +16,75 @@
   {/if}
   <hbox class="buttons">
     {#if selectedCalendar}
-      <AccountDropDown selectedAccount={selectedCalendar} accounts={calendars} withIcon={true} withLabel={true} disabled={calendars.length < 2} />
+      <AccountDropDown selectedAccount={selectedCalendar} accounts={calendars} withIcon={true} withLabel={true} disabled={calendars.length < 2} on:Select={selectCalendar} />
     {/if}
     {#if message.invitationMessage == InvitationMessage.Invitation}
-      <Button
-        label={$t`Confirm *=> Confirm to attend the meeting`}
-        onClick={onAccept}
-        icon={AcceptIcon}
-        classes="accept font-normal" />
-      <Button
-        label={$t`Reject *=> Decline to attend the meeting`}
-        onClick={onDecline}
-        icon={DeclineIcon}
-        classes="decline secondary font-normal" />
-      <ButtonMenu buttonIcon={ChevronDownIcon}>
-        <MenuItem
-          label={$t`Maybe *=> Maybe attend the meeting`}
-          onClick={onTentative}
-          icon={MaybeIcon}
-          classes="maybe font-normal" />
-      </ButtonMenu>
-    {:else if message.invitationMessage}
-      <Button label={$t`Update calendar`} disabled={updateDisabled} onClick={onUpdate} classes="font-normal" />
-    {:else if false}
-      <Button
-        label={$t`Confirmed *=> A meeting request has been confirmed by you`}
-        icon={AcceptIcon}
-        selected={true}
-        disabled={true}
-        classes="accept done font-normal" />
-      <ButtonMenu buttonIcon={ChevronDownIcon}>
-        <MenuItem
-          label={$t`Change to Reject *=> not attend the meeting`}
-          onClick={onDecline}
+      {#if myParticipation == InvitationResponse.Accept}
+        <Button
+          label={$t`Confirmed *=> A meeting request has been confirmed by you`}
+          icon={AcceptIcon}
+          selected={true}
+          disabled={true}
+          classes="accept done font-normal" />
+        <ButtonMenu buttonIcon={ChevronDownIcon}>
+          <MenuItem
+            label={$t`Change to Reject *=> not attend the meeting`}
+            onClick={onDecline}
+            icon={DeclineIcon}
+            classes="decline font-normal" />
+          <MenuItem
+            label={$t`Change to Maybe *=> Maybe attend the meeting`}
+            onClick={onTentative}
+            icon={MaybeIcon}
+            classes="maybe font-normal" />
+        </ButtonMenu>
+      {:else if myParticipation == InvitationResponse.Decline}
+        <Button
+          label={$t`Rejected *=> A meeting request has been declined by you`}
           icon={DeclineIcon}
-          classes="decline font-normal" />
-        <MenuItem
-          label={$t`Change to Maybe *=> Maybe attend the meeting`}
-          onClick={onTentative}
-          icon={MaybeIcon}
-          classes="maybe font-normal" />
-      </ButtonMenu>
-    {:else if false}
-      <Button
-        label={$t`Rejected *=> A meeting request has been declined by you`}
-        icon={DeclineIcon}
-        selected={true}
-        disabled={true}
-        classes="decline done font-normal" />
-      <ButtonMenu buttonIcon={ChevronDownIcon}>
-        <MenuItem
-          label={$t`Change to Accept *=> attend the meeting`}
+          selected={true}
+          disabled={true}
+          classes="decline done font-normal" />
+        <ButtonMenu buttonIcon={ChevronDownIcon}>
+          <MenuItem
+            label={$t`Change to Accept *=> attend the meeting`}
+            onClick={onAccept}
+            icon={AcceptIcon}
+            classes="accept font-normal" />
+          <MenuItem
+            label={$t`Change to Maybe *=> Maybe attend the meeting`}
+            onClick={onTentative}
+            icon={MaybeIcon}
+            classes="maybe font-normal" />
+        </ButtonMenu>
+      {:else}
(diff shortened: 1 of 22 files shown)


app/frontend/Calendar/DisplayEvent/Invitation.svelte
```diff
@@ -14,88 +14,67 @@
...
         icon={DeclineIcon}
         classes="decline secondary" />
-      <ButtonMenu buttonIcon={ChevronDownIcon}>
-        <MenuItem
-          label={$t`Maybe *=> Maybe attend the meeting`}
-          onClick={onTentative}
-          icon={MaybeIcon}
-          classes="maybe" />
-      </ButtonMenu>
-    </hbox>
```
benbucksch, line 36: Please keep the UI as-is. We had intentionally put this into a menu.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
```diff
@@ -14,88 +14,67 @@
...
-          icon={AcceptIcon}
-          classes="accept" />
-        <MenuItem
-          label={$t`Change to Maybe *=> Maybe attend the meeting`}
-          onClick={onTentative}
-          icon={MaybeIcon}
-          classes="maybe" />
-      </ButtonMenu>
-    </hbox>
-  {/if}
```
benbucksch, line 79: We want to show how our user answered to a given invitation email. Please keep the UI.

app/frontend/Calendar/DisplayEvent/InvitationDisplay.svelte
```diff
@@ -6,7 +6,6 @@
       {/if}
     </hbox>
     <hbox flex />
-    <AccountDropDown selectedAccount={calendar} accounts={appGlobal.calendars} withIcon={true} withLabel={false} />
```
benbucksch, line 9: The user can select which calendar to put the event in. That is particularly needed for IMAP account in combination with multiple local or CalDAV calendars.

benbucksch (changes requested):
This PR is removing significant parts of the UI had we need.

Also the APIs are not right yet. As mentioned, I would like the Invitation handling functions on a separate OutgoingInvitation (and IncomingInvitation) class.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
NeilRashbrook: Sadly it doesn't actually work, at least, not in the state it's in right now. Leaving it here would also break the fix I'm trying to make. And fixing it is beyond the scope of this PR.

app/frontend/Calendar/DisplayEvent/InvitationDisplay.svelte
NeilRashbrook: It's not gone, it just got moved somewhere where it was easier for me to populate it correctly.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
NeilRashbrook: This was the result of a merge conflict. At least I got the right label.

NeilRashbrook (commented):
> Also the APIs are not right yet. As mentioned, I would like the Invitation handling functions on a separate OutgoingInvitation (and IncomingInvitation) class.

I thought about that, but I was too tired at the time to refactor it.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: > Leaving it here would also break the fix I'm trying to make.

How it would break with this PR, and why?

We've worked on this UI quite a lot. In fact, 3 people in the team worked together on this. The translated labels were pain to get right, and needed several rounds to be correctly translated in various languages. I do no want that work to get lost. Please consulting with me before making such drastic changes to the UI.

At the very least, please wrap them in an `{#if false}`, and then follow up with another PR to fix it and make them functional.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: So, could you please restore it?

benbucksch: Please also check whether there were other "merge errors".

benbucksch (changes requested):

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: FWIW, the UI specs are here: https://parula.cloud.xwiki.com/xwiki/bin/view/Main/Calendar/Email%20invitation/

NeilRashbrook: Because Scheduling.Accepted and Scheduling.Declined have always referred to incoming messages from attendees that you invited and have nothing whatsoever to do with how you responded to an invitation.

benbucksch: Ah, OK, then I misunderstood the API. We need to straighten that out, to make it clearer.
`{#if false}` is fail, then. Thanks that you did that and restored the buttons.

NeilRashbrook:
> I would like the Invitation handling functions on a separate class.

I've refactored them again. And while I was there, I mostly fixed the invitation UI, so that for IMAP accounts you can see how you responded to the invitation (if it's still in your calendar), even after the fact.

benbucksch:
Thank you!

app/logic/Calendar/IncomingActions.ts
```diff
@@ -0,0 +1,64 @@
+import type { Calendar } from "./Calendar";
+import type { Event } from "./Event";
+import { Scheduling, ResponseType, type Responses } from "./Invitation";
+import type { EMail } from "../Mail/EMail";
+import { assert } from "../util/util";
+
+export class IncomingActions {
```
benbucksch, line 7: Please rename to `IncomingInvitation` and move to app/logic/Calendar/Invitation/IncomingInvitation.ts . Ditto for `ActiveSyncIncomingInvitation` etc.

app/logic/Calendar/IncomingActions.ts
```diff
@@ -0,0 +1,64 @@
...
+    let participant = event.participants.find(participant => participant.emailAddress == this.message.folder.account.emailAddress);
+    if (participant) {
+      event.response = participant.response = response;
+      await event.save();
+      await this.message.folder.account.sendInvitationResponse(this.event, response);
+    }
+    /* else add participant? */
+  }
+
+  async updateFromResponse() {
```
benbucksch, line 46: The function name is not clear to me. Are these responses from other participants (or the organizer) about whether the other people are going to attend? If so, can you rename it to `updateFromOtherParticipant()` ?

benbucksch (changes requested):

app/logic/Calendar/IncomingActions.ts
NeilRashbrook: It actually handles two cases:
- Cancellations from meeting organisers from meetings they invited you to
- Replies from participants to meetings you organised

But then again you knew I would name it badly...

NeilRashbrook:
I've rebased everything onto master again, as there were a boatload of merge conflicts, and while I was there I also addressed your review comments.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
```diff
@@ -14,88 +14,119 @@
...
+        <ButtonMenu buttonIcon={ChevronDownIcon}>
+          <MenuItem
+            label={$t`Maybe *=> Maybe attend the meeting`}
+            onClick={onTentative}
+            icon={MaybeIcon}
+            classes="maybe font-normal" />
+        </ButtonMenu>
+      {/if}
+    {:else if message.invitationMessage}
+      <Button label={$t`Update calendar`} disabled={updateDisabled} onClick={onUpdate} classes="font-normal" />
```
benbucksch, line 80: I don't understand what this button does. (And if I don't understand, how will a user understand?)

app/frontend/Calendar/DisplayEvent/Invitation.svelte
```diff
@@ -14,88 +14,119 @@
...
+    calendars = message.getUpdateCalendars();
+    selectedCalendar = calendars.first;
+    updateDisabled = calendars.length ? false : gt`This event is not in your calendar`;
+    if (selectedCalendar) {
+      selectCalendar();
+    }
+  }
+  function selectCalendar() {
+    incomingInvitation = selectedCalendar.getIncomingInvitationFor(message);
+    myResponse = incomingInvitation.myResponse;
```
benbucksch, line 124: This is strange. 

The account selector is there to allow the user to select which calendar the meeting will be placed in, at the moment when he clicks Accept. Once the user selected a response, we show that state when the user clicks on the message again (quite possibily on another day).

Am I reading that right that it will clear the response display once the user changed the calendar in the calendar selector? That's confusing, because the user might think that it cleared his response at the organizer, but it did not.

app/logic/Mail/EMail.ts
```diff
@@ -201,36 +202,11 @@ export class EMail extends Message {
...
-      await this.sendInvitationResponse(response);
-    }
-    /* else add participant? */
-  }
-
-  protected async sendInvitationResponse(response: InvitationResponseInMessage): Promise<void> {
-    await this.folder.account.sendInvitationResponse(this.event, response);
+  getUpdateCalendars(): Collection<Calendar> {
+    assert(this.invitationMessage && this.event, "Must have event to find calendar");
+    let validCalendars = appGlobal.calendars.filter(calendar => !calendar.mainAccount);
```
benbucksch, line 207: Why this condition that the calendar cannot be related to an email account?

Note that we will have CalDAV accounts that will have an IMAP account as mainAccount.

app/logic/Mail/EMail.ts
```diff
@@ -201,36 +202,11 @@ export class EMail extends Message {
...
+    let foundCalendars = validCalendars.filter(calendar => calendar.events.some(event => event.calUID == this.event.calUID));
```
benbucksch, line 208: It seems that you're primarily looking for the calendar that has this event. If so, why not return the Event in the Calendar directly?

(validCalendars looks bogey to me anyways.)

If we can move this function somewhere into Calendar/, that would be better.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -31,6 +31,7 @@ export class EWSAccount extends MailAccount {
   readonly tls = TLSSocketType.TLS;
   readonly canSendInvitations: boolean = false;
   readonly folderMap = new Map<string, EWSFolder>;
+  calendar: EWSCalendar | void;
```
benbucksch, line 34: Can you please remove this?

I'm trying to keep the zig-zack dependencies at bay. Email should not depend on Calendar.

(The code below that creates the calendar should also eventually move out of here. E.g. into setup. One hack doesn't justify another.)

benbucksch (changes requested):
This is going in the right direction, but not there yet.

benbucksch:
Please continue work on CardDAV for now.

app/frontend/Shared/Button.svelte
```diff
@@ -89,7 +89,9 @@
       clearTimeout(loadTimeout);
       loading = false;
     }
-    disabled = previousDisabled;
+    if (disabled === true) {
```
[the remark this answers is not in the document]
benbucksch: `previousDisabled` was supposed to restore it to the previous state before onClick ran, whether it's disabled or enabled.

The only condition that I see where your change make a difference is if the `disabled` Svelte property changed while the onClick command was running.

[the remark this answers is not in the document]
NeilRashbrook: Well, yes, I wanted the callback to be able to keep the button disabled.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: Can you please do that automatically and remove the button?

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: Can we move the event to the other calendar, using `event.moveToCalendar()`?
If that is much more difficult, then disable the calendar selector after response, yes.

app/frontend/Calendar/EditEvent/InvitationResponseButtons.svelte
```diff
@@ -1,5 +1,5 @@
 <hbox class="buttons">
-  {#if event.response == InvitationResponse.Unknown || event.response == InvitationResponse.Organizer}
+  {#if event.response > InvitationResponse.Organizer && event.participants.some(participant => participant.response == InvitationResponse.Organizer)}
```
benbucksch, line 2: `>`: Please avoid such order comparisons for enums. See #534

app/frontend/Shared/Button.svelte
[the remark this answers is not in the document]
benbucksch: So, the `onClick` function changes it to disabled. OK, makes sense.

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Yes, please. Shared calendars are coming right after.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -34,6 +34,7 @@ export class OWAAccount extends MailAccount {
   readonly tls = TLSSocketType.TLS;
   readonly canSendInvitations: boolean = false;
   readonly folderMap = new Map<string, OWAFolder>;
+  calendar: OWACalendar | void;
```
benbucksch, line 37: ditto

benbucksch (changes requested):

app/frontend/Calendar/DisplayEvent/Invitation.svelte
NeilRashbrook: I can look into that.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: Good. Nice. Do that.

benbucksch: Please do that in a new PR, so that this here can land.

NeilRashbrook:
(Converted to draft as this temporarily depends on PR #604)

benbucksch:
Hey Neil, could you please now come back to this and finish this, addressing the review comments? Also, adapt to the changes that happened on master, including the renames etc. I would like to land this tomorrow (Tuesday), if possible.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: Filed #619

app/frontend/Calendar/DisplayEvent/InvitationDisplay.svelte
```diff
@@ -6,22 +6,26 @@
...
+    {#if selectedCalendar}
+      <AccountDropDown selectedAccount={selectedCalendar} accounts={calendars} withIcon={true} withLabel={true} disabled={calendars.length < 2} on:Select={selectCalendar} />
```
benbucksch, line 10: Simply don't show the dropdown, if there's only 1 calendar.

app/frontend/Shared/AccountDropDown.svelte
```diff
@@ -24,6 +24,7 @@
   export let accounts: Collection<Account>;
   export let withIcon: boolean = false;
   export let withLabel: boolean = true;
+  export let disabled: boolean = false;
```
benbucksch, line 27: Not needed, simply don't show, see above.

app/logic/Calendar/Calendar.ts
```diff
@@ -1,20 +1,27 @@
...
+import { IncomingInvitation } from "./Invitation/IncomingInvitation";
+import type { EMail } from "../Mail/EMail";
 import { appGlobal } from "../app";
 import { Collection, ArrayColl } from "svelte-collections";
 import { ICalEMailProcessor } from "./ICal/ICalEMailProcessor";
 
 export class Calendar extends Account {
   readonly protocol: string = "calendar-local";
   readonly events = new ArrayColl<Event>();
+  readonly canAcceptAnyInvitation: boolean = true;
```
benbucksch, line 13: Please document what you mean with "any".

app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (participant) {
+      event.myParticipation = participant.response = response;
+      await event.save();
+      await this.event.sendInvitationResponse(response, this.message.folder.account);
+    }
+    /* else add participant? */
+  }
+
+  /** Handle the other two InvitationMessage values, as
+   *  respondToInvitation only handles InvitationMessage.Invitation */
```
benbucksch, line 46:
```suggestion
  /** It handles two cases:
   * - Cancellations from meeting organisers for `IncomingInvitation`
   * - Replies from participants in response to `OutgoingInvitation`
   *   TODO Move second case to `OutgoingInvitation`
   */
```

benbucksch (changes requested):

NeilRashbrook:
I've fixed two bugs in the changes that I'd made (it's always possible there are more but ActiveSync keeps timing out on me and waiting 20 minutes would not be good use of my time) and I've also cherry-picked your commit a6c2cb6 and then I've force-pushed my branch (because I've already rebased it).

app/logic/Calendar/Calendar.ts
```diff
@@ -1,20 +1,28 @@
...
+import { IncomingInvitation } from "./Invitation/IncomingInvitation";
+import type { EMail } from "../Mail/EMail";
 import { appGlobal } from "../app";
 import { Collection, ArrayColl } from "svelte-collections";
 import { ICalEMailProcessor } from "./ICal/ICalEMailProcessor";
 
 export class Calendar extends Account {
   readonly protocol: string = "calendar-local";
   readonly events = new ArrayColl<Event>();
+  /** Can this calendar accept incoming invitations from any source? */
```
benbucksch, line 13:
```suggestion
  /** Can this calendar accept incoming invitations from inboxes in other accounts? */
```

benbucksch (approved):

app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -0,0 +1,63 @@
...
+    assert(this.invitationMessage && this.invitationMessage != InvitationMessage.Invitation, "can't update from an invitation");
+    let event = this.calendar.events.find(p => p.calUID == this.event.calUID);
+    if (this.invitationMessage == InvitationMessage.CancelledEvent) {
+      let organizer = event.participants.find(participant => participant.response == InvitationResponse.Organizer);
+      if (organizer) {
+        organizer.response = InvitationResponse.Decline;
+        await event.save();
+      }
+    } else {
+      let invitee = this.event.participants.find(participant => participant.response != InvitationResponse.Organizer);
```
benbucksch, line 55: You're assuming that there will be only 2 participants in the response: The organizer and the sender. Is that mandated by the spec? If not, it's conceivable that there is some software that sends *all* participants. (Even if you have not actually encountered such a message. We have not seen all software that sends iCal.)

I think there should be an extra check that either a) there are <= 2 participants in the list or b) the participant email address matches the From email address of the email.

benbucksch (approved):

app/logic/Calendar/Invitation/IncomingInvitation.ts
NeilRashbrook: in theory there is apparently something called delegation - you receive an invitation, but you want someone else to take your place. In this case you're supposed to send the organiser a message that you want to delegate the invitation and the other person sends to both people whether they can attend, and the resulting message has three participants.

Unfortunately whenever I try to search for support for this feature I only get the completely different feature of Office 365/Exchange whereby you can look at another user's calendar, inbox etc.

benbucksch: Yes, I don't think Exchange does it this way.

benbucksch: I was thinking about the case where some calendar client on the other end simply copies *all* participants from the original invitation that they received, copied all their status, and changed only their own status. I am not sure such a software exists, but it's conceivable, and we should not break in this case.

benbucksch: Also, this needs a null check. If there is no other participant is found, `invitee` will be null, and `invitee.emailAddress` and `invitee.response` below will throw.

NeilRashbrook: Given that, how should we notify the user that the reply is malformed?

benbucksch: I guess the best would be an inline error message inside the invitation display.

Issue #456, opened 2025-02-25
Calendar: ICS: Unit tests


benbucksch:
Create unit tests to check whether we can read the iCal/ICS files correctly and correctly interpret the data in it.

That means reading it in an `Event` object, writing it out as JSCalendar JSON object, and comparing the result with the expected result.
This should be based on test cases supplied by others, en masse.

Potential sources:
* https://github.com/jens-maus/node-ical/tree/master/test
* https://github.com/libical/libical/tree/master/test-data
* https://github.com/icalendar/icalendar/tree/main/spec/fixtures
* https://devguide.calconnect.org/Appendix/Examples/

Pay attention for licenses, please.

Other parsers or resources:
* https://git.sr.ht/~whynothugo/vparser
* https://github.com/CalConnect/DEVGUIDE-EXAMPLES/tree/master/

fbartels:
Here are some more test sources:

- https://github.com/kewisch/ical.js/tree/main/samples
- https://github.com/peterbraden/ical.js/tree/master/test
- https://github.com/Kopano-dev/kopano-core/tree/master/libicalmapi/tests/ics
- https://github.com/Kopano-dev/kopano-core/tree/master/libicalmapi/tests/issues

NeilRashbrook:
Of 45 request, reply and cancel messages I found from the files in https://github.com/libical/libical/tree/master/test-data
- I haven't looked to see how the messages are used, so it's possible that some of them have deliberate errors.
  - 1 has a malformed METHOD
  - 2 have a malformed DTSTART
  - 3 have a malformed DTEND
- None of the reply and cancel messages bother filling in the whole event; they only provide the UID and ATTENDEEs. I haven't checked to see whether this is permitted.
- I found some code errors; I don't know whether it's worth creating individual PRs for each of these.
  - We aren't parsing the recurrence rule case-insensitively (as it happens all of the test cases were already in upper case)
  - Stringifying the recurrence rule should use the `allDay` format as appropriate
  - The wrong variable is being used to check for the organiser being a participant
  - It is possible for the organiser to match against multiple participants

NeilRashbrook:
Of the rest of the sample files, most weren't in a format that looked like an invitation, so I'm not sure how useful they are as a test, but I was able to adapt a few more, which led me to discover that all day events weren't being properly recognised,

benbucksch:
> Pay attention for licenses, please.

BSD, MIT, and Apache 2 licenses are OK

[...]

PR #558, opened 2025-04-16
ICS: Unit tests #456

NeilRashbrook:
Includes some icaljs tests and somee tests extracted from rfc6047 as I'm less unsure about the licences on those tests.


benbucksch (approved):
Nice, thank you!

Issue #556, opened 2025-04-16
UI: Change font sizes into CSS classes


benbucksch:
Across the app, in the CSS, we set the `font-size` to an absolute size, over 130 times. This is a problem when building the mobile UI, where the font size needs to be bigger.

Most of the places are font size 12px, 14px or 16px. So, we can make 3 classes for them: `font-smallest`, `font-small`, and `font-normal`, respectively. That would allow us to change the font size across the app, by only adapting the style sheet.

git commit 2e6b95d6 adds the classes. Please search through the app for wherever we set the font-size, remove it, find the HTML element where we set the font-size, and set the corresponding class on that element.

* 9, 10, 11, 12px -> `font-smallest`
* 13, 14px -> `font-small`
* 15, 16px -> `font-normal`

Then, make a second pass for the odd cases where we set the `font-size` using `:global()` or similar, which cannot be fixed this way.
Then, make a third pass, together with Elena, about the font sizes which are larger than 16px.

benbucksch:
> How should we fix this case? Specifically, cases where the selector is selecting an element in a <slot/> component?

I don't know yet. We'll look at those individually.

> I've added the class for selectors that were selecting components that export a classes prop.

Nice

PR #562, opened 2025-04-18
UI: Change font sizes into CSS classes - Pass 2 (#556)

jermy-c:
- Replaced instances of `font-size` in CSS rules with classes: `font-smallest`, `font-small`, `font-normal`
- Instances where components export a `classes` prop and select the correct element are replaced
- Instances where the CSS rule is in a tag select are replaced
- Instances where there's a `!important` is still not replaced

### DIFF
--- a/app/frontend/Calendar/DisplayEvent/Invitation.svelte
+++ b/app/frontend/Calendar/DisplayEvent/Invitation.svelte
@@ -20,18 +20,18 @@
         label={$t`Confirm *=> Confirm to attend the meeting`}
         onClick={onAccept}
         icon={AcceptIcon}
-        classes="accept" />
+        classes="accept font-normal" />
       <Button
         label={$t`Reject *=> Decline to attend the meeting`}
         onClick={onDecline}
         icon={DeclineIcon}
-        classes="decline secondary" />
+        classes="decline secondary font-normal" />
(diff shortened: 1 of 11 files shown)


app/frontend/Calendar/DisplayEvent/Invitation.svelte
```diff
@@ -20,18 +20,18 @@
         label={$t`Confirm *=> Confirm to attend the meeting`}
         onClick={onAccept}
         icon={AcceptIcon}
-        classes="accept" />
+        classes="accept font-normal" />
```
benbucksch, line 23: Does it work to put the font class on the `<hbox class="buttons font-normal">` instead of the individual buttons`?

benbucksch (approved):

app/frontend/Calendar/DisplayEvent/Invitation.svelte
jermy-c: No, it doesn't work. I've just tested it. The `user agent stylesheet` overwrites the style for `buttons`.

benbucksch: Thanks for checking. Given that it landed as-us, it's OK as-is.

PR #570, opened 2025-04-21
Clarify invitation types

NeilRashbrook:
Don't ask me where those values in the comment in `Mail/SQL/createDatabase` came from...

### DIFF
--- a/app/logic/Calendar/Invitation.ts
+++ b/app/logic/Calendar/Invitation.ts
@@ -5,9 +5,17 @@ export type iCalMethod = "CANCEL" | "REQUEST" | "REPLY";
  */
 export enum InvitationMessage {
   None = 0,
-  Invitation = 4, // METHOD:REQUEST, an incoming invitation
-  CancelledEvent = 5, // METHOD:CANCEL, a follow-up cancellation
-  ParticipantReply = 6, // METHOD:REPLY, a reply to your outgoing invitation
+  /** Creates an incoming invitation.
+   * iCal METHOD: REQUEST */
+  Invitation = 4,
+  /** In incoming invitation is being nullified.
+   * Only valid for incoming invitations.
+   * iCal METHOD: CANCEL */
+  CancelledEvent = 5,
+  /** A reply to your outgoing invitation.
+   * Only valid for outgoing invitations.
+   * iCal METHOD: REPLY */
+  ParticipantReply = 6,
 }
 
 /**
@@ -19,11 +27,16 @@ export enum InvitationMessage {
  * Note: These are EWS/OWA names and ActiveSync values.
  */
 export enum InvitationResponse {
+  /** The event is not an invitation, neither incoming nor outgoing. */
   Unknown = 0,
   Organizer = 1,
+  /** The participant has notified us that he is unsure whether he will attend or not.
+   For incoming invitations only. */
   Tentative = 2,
   Accept = 3,
   Decline = 4,
+  /** The participant has not yet answered.
+   For incoming invitations only. */
   NoResponseReceived = 5,
 }
 
--- a/app/logic/Calendar/Invitation.ts
+++ b/app/logic/Calendar/Invitation.ts
@@ -1,20 +1,24 @@
 export type iCalMethod = "CANCEL" | "REQUEST" | "REPLY";
 
 /**
- * For an inbox item that represents a scheduling message, the type of message:
- * Accepted/Tentative/Declined responses, invitations, or cancellations.
+ * For an inbox item that represents a scheduling message, the type of message.
  */
-export enum Scheduling {
+export enum InvitationMessage {
   None = 0,
-  Accepted = 1,
-  Tentative = 2,
-  Declined = 3,
-  Request = 4,
-  Cancellation = 5,
+  Invitation = 4, // METHOD:REQUEST, an incoming invitation
+  CancelledEvent = 5, // METHOD:CANCEL, a follow-up cancellation
+  ParticipantReply = 6, // METHOD:REPLY, a reply to your outgoing invitation
 }
 
-/* Note: These are EWS/OWA names and ActiveSync values. */
-export enum ResponseType {
+/**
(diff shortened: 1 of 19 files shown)


app/logic/Calendar/Invitation.ts
```diff
@@ -1,20 +1,24 @@
...
+export enum InvitationMessage {
   None = 0,
-  Accepted = 1,
-  Tentative = 2,
-  Declined = 3,
-  Request = 4,
-  Cancellation = 5,
+  Invitation = 4, // METHOD:REQUEST, an incoming invitation
+  CancelledEvent = 5, // METHOD:CANCEL, a follow-up cancellation
+  ParticipantReply = 6, // METHOD:REPLY, a reply to your outgoing invitation
```
benbucksch, line 10: Make these JSDoc, please. Like so:
```suggestion
  /** Creates an incoming invitation.
   * iCal METHOD: REQUEST */
  Invitation = 4,
  /** In incoming invitation is being nullified.
   * Only valid for incoming invitations.
   * iCal METHOD: CANCEL */
  CancelledEvent = 5,
  /** A reply to your outgoing invitation.
   * Only valid for outgoing invitations.
   * iCal METHOD: REPLY */
  ParticipantReply = 6,
```

app/logic/Calendar/Invitation.ts
```diff
@@ -23,12 +27,12 @@ export enum ResponseType {
   NoResponseReceived = 5,
```
benbucksch, line 27:
```suggestion
  /** The participant has not yet answered.
   For incoming invitations only. */
  NoResponseReceived = 5,
```

app/logic/Calendar/Invitation.ts
```diff
@@ -1,20 +1,24 @@
...
+ * For meetings only, these values indicate the status of either:
+ * - For any given participant, their participation in an event
+ * - The user's own participation in an event
+ *   In particular, this identifies whether the user organised this meeting.
+ *
+ * Note: These are EWS/OWA names and ActiveSync values.
+ */
+export enum InvitationResponse {
   Unknown = 0,
   Organizer = 1,
```
benbucksch, line 23:
```suggestion
  /** Our user is the organizer of this event.
   Signals that this is an outgoing invitation.
   For outgoing invitations only. */
  Organizer = 1,
```

app/logic/Calendar/Invitation.ts
```diff
@@ -1,20 +1,24 @@
...
+/**
+ * For meetings only, these values indicate the status of either:
+ * - For any given participant, their participation in an event
+ * - The user's own participation in an event
+ *   In particular, this identifies whether the user organised this meeting.
+ *
+ * Note: These are EWS/OWA names and ActiveSync values.
+ */
+export enum InvitationResponse {
   Unknown = 0,
```
benbucksch, line 22:
```suggestion
  /** The event is not an invitation, neither incoming nor outgoing. */
  Unknown = 0,
```

app/logic/Calendar/Invitation.ts
```diff
@@ -1,20 +1,24 @@
...
   Organizer = 1,
   Tentative = 2,
```
benbucksch, line 24:
```suggestion
  /** The participant has notified us that he is unsure whether he will attend or not.
   For incoming invitations only. */
  Tentative = 2,
```

benbucksch (changes requested):

benbucksch (approved):

app/logic/Calendar/Invitation.ts
NeilRashbrook: (iCal does not allow extra whitespace in its values)

app/logic/Calendar/Invitation.ts
NeilRashbrook: As a status on an event, this shows that you were the organiser. (Your participant object will also have the Organiser status.)

For meetings organised by someone else, this is set on the organising attendee, unless the meeting was cancelled.

app/logic/Calendar/Invitation.ts
NeilRashbrook: As a status on an event, this shows your latest response back to the organiser. (Your participant object will also have that status.)

For meetings you organise, this is set on attendees to show that you have seen a response from that attendee (and what it is).

app/logic/Calendar/Invitation.ts
NeilRashbrook: I'd have to double-check to see whether this status can appear directly on events. It appears on a participant of a meeting you organise to show that you have not yet seen a response from that participant.

It's also used for your status as an attendee on incoming meeting requests, since at this point you haven't responded yet.

app/logic/Calendar/Invitation.ts
benbucksch: You can change `METHOD` to lowercase `method`.

app/logic/Calendar/Invitation.ts
benbucksch: OK, can you please update the JSDoc definition?

If possible, I'd prefer to have 2 enums: one for incoming invitations, and one for outgoing invitations. The code should not be shared between them, so the enum isn't shared, either. Having a separate enum would make the cases clearer.

--- 
Also, please add JSDoc to `Accept` and `Decline` below.

NeilRashbrook: I don't understand.

Tom uses Mustang to invite Harry to a meeting. Harry uses Mustang to accept the invitation. This means that both Tom and Harry have a meeting where Tom is the organiser and Harry is an accepted invitee. Why would they use different enums?

benbucksch: You're talking about Accept/Decline now. I was referring to the Organizer state that [you explained above](https://github.com/mustang-im/mustang/pull/570#discussion_r2052697983).

But if it makes more sense, you can keep them in the same enum. Just please document for `Organizer` what you said in your previous comment, what it means for incoming and outgoing invitations.

PR #571, opened 2025-04-21
SQL: Add json column to most tables, for supplemental properties

benbucksch:
* Add `json` column to most tables, for supplemental properties. This allows to add message-specific additional data, e.g. from addons or for calendar, without changing the DB schema.
* Move email.scheduling into the JSON, given that it's used only for emails that contain invitations
* Move folder.uidvalidity into JSON, given that it's needed only for IMAP folders.
* Keep syncState as separate column, not in JSON, given that it's needed for most protocols, and will be written often.
* Add `meet.call` table, for a call log

### DIFF
--- a/app/logic/Mail/SQL/SQLEMail.ts
+++ b/app/logic/Mail/SQL/SQLEMail.ts
@@ -45,13 +45,11 @@ export class SQLEMail {
           INSERT INTO email (
             messageID, folderID, pID, parentMsgID,
             size, dateSent, dateReceived,
-            scheduling,
             outgoing, contactEmail, contactName,
             subject, plaintext, html
           ) VALUES (
             ${email.id}, ${email.folder.dbID}, ${email.pID}, ${email.inReplyTo},
             ${email.size}, ${email.sent.getTime() / 1000}, ${email.received.getTime() / 1000},
-            ${email.scheduling},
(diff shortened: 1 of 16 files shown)


benbucksch:
@NeilRashbrook Could you please to a post-review, and test it carefully, including all the previous SQL changes I did yesterday?

app/logic/Mail/SQL/SQLEMail.ts
```diff
@@ -363,6 +365,8 @@ export class SQLEMail {
     email.isSpam = sanitize.boolean(row.isSpam, false);
     email.threadID = sanitize.string(row.threadID ?? row.parentMsgID, null);
     email.downloadComplete = sanitize.boolean(row.downloadComplete, false);
+    let json = sanitize.json(row.json, {});
+    email.scheduling = sanitize.integer(json.scheduling, 0);
```
NeilRashbrook, line 369: This is incorrect, it should say
```suggestion
    email.scheduling = sanitize.integer(json.invitationMessage, 0);
```
So far I have not found any other problems.

benbucksch: Thanks. Fixed in git commit 825bb5ce

NeilRashbrook: Yes I had suspected you would do that :wink:

PR #573, opened 2025-04-22
Account: Add mainAccount and dependent accounts - fixes #155

### LINKED ISSUE #155: Dependent/main accounts
### Situation
When the user sets up an Office365, or Google, or Mustang account, the mail account also comes with address book sync, calendar sync, file sharing, and video conference. We are creating corresponding "accounts" in Mustang, one for each address book, one for each calendar etc.

### Problem

Their login is directly dependent on the login of the main mail account. I.e. if I didn't log in to the mail account, then saving a contact to the server will fail.

#### Link

Create an explicit "dependent" and "main" account link. I.e. the address book has a "mainAccount" property which points to the mail account. And the mail account has a "dependentAccounts" array, which points to the address books, calendars, file sharing, video conference accounts.

#### isLoggedIn
The address book's `isLoggedIn` will simply return `this.mainAccount.isLoggedIn`. For that to work, the address book needs to subscribe to the main account changes on load (in the moment when the `mainAccount` property is set), so that listeners to the address book `isLoggedIn` state are updated as soon as the mail account login state changes.

#### Login
Likewise, the address book's `login()` function will simply delegate to `this.mainAccount.login()`.

When a contact needs to be saved to the server, and the main account is not logged in, the dependent account needs to trigger the login. Whether an interactive login is requested depends on whether the user himself started the modification directly, or whether it's a background process.

#### Save link on disk

The link needs to be saved on disk, and restored on load.

benbucksch:
* Add `Account.mainAccount` and `Account.dependentAccounts()`
* Persist it to DB and restore it after load
* Use it for addresssbook and calendar to email account in Exchange protocols
* TODO: SMTP to IMAP

### DIFF
--- a/app/logic/Mail/AccountsList/MailAccounts.ts
+++ b/app/logic/Mail/AccountsList/MailAccounts.ts
@@ -16,6 +16,7 @@ import { OWAAccount } from '../OWA/OWAAccount';
 import { ActiveSyncAccount } from '../ActiveSync/ActiveSyncAccount';
 import { GraphAccount } from '../Graph/GraphAccount';
 // #endif
+import { getAllAccounts } from '../../Abstract/Account';
 import { setStorage } from '../Store/setStorage';
 import { NotReached } from '../../util/util';
 import type { Collection } from 'svelte-collections';
@@ -67,7 +68,6 @@ export async function readMailAccounts(): Promise<Collection<MailAccount>> {
   return await SQLMailStorage.readMailAccounts();
 }
 // #endif
-
 
 const kProtocolLabel = {
   // Displayed to user, but do not translate (technical term)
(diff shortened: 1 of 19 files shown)


benbucksch:
@NeilRashbrook Could you please review and test this? If you find problems, please note it here, and submit a PR with the fix.

app/logic/Mail/AccountsList/MailAccounts.ts
```diff
@@ -16,6 +16,7 @@ import { OWAAccount } from '../OWA/OWAAccount';
 import { ActiveSyncAccount } from '../ActiveSync/ActiveSyncAccount';
 import { GraphAccount } from '../Graph/GraphAccount';
 // #endif
+import { getAllAccounts } from '../../Abstract/Account';
```
NeilRashbrook, line 19: Is this import still needed?

benbucksch: No. Will fix in #574

PR #574, opened 2025-04-22
SMTP: Use mailAccount to link to IMAP/POP3 account - fixes #155

### LINKED ISSUE #155: Dependent/main accounts

benbucksch:
Part of #155. Followup to #573.

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -441,8 +441,7 @@ export class ActiveSyncAccount extends MailAccount {
             break;
           case FolderType.Calendar:
           case FolderType.UserCalendar:
-            // TODO compare calendar ID
-            let calendar = appGlobal.calendars.find((calendar: ActiveSyncCalendar) => calendar.mainAccount == this) as ActiveSyncCalendar | null;
+            let calendar = appGlobal.calendars.find(calendar => calendar.mainAccount == this && calendar.url == url.toString()) as ActiveSyncCalendar | null;
             console.log("found the ActS cal again", calendar?.name);
             if (calendar) {
               calendar.name = change.DisplayName;
@@ -450,6 +449,7 @@ export class ActiveSyncAccount extends MailAccount {
               calendar = newCalendarForProtocol("calendar-activesync") as ActiveSyncCalendar;
               calendar.name = change.DisplayName;
               calendar.url = url.toString();
+              // TODO save calendar ID
               calendar.username = this.username;
               calendar.workspace = this.workspace;
               calendar.mainAccount = this;
@@ -458,8 +458,7 @@ export class ActiveSyncAccount extends MailAccount {
             break;
           case FolderType.Contacts:
           case FolderType.UserContacts:
-            // TODO compare addressbook ID
-            let addressbook = appGlobal.addressbooks.find((addressbook: ActiveSyncAddressbook) => addressbook.mainAccount == this) as ActiveSyncAddressbook | null;
+            let addressbook = appGlobal.addressbooks.find(addressbook => addressbook.mainAccount == this && addressbook.url == url.toString()) as ActiveSyncAddressbook | null;
             console.log("found the ActS AB again", addressbook?.name);
             if (addressbook) {
               addressbook.name = change.DisplayName;
@@ -467,6 +466,7 @@ export class ActiveSyncAccount extends MailAccount {
               addressbook = newAddressbookForProtocol("addressbook-activesync") as ActiveSyncAddressbook;
               addressbook.name = change.DisplayName;
               addressbook.url = url.toString();
+              // TODO save addressbook ID
               addressbook.username = this.username;
               addressbook.workspace = this.workspace;
               addressbook.mainAccount = this;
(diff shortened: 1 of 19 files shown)


app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -441,15 +441,15 @@ export class ActiveSyncAccount extends MailAccount {
             break;
           case FolderType.Calendar:
           case FolderType.UserCalendar:
-            // TODO compare calendar ID
-            let calendar = appGlobal.calendars.find((calendar: ActiveSyncCalendar) => calendar.mainAccount == this) as ActiveSyncCalendar | null;
+            let calendar = appGlobal.calendars.find(calendar => calendar.mainAccount == this && calendar.url == url.toString()) as ActiveSyncCalendar | null;
```
NeilRashbrook, line 444: Aha, I had been wondering about this!

benbucksch: Yes, putting the serverID into the URL was mighty subtle!

We should put it into a separate field. It wasn't possible before, but now with the JSON properties, we can store arbitrary data for each account.

benbucksch: Thanks for having noticed this. This is the level of review that I needed.

benbucksch:
@NeilRashbrook Please review

app/logic/Mail/AutoConfig/manualConfig.ts
```diff
@@ -16,7 +16,7 @@ export function makeManualConfig(emailAddress: string, password: string): MailAc
   config.authMethod = AuthMethod.Password;
   config.source = "manual";
 
-  let outgoing = new SMTPAccount();
+  let outgoing = newAccountForProtocol("smtp") as any as SMTPAccount;
```
NeilRashbrook, line 19: Was account setup broken because it wasn't saving the account or something? (I'm worried that these changes relate to the "FOREIGN KEY constraint failed" errors I was seeing.)

(The `as any` shouldn't be necessary, but I've had similar problems before, so I'll take your word for it.)

app/logic/Mail/SQL/SQLMailAccount.ts
```diff
@@ -85,13 +79,10 @@ export class SQLMailAccount {
...
-    for (let account of accounts) {
-      let id = account.outgoingAccountID;
-      delete account.outgoingAccountID;
-      if (id) {
-        account.outgoing = smtpAccounts.find(acc => acc.id == id);
-      }
+    // Need to special-case it here, because SMTP accounts
+    // aren't in `appGlobal.emailAccounts`, so `setMainAccounts()` won't find them.
+    for (let smtp of smtpAccounts) {
+      smtp.mainAccount = accounts.find(acc => acc.id == smtp._mainAccountID);
```
NeilRashbrook, line 85: (I was wondering how this worked with other cases, but I think what happens is that this causes the account to be the outgoing account of its main account, at which point it will start appearing in `allAccounts()`.)

NeilRashbrook (commented):
I had originally set up an EWS account using a previous version of this patch (commit 3c5aa46), which continued to work, however for rest of the review I used master (commit 844ab27).

When I tried to create an OWA account I got "FOREIGN KEY constraint failed". After cancelling setup I noticed that the calendar and addressbook had already been added, although the mail account had not actually been saved at this point.

(I was also confused by there being two logins, one using the EWS/ActiveSync webview, one using the OWA popup.)

I was then able to create an ActiveSync account, which seemed to be working normally, e.g. all of the calendars were discovered as expected.

I then tried to create an IMAP account and also a second EWS account but each time I also got "FOREIGN KEY constraint failed". In the case of the EWS account a phantom address book had been created (I couldn't delete it but it went away after I restarted).

I was then able to create a second ActiveSync account, which also seemed to be working normally.

app/logic/Mail/MailAccount.ts
```diff
@@ -118,13 +115,16 @@ export class MailAccount extends TCPAccount {
   }
 
   async save(): Promise<void> {
+    if (this.outgoing) {
+      this.outgoing.mainAccount = this;
+    }
+    console.log("mail account save", this.protocol, this.name, this, "outgoing", this.outgoing, "mainaccount", this.mainAccount, "storage", this.storage);
+    await super.save();
     await this.storage?.saveAccount(this);
```
NeilRashbrook, line 123: The problem here is that `super.save` tries to save all dependent accounts, which doesn't work until this account has been saved. This causes the creation of IMAP and POP3 accounts to fail, because they try to save the SMTP account first.

NeilRashbrook (commented):
I've since found that ActiveSync works because it has a custom `verifyLogin` which doesn't try to create and save any dependent accounts, but EWS and OWA fail because `verifyLogin` just defaults to calling `login(true);` which tries to create dependent addressbook and calendar accounts, and this fails because the main account hasn't been saved yet. (Presumably under the old code, these accounts would simply be saved, and would linger if you cancelled account setup at the last second.)

benbucksch:
> I've since found that ActiveSync works because it has a custom `verifyLogin` which doesn't try to create and save any dependent accounts, but EWS and OWA fail because `verifyLogin` just defaults to calling `login(true);` which tries to create dependent addressbook and calendar accounts, and this fails because the main account hasn't been saved yet. (Presumably under the old code, these accounts would simply be saved, and would linger if you cancelled account setup at the last second.)

Can you please make a PR with a fix?

NeilRashbrook:
I'm used to GitHub auto-linking in both directions automatically (`<user> mentioned this pull request <date>`), but I assume you meant in email too?

PR #578, opened 2025-04-24
Fix various constraint failures during account setup

NeilRashbrook:
As mentioned in #574:
- accounts are trying to save their dependent accounts first, which doesn't work
- EWS and OWA accounts need a "light" `verifyLogin` which doesn't try to create dependent accounts

### DIFF
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -110,6 +110,12 @@ export class Account extends Observable {
   /** Saves the config in this account to disk.
    * Does not save the contents, e.g. messages. */
   async save(): Promise<void> {
+    throw new AbstractFunction();
+  }
+
+  /** Saves the config in this account and its dependent accounts to disk. */
+  async saveAll(): Promise<void> {
+    await this.save(); // must be first, because of foreign key constraint
     for (let dependent of this.dependentAccounts()) {
       await dependent.save();
     }
--- a/app/logic/Chat/ChatAccount.ts
+++ b/app/logic/Chat/ChatAccount.ts
@@ -29,7 +29,6 @@ export class ChatAccount extends TCPAccount {
 
   async save(): Promise<void> {
     await this.storage?.saveAccount(this);
-    await super.save();
   }
 
   async deleteIt(): Promise<void> {
--- a/app/logic/Chat/ChatAccount.ts
+++ b/app/logic/Chat/ChatAccount.ts
@@ -28,8 +28,8 @@ export class ChatAccount extends TCPAccount {
   }
 
   async save(): Promise<void> {
-    await super.save();
     await this.storage?.saveAccount(this);
+    await super.save();
   }
 
   async deleteIt(): Promise<void> {
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -51,7 +51,7 @@ export class EWSAccount extends MailAccount {
     await this.commonLogin(true);
   }
 
-  async commonLogin(interactive): Promise<void> {
+  private async commonLogin(interactive): Promise<void> {
     if (this.authMethod == AuthMethod.OAuth2) {
       if (!this.oAuth2) {
         let urls = OAuth2URLs.find(a => a.hostnames.includes(this.hostname));
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -47,7 +47,11 @@ export class EWSAccount extends MailAccount {
     return this.authMethod != AuthMethod.OAuth2 || this.oAuth2?.isLoggedIn;
   }
 
-  async verifyLogin(interactive = true): Promise<void> {
+  async verifyLogin(): Promise<void> {
+    await this.commonLogin(true);
+  }
+
+  async commonLogin(interactive): Promise<void> {
     if (this.authMethod == AuthMethod.OAuth2) {
       if (!this.oAuth2) {
         let urls = OAuth2URLs.find(a => a.hostnames.includes(this.hostname));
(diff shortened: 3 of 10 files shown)


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -61,6 +59,12 @@ export class EWSAccount extends MailAccount {
       await this.oAuth2.login(interactive);
     }
     await this.listFolders();
+  }
+
+  async login(interactive: boolean): Promise<void> {
+    await ensureLicensed();
+    await super.login(interactive);
+    await this.verifyLogin(interactive);
```
benbucksch, line 67: Please make another function for that (which will be called by `login()` and `verifyLogin()`). Don't re-use `verifyLogin()`

app/logic/Chat/ChatAccount.ts
```diff
@@ -28,8 +28,8 @@ export class ChatAccount extends TCPAccount {
   }
 
   async save(): Promise<void> {
-    await super.save();
     await this.storage?.saveAccount(this);
+    await super.save();
```
benbucksch, line 32: I understand that this change was necessary to make it work, and that it works.

But this makes assumptions that shouldn't be there. This relies on subtle ordering, due to completely unrelated code which doesn't appear here. This is very likely to regress.

It also makes assumptions about the mail account being the primary account and the chat account being the secondary one. We should avoid such assumptions, if possible. And particularly, such assumptions must never be implicit, but either clearly obvious from the code, or documented on both sides of the assumption/link.

Can you find a fix that does not rely on this ordering, please?

benbucksch (changes requested):

app/logic/Chat/ChatAccount.ts
NeilRashbrook: There will always be an ordering, since the database won't allow a dependent account to be saved before its main account, so the only solution I could come up with was to make the ordering more explicit.

(Note that an ordering also applies to `deleteIt()` in that the `super.deleteIt();` must appear first.)

app/logic/Abstract/Account.ts
```diff
@@ -110,6 +110,12 @@ export class Account extends Observable {
   /** Saves the config in this account to disk.
    * Does not save the contents, e.g. messages. */
   async save(): Promise<void> {
+    throw new AbstractFunction();
+  }
+
+  /** Saves the config in this account and its dependent accounts to disk. */
+  async saveAll(): Promise<void> {
```
benbucksch, line 117: Good idea. In most cases when we save, we want to save only that one account, not the dependent accounts.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -47,9 +47,11 @@ export class EWSAccount extends MailAccount {
...
   }
 
-  async login(interactive: boolean): Promise<void> {
-    await ensureLicensed();
-    await super.login(interactive);
+  async verifyLogin(): Promise<void> {
+    await this.commonLogin(true);
+  }
+
+  async commonLogin(interactive): Promise<void> {
```
benbucksch, line 54:
```suggestion
  protected async loginCommon(interactive): Promise<void> {
```
(same elsewhere)

benbucksch (approved):
Thanks for this nice solution

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -47,9 +47,11 @@ export class EWSAccount extends MailAccount {
...
   }
 
-  async login(interactive: boolean): Promise<void> {
-    await ensureLicensed();
-    await super.login(interactive);
+  async verifyLogin(): Promise<void> {
+    await this.commonLogin(true);
+  }
+
+  private async commonLogin(interactive): Promise<void> {
```
benbucksch, line 54:
```suggestion
  private async loginCommon(interactive): Promise<void> {
```

benbucksch (changes requested):

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Can you please also change the function name?

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: * function name
* protected, not private. Subclasses are allowed to use it.

benbucksch: * the parameter type was missing, too.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: Sorry, I overlooked that you wanted the function name changed.

app/logic/Abstract/Account.ts
```diff
@@ -110,6 +110,12 @@ export class Account extends Observable {
...
+    await this.save(); // must be first, because of foreign key constraint
     for (let dependent of this.dependentAccounts()) {
       await dependent.save();
```
NeilRashbrook, line 120:
```suggestion
      await dependent.saveAll();
```

Bah, at one point I had this change locally, but I couldn't decide whether or not `saveAll` was a good idea, and removed the change, only to put `saveAll` back, but without the change. Sigh...

benbucksch: Are you saying that it's broken as landed?

benbucksch: If so, can you please make a PR to fix it?

NeilRashbrook: It doesn't do what the old code does, which is to recursively save all dependent accounts. (We don't actually have more than one nesting of account, especially not at the point of the only caller, so it's not broken in that sense.)

benbucksch: Oh, I see, you're thinking of multiple levels of dependent accounts.
Yes, it would be a nice change to make, but I agree: We don't have that case right now.

PR #582, opened 2025-04-26
Mobile: Backend: Vite config target for Node.js & ESM

jermy-c:
**Errors**

1. Our code is ESM and needs to be compiled to ESM
2. Vite lib mode targets to browser

**Notes**

- I remembered that we changed all modules to ESM so the build needs to target ESM otherwise there would be an error.
- I tried the `cjs` format but I kept getting errors. Was there a reason for setting the build to CJS?
- Another error, was that vite in `lib` mode only targets browser so it was bundling for the browser. [link](https://github.com/vitejs/vite/issues/8910#issuecomment-1174779427)
- I tried another fix by removing the `browser` from `resolve.mainFields`. [link](https://github.com/vitejs/vite/issues/8910#issuecomment-2128107137)
- So the fix that worked for me was setting `build.ssr` to true and` ssr: { noExternal: true }`. [link](https://github.com/vitejs/vite/issues/8910#issuecomment-1999152627)

### DIFF
--- a/mobile/backend/vite.config.ts
+++ b/mobile/backend/vite.config.ts
@@ -2,17 +2,19 @@ import { defineConfig } from 'vite';
 import nodeExternals from 'rollup-plugin-node-externals';
 
 export default defineConfig({
+  ssr: { noExternal: true },
   build: {
     target: 'node18',
     lib: {
       name: 'index',
       fileName: () => 'index.js',
-      formats: ['cjs'],
+      formats: ['es'],
       entry: './index.ts'
     },
     outDir: '../android/app/src/main/assets/public/nodejs',
     emptyOutDir: false,
-    minify: false
+    minify: false,
+    ssr: true,
   },
   plugins: [
     nodeExternals({


benbucksch:
Great! Thank you!!!

No, there was no particular reason for lib and cjs. The combination of lib and es didn't work, but I didn't realize that lib was for browser, due to its name I thought it was for node.js modules. That was my mistake. Thank you!

benbucksch:
Hey Jeremy, this runs when started directly from node on the commandline, as index.mjs file. However, capacitory-nodejs doesn't start it:
1. If I compile it as ESM to index.js, node complains that it's not a CJS module:
```
(node:4149) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
  (Use `node --trace-warnings ...` to show where the warning was created)
nodejs                  im.mustang.capa                      E  /data/data/im.mustang.capa/files/nodejs/public/index.js:1
  import require$$1$3 from "node:path";
  ^^^^^^                                                                                                    
SyntaxError: Cannot use import statement outside a module
...
nodejs                  im.mustang.capa                      E  Node.js v18.17.1
```
2. If I rename the file to index.mjs, it doesn't find it anymore:
```
Unable to access main script of the Node.js project. (No such file)
```
(Debug message only, in logcat)

benbucksch:
We're using https://github.com/hampoelz/Capacitor-NodeJS to start NodeJS on mobile. It may be that the fix is there.

benbucksch:
It starts node.js here:
https://github.com/hampoelz/Capacitor-NodeJS/blob/main/android/src/main/java/net/hampoelz/capacitor/nodejs/CapacitorNodeJS.java#L144

benbucksch:
I've seen that it reads the filename from a package.json file in the same directory, so I put this in there:
```
{
  "main": "./index.mjs"
}
```
and it starts the backend

PR #584, opened 2025-04-26
Mobile Backend: Make capacitor-nodejs start an ESM module

### LINKED ISSUE #583: Mobile: Backend: index.js interpreted as CommonJS, although it's an ESM
#582 runs when started directly from node on the commandline, as index.mjs file. 
However, capacitory-nodejs doesn't start it:

1. If I compile it as ESM to index.js, node complains that it's not a CJS module:
```
(node:4149) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
  (Use `node --trace-warnings ...` to show where the warning was created)
nodejs                  im.mustang.capa                      E  /data/data/im.mustang.capa/files/nodejs/public/index.js:1
  import require$$1$3 from "node:path";
  ^^^^^^                                                                                                    
SyntaxError: Cannot use import statement outside a module
...
nodejs                  im.mustang.capa                      E  Node.js v18.17.1
```

2. If I rename the file to index.mjs, it doesn't find it anymore:
```
Unable to access main script of the Node.js project. (No such file)
```
(Debug message only, in logcat)

_Originally posted by @benbucksch in https://github.com/mustang-im/mustang/issues/582#issuecomment-2832128588_

benbucksch:
This is an attempt to fix #583, but it breaks the gradle build in Android Studio.
It's a hack anyways. Just FYI.

### DIFF
--- a/mobile/backend/vite.config.ts
+++ b/mobile/backend/vite.config.ts
@@ -27,12 +27,12 @@ export default defineConfig({
       targets: [
         {
           src: 'dist/index.js',
-          dest: '../../android/app/src/main/assets/public/nodejs',
+          dest: '../../../app/dist/nodejs',
           rename: 'index.mjs',
         },
         {
           src: 'package-deploy.json',
-          dest: '../../android/app/src/main/assets/public/nodejs',
+          dest: '../../../app/dist/nodejs',
           rename: 'package.json',
         },
       ]
--- a/mobile/backend/vite.config.ts
+++ b/mobile/backend/vite.config.ts
@@ -1,5 +1,6 @@
 import { defineConfig } from 'vite';
 import nodeExternals from 'rollup-plugin-node-externals';
+import { viteStaticCopy } from 'vite-plugin-static-copy';
 
 export default defineConfig({
   ssr: { noExternal: true },
@@ -7,12 +8,13 @@ export default defineConfig({
     target: 'node18',
     lib: {
       name: 'index',
-      fileName: () => 'indexs.js',
       formats: ['es'],
       entry: './index.ts'
     },
-    outDir: '../android/app/src/main/assets/public/nodejs',
-    emptyOutDir: false,
+    //outDir: '../android/app/src/main/assets/public/nodejs',
+    //emptyOutDir: false,
+    outDir: 'dist',
+    emptyOutDir: true,
     minify: false,
     ssr: true,
   },
@@ -20,6 +22,20 @@ export default defineConfig({
     nodeExternals({
       deps: false,
       devDeps: true, // Use node.js internal modules
-    })
+    }),
+    viteStaticCopy({
+      targets: [
+        {
+          src: 'dist/index.js',
+          dest: '../../android/app/src/main/assets/public/nodejs',
+          rename: 'index.mjs',
+        },
+        {
+          src: 'package-deploy.json',
+          dest: '../../android/app/src/main/assets/public/nodejs',
+          rename: 'package.json',
+        },
+      ]
+    }),
   ],
 });
--- a/mobile/android/app/src/main/AndroidManifest.xml
+++ b/mobile/android/app/src/main/AndroidManifest.xml
@@ -7,7 +7,8 @@
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme"
+        android:usesCleartextTraffic="true">
 
         <activity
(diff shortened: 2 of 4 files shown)


mobile/backend/vite.config.ts
```diff
@@ -1,25 +1,41 @@
...
     nodeExternals({
       deps: false,
       devDeps: true, // Use node.js internal modules
-    })
+    }),
+    viteStaticCopy({
+      targets: [
+        {
+          src: 'dist/index.js',
+          dest: '../../../app/dist/nodejs',
```
jermy-c, line 30: This works but it needs to be copied to the specified webDir.

jermy-c:
The config with the `index.mjs` works but the problem was that before `cap run android` actually runs, `cap sync` is executed which overwrites the `public` dir with the `webDir`.

https://github.com/mustang-im/mustang/blob/ee0ff84a30cac8ae432b8dafb6bc0923b553c56e/mobile/capacitor.config.json#L4

So the solution was to just copy the `nodejs` content to the same `webDir`

jermy-c:
There was a `NET::ERR_CLEARTEXT_NOT_PERMITTED` Error which I added the permission to allow clear text as explained [here](https://kinsta.com/knowledgebase/net-err_cleartext_not_permitted/) but that may pose security issues in production. Maybe solution [2](https://kinsta.com/knowledgebase/net-err_cleartext_not_permitted/#2-allowlist-the-domain-using-network_security_configxml) would be better, just allow certain domains for cleartext.

jermy-c:
I just managed to connect to the backend and also check the frontend console.log and got the following error:
```
error.ts:6 Error: Error while trying to report error
    at message.js:155:16
    at new Promise (<anonymous>)
    at MLQ.makeCall (message.js:150:12)
    at qLQ.callRemote (protocol.js:184:31)
    at obj.js:213:55
    at UI (SQLDatabase.ts:14:24)
    at BYQ (SQLAccountsMigrate.ts:12:25)
    at wD (SQLDatabase.ts:17:9)
    at kB.readAll (SQLAccount.ts:67:29)
    at fr.readAll (SQLMailAccount.ts:67:33)
```
I think this should be an error because of the prebuild binary for `bettersqlite`.

jermy-c:
Note: I'm not sure if setting the env `WS_NO_BUFFER_UTIL` to true was part of the fix but I did that also. Because I had an `bufferutil.[some function] is not a function` error. https://github.com/websockets/ws/blob/master/doc/ws.md#ws_no_buffer_util

benbucksch:
> There was a `NET::ERR_CLEARTEXT_NOT_PERMITTED` Error which I added the permission to allow clear text as explained [here](https://kinsta.com/knowledgebase/net-err_cleartext_not_permitted/) but that may pose security issues in production. Maybe solution [2](https://kinsta.com/knowledgebase/net-err_cleartext_not_permitted/#2-allowlist-the-domain-using-network_security_configxml) would be better, just allow certain domains for cleartext.

Normally, `http://localhost` and `ws://localhost` should be exempted from this rule. I can see 3 reasons for this error:
* We are not using `localhost` (literally, not `ws://127.0.0.1` etc.). In this case, we need to fix that.
* Android complains also about localhost. In this case, we need to override this.
* The error is actually about the IMAP connection. We do need to make cleartext IMAP connections. STARTTLS starts as cleartext, then switches to encrypted later.

benbucksch:
> Error while trying to report error

This is a problem on JPC level. There was an exception, and the side which had the exception was not able to send the error over JPC to the other side.
This needs to be fixed either way, independent from the actual error. Error reporting has to work.

> this should be an error because of the prebuild binary for bettersqlite.

If so, that's a separate problem.

jermy-c:
> ```
> error.ts:6 Error: Error while trying to report error
>     at message.js:155:16
>     at new Promise (<anonymous>)
>     at MLQ.makeCall (message.js:150:12)
>     at qLQ.callRemote (protocol.js:184:31)
>     at obj.js:213:55
>     at UI (SQLDatabase.ts:14:24)
>     at BYQ (SQLAccountsMigrate.ts:12:25)
>     at wD (SQLDatabase.ts:17:9)
>     at kB.readAll (SQLAccount.ts:67:29)
>     at fr.readAll (SQLMailAccount.ts:67:33)
> ```

This was an error with:

- `__filename` still remaining in the build, we had error similar for our electron build with `__dirname` but we did a replace for that. I did manage to fix this with `@rollup/plugin-esm-shim` same first proposed solution we for our electron build.
- Another problem appears after fixing the previous point where the node native modules are not bundled into the build. `electron-builder` has `install-app-deps` which installs the node natives.
- packages affected `better-sqlite3`, `bufferutil`

benbucksch:
Great that you found this out! Looking forward to the fixes.

benbucksch:
Superceded by #590.

PR #586, opened 2025-04-29
Create explicit property to check for incoming meetings

NeilRashbrook:
I am confused as to why `EditEvent` has a check, then an `<hbox>`, then an `InvitationResponseButtons`, which itself then contains an `<hbox>` and a check, but I wasn't sure which could be removed.

### DIFF
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -13,7 +13,7 @@
             iconSize="16px"
             />
         {/if}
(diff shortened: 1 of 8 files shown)


benbucksch (approved):
Thanks

benbucksch:
> EditEvent has ... a check, then an InvitationResponseButtons, which itself then contains ... a check, but I wasn't sure which could be removed.

Each component lives on its own. A component should not assume anything outside itself, other than the defined input/output types. (Exceptions must be documented and should be avoided.)

"My component is my castle" ;-)

NeilRashbrook:
> ...

(Actually I was also wondering about the double `<hbox>`es.)

PR #587, opened 2025-04-29
VCard parser tests, part 2

NeilRashbrook:
vCard allows a TEL to have the tel: prefix, which I don't currently strip off. Should I do that?


benbucksch (approved):

benbucksch:
> vCard allows a TEL to have the tel: prefix, which I don't currently strip off. Should I do that?

You mean when you're writing vCards out?
Do whatever is the most compatible with the important vCard clients.

PR #589, opened 2025-04-30
Strip leading tel: when parsing vCards

NeilRashbrook:
Also added another test case.


benbucksch (approved):
I had to read that code 5 times to understand that you overwrite the `protocol` function parameter.

PR #590, opened 2025-04-30
Mobile Backend: Make capacitor-nodejs start an ESM module (#583)

jermy-c:
- Make capacitor-nodejs start an ESM module
- Allow clear text traffic for now to connect the frontend to the backend
- Copy the backend files to the `webDir` so it's copied to the android build files also and it prevents it from getting deleted by an overwrite
- Based on #584

### DIFF
--- a/mobile/backend/vite.config.ts
+++ b/mobile/backend/vite.config.ts
@@ -27,12 +27,12 @@ export default defineConfig({
       targets: [
         {
           src: 'dist/index.js',
-          dest: '../../android/app/src/main/assets/public/nodejs',
+          dest: '../../../app/dist/nodejs',
           rename: 'index.mjs',
         },
         {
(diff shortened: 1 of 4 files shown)


benbucksch (approved):

mobile/backend/vite.config.ts
```diff
@@ -1,25 +1,41 @@
...
     lib: {
       name: 'index',
-      fileName: () => 'indexs.js',
       formats: ['es'],
       entry: './index.ts'
     },
-    outDir: '../android/app/src/main/assets/public/nodejs',
-    emptyOutDir: false,
+    //outDir: '../android/app/src/main/assets/public/nodejs',
+    //emptyOutDir: false,
```
benbucksch, line 15:
```suggestion
```

Issue #588, opened 2025-04-30
Calendar: <WeekView start=> is null


benbucksch:
### Reproduction
1. Go to Contacts and select any contact
2. Go to Meet and start a test meeting, e.g. via [Start fake Meeting]. Enter the meeting.
3. Go to Calendar

### Actual result
* The Meet sidebar appears, because you are in a meeting and you switch to another app. (Correct)
* The Calendar app starts to appear, but freezes
* The app is completely frozen. No UI reacts anymore at all.
* On the console, you see error:
```
[HMR][Svelte] Unrecoverable HMR error in <WeekView>: next update will trigger a full reload
Proxy<WeekView> @ proxy.js?v=58d4fa39:380
create_else_block @ MainView.svelte:45
update @ MainView.svelte:26
update @ index.mjs:1347
flush @ index.mjs:1307
Promise.then
schedule_update @ index.mjs:1258
make_dirty @ index.mjs:2133
(anonymous) @ index.mjs:2171
(anonymous) @ MainWindow.svelte:6
set @ index.mjs:34
appbar_selectedApp_binding @ MainWindow.svelte:9
appbar_selectedApp_binding @ MainWindow.svelte:7
(anonymous) @ index.mjs:2169
click_handler @ AppBar.svelte:26
click_handler @ AppBar.svelte:6
(anonymous) @ index.mjs:1244
bubble @ index.mjs:1244
click_handler @ AppButton.svelte:12

WeekView.svelte:63 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toDateString')
    at $$self.$$.update (WeekView.svelte:63:24)
    at init (index.mjs:2176:8)
    at new WeekView (WeekView.svelte:101:22)
    at createProxiedComponent (svelte-hooks.js?v=58d4fa39:341:9)
    at new ProxyComponent (proxy.js?v=58d4fa39:242:7)
    at new Proxy<WeekView> (proxy.js?v=58d4fa39:349:11)
    at Array.create_else_block (MainView.svelte:45:24)
    at Object.update [as p] (MainView.svelte:26:28)
    at update (index.mjs:1347:36)
    at flush (index.mjs:1307:17)
```

### Expected result
* The Meet sidebar appears, because you are in a meeting and you switch to another app. (Correct)
* The Calendar app appears and works
* You can use Mustang normally

benbucksch:
I've added a `FakeMeeting`, which should make it trivial to reproduce this bug.

[...]

PR #596, opened 2025-05-01
Calendar: WeekView: Only use start if it's not undefined (#588)

jermy-c:
- Calendar: WeekView: Only use start if it's not undefined
- This seems to be the bug where when you switch apps the variable binding happens after mount so it's intially `undefined`


benbucksch (approved):
This is just papering over the issue. We need to find and fix the root cause, in #588.

However, until we found it, this is a fair workaround.
Note that this does not work correctly, either: Because `dateInterval` is undefined, it loads the WeekView, although it should load the DualView (called Day/Month in the UI).

After the root cause is found and fixed, we will need to revert this change here.

Issue #598, opened 2025-05-03
Meet: Gallery: Scroll snap into top of video


benbucksch:
### Reproduction
* Open Meet and start a fake meeting
* Add 20 participants
* Switch to Gallery view
* Switch to 2x2
* Scroll down

### Actual result
* The viewport scrolls so that you see only the bottom of the top videos, one row of full videos, and and the top of the bottom videos
* I.e. you see 4 half-videos and 2 full videos

### Expected result
* The top video should always start at the top of the viewport
* You do not see half videos
* You see 4 full videos

### Implementation
* Try to use CSS `scroll-snap-type: y mandatory` and `scroll-snap-align: start`. They are made for exactly this usecase. I couldn't get them to work, though. Maybe it's a browser implementation limitation connected to `flex` box? See git branch `ben/meet-scroll-snap` (git commit d18b77ef)

PR #599, opened 2025-05-05
Meet: Gallery: Scroll snap into top of video (#598)

jermy-c:
- The inner name label element also had a participant class so the `scroll-snap-align` was also being applied to that.
- Even with this fixed the there's still a little half video at the bottom.
- Because the `clientWidth` and `clientHeight` of the gallery includes the scroll it causes the size to be miscalculated.
- There's a way to fix but it's not ideal you would need to get the `clientWidth` and `clientHeight` of the `<Scroll>` component.

### DIFF
--- a/app/frontend/Meet/View/Gallery.svelte
+++ b/app/frontend/Meet/View/Gallery.svelte
@@ -80,7 +80,7 @@
   .gallery :global(.scroll) {
     scroll-snap-type: y mandatory;
   }
-  .participants :global(.participant) {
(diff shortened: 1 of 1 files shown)


benbucksch (approved):

benbucksch:
> Because the clientWidth and clientHeight of the gallery includes the scroll it causes the size to be miscalculated.

Ah, thanks for finding that. I guessed it was something like this, but didn't find the cause.

PR #602, opened 2025-05-08
Mobile: Build config

jermy-c:
- Configured Vite to build to `mobile/dist`
- Configured scripts to build node natives to desired architecture
- Added docs on how to build for Android
- Sqlite still doesn't work after this because the OS dir path is wrong


docs/mobile/setup.md
```diff
@@ -0,0 +1,12 @@
+### Environment
+
+1. Node.js v18, `node-gyp` will have issues with anything above
+
+
+### Android
+
+1. Install [Android NDK](https://developer.android.com/ndk/index.html)
+2. Set `$ export ANDROID_NDK_HOME=/Users/username/Library/Android/sdk/ndk-bundle`
+3. Make sure have python2 installed with the paths set, [pyenv](https://github.com/pyenv/pyenv) can with that
```
benbucksch, line 10: "pyenv can with that"?


```suggestion
3. Make sure have python2 installed with the paths set, [pyenv](https://github.com/pyenv/pyenv) can help with that
```

docs/mobile/structure.md
```diff
@@ -0,0 +1,3 @@
+### Android
+
+The app files should be in `[homedir]/data/im.mustang.capa`. Writing files outside of this directory cause problems and requires extra permissions.
```
benbucksch, line 3: What is [homedir]? You mean my user dir on Unix = ~ = /home/jeremy/ on my development machine?
That won't fly. You need to make all paths relative to the source directory.

Or do you mean within the app bundle? or the user's drive? please specify.

mobile/src/index.html
```diff
@@ -1,32 +1,17 @@
...
-    <script
-      nomodule
-      src="https://unpkg.com/@ionic/pwa-elements@latest/dist/ionicpwaelements/ionicpwaelements.js"
-    ></script>
+<head>
+  <title>Mustang</title>
+  <meta charset="UTF-8" />
+  <link rel="icon" type="image/svg+xml" href="/icon/logo.svg" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="Content-Security-Policy" content="default-src 'none'; connect-src * data:; script-src 'self'; frame-src 'self' data:; style-src 'self' 'unsafe-inline'; img-src 'self' blob: data: https:; media-src 'self' blob: https://www.mustang.im; font-src 'self';">
```
benbucksch, line 9: Can we use the one from app/index.html ? Isn't that in dist/ anyway?

We may change the CSP or other parts. Duplication is a problem.

mobile/vite.config.ts
```diff
@@ -1,11 +1,35 @@
...
-  publicDir: './njs',
+  publicDir: '../../app/public',
   build: {
     outDir: '../dist',
     minify: false,
     emptyOutDir: true,
+    sourcemap: production,
   },
+  plugins: [
+    conditionalCompile({
```
benbucksch, line 18: Avoid copying. Can you include the other vite config ? Like (similar to):
```
import appViteConfig from "../app/vite.config";

defineConfig({
  plugins: appViteConfig.plugins,
});
```

benbucksch (changes requested):

docs/mobile/structure.md
jermy-c: Sorry, I meant on the Android device, it should be `data/data/im.mustang.capa`. But the first `data` dir is return by `os.homedir()`. According to the [docs](https://nodejs-mobile.github.io/docs/api/differences#oshomedir)

mobile/vite.config.ts
jermy-c: Importing doesn't work. I reverted it to just execute the command in `../app` and then just copy the files over to `mobile/dist`. I moved everything over to make it similar to `e2` but that doesn't work well. Right now, it just executes and copies to `mobile/dist` to make less confusing and `mobile/vite.config.ts` is not used now.

mobile/src/index.html
jermy-c: Yes, changed it back to use the one in `/app`.

docs/mobile/structure.md
```diff
@@ -0,0 +1,13 @@
+### Android
+
+The app config directory and files directory should be in `/data/data/im.mustang.capa` on the Android device. Writing files outside of this directory cause problems and requires extra permissions. Files directory could be outside but we need to find the API to get the specific path.
```
benbucksch, line 3: Thank you for documenting that! This is great (and needed).

benbucksch (approved):

PR #603, opened 2025-05-09
Mobile backend: Point to app dir

jermy-c:
- Access the app directory only, app data should be saved within the app directory only

### DIFF
--- a/backend/backend.ts
+++ b/backend/backend.ts
@@ -398,10 +398,14 @@ function directory(type: string): string {
  * Mac OS: /Users/USER/Library/Application Support/Mustang
  */
 function getConfigDir(): string {
+  let dir: string = os.homedir();
   if (os.platform() == "android") {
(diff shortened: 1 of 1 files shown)


backend/backend.ts
```diff
@@ -414,7 +421,14 @@ function getConfigDir(): string {
  * Mac OS: /Users/USER/Library/Mustang
  */
 function getFilesDir(): string {
-  return os.homedir();
+  let dir: string = os.homedir();
```
benbucksch, line 424: Can you cache it, please?

benbucksch (changes requested):

benbucksch (approved):
Normally, we write this as "early exit", i.e.
```
if (configDir) {
  return configDir;
}
```

That saves 1 indention level for the entire function and is easier to understand.

But we can fix that later.

PR #608, opened 2025-05-12
Don't restart editing an event when the editor is refeshed

NeilRashbrook:
Certain actions can cause the dialog header element to refresh, which will then attempt to restart editing of the event.

### DIFF
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -241,9 +241,11 @@ export class Event extends Observable {
   }
 
   /**
-   * Starts editing an event, if it hasn't already started being edited.
-   * This can happen when switching between the event editor and another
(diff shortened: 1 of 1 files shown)


app/logic/Calendar/Event.ts
```diff
@@ -240,7 +240,15 @@ export class Event extends Observable {
     );
   }
 
+  /**
+   * Starts editing an event, if it hasn't already started being edited.
+   * This can happen when switching between the event editor and another
+   * screen, or certain updates within editing events can also trigger it.
```
benbucksch, line 246:
```suggestion
   * Starts editing an event.
   *
   * If it is already being edited, this is a no-op. This can happen when switching
   * between the event editor and another screen, or certain updates within
   * editing events can also trigger it.
```

benbucksch (approved):

PR #611, opened 2025-05-14
Calendar: Edit: Reminder: Fix to allow editing it

NeilRashbrook:
The duration in unit was not calling the correct function to notify the duration unit widget of the change.

Furthermore, there was an update loop that prevented you from change the duration after changing the duration unit.

Additionally the `min={0}` was not actually achievable.

This meant that the only reminders possible were 5 minutes, 1 hour and 1 day.

I also checked to ensure that the event duration UI was unaffected.

### DIFF
--- a/app/frontend/Calendar/EditEvent/ReminderBox.svelte
+++ b/app/frontend/Calendar/EditEvent/ReminderBox.svelte
@@ -1,6 +1,6 @@
 <hbox title={$t`Reminder`} flex>
-  <input class="duration" type="number" bind:value={durationInUnit} on:input={durationUnit.onDurationChanged} min={0} />
-  <DurationUnit bind:durationInSeconds={beforeInSec} bind:durationInUnit bind:this={durationUnit} on:change={onChanged} />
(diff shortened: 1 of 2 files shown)


app/frontend/Calendar/EditEvent/ReminderBox.svelte
```diff
@@ -27,16 +27,9 @@
...
+  let beforeInSec = $event.alarm ? ($event.startTime.getTime() - $event.alarm.getTime()) / 1000 : 0;
 
-  function onChanged(ev: CustomEvent<number>) {
-    let seconds = ev.detail;
-    if (!event.alarm) {
-      event.alarm = new Date();
-    }
-    event.alarm.setTime(event.startTime.getTime() - seconds * 1000);
-    event.notifyObservers();
-  }
```
benbucksch, line 39: hm, this is some very odd code, so this was definitely deliberately there, to fix a certain case/bug.
Are you sure you tested all cases, and that this is not needed? If so, all the better. I Just want to make sure that this is not regressing a bugfix.

benbucksch (approved):

PR #617, opened 2025-05-16
Windows: Code signing

jermy-c:
- Here are changes for CI to code sign our app
- We need to test if setting these env variables on a local machine code signs successfully and doesn't break the code
- I've set different CSC_* secret names so we can have one for Parula and one for Mustang
- If we're setting the CSC_LINK directly in base64-encoding then we might get a character limit error, then we need fix that by creating a file and writing the certificate there and link to that

### DIFF
--- a/.github/workflows/installer-windows-parula.yml
+++ b/.github/workflows/installer-windows-parula.yml
@@ -26,6 +26,8 @@ jobs:
     - run: cd e2; yarn run build:release:win
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        CSC_LINK: ${{ secrets.CSC_LINK_WIN_PARULA }}
+        CSC_KEY_PASSWORD: ${{ secrets.CSC_KEY_PASSWORD_WIN_PARULA }}
         SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
         # For Sentry source maps. Otherwise, vite build is out of memory
         NODE_OPTIONS: '--max-old-space-size=32768'
--- a/.github/workflows/installer-windows.yml
+++ b/.github/workflows/installer-windows.yml
@@ -26,6 +26,8 @@ jobs:
     - run: cd e2; yarn run build:release:win
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        CSC_LINK: ${{ secrets.CSC_LINK_WIN_MUSTANG }}
+        CSC_KEY_PASSWORD: ${{ secrets.CSC_KEY_PASSWORD_WIN_MUSTANG }}
         SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
         # For Sentry source maps. Otherwise, vite build is out of memory
         NODE_OPTIONS: '--max-old-space-size=32768'


benbucksch:
Where can we put the certificate? Is my understanding correct that it contains the private key used for signing? I am not comfortable putting that on a publicly accessible HTTP URL, even if it's encrypted and you need the password to use it.

benbucksch:
[Options for certs](https://learn.microsoft.com/en-us/windows-hardware/drivers/dashboard/code-signing-reqs#where-to-get-ev-code-signing-certificates):
* [ssl.com Normal, in Cloud or with app](https://www.ssl.com/certificates/code-signing/): 65 USD/year
* [ssl.com EV, in Cloud or with app](https://www.ssl.com/certificates/ev-code-signing/): 250 USD/year -- 
* [Certum Normal in Cloud, for Open Source](https://shop.certum.eu/open-source-code-signing-on-simplysign.html): 50 USD/year

* In cloud =  using a HTTP API
* In app = custom, CA-specific Linux commandline app to install and run
* Cert = We download the cert as file, put it somewhere for the CI process to use (main question: where?)
* Hardware = The key is stored on a hardware device. Not an option for CI.

both in cloud and in app require would be a few lines of scripting to integrate into CI

jermy-c:
> Where can we put the certificate?

We'd have to put it as a Github secret. It should not be publicly seen but it'll be on Github similar to the Mac builds.

benbucksch:
Another option is that we use the cloud signing of ssl.com . That would allow us to use their EV certificate or the cheaper normal one.

jermy-c:
## CI setup

### Local file Setup

A local file setup would look something like what we have for Mac:

https://github.com/mustang-im/mustang/blob/59a08095fc55c55bf95eabe5dcf9d03725091672/.github/workflows/installer-mac-parula.yml#L27-L37

1. We write the file for `CSC_LINK`, for our Mac OS example we have `~/private_keys/AuthKey_${{ secrets.APPLE_API_KEY_ID }}.p8`
2. And then we had `APPLE_API_KEY` linking to that written file
3. And the password as `CSC_KEY_PASSWORD`.
4. Since it's not used an ENV then there wouldn't be any character limit. The Github secrets character limit is 48KB and the Windows ENV character limit is 8192 characters.

I would need find the equivalent command for Windows for that to work.

### Cloud Setup

For a cloud setup it would be the HTTPS link for `CSC_LINK` instead of the local file location. And the password `CSC_KEY_PASSWORD`.

## Github Secrets Setup

We'd need to setup secrets for the Repository.

https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository

benbucksch:
Ah, nice, I didn't know that you can map GitHub Secrets to files in CI.

> For a cloud setup it would be the HTTPS link for CSC_LINK

Does that work with the ssl.com REST API for signing? Can you please check that?

benbucksch:
How would we do 2 different signing keys for Mustang and Parula?

benbucksch:
UPDATE: I see that you already did it for Mustang and Parula brands in the PR. What you did makes sense. Thank you!

benbucksch:
Given that ssl.com is deceptive about the price (the price for 1 year is actually twice as high as adverized), I decided to go with certum.eu .

jermy-c:
I've been searching if there's a command line tool or workflow provided by Certum to enable signing on CI but so far I haven't found any. They require you to use [SimplySign](https://support.certum.eu/en/installation-of-the-simplysign-application/) to install the certificate on the system and then use SignTool to sign the app.

benbucksch:
From what I understood, they allow you to download the certificate, so it's a file on the local machine. The certificate is a standardized format, normally `.p12` , and standard Open-Source signing tools can be used.

jermy-c:
> From what I understood, they allow you to download the certificate, so it's a file on the local machine.

Open Source Code Signing - set

- Uses a cryptographic card
- Downloads to your local machine

Open Source Code Signing in the Cloud

- SimplySign Desktop application for computer to connect to virtual cryptographic card in the cloud
- SimplySign installs and manages certificates on your machine

benbucksch:
I got the "Open Source Code Signing in the Cloud".

Are you sure you need *their* SimplySign tool? If we get a standard x509 certificate as file, then we can use *any* Open-Source sign tool.

jermy-c:
> Are you sure you need their SimplySign tool? If we get a standard x509 certificate as file, then we can use any Open-Source sign tool.

I'm hoping we can just Sign in with their tool and export the certificate as a file and also have the password for it, then just upload it at the minimum.

However, looking at the [requirements](https://cabforum.org/working-groups/code-signing/requirements/#122-relevant-dates) set by CA/Browser forum which are the standards they follow, it seems like the private key must be on a Hardware module or at least something similar.

> Effective June 1, 2023, for Code Signing Certificates, CAs SHALL ensure that the Subscriber’s Private Key is generated, stored, and used in a suitable Hardware Crypto Module that meets or exceeds the requirements specified in section 6.2.7.4.1 using one of the methods in 6.2.7.4.2.

If we have to use their tool, then we might need to either:

- Code sign manually on a local machine
- Host our own runner with the SimplySign tool installed and keep the certificate there
- Use another service which supports CI
- Use Azure Trusted Signing

benbucksch:
Running **their** "desktop software" is not an option. Let's forget about Certum, then. I asked for a refund.

jermy-c:
## Azure Trusted Signing

### Pricing

$9.99/month for 5000 signatures

image

### CI Process

`electron-builder` offers this [feature](https://www.electron.build/code-signing-win#using-azure-trusted-signing-beta) natively but it's still in beta.

1. We'd have to fill out the fields listed here: https://www.electron.build/app-builder-lib.interface.windowsazuresigningconfiguration
2. We'd need fill out the Azure tenant environment variables for the minimal App registration setup, `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`

## SSL.com

### Pricing

For CI to work, we'd need 2 services: the certificate and the eSigner service(so you don't need a physical USB)

**Certificate Pricing**

image

**eSigner Pricing**

image

Only the first 30 days include eSigner for free

image

### CI Process

They have an [article](https://www.ssl.com/how-to/cloud-code-signing-integration-with-github-actions/) for the using eSigner with GitHub Actions.

benbucksch:
I understood that the "cloud signing" is included in the ssl.com code signing cert. If this costs $1 USD per signed file on top, that's too expensive.

Let's try to use Azure, then.

Issue #621, opened 2025-05-17
Invitation: Response is not saved


benbucksch:
### Reproduction
1. neil@be as EWS account
2. Inbox, show email "Test meeting request"
3. Click "Confirm"
4. Click on other email
5. Click on the email from step 2/3 again
6. Go to Sent folder, open "Accepted: ..." email
7. Click "Update calendar"
8. Go back to Inbox, click on email from step 2/3 again

### Actual result
* After step 3, the invitation shows [Confirmed] (correct)
* After step 5, the invitation shows [Confirm] [Reject] (that's the bug)
* After step 8, the invitation shows [Confirmed] (correct)

### Expected result
* The invitation shows [Confirmed] in steps 3, 5, and 8

![Image](https://github.com/user-attachments/assets/294823ac-9469-476d-a60d-888e2b7a55d2)

benbucksch:
This is similar to #619, but not the same.

NeilRashbrook:
This is because at step 5 we haven't actually synced the calendar yet, so we don't find the meeting there. We should probably do that.

benbucksch:
We should sync the calendar (to the server) in step 3, when I confirm, not when loading the email.

In step 5, we should make sure that the calendar is loaded (from the DB).

PR #625, opened 2025-05-19
Sync calendar after accepting invitation #621

NeilRashbrook:


app/logic/Calendar/ActiveSync/ActiveSyncIncomingInvitation.ts
```diff
@@ -39,6 +39,7 @@ export class ActiveSyncIncomingInvitation {
     this.event.myParticipation = response;
     await this.event.respondToInvitation(response, this.calendar.account); // needs 16.x to do this automatically
     await this.message.deleteMessageLocally(); // Exchange deletes the message from the inbox
+    await this.calendar.listEvents();
```
benbucksch, line 42: 1. Shall we be more specific and use the new `fetchFromServer()` function? Or do we not know the evffefnt yet?
2. How does the UI get updated? If the new event comes in, how does the Invitation here know to find the event and update the UI?

app/logic/Calendar/OWA/OWAIncomingInvitation.ts
```diff
@@ -38,8 +38,18 @@ export class OWAIncomingInvitation {
...
     await this.message.deleteMessageLocally(); // Exchange deletes the message from the inbox
+    let itemID = result.Items.find(item => item?.__type == "CalendarItem:#Exchange")?.ItemId?.Id;
+    if (itemID) {
+      let events = new ArrayColl<OWAEvent>();
+      await this.calendar.getEvents([itemID], events);
+      for (let event of events) {
+        if (!this.calendar.events.contains(event)) {
+          this.calendar.events.add(event);
+        }
+      }
```
benbucksch, line 51: 1. Move into a function with an appropriate name, like `getEventFromServerByID()` or similar.
2. So, we add it to the calendar. And then? How does it update the Invitation (same as question 2 above).

app/logic/Calendar/ActiveSync/ActiveSyncIncomingInvitation.ts
NeilRashbrook: 1. We're going to have to Sync the calendar sooner or later, so we might as well do it now.
2. While you're still viewing the email (which should never happen, because Exchange deletes it) the UI keeps track of the response. If you get an update for that event, and view that email, the event in the calendar will be found due to its matching calUID. (This applies to all protocols.)

benbucksch (approved):

app/logic/Calendar/OWA/OWAIncomingInvitation.ts
benbucksch: Thanks

PR #629, opened 2025-05-20
IMAP: Gmail: Get new access on reconnect (#350)

jermy-c:
- Gmail access tokens are only valid for a single continuous connection, therefore we getting an error whenever reconnecting without getting a new access token
- You have to set `alwaysNewToken` after or during login because on initialization `hostname` is not set yet

### DIFF
--- a/app/logic/Auth/OAuth2.ts
+++ b/app/logic/Auth/OAuth2.ts
@@ -296,12 +296,8 @@ export class OAuth2 extends Observable {
     if (this.expiryTimout) {
       clearTimeout(this.expiryTimout);
     }
-    try {
-      if (this.refreshToken) {
-        await this.getAccessTokenFromRefreshToken(this.refreshToken);
-      }
-    } catch (ex) {
-      this.refreshErrorCallback(ex);
+    if (this.refreshToken) {
+      await this.getAccessTokenFromRefreshToken(this.refreshToken);
     }
   }
 
--- a/app/logic/Auth/OAuth2.ts
+++ b/app/logic/Auth/OAuth2.ts
@@ -293,9 +293,6 @@ export class OAuth2 extends Observable {
   }
 
   async refreshImmediately() {
-    if (this.expiryTimout) {
-      clearTimeout(this.expiryTimout);
-    }
     if (this.refreshToken) {
       await this.getAccessTokenFromRefreshToken(this.refreshToken);
     }
--- a/app/logic/Auth/OAuth2.ts
+++ b/app/logic/Auth/OAuth2.ts
@@ -292,6 +292,19 @@ export class OAuth2 extends Observable {
     }
   }
 
+  async refreshImmediately() {
+    if (this.expiryTimout) {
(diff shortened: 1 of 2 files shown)


jermy-c:
I've tested this fix on my machine and it seems to work for me

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -217,7 +224,7 @@ export class IMAPAccount extends MailAccount {
     this.notifyObservers();
 
     if (this.authMethod == AuthMethod.OAuth2 && this.oAuth2 &&
-        !this.oAuth2?.isLoggedIn) {
+        !this.oAuth2?.isLoggedIn || this.alwaysNewToken) {
```
benbucksch, line 227: I don't think this can work reliably, because `OAuth2.login()` does **not** refresh the access token:
```
async login(interactive: boolean): Promise<string> {
  if (this.accessToken) {
    return this.accessToken;
  }
```

benbucksch (changes requested):

app/logic/Mail/IMAP/IMAPAccount.ts
jermy-c: Yeah, you're right. I changed it to `refreshInSeconds(0)`. I'm going to test if it works.

jermy-c:
The solution with `refreshInseconds(0)` doesn't work.

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -224,9 +228,12 @@ export class IMAPAccount extends MailAccount {
     this.notifyObservers();
 
     if (this.authMethod == AuthMethod.OAuth2 && this.oAuth2 &&
-        !this.oAuth2?.isLoggedIn || this.alwaysNewToken) {
+        !this.oAuth2?.isLoggedIn) {
       await this.oAuth2.login(false);
     }
+    if (this.isGMail()) {
```
benbucksch, line 234:
```suggestion
    if (this.alwaysNewOAuth2AccessToken) {
```

app/logic/Auth/OAuth2.ts
```diff
@@ -292,6 +292,19 @@ export class OAuth2 extends Observable {
...
+  async refreshImmediately() {
+    if (this.expiryTimout) {
+      clearTimeout(this.expiryTimout);
+    }
+    try {
+      if (this.refreshToken) {
+        await this.getAccessTokenFromRefreshToken(this.refreshToken);
+      }
+    } catch (ex) {
+      this.refreshErrorCallback(ex);
```
benbucksch, line 304: Throw into caller

jermy-c: Ok, thanks for checking that.

benbucksch:
I am not sure that this will work, because the logs in #350 show that we're not getting a refresh token, so we cannot refresh. Are you sure that this fix works?

benbucksch:
The fix was in PR #635

PR #635, opened 2025-05-22
OAuth2: Use refresh token from storage, not only in first call, but also after expiry #350

benbucksch:


jermy-c (approved):
I see this.refreshToken defined with a token and refreshing being executed by setting console.log() and setting the refresh time to 20 seconds.

PR #647, opened 2025-05-30
MailComposer: Save draft in correct folder (#646)

### LINKED ISSUE #646: Mail: Draft is not save in the selected identity folder
### Reproduction

1. Write an email
2. Change the identity
3. Click `Save`

### Actual Result

1. Draft is saved in the initial identity mail box not to the changed identity
2. E.g. initial identity is `john@example.org`, changed identity to `jane@example.com` but mail is still saved to `john@example.org`

### Expected Result

1. Draft is saved to the changed identity
2. E.g. initial identity is `john@example.org`, changed identity to `jane@example.com` but mail is still saved to `jane@example.com`

jermy-c:
- Emails were saving to the initial identity even when the identity is changed
- The `mail.folder` is now set before saving

### DIFF
--- a/app/frontend/Mail/Composer/MailComposer.svelte
+++ b/app/frontend/Mail/Composer/MailComposer.svelte
@@ -238,7 +238,7 @@
   $: fromIdentity && setAuthor()
   function setAuthor() {
     mail.identity = fromIdentity;
-    mail.folder = fromIdentity.account.getSpecialFolder(SpecialFolder.Sent)
+    mail.folder ??= fromIdentity.account.getSpecialFolder(SpecialFolder.Sent)
       ?? fromIdentity.account.inbox;
     if (fromIdentity == mail.identity) {
       return; // don't overwrite concrete email address for catch-all identity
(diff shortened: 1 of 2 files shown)


benbucksch (changes requested):
Instead, in line 241, change `??=` to `=` .

Test various scenarios to make sure it doesn't regress something.

jermy-c:
> Instead, in line 241, change ??= to = .

This works.

I'm getting some errors when deleting the saved draft for EWS. And it seems to be saving a new copy if you're editing the draft.

1. `this.folder.account.callEWS is not a function`
2. `The specified object was not found in the store., The process failed to get the correct properties.`

benbucksch:
Are these bugs new regressions from this change here?

I suspect that we need to check whether the folder changed, and delete the old draft from the old folder.

NeilRashbrook:
Once you edit a draft, you're changing an email which thinks it knows what its folder and account are, and doesn't take very kindly to having these things changed for it. For instance, the draft folder still has the draft email in its messages list.

From my point of view the ideal solution would be to clone the email and edit that instead, keeping a reference to the original email so that it can be deleted (seems simpler than looking for it).

PR #651, opened 2025-06-11
Test: FakeData

jermy-c:
- FakeData for Email works, moving, saving, editing emails work
- FakeData for Files work
- FakeData for Event not working yet, missing organizer error

### DIFF
--- a/app/logic/testData.ts
+++ b/app/logic/testData.ts
@@ -253,12 +253,19 @@ class FakeEMail extends EMail {
   get compose() {
     return new FakeComposeActions(this);
   }
+  async download() {}
 }
 
 class FakeComposeActions extends ComposeActions {
+  declare email: FakeEMail;
   async saveAsDraft(): Promise<void> {
     this.email.sent = new Date();
     await super.saveAsDraft();
+  }
+  async send() {
+    this.email.sent = new Date();
+    await this.email.identity.account.send(this.email);
+    await this.deleteDrafts();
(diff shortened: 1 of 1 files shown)


app/logic/testData.ts
```diff
@@ -23,6 +23,7 @@ import { DummyAddressbookStorage } from './Contacts/SQL/DummyAddressbookStorage'
 import { notifyChangedProperty } from './util/Observable';
 import { ArrayColl, type Collection } from 'svelte-collections';
 import { faker } from '@faker-js/faker';
+import { ComposeActions } from './Mail/ComposeActions';
```
benbucksch: Please move this below `EMail` import

app/logic/testData.ts
```diff
@@ -201,6 +202,17 @@ class FakeFolder extends Folder {
   newEMail(): FakeEMail {
     return new FakeEMail(this);
   }
+  async addMessage(message: EMail): Promise<void> {
```
benbucksch:
```suggestion
  async addMessage(message: EMail) {
```

app/logic/testData.ts
jermy-c: Ok, done in commit https://github.com/mustang-im/mustang/pull/651/commits/3b3595174be48ba2c2d8f4218b42644e066f64c1

benbucksch (approved):

PR #652, opened 2025-06-11
Files: Set icon size to normal for MacOS

jermy-c:
- Setting size to `large` was causing the Files app to crash whenever there are files to display
- size `large` is not supported on MacOS

https://www.electronjs.org/docs/latest/api/app#appgetfileiconpath-options

### DIFF
--- a/backend/backend.ts
+++ b/backend/backend.ts
@@ -469,7 +469,9 @@ async function getIconForFileType(ext: string, mimetype: string): Promise<string
 
 /** @returns data: URL */
 async function getIconForLocalFile(fullPath: string): Promise<string> {
-  let image = await app.getFileIcon(fullPath, { size: 'large' });
+  let platform = os.platform();
+  // size `large` not supported on MacOS, it causes the app to crash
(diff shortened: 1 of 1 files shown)


backend/backend.ts
```diff
@@ -469,7 +469,9 @@ async function getIconForFileType(ext: string, mimetype: string): Promise<string
 
 /** @returns data: URL */
 async function getIconForLocalFile(fullPath: string): Promise<string> {
-  let image = await app.getFileIcon(fullPath, { size: 'large' });
+  let platform = os.platform();
+  // size `large` not supported on MacOS, it causes the app to crash
+  let image = await app.getFileIcon(fullPath, { size: platform == 'darwin' ? 'normal' : 'large' });
```
benbucksch:
```suggestion
  // size `large` not supported on MacOS, it causes the app to crash
  let isMac = os.platform() == "darwin";
  let image = await app.getFileIcon(fullPath, { size: isMac ? "normal" : "large" });
```

benbucksch (approved):

Issue #654, opened 2025-06-16
Calendar: Recurrence: Fix bugs


benbucksch:
* Fix Neil's comments from https://github.com/mustang-im/mustang/commit/f71507fc84d5fadcfad8a94ecb4b66da7bc528d9#commitcomment-159233702
* Test and fix other recurrence bugs
* Make it stable and reliable

FYI, the change in git commit f71507fc was necessary, because the `calendar.events` were not observed anymore and therefore events added/remove were not showing up. This change fixed a good number of bugs, and should stay. Please fix the current code based on that.

benbucksch:
* `Event.save()` should invalidate and regenerate outdated instances in `this.instances`
* `SQLEvent.read()` should fill all recurrences, including normal instances.

benbucksch:
collection.flatMap() is now observable and also observes the source collections.

NeilRashbrook:
I've added six commits to branch `cal-recur`. It's now too late for me to work out how to fix `truncateRecurrence`.

PR #655, opened 2025-06-16
Don't regenerate instances in UI code

NeilRashbrook:
Well, the good news is that clearing exceptions when recurrence rules change is now centralised in save like you asked... but the bad news is that this means that all callers need to start editing the event so that save knows what the old rule was...

### DIFF
--- a/app/logic/Calendar/RecurrenceRule.test.ts
+++ b/app/logic/Calendar/RecurrenceRule.test.ts
@@ -2,7 +2,7 @@ import { expect, test } from 'vitest';
 import { Frequency, Weekday, RecurrenceInit, RecurrenceRule } from './RecurrenceRule';
 
 function check(calString: string, data: RecurrenceInit, expected: [number, number, number][]) {
-  data.duration = 36e5;
+  data.masterDuration = 36e5;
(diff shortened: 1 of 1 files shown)


benbucksch:
The API for `save()` (or `saveLocally()`) should not change. `startEditing()` is only so that we can revert back to original state, and check whether there are changes. The protocol implementations shouldn't use it.

I see why you do that, in saveLocally(). However, we should be able to simply delete the cached instances and re-create them. See `recurrenceCache` described in https://github.com/mustang-im/mustang/issues/654#issuecomment-2977309371

NeilRashbrook:
I've finished my new approach to moving the logic for regenerating recurrences out of the UI.

benbucksch:
* Make `recurrenceRule` a setter. It should accept null, which is your clear(). This simplifies a lot of the callers. Use `?:` operator where necessary, e.g. in OWAEvent
* I am fuzzy about the duration. Why do you check only the duration and not the master starttime and end time?
* One alternative to storing duration on the recurrencerule would be to pass in as function arguments the old and new master event, if that's readily available in the callers of `isCompatible()`.

PR #656, opened 2025-06-17
Allow to change just the recurrence of a series

NeilRashbrook:
Steps to reproduce problem:
1. Create a new event recurring every Tuesday
2. Edit the event and try to change only the series to recur every Tuesday and Thursday

Expected result: Menu option to change entire series

Actual result: Only cancel option available

This change makes the menu option appear whenever you have a suitable instance open, although if you haven't made any other changes then the "Change only this instance" becomes "Cancel".

I've also made the "entire series" options appear on any generated instance, not just the first and middle ones.


benbucksch:
* I've done the delete button change on master
* I've changed (on master) `canSaveSeries` into `hasMinimalProps` , allowing you to adapt `canSaveSeries` to how it make sense for the UI buttons. Please adapt `canSaveSeries`  accordingly.
* When the user changed a single event into a recurrence, the [Revert] and [Save] buttons should show, not [Cancel]. That's already working.
* If the recurrence of an instance is changed, it actually changes the master, i.e. the entire series, so the buttons for changing the series (rest of series, or entire series) should show, and a [Revert] button, not a [Cancel] button

NeilRashbrook:
Good news, showing [Revert] and [Save] (with menu) simplifies the diff considerably. The annoying thing is that there's no easy way to tell whether the recurrence has changed, so that the buttons always show for instances. Maybe there needs to be a recurrence rule variable in `EditEvent` which the recurrence box keeps up-to-date (when it's open)?

benbucksch (approved):

PR #657, opened 2025-06-17
Observe when events are added or removed

NeilRashbrook:
> FYI, the change in git commit f71507f was necessary, because the calendar.events were not observed anymore and therefore events added/remove were not showing up. This change fixed a good number of bugs, and should stay. Please fix the current code based on that.

Sorry, but this was never true; the events collection had never been observed which is why nothing happened when events were asynchronously added or deleted.

With these changes, when I click the Sync button all of the additions and deletions show up as expected. (Other changes sometimes show up and sometimes don't, so I suspect there are more places where events aren't being properly observed.)


benbucksch (approved):

NeilRashbrook:
Ben says that even if we're only changing the recurrence of a series there should still be a revert button, this should replace the cancel option that I had added.

Issue #361, opened 2025-01-08
FastList: Scrolling too fast when using a touchpad


benbucksch:
### Reproduction
1. Use a MacBook, or maybe other notebook with a touchpad
2. Open a mail folder, and mouse over the message list
3. On the touchpad, move 2 fingers up and down

### Actual result
* It scrolls really fast. With a small movement, you scroll 6 months back
* It works fine with a mouse scroll wheel: One move with the scroll wheel moves 3 messages up/down.

### Expected
* When usign a touchpad, it scrolls roughly as fast as it does with a mouse scroll wheel.

PR #658, opened 2025-06-17
FastList: Scroll speed normalized (#361)

jermy-c:
- the wheel event seems to be triggering a lot per second which caused the scrolling to be too fast, however, the deltaY was only 1 or 2
- when you use a scroll wheel there's less events triggered per second but the deltaY was much more
- I'm proposing throttling onWheel callback to prevent scrolling too fast for trackpads but also we need to adjust the scroll based on the deltaY for scroll wheels otherwise it would be too slow for scroll wheels
- this will be a little different since we're not scrolling 3 items on each scroll

### DIFF
--- a/app/frontend/Shared/FastList.svelte
+++ b/app/frontend/Shared/FastList.svelte
@@ -1,5 +1,5 @@
 <hbox class="fast-list"
-  on:wheel={event => onScrollWheelThrottled(event)}
+  on:wheel={event => onScrollWheel(event)}
   on:keydown={event => onKey(event)}
   tabindex={0}
   bind:this={listE}>
@@ -57,7 +57,6 @@
   import { Collection, CollectionObserver, ArrayColl } from "svelte-collections"
   import { onMount, tick } from "svelte";
   import debounce from "lodash/debounce";
-  import { useThrottle } from "@svelteuidev/composables";
 
   type T = $$Generic;
 
@@ -232,7 +231,6 @@
   }
 
   // Throttle scroll speed for trackpad scrolling
-  const onScrollWheelThrottled = useThrottle(onScrollWheel, 30);
   function onScrollWheel(event: WheelEvent) {
     let scrollRows = Math.round(Math.abs(event.deltaY * 0.1)); // How many rows to scroll each time
     if (event.deltaY > 0) {
--- a/app/frontend/Shared/FastList.svelte
+++ b/app/frontend/Shared/FastList.svelte
@@ -227,16 +227,14 @@
     if (index >= scrollPos && index < scrollPos + showRows - 1) {
       return;
     }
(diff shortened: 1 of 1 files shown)


app/frontend/Shared/FastList.svelte
```diff
@@ -57,6 +57,7 @@
   import { Collection, CollectionObserver, ArrayColl } from "svelte-collections"
   import { onMount, tick } from "svelte";
   import debounce from "lodash/debounce";
+  import { useThrottle } from "@svelteuidev/composables";
```
benbucksch: This is a generic component, usable outside Mustang. It should not depend on svelteui.

Is the throttle necessary, given that your now scroll based on deltaY? For what exactly?

benbucksch (changes requested):

app/frontend/Shared/FastList.svelte
jermy-c: You're right the throttle is not necessary. I thought the triggering more wheelEvents would still cause the issue but it seems like it doesn't cause any issue. I've removed the throttle now.

benbucksch:
Hey Jeremy, I've made a few changes:
* Remove `abs()`
* Use min()/max() instead of `<`
* Fix buggy min()/max() in another place
* Consider `rowHeight`, because `deltaY` is in pixels and the height of each row in pixels matters here.
* The `|| 20` is a fallback, in case `rowHeight` has not been inited, to avoid "division by 0" error.

On my computer, with a mouse wheel, it scrolls exactly 3 rows, in the email list. Can you please check on a touch pad?

benbucksch:
Note that there is no special number anymore. I compared with other apps (a text editor and a browser) on my machine, and they also scroll about 3 lines with my mouse wheel, so this seems correct.

jermy-c:
There's a little unresponsiveness feeling when using the trackpad compared to a mouse, especially when scrolling slowly. Below is the log I get when using a Trackpad compared to a mouse. The trackpad triggers more events but less deltaY change but the mouse triggers less events but more deltaY.

Trackpad
```
scrollsRows 0 deltaY 1
scrollsRows 0 deltaY 7
scrollsRows 0 deltaY 13
scrollsRows 1 deltaY 23
scrollsRows 1 deltaY 30
scrollsRows 1 deltaY 33
scrollsRows 1 deltaY 24
scrollsRows 1 deltaY 27
scrollsRows 0 deltaY 17
scrollsRows 0 deltaY 12
scrollsRows 0 deltaY 8
scrollsRows 0 deltaY 5
scrollsRows 0 deltaY 2
scrollsRows 0 deltaY 1
```

Mouse
```
scrollsRows 0 deltaY 4.000244140625
scrollsRows 1 deltaY 33.0340576171875
scrollsRows 3 deltaY 124.0582275390625
scrollsRows 4 deltaY 173.455810546875
```

PR #661, opened 2025-06-18
FastList: Wheel event with touchpad: Throttle (#361)

benbucksch:
Jeremy wrote:
There's a little unresponsiveness feeling when using the trackpad compared to a mouse, especially when scrolling slowly. Below is the log I get when using a Trackpad compared to a mouse. The trackpad triggers more events but less deltaY change but the mouse triggers less events but more deltaY.

---

This PR adds a throttle function for the wheel event, hoping that this will fix the unresponsiveness.

### DIFF
--- a/app/frontend/Shared/FastList.svelte
+++ b/app/frontend/Shared/FastList.svelte
@@ -1,5 +1,5 @@
 <hbox class="fast-list"
-  on:wheel={event => onScrollWheel(event)}
+  on:wheel={event => onScrollWheelDebounced(event)}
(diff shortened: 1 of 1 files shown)


jermy-c:
It feels almost the same as before where it scrolls too fast. There’s about 10 scroll events with touchpad compared to 4 scroll events with mouse per scroll.

benbucksch:
No longer relevant, now that the FastList scrolling is completely rewritten and based on native scrolling

PR #663, opened 2025-06-19
Calendar: Recurrence: Rewrite logic to separate instances, exceptions and exclusions

benbucksch:


app/logic/Calendar/Event.ts
```diff
@@ -481,18 +536,17 @@ export class Event extends Observable {
   async deleteLocally() {
     assert(this.calendar, "To delete an event, it needs to be in a calendar first");
     assert(this.calendar.storage, "To delete an event, the calendar needs to be saved first");
+    // this.recurrenceRule = null; is overkill for removing exceptions; the database already cascades the delete
+    this.calendar.events.remove(this.exceptions);
```
NeilRashbrook:
```suggestion
    this.calendar.events.removeAll(this.exceptions);
```

app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
```diff
@@ -145,16 +145,15 @@ export class ActiveSyncCalendar extends Calendar implements ActiveSyncPingable {
           // Exceptions must be handled after the master event has been saved.
           for (let exception of ensureArray(item.ApplicationData.Exceptions?.Exception)) {
             if (exception.Deleted != "1") {
-              let occurrences = event.recurrenceRule.getOccurrencesByDate(fromCompact(exception.ExceptionStartTime));
-              let instance = event.instances.get(occurrences.length - 1);
-              if (instance) {
-                instance.fromWBXML(exception);
-                await instance.saveLocally();
+              let exceptionTime = fromCompact(exception.ExceptionStartTime);
+              let existing = event.exceptions.find(event => event.recurrenceStartTime.getDate() == exceptionTime.getTime());
```
NeilRashbrook:
```suggestion
              let existing = event.exceptions.find(event => event.recurrenceStartTime.getTime() == exceptionTime.getTime());
```

app/logic/Calendar/SQL/SQLEvent.ts
```diff
@@ -156,15 +155,15 @@ export class SQLEvent extends Event {
...
-        // TODO Expensive
-        let occurrences = event.parentEvent.recurrenceRule.getOccurrencesByDate(event.recurrenceStartTime);
-        // TODO assumes that the master is read before the exception
-        event.parentEvent.instances.set(occurrences.length - 1, event);
+        parentEvent.exceptions.add(event);
+        parentEvent.instances.clear();
+        parentEvent.fillRecurrences();
       }
     }
     if (row.recurrenceRule) {
```
NeilRashbrook, line 168: (I can't comment on the next two lines, so here they are:)
```typescript
      event.recurrenceRule = RecurrenceRule.fromCalString(event.duration, event.startTime, row.recurrenceRule);
      await SQLEvent.readExclusions(event);
```
Setting the recurrence rule sets the instances, before we've had a chance to load the exclusions. Swapping the lines around won't work because setting the recurrence rule also clears any existing exceptions...

app/logic/Calendar/Event.ts
```diff
@@ -153,14 +153,12 @@ export class Event extends Observable {
...
-    if (this.calendar && !this._recurrenceRule?.timesMatch(rule)) {
+    if (!this._recurrenceRule?.timesMatch(rule)) {
       this.clearExceptions();
     }
     this._recurrenceRule = rule;
     this.recurrenceCase = RecurrenceCase.Master; // notifies
-    if (this.calendar) {
-      this.fillRecurrences();
-    }
+    this.instances.clear();
```
NeilRashbrook: You'll want to generate the instances at some point; currently you're not doing it at all.

NeilRashbrook:
FYI when I delete an instance from the sidebar, the `remove` notifies an `AdditionCollectionDedupObserver` (whose `_mapFunc` is `mapToEvents`) which notifies another `AdditionCollectionDedupObserver` (whose `_mapFunc` is `cal => cal.eventsWithRecurrences`) but that one doesn't have anything observing it.

If I remove the `.sortBy(ev => ev.startTime)` then the delete happens instantly. It looks as if the observers are getting garbage collected due to `svelte-collections` version 1.1.23 switching to `WeakSetIterable` for its _observers. (Some observers are set as properties on their collection which stops them from getting garbage collected.)

benbucksch:
> It looks as if the observers are getting garbage collected due to svelte-collections version 1.1.23 switching to WeakSetIterable for its _observers.

Thank you! I would have hunted that bug for quite a while.
Indeed, given that most observers are only set as observers to the collection, it makes no sense to make them weak refs. That was a bad change. I've reverted that, released it as svelte-collections 1.2.11, and added it to the branch.

app/logic/Calendar/Event.ts
benbucksch: Indeed. Fixed in latest commit.

app/logic/Calendar/SQL/SQLEvent.ts
benbucksch: I'm now calling generate() after both lines. That should fix it, right?

app/logic/Calendar/Event.ts
```diff
@@ -153,12 +153,12 @@ export class Event extends Observable {
...
     assert(this.recurrenceCase == RecurrenceCase.Normal || this.recurrenceCase == RecurrenceCase.Master, "Instances can't themselves recur");
+    this._recurrenceRule = rule;
+    this.recurrenceCase = RecurrenceCase.Master; // notifies
     if (!this._recurrenceRule?.timesMatch(rule)) {
       this.clearExceptions();
     }
-    this._recurrenceRule = rule;
-    this.recurrenceCase = RecurrenceCase.Master; // notifies
-    this.instances.clear();
+    this.generateRecurringInstances();
```
NeilRashbrook: Oops, you're comparing the rule after you replace it...

benbucksch: Fixed by Neil in #671

PR #664, opened 2025-06-19
Support RECURRENCE-ID in iCal parsing and generation

NeilRashbrook:
I had to tweak some `Event` methods to get the tests to pass.

Note that the CalDAV calendar, when it finds an event with a `recurrenceStartTime`, probably needs to find the master with the same `calUID` so that it can set it as its parent event and tell it that it has an exception.


benbucksch:
> CalDAV calendar, when it finds an event with a recurrenceStartTime, ... needs to find the master with the same calUID

OK. Can you do that quickly?

NeilRashbrook:
well, not *that* quickly...

benbucksch:
> well, not _that_ quickly...

OK, then finish the rest of recur first, then come back to this tonight, please

PR #665, opened 2025-06-19
Don't try to unset the recurrence of an instance

NeilRashbrook:
OWA doesn't seem to like it any more either.

Compare #536.


NeilRashbrook:
I think ActiveSync works differently, it expects you to send the new object rather than commands to update the object. And exceptions will never have recurrences, so there's nothing to send.

I don't know how CalDAV handles exceptions.

NeilRashbrook:
So, like a big comment on `ActiveSyncEvent.toFields` explaining that ActiveSync 14.x expects us to send only the fields that have values, and it will delete the remaining fields, and `[]` is used by the WBXML converter to signify a lack of value. except for Body, which needs to be explicitly set to the empty string, and exceptions (and exclusions) are optional those not specified will not be affected?

PR #669, opened 2025-06-26
vCard test fixes

NeilRashbrook:
- Commit 7079837 broke parsing of `type=pref`, and also the default preference values needed to be updated in tests.
- Commit 6932879 broke parsing of `x-custom` properties.
- Commit 7079837 broke tests.

For each of the above, there is a commit in this PR that fixes specifically the bustage from that commit only.

I have also added a commit to test round-tripping of the PO Box and Suite entries of a vCard.


app/logic/Contacts/JSON/JSONPerson.ts
```diff
@@ -101,4 +114,18 @@ export class JSONPerson {
...
+  protected static readStreetAddresses(addresses: any[]): ArrayColl<ContactEntry> {
+    return new ArrayColl(contacts.map(c => this.readAddress(c)));
+  }
+
+  protected static readAddress(json: any): ContactEntry {
+    let purpose = sanitize.label(json.purpose);
+    let address = new StreetAddress();
+    address.fromJSON(json);
+    let contact = new ContactEntry(address.toString(), purpose);
+    contact.preference = sanitize.integer(json.preference, 100);
```
benbucksch:
```suggestion
    contact.preference = sanitize.integer(json.preference, defaultPreference);
```

app/logic/Contacts/JSON/JSONPerson.ts
```diff
@@ -45,6 +46,18 @@ export class JSONPerson {
     return json;
   }
 
+  static saveStreetAddresses(addresses: Collection<ContactEntry>): any[] {
+    return addresses.contents.map(c => this.saveAddress(c));
+  }
+
+  static saveAddress(contact: ContactEntry): any {
+    let json: any = new StreetAddress(contact.value).toJSON();
```
benbucksch: Make this:
`json.value = new StreetAddress(contact.value).toJSON();`
please, so that the address is in its own JSON object within property `value`, not on the same level as `purpose`.

app/test/logic/Contacts/VCard/TestData/Thunderbird.json
```diff
@@ -28,7 +28,12 @@
     "preference": 100
   }],
   "streetAddresses": [{
-    "value": "My Street\nMy City\n00000\nMy State\nGermany",
+    "instructions": null,
+    "street": "My Street",
+    "city": "My City",
+    "state": "My State",
+    "postCode": "00000",
```
benbucksch: `postalCode` (with your coming fix)

app/logic/Contacts/VCard/VCard.ts
```diff
@@ -196,7 +196,7 @@ function makeContactEntry(entry: ICalEntry, protocol?: string, isURI = true, str
     value,
     /work/i.test(entry.properties.type) ? "work" : /home/i.test(entry.properties.type) ? "home" : "other",
     protocol,
-    sanitize.integer(entry.properties.pref, ContactEntry.defaultPreference));
+    sanitize.integer(entry.properties.pref, /pref/i.test(entry.properties.type) ? 1 : ContactEntry.defaultPreference));
```
benbucksch, line 199: Q: In modern browser javascript/typescript, what's the fastest way to compare 2 strings case-insensitive? Which is the most readable way? Which way has the best balance between runtime speed and code readability?
A: https://www.perplexity.ai/search/what-s-the-fastest-way-to-comp-UzeDgN5aQV673aL7kq1.OA

toLowerCase() / toUpperCase()	Fastest	Most clear
RegExp with /i flag	Slowest	Less clear

benbucksch: I see that you use the same method above, so leave it here for now.

app/logic/Contacts/JSON/JSONPerson.ts
NeilRashbrook: Ugh, that's going to make those test files really ugly...

app/logic/Contacts/JSON/JSONPerson.ts
NeilRashbrook: That's what you get for copying code...

app/logic/Contacts/VCard/VCard.ts
NeilRashbrook: I'm not comparing strings. And the best way if I was would be to use `localeCompare`, of course.

app/test/logic/Contacts/VCard/TestData/Thunderbird.json
NeilRashbrook: So which PR do I put the `postalCode` fix in? Probably this one, actually.

benbucksch (approved):

PR #672, opened 2025-06-28
Default mail client / mailto: handling

jermy-c:
This is just working code that works and tested on MacOS

**Working**

- MacOS `mailto` link when Mustang is running and not running

**Not Working**

- No option to not set Mustang as defualt mail client
- No option to explicitly set Mustang as default mail client
- Might have an error if the users machine is slow because everything is based on `setTimeout()` at the moment

### DIFF
--- a/app/logic/startup.ts
+++ b/app/logic/startup.ts
@@ -16,6 +16,8 @@ import { type Account, setMainAccounts } from './Abstract/Account';
 import JPCWebSocket from '../../lib/jpc-ws';
 import { production } from './build';
 import { logError } from '../frontend/Util/error';
+import { mailMustangApp } from '../frontend/Mail/MailMustangApp';
+import { findOrCreatePersonUID } from './Abstract/PersonUID';
 
 const kSecret = 'eyache5C'; // TODO generate, and communicate to client, or save in config files.
 
@@ -38,6 +40,9 @@ export async function getStartObjects(): Promise<void> {
   appGlobal.collectedAddressbook = appGlobal.addressbooks.get(1);
   readSavedSearches();
   await loadTagsList();
+
+  await appGlobal.remoteApp.setAsDefaultMailClient();
+  await listenForURL();
 }
 
 /**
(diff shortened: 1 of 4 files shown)


app/logic/startup.ts
```diff
@@ -119,6 +124,27 @@ async function fileShareLogin(account: FileSharingAccount) {
...
+    writeMail(url);
+  });
+  console.log(ipcRenderer);
+  await appGlobal.remoteApp.listenForURL((_, url) => writeMail(url));
+}
+
+function writeMail(url: string) {
+  let selectedAccount = appGlobal.emailAccounts.find((acc) => acc.isLoggedIn);
+  if (!selectedAccount) {
+    return setTimeout(() => writeMail(url), 1000);
```
jermy-c, line 140: There's an `undefined` error for the selectedAccount

e2/src/main/index.ts
```diff
@@ -72,6 +78,17 @@ function createWindow(): void {
       });
     });
 
+    mainWindow.webContents.on('did-finish-load', () => {
+      if (deeplinkUrl) {
+        // Requires a delay after onload is emitted otherwise listener
+        // doesn't get the message
+        setTimeout(() => {
+          mainWindow.webContents.send('deeplink', deeplinkUrl);
```
jermy-c, line 86: Even `did-finish-load` was too early for the frontend listener

e2/src/main/index.ts
```diff
@@ -126,6 +143,41 @@ app.whenReady().then(async () => {
   }
 })
 
+let deeplinkUrl: string | null = null;
+
+recordDeepLink();
+
+function recordDeepLink() {
```
jermy-c, line 150: We need to capture the URL link before the frontend and backend start running when the app is not running yet.

benbucksch:
Here's how you can get rid of the setTimeout:
1. Create a class in the backend.ts file:
```js
class StartupURL extends Observable {
   @notifyObservableProperty
  url: URLString
}
const startupURL = new StartupURL;
```
2. In the Electron main file, capture/read the URL using the Electron methods and set it in the backend.ts using `startupURL.url = url;`
3. The frontend adds a observer to `startupURL`, like so:
```
$: runStartupURL($startupURL.url);
function runStartupURL(url: URLString) {
  if (!url) {
    return;
  }
  if (url.startsWith("mailto:")) {
    // open composer ...
  }
}
```
3.a. This allows the frontend to handle the URL whenever the frontend is ready. This avoids the setTimeout()
3.b. Because this adds an observer, this should also handle the case when a URL is opened after the app has already started. (If this doesn't work, then there's likely something wrong with the observer, and you need to debug it.)

4. (You don't need to wait for the user to select an account before opening the mail composer. Simply use the first mail account. The user can change it in the mail composer.)

benbucksch:
Implemented in git commit fd32236

Issue #673, opened 2025-06-30
Calendar: Recurrence: Invitation


benbucksch:
Testcase in the neil@beonex account (here as EWS account):

![Image](https://github.com/user-attachments/assets/879a8f7d-320e-4ce7-9433-75d541587252)

```
TypeError: Cannot read properties of undefined (reading 'newEvent')
    at _Event.fillRecurrences (Event.ts:662:36)
    at _Event.generateRecurringInstances (Event.ts:439:10)
    at _Event.clearExceptions (Event.ts:689:10)
    at _Event.setRecurrenceRule (Event.ts:162:12)
    at set recurrenceRule (Event.ts:135:12)
    at convertICalParserToEvent (ICalToEvent.ts:51:11)
    at ICalEMailProcessor.process (ICalEMailProcessor.ts:19:19)
    at async EWSEMail.parseMIME (EMail.ts:300:7)
    at async EWSEMail.download (EWSEMail.ts:53:5)
    at async EWSEMail.loadMIME (EMail.ts:358:5)
```

NeilRashbrook:
Ah, this is fallout from #663, specifically commit 568b713 which changed a code path to bypass an `if (calendar)` check.

benbucksch:
Are recurrences in invitations working otherwise? Including showing their recurrence pattern in the display dialog before I accept the invitation?

NeilRashbrook:
`DisplayEvent.svelte` has code for showing recurrence patterns.

PR #674, opened 2025-06-30
Don't try to generate instances of pseudo events in messages #673

NeilRashbrook:


app/logic/Calendar/Event.ts
```diff
@@ -647,7 +647,7 @@ export class Event extends Observable {
    */
   fillRecurrences(seriesEndTime: Date = new Date(Date.now() + 1e11)): Collection<Event> {
     assert(this.recurrenceCase == RecurrenceCase.Master, "Not a recurrence master");
-    if (this.instances.hasItems) {
+    if (!this.calendar || this.instances.hasItems) {
```
benbucksch:
```suggestion
    if (this.instances.hasItems || // cached
        !this.calendar) { // Not for pseudo events in messages
```

PR #675, opened 2025-06-30
Allow an observable to propagate notifications from its property

NeilRashbrook:
If you have an existing meeting and you try to edit just the participant list, you can't save the event; this is because the notifications for changes to the participant list don't go anywhere. Now, you could add `$: participants = event.participants;` and reference `$participants` in `canSave`... but it's probably better to propagate the notification.

(`notifyChangedProperty` is also fairly useless on a `readonly` property...)

### DIFF
--- a/app/logic/util/Observable.ts
+++ b/app/logic/util/Observable.ts
@@ -110,7 +110,8 @@ export function notifyChangedObservable<T extends Observable>(obj: T, propertyNa
     this._properties[propertyName] = val;
     // subscribe() calls the subscriber once immediately at subscription time,
     // so notifyObservers() is called immediately, and we need that.
-    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() => this.notifyObservers(propertyName));
+    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() =>
+      this.notifyObservers(propertyName));
   }
   descriptor.get = function (this: T) {
     return this._properties[propertyName];
--- a/lib/util/Observable.ts
+++ b/lib/util/Observable.ts
@@ -108,8 +108,10 @@ export function notifyChangedObservable<T extends Observable>(obj: T, propertyNa
     }
     this._properties["_unsubscribe_" + propertyName]?.();
     this._properties[propertyName] = val;
-    this.notifyObservers(propertyName, oldValue);
-    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() => this.notifyObservers(propertyName, oldValue));
+    // subscribe() calls the subscriber once immediately at subscription time,
+    // so notifyObservers() is called immediately, and we need that.
+    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() =>
+      this.notifyObservers(propertyName));
   }
   descriptor.get = function (this: T) {
     return this._properties[propertyName];


app/logic/util/Observable.ts
```diff
@@ -81,4 +81,40 @@ export function notifyChangedProperty<T extends Observable>(obj: T, propertyName
...
+    };
+  }
+  descriptor.set = function (this: T, val: IObservable | undefined): void {
+    let oldValue: IObservable | undefined = this._properties[propertyName];
+    if (oldValue === val) {
+      return;
+    }
+    this._properties["_unsubscribe_" + propertyName]?.();
+    this._properties[propertyName] = val;
+    this.notifyObservers(propertyName, oldValue);
```
benbucksch:
```suggestion
    // subscribe() calls the subscriber once immediately at subscription time,
    // so notifyObservers() is called immediately, and we need that.
```

Can you please check whether it works without this line?

To my knowledge, the `subscribe()` API mandates that the subscriber is called exactly once at the time of subscription, so this function would be called anyways in the next time. This is because your usecase is very common and the defining it this way avoids code duplication.

Can you please double-check that this is the case? And if so, apply the suggestion.

app/logic/util/Observable.ts
```diff
@@ -81,4 +81,40 @@ export function notifyChangedProperty<T extends Observable>(obj: T, propertyName
   Object.defineProperty(obj, propertyName, descriptor);
 }
 
-// <copied to="../../lib/util/Observable.ts" />
+/** Decorator for object properties.
+ * Propagates notifications. */
```
benbucksch:
```suggestion
 * Subscribes to the property and propagates its notifications. */
```

app/logic/util/Observable.ts
```diff
@@ -81,4 +81,40 @@ export function notifyChangedProperty<T extends Observable>(obj: T, propertyName
...
+  }
+  descriptor.set = function (this: T, val: IObservable | undefined): void {
+    let oldValue: IObservable | undefined = this._properties[propertyName];
+    if (oldValue === val) {
+      return;
+    }
+    this._properties["_unsubscribe_" + propertyName]?.();
+    this._properties[propertyName] = val;
+    this.notifyObservers(propertyName, oldValue);
+    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() => this.notifyObservers(propertyName, oldValue));
```
benbucksch:
```suggestion
    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() => this.notifyObservers(propertyName));
```

app/logic/util/Observable.ts
```diff
@@ -81,4 +81,41 @@ export function notifyChangedProperty<T extends Observable>(obj: T, propertyName
...
+  descriptor.set = function (this: T, val: IObservable | undefined): void {
+    let oldValue: IObservable | undefined = this._properties[propertyName];
+    if (oldValue === val) {
+      return;
+    }
+    this._properties["_unsubscribe_" + propertyName]?.();
+    this._properties[propertyName] = val;
+    // subscribe() calls the subscriber once immediately at subscription time,
+    // so notifyObservers() is called immediately, and we need that.
+    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() => this.notifyObservers(propertyName));
```
benbucksch:
```suggestion
    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() =>
      this.notifyObservers(propertyName));
```

lib/util/Observable.ts
```diff
@@ -81,6 +81,42 @@ export function notifyChangedProperty<T extends Observable>(obj: T, propertyName
...
+  }
+  descriptor.set = function (this: T, val: IObservable | undefined): void {
+    let oldValue: IObservable | undefined = this._properties[propertyName];
+    if (oldValue === val) {
+      return;
+    }
+    this._properties["_unsubscribe_" + propertyName]?.();
+    this._properties[propertyName] = val;
+    this.notifyObservers(propertyName, oldValue);
+    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() => this.notifyObservers(propertyName, oldValue));
```
benbucksch:
```suggestion
    // subscribe() calls the subscriber once immediately at subscription time,
    // so notifyObservers() is called immediately, and we need that.
    this._properties["_unsubscribe_" + propertyName] = val?.subscribe(() =>
      this.notifyObservers(propertyName));
```

benbucksch (approved):

benbucksch:
@NeilRashbrook Can you please test whether it works correctly with the discussed code changes?

app/logic/util/Observable.ts
NeilRashbrook: It works in our case because the property is readonly and gets assigned exactly once, but in general, were you to want to allow setting the property to null, you would not actually notify at when that happened.

benbucksch: Because of the `?` before the `subscribe()`. Good point.

benbucksch: Fixed in git commit cf9bbf494

PR #678, opened 2025-06-30
Invitation: Send only when specifically needed

NeilRashbrook:
- Don't send invitations when changing, say, the reminder
- If only adding/removing participants, only send them invitations/cancellations
- But if major event details change, send to all participants

### DIFF
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -397,26 +397,6 @@ export class Event extends Observable {
   }
 
   /**
-   * Has a property changed that we should notify invitees about?
-   */
-  hasChangedInvitation(): boolean {
-    return !!this.unedited && (
-      this.startTime?.getTime() != this.unedited.startTime?.getTime() ||
-      this.endTime?.getTime() != this.unedited.endTime?.getTime() ||
-      this.timezone != this.unedited.timezone ||
-      this.allDay != this.unedited.allDay ||
-      this.title != this.unedited.title ||
-      this.rawText != this.unedited.rawText ||
-      this.rawHTMLDangerous != this.unedited.rawHTMLDangerous ||
-      this.location != this.unedited.location ||
-      this.isOnline != this.unedited.isOnline ||
-      this.onlineMeetingURL != this.unedited.onlineMeetingURL ||
-      this.recurrenceCase != this.unedited.recurrenceCase ||
-      this.recurrenceRule != this.unedited.recurrenceRule
-    );
-  }
-
-  /**
    * Starts editing an event.
    *
    * If it is already being edited, this is a no-op. This can happen when switching
@@ -538,15 +518,10 @@ export class Event extends Observable {
   }
 
   async saveToServer(): Promise<void> {
-    if (this.isIncomingMeeting || !this.unedited) {
-      return;
+    if (!this.isIncomingMeeting && this.outgoingInvitation && this.participants.hasItems && this.hasChanged()) {
+      this.calUID ??= crypto.randomUUID();
+      await this.outgoingInvitation.sendInvitations();
     }
-    this.calUID ??= crypto.randomUUID();
-    let removed = this.unedited.participants.subtract(this.participants);
-    // Use the original event when sending cancellations
-    this.unedited.outgoingInvitation.sendCancellations(removed);
-    let notify = this.hasChangedInvitation() ? this.participants : this.participants.subtract(this.unedited.participants);
-    await this.outgoingInvitation.sendInvitations(notify);
   }
 
   /**
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -397,6 +397,26 @@ export class Event extends Observable {
   }
 
   /**
+   * Has a property changed that we should notify invitees about?
+   */
+  hasChangedInvitation(): boolean {
+    return !!this.unedited && (
+      this.startTime?.getTime() != this.unedited.startTime?.getTime() ||
+      this.endTime?.getTime() != this.unedited.endTime?.getTime() ||
+      this.timezone != this.unedited.timezone ||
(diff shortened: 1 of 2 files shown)


app/logic/Calendar/Event.ts
```diff
@@ -396,6 +396,26 @@ export class Event extends Observable {
     );
   }
 
+  /**
+   * Has a property changed that we should notify invitees about?
+   */
+  hasChangedInvitation(): boolean {
```
benbucksch: Move to `OutgoingInvitation.changesNeedNotify(): boolean`

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -28,17 +29,17 @@ export default class OutgoingInvitation {
     return account.identities.first;
   }
 
-  async sendInvitations() {
-    for (let participant of this.event.participants) {
+  async sendInvitations(participants: Collection<Participant>) {
```
```suggestion
  async sendInvitationsTo(participants: Collection<Participant>) {
```

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -28,17 +29,17 @@ export default class OutgoingInvitation {
...
       if (participant.response != InvitationResponse.Organizer) {
         participant.response ||= InvitationResponse.NoResponseReceived;
         await this.send("REQUEST", participant);
       }
     }
   }
 
-  async sendCancellations() {
-    for (let participant of this.event.participants) {
+  async sendCancellations(participants: Collection<Participant>) {
```
```suggestion
  async sendCancellationsTo(participants: Collection<Participant>) {
```

app/logic/Calendar/Event.ts
```diff
@@ -518,10 +538,15 @@ export class Event extends Observable {
...
-      await this.outgoingInvitation.sendInvitations();
+    if (this.isIncomingMeeting || !this.unedited) {
+      return;
     }
+    this.calUID ??= crypto.randomUUID();
+    let removed = this.unedited.participants.subtract(this.participants);
+    // Use the original event when sending cancellations
+    this.unedited.outgoingInvitation.sendCancellations(removed);
+    let notify = this.hasChangedInvitation() ? this.participants : this.participants.subtract(this.unedited.participants);
+    await this.outgoingInvitation.sendInvitations(notify);
```
benbucksch: Move this logic into `outgoingInvitation.sendCancellations()`, and keep the old code (including the old `if` block), only adapting the condition.

app/logic/Calendar/Event.ts
```diff
@@ -557,7 +582,11 @@ export class Event extends Observable {
       return;
     }
     if (this.myParticipation == InvitationResponse.Organizer) {
-      await this.outgoingInvitation.sendCancellations();
+      if (this.unedited) {
+        await this.unedited.outgoingInvitation.sendCancellations(this.unedited.participants);
+      } else {
+        await this.outgoingInvitation.sendCancellations(this.participants);
+      }
```
benbucksch: ditto

benbucksch (changes requested):

app/logic/Calendar/Event.ts
```diff
@@ -518,7 +518,7 @@ export class Event extends Observable {
   }
 
   async saveToServer(): Promise<void> {
-    if (!this.isIncomingMeeting && this.participants.hasItems && this.hasChanged()) {
+    if (!this.isIncomingMeeting && this.outgoingInvitation && this.participants.hasItems && this.hasChanged()) {
```
benbucksch: Is this condition correct? It's based on the old condition, and it makes sense to me, but you changed it to a different condition, and I didn't understand why. Please adapt it as *necessary* (but not merely because of preference).

benbucksch (approved):
Hey Neil, I've made the review changes.

I would really like to keep all code related to invitations in the specific classes and outside of the `Event` class. Otherwise, `Event` becomes a monster class that implements half of the calendar.

benbucksch:
Could you please review and test my changes? If not good, please fix them. Once good, please tell me, so that I can merge.

app/logic/Calendar/Event.ts
benbucksch: Neil confirms that this is still correct, just does unnecessary checks that are done in the function, too.

benbucksch:
@NeilRashbrook Could you please test tomorrow whether this still works correctly and as you intended, after my changes?

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -64,4 +86,26 @@ export default class OutgoingInvitation {
...
     return email;
   }
+
+  /**
+   * Has a property changed that we should notify invitees about?
+   */
+  changesNeedToNotify(): boolean {
+    let event = this.event;
+    let unedited = this.event.unedited;
+    return !!event.unedited && (
```
NeilRashbrook, line 96:
```suggestion
    return !!unedited && (
```

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -64,4 +86,26 @@ export default class OutgoingInvitation {
...
+      event.endTime?.getTime() != unedited.endTime?.getTime() ||
+      event.timezone != unedited.timezone ||
+      event.allDay != unedited.allDay ||
+      event.title != unedited.title ||
+      event.rawText != unedited.rawText ||
+      event.rawHTMLDangerous != unedited.rawHTMLDangerous ||
+      event.location != unedited.location ||
+      event.isOnline != unedited.isOnline ||
+      event.onlineMeetingURL != unedited.onlineMeetingURL ||
+      event.recurrenceCase != unedited.recurrenceCase ||
```
NeilRashbrook, line 107: Thinking about it, this line should be removed, because we shouldn't notify all existing participants on a recurring meeting for adding a participant as a new exception.
```suggestion
```

Issue #314, opened 2024-12-14
Mac: Default mail client


benbucksch:
Need special entitlement:
https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.developer.mail-client

benbucksch:
Property List Key
com.apple.developer.mail-client
A Boolean that indicates whether the app can act as a user’s default email client.

The system launches the default mail client in iOS whenever a user opens a mailto: link. Since email is a critical avenue for communication, Apple requires that email apps must meet specific functional criteria aimed at ensuring private and accurate access for users.

Apps signal their intent to be used as a default mail app by using this managed entitlement. Request permission to use this entitlement by emailing default-mail-app-requests@apple.com.

Any app that registers as a default email client option must:
- [ ] Specify the mailto: scheme in its Info.plist file. #315
- [x] Be able to send a message to any valid email recipient.
- [x] Be able to receive a message from any email sender. Apps that provide user-controlled incoming mail screening features are permitted.

benbucksch:
@jermy-c mailto: on Linux now works. Can you please test whether it also works on Mac? If not, make it work?

benbucksch:
It works here: The observers fire, and the mail composer comes up when I click on mailto: URL in the browser or start `xdg-open mailto:foo@example.com` on Linux. The observers are not fired automatically just by setting `.url = `, so make sure to call `.notifyObservers()`, like I do in my code.

benbucksch:
These are the errors that you run into:

```
TypeError: Cannot read properties of undefined (reading 'url')
  at $$self.$$.update (/frontend/Mail/MailInBackground.svelte:165:49)
  at new MailInBackground (/frontend/Mail/MailInBackground.svelte:189:3)
  at create_fragment (/frontend/MainWindow/MainWindow.svelte:403:21)
  at new MainWindow (/frontend/MainWindow/MainWindow.svelte:762:3)
  at None (/frontend/main.ts:22:13)
...
(8 additional frame(s) were not displayed)
```

```
TypeError: store.subscribe is not a function
  at $$subscribe_startupArgs (/frontend/Mail/MailInBackground.svelte:68:93)
  at $$self.$$.update (/frontend/Mail/MailInBackground.svelte:156:7)
```

benbucksch:
> I clicking a mailto link works now, while Mustang is running.

Great!

[...]

PR #679, opened 2025-06-30
Mac: Default mail client (#314)

jermy-c:
- Open window if Mustang is running but no windows are open
- notifyObservers after setting the URL
- Not working when the app is not running


e2/src/main/index.ts
```diff
@@ -150,7 +150,11 @@ app.on('window-all-closed', () => {
 function registerURLOpen() {
   // macOS: Capture URL during launch
   app.on("open-url", (_event, url) => {
+    if (BrowserWindow.getAllWindows().length === 0) {
+      createWindow();
+    }
```
benbucksch: This to move this after you set the URL. (Probably won't help, but worth a try)

(I forgot to submit this comment - sorry, my fault.)

jermy-c:
I couldn't find a way to read the console log of the frontend with the URL passed. However, I tried recording the logs and sent them to the frontend. Based on that result, I didn't see any `open-url` event.

e2/src/main/index.ts
benbucksch: I made the change and commited it as git commit 7fed1478

benbucksch (changes requested):

PR #687, opened 2025-07-01
Update UI after simple event is deleted or made recurring

NeilRashbrook:
`mapToCollection` has a problem, and that is that it assumes that the mapping function is idempotent, i.e. it always returns the same value each time.

This is not true with our current `mapToEvents` function, except in the edge case where you only create and delete recurring events (which does work on current code).

Now there may be another way to achieve this which allows a dynamic mapping function, but I chose to go the other way and make the mapping function idempotent.

To this end, the `instances` collection of the event should contain itself if it's a single event or an exception otherwise it should contain its occurrences.

While I was there I noticed that changing a recurring event to a single event didn't remove its occurrences (because clearExceptions regenerates them) so I rearranged the code so that this works too.

I was tempted to remove `RecurrenceColl.ts`...

### DIFF
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -146,6 +146,7 @@ export class Event extends Observable {
       this.clearExceptions();
       this._recurrenceRule = null;
       this.recurrenceCase = RecurrenceCase.Normal; // notifies
+      this.instances.replaceAll([this]);
     }
   }
   /**
@@ -157,11 +158,10 @@ export class Event extends Observable {
     let timesMatch = this._recurrenceRule?.timesMatch(rule);
     this._recurrenceRule = rule;
     this.recurrenceCase = RecurrenceCase.Master; // notifies
-    if (timesMatch) {
-      this.generateRecurringInstances();
-    } else {
+    if (!timesMatch) {
       this.clearExceptions();
     }
+    this.generateRecurringInstances();
   }
   /** Links back to the recurring master.
    * Only for RecurrenceCase == Instance or Exception */
@@ -179,7 +179,9 @@ export class Event extends Observable {
    * For RecurrenceCase == Instance, it's identical to `startTime`.
    */
   recurrenceStartTime: Date | null = null;
-  /** Only for recurringCase == Master.
+  /** Contains all instances that should be displayed for this event.
+   *
+   * For master events only:
    *
    * Contains all Instances generated from the master.
    * Does *not* contain:
@@ -187,10 +189,13 @@ export class Event extends Observable {
    * - Exceptions
    * - Exclusions
    *
+   * For other events:
+   *   Contains the event itself.
+   *
    * This is a dynamic collection, and will be updated automatically
    * when the master changes or the recurrence rule changes.
    */
-  readonly instances = new ArrayColl<Event>;
+  readonly instances = new ArrayColl<Event>([this]);
   /**
    * Only for RecurrenceCase == Master
    *
@@ -702,7 +707,6 @@ export class Event extends Observable {
     }
     this.exceptions.clear();
     this.exclusions.clear();
-    this.generateRecurringInstances();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


app/logic/Calendar/Event.ts
```diff
@@ -179,18 +179,23 @@ export class Event extends Observable {
    * For RecurrenceCase == Instance, it's identical to `startTime`.
    */
   recurrenceStartTime: Date | null = null;
-  /** Only for recurringCase == Master.
+  /** Contains all instances that should be displayed for this event.
+   *
+   * For master events only:
```
benbucksch:
```suggestion
   * For recurringCase == Master:
```

app/logic/Calendar/Event.ts
```diff
@@ -179,18 +179,23 @@ export class Event extends Observable {
...
    *
    * Contains all Instances generated from the master.
    * Does *not* contain:
    * - The master itself
    * - Exceptions
    * - Exclusions
    *
+   * For other events:
```
benbucksch:
```suggestion
   * For recurringCase == Normal and Exception:
```

Not for instances

benbucksch:
It's a pity that you went this route, because it's a hack that I added arrays with a single event. I was meaning to fix `mergeColls()` to also accept individual items instead of collections, to avoid all these unneccessary arrays.

But maybe it's not important and this is premature optimization...

app/logic/Calendar/Event.ts
```diff
@@ -157,11 +158,10 @@ export class Event extends Observable {
     let timesMatch = this._recurrenceRule?.timesMatch(rule);
     this._recurrenceRule = rule;
     this.recurrenceCase = RecurrenceCase.Master; // notifies
-    if (timesMatch) {
-      this.generateRecurringInstances();
-    } else {
+    if (!timesMatch) {
       this.clearExceptions();
     }
+    this.generateRecurringInstances();
```
benbucksch, line 164: Why this change? (see also clear())

app/logic/Calendar/Event.ts
```diff
@@ -702,7 +707,6 @@ export class Event extends Observable {
     }
     this.exceptions.clear();
     this.exclusions.clear();
-    this.generateRecurringInstances();
```
benbucksch, line 705: Why this change? (see also above)

benbucksch (approved):

app/logic/Calendar/Event.ts
NeilRashbrook: `clearExceptions` is used by `clearRecurrenceRule`, but we don't want instances generated in this case. We haven't actually cleared the rule yet, because `generateRecurringInstances` used to assert. I guess we could change `clearRecurrenceRule` to call `clearExceptions` last, and then add a check that we're still a master before generating instances?

app/logic/Calendar/Event.ts
NeilRashbrook: Because we only want to generate instances after setting a rule, not after clearing it.

app/logic/Calendar/Event.ts
NeilRashbrook: (It's called `recurrenceCase`)

app/logic/Calendar/Event.ts
NeilRashbrook: Well, it's unused for instances, it was just easier to combine them with other non-master events.

app/logic/Calendar/Event.ts
benbucksch: We can just mark clearExceptions() as protected and then it's fine.

Public APIs *must* have an API that makes sense without explaining, and *must* protect themselves from mis-use, i.e. not depend on the caller to call them in a certain order, in a certain state etc. Ideally, they ensure that state, instead of asserting. In other words, the public API of each class stands on its own, a class is its own "castle". For protected functions, this does not apply (but must still be documented as code comments).

benbucksch: Marked `protected` in git commit f21839498b

PR #689, opened 2025-07-01
Calendar: HTML descriptions

benbucksch:
* Read HTML description from iCal
* Save HTML description to iCal

In 3 formats:
* [RFC 2445 4.2.1](https://datatracker.ietf.org/doc/html/rfc2445#section-4.2.1), 4.2, [RFC 5545 3.2.1](https://www.rfc-editor.org/rfc/rfc5545#section-3.2.1) ([iCal.org descr](https://icalendar.org/iCalendar-RFC-5545/3-2-1-alternate-text-representation.html)) and [Thunderbird](https://bugzilla.mozilla.org/show_bug.cgi?id=1607834) format: `DESCRIPTION` property with parameter `ALTREP` with `data:text/html` URL as value of the parameter
* [RFC 9073 6.5](https://www.rfc-editor.org/rfc/rfc9073.html#name-styled-description) format: `STYLED-DESCRIPTION` property with parameter `FMTTYPE="text/html"` and HTML as value of the property
* Outlook format: `X-ALT-DESC` property with parameter `FMTTYPE="text/html"` and HTML as value of the property

Preference order: 1. RFC 9075, 2. Thunderbird, 3. Outlook

### DIFF
--- a/app/frontend/Util/util.ts
+++ b/app/frontend/Util/util.ts
@@ -40,6 +40,27 @@ export async function stringToDataURL(mimetype: string, content: string): Promis
   });
 }
 
+export function stringFromDataURL(dataURL: URLString, mimetype: string): string | null {
+  console.log("altrep URL", dataURL);
+  if (!dataURL || typeof (dataURL) != "string" ||
+      dataURL.substring(0, "data:".length + mimetype.length).toLowerCase() != "data:text/html") {
+    return null;
+  }
+  // There might be `;charset=utf-8` in-between
+  let comma = dataURL.indexOf(",");
+  console.log("comma", comma);
+  if (comma < 1) {
+    return null;
+  }
+  let htmlEncoded = dataURL.substring(comma + 1);
+  console.log("html encoded", htmlEncoded);
+  if (!htmlEncoded) {
+    return null;
+  }
+  console.log("html decoded", decodeURIComponent(htmlEncoded));
+  return decodeURIComponent(htmlEncoded);
+}
+
 /**
  * Gets the OS using the userAgent string.
  * @returns os name
--- a/app/logic/Calendar/ICal/ICalEMailProcessor.ts
+++ b/app/logic/Calendar/ICal/ICalEMailProcessor.ts
@@ -21,7 +21,7 @@ export class ICalEMailProcessor extends EMailProcessor {
       return;
     }
     if (email.html) {
-      event.rawHTMLDangerous = email.html;
+      event.rawHTMLDangerous = email.rawHTMLDangerous;
     }
     email.event = event;
   }
--- a/app/logic/Calendar/ICal/ICalToEvent.ts
+++ b/app/logic/Calendar/ICal/ICalToEvent.ts
@@ -5,6 +5,7 @@ import { ParticipationStatus, InvitationResponse } from "../Invitation/Invitatio
 import ICalParser from "./ICalParser";
 import WindowsToIANATimezone from "./WindowsToIANATimezone";
 import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
+import { stringFromDataURL } from "../../../frontend/Util/util";
 
 /**
  * @param ics iCal / ICS contents to be parsed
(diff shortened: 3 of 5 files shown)


app/logic/Calendar/ICal/ICalToEvent.ts
```diff
@@ -33,7 +34,31 @@ export function convertICalParserToEvent(ics: ICalParser, event: Event): boolean
     event.title = vevent.entries.summary[0].value;
   }
   if (vevent.entries.description) {
+    // Plaintext
     event.descriptionText = vevent.entries.description[0].value;
+    // HTML RFC 2445 4.2.1, 4.2, RFC 5545 3.2.1 and Thunderbird
+    // <https://datatracker.ietf.org/doc/html/rfc2445#section-4.2.1>
+    // <https://bugzilla.mozilla.org/show_bug.cgi?id=1607834>
+    let altrep = vevent.entries.description[0].properties.altrep;
```
benbucksch, line 42: @NeilRashbrook BTW You got the nomenclature wrong. The DESCRIPTION is a "property", not an "entry", and the ALTREP is a "parameter" of the property, not a "property". See [RFC 2445](https://www.rfc-editor.org/rfc/rfc2445.html#section-4.2.1).

NeilRashbrook: Ugh, naming things... well, I know you can easily rename `properties` to `parameters` and `entries` to `properties`...

app/frontend/Util/util.ts
```diff
@@ -40,6 +40,27 @@ export async function stringToDataURL(mimetype: string, content: string): Promis
...
+      dataURL.substring(0, "data:".length + mimetype.length).toLowerCase() != "data:text/html") {
+    return null;
+  }
+  // There might be `;charset=utf-8` in-between
+  let comma = dataURL.indexOf(",");
+  console.log("comma", comma);
+  if (comma < 1) {
+    return null;
+  }
+  let htmlEncoded = dataURL.substring(comma + 1);
```
NeilRashbrook, line 55: Prefer to use `slice` rather than `substring`; it has the same behaviour for arrays and strings, so you don't have to worry about any of its gotachs (did you know that `"string".substring(4, 3) == "i"`?)

Also, why is this in `frontend`? There shouldn't be any dependencies of `logic` on `frontend`, should there?

app/frontend/Util/util.ts
```diff
@@ -40,6 +40,27 @@ export async function stringToDataURL(mimetype: string, content: string): Promis
...
+export function stringFromDataURL(dataURL: URLString, mimetype: string): string | null {
+  console.log("altrep URL", dataURL);
+  if (!dataURL || typeof (dataURL) != "string" ||
+      dataURL.substring(0, "data:".length + mimetype.length).toLowerCase() != "data:text/html") {
+    return null;
+  }
+  // There might be `;charset=utf-8` in-between
+  let comma = dataURL.indexOf(",");
+  console.log("comma", comma);
+  if (comma < 1) {
```
NeilRashbrook, line 52: Oh, and this is an unusual test.
```suggestion
  if (comma == -1) {
```

NeilRashbrook:
I tried this out in the debugger, setting a breakpoint on line 110 of `Invitation.svelte`, and `message.event.rawHTMLDangerous` was set to the formatted description as expected.

app/frontend/Util/util.ts
benbucksch: > Also, why is this in frontend? There shouldn't be any dependencies of logic on frontend, should there?

Oops, I didn't notice that. Thanks! I need to move it to logic.

app/logic/Calendar/ICal/ICalEMailProcessor.ts
```diff
@@ -21,7 +21,7 @@ export class ICalEMailProcessor extends EMailProcessor {
       return;
     }
     if (email.html) {
```
NeilRashbrook, line 23: This is my fault, I didn't fix it hard enough in my POC.
```suggestion
    if (email.rawHTMLDangerous) {
```

benbucksch: Works

Issue #619, opened 2025-05-16
Auto-process updates


benbucksch:

NeilRashbrook:
In particular, each participant needs to have a DTSTAMP, and ideally each event would also have its own DTSTAMP, and the RFC would like events to have a SEQUENCE as well.

benbucksch:
FIXED most of it, pair programming together with Neil

@NeilRashbrook Cancellation update is not being processed.

### Reproduction
1. Create event in TB
2. Invite Mustang email account
3. Accept in Mustang
4. In TB, delete the event and confirm notification
5. In Mustang, click on the "Cancelled" email

### Actual result
* The event is still in the calendar in Mustang

### Expected result
* The event disappeared from the calendar in Mustang

PR #696, opened 2025-07-02
Invitation: Auto-process replies and cancellations #619

benbucksch:
When we receive:
* A reply from a participant to an meeting that we organize
* A cancellation from an organizer

we should automatically update the information in the event in our calendar, so that the event has the most up-to-date status about all participants that responded, and can show that an event that we were previously invited to was now cancelled.

The processing should happen
a) immediately once we received the email
b) when the user reads the email

### DIFF
--- a/app/frontend/Calendar/DisplayEvent/Invitation.svelte
+++ b/app/frontend/Calendar/DisplayEvent/Invitation.svelte
@@ -75,7 +75,7 @@
             classes="maybe font-normal" />
         </ButtonMenu>
       {/if}
-    {:else if $message.invitationMessage && incomingInvitation}
+    {:else if ($message.invitationMessage == InvitationMessage.ParticipantReply || $message.invitationMessage == InvitationMessage.CancelledEvent) && incomingInvitation}
       {#await onUpdate()}
         <!-- Update processing -->
       {:then}
@@ -109,7 +109,6 @@
   let calendars: Collection<Calendar>;
   let selectedCalendar: Calendar | undefined;
   let myParticipation: InvitationResponse = InvitationResponse.NoResponseReceived;
-  let updateDisabled: boolean | string = false;
   let incomingInvitation: IncomingInvitation;
   let event: Event | undefined;
 
@@ -124,7 +123,6 @@
   async function loadCalendars() {
     calendars = message.getUpdateCalendars();
     selectedCalendar = calendars.find(calendar => calendar.events.some(event => event.calUID == message.event.calUID)) || calendars.first;
-    updateDisabled = calendars.length ? false : gt`This event is not in your calendar`;
     if (selectedCalendar) {
       await selectCalendar();
     }
@@ -153,7 +151,11 @@
     await respond(InvitationResponse.Decline);
   }
   async function onUpdate() {
-    await incomingInvitation.updateFromOtherInvitationMessage();
+    let foundEventInCalendars = message.getUpdateCalendars();
+    for (let calendar of foundEventInCalendars) {
+      let incomingInvitation = calendar.getIncomingInvitationFor(message);
+      await incomingInvitation.updateFromOtherInvitationMessage();
+    }
   }
 </script>
 
--- a/app/frontend/Calendar/DisplayEvent/Invitation.svelte
+++ b/app/frontend/Calendar/DisplayEvent/Invitation.svelte
@@ -76,7 +76,13 @@
         </ButtonMenu>
       {/if}
     {:else if $message.invitationMessage && incomingInvitation}
-      <Button label={$t`Update calendar`} disabled={updateDisabled} onClick={onUpdate} classes="font-normal" />
+      {#await onUpdate()}
+        <!-- Update processing -->
+      {:then}
+        <!-- Update processed -->
+      {:catch ex}
+        <ErrorMessageInline {ex} />
+      {/await}
     {/if}
   </hbox>
 </vbox>
@@ -88,6 +94,7 @@
   import type { IncomingInvitation } from "../../../logic/Calendar/Invitation/IncomingInvitation";
   import { InvitationMessage, InvitationResponse, type InvitationResponseInMessage } from "../../../logic/Calendar/Invitation/InvitationStatus";
   import InvitationDisplay from "./InvitationDisplay.svelte";
+  import ErrorMessageInline from "../../Shared/ErrorMessageInline.svelte";
   import ButtonMenu from "../../Shared/Menu/ButtonMenu.svelte";
   import MenuItem from "../../Shared/Menu/MenuItem.svelte";
   import Button from "../../Shared/Button.svelte";
@@ -147,7 +154,6 @@
   }
   async function onUpdate() {
     await incomingInvitation.updateFromOtherInvitationMessage();
-    updateDisabled = gt`This update has already been processed`;
   }
 </script>
 
--- a/app/logic/Abstract/PersonUID.ts
+++ b/app/logic/Abstract/PersonUID.ts
@@ -3,7 +3,6 @@ import { Group } from "./Group";
 import { ContactEntry, Person } from "./Person";
 import { appGlobal } from "../app";
 import { Observable, notifyChangedProperty } from "../util/Observable";
-import { NotReached } from "../util/util";
 import { ArrayColl } from "svelte-collections";
 
 export class PersonUID extends Observable {
@@ -14,6 +13,10 @@ export class PersonUID extends Observable {
   emailAddress: string;
   @notifyChangedProperty
   person?: Person;
+  // Used so that autocompleted meeting participants have a default value.
+  // They should really be Participant objects, but I'm cheating for now.
+  @notifyChangedProperty
+  protected response = 0;
 
   constructor(emailAddress?: string, name?: string) {
     super();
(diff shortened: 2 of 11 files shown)


benbucksch:
This should fix #619, but some parts are incomplete. @NeilRashbrook Can you help with the TODOs, please?

Also, we need to download and parse new incoming emails for this fix to take effect.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
```diff
@@ -147,7 +154,6 @@
   }
   async function onUpdate() {
     await incomingInvitation.updateFromOtherInvitationMessage();
-    updateDisabled = gt`This update has already been processed`;
```
NeilRashbrook, line 150: (Should the rest of the `updateDisabled` code be removed as well?)

app/logic/Abstract/PersonUID.ts
```diff
@@ -61,6 +58,15 @@ export class PersonUID extends Observable {
...
 
+  /** Used so that autocompleted meeting participants have a default value.
+   * They should really be Participant objects, but I'm cheating for now. */
+  get response() {
+    return 0;
+  }
+  set response(val: any) {
+    throw new NotReached("Only valid on Participant class, not PersonUID");
+  }
+
```
NeilRashbrook: (I'm not exactly sure when meeting participants become actual Participant objects...)

app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -35,25 +40,40 @@ export class IncomingInvitation {
...
   }
 
-  /** Handle the other two InvitationMessage cases:
-   *  CancelledEvent: the organiser cancelled an incoming meeting
-   *  ParticpantReply: an invitee replied to your outgoing invitation */
+  /** CancelledEvent: the organiser cancelled an incoming meeting */
+  async updateCancelled() {
+    assert(this.invitationMessage && this.invitationMessage != InvitationMessage.Invitation, "Can't update from an invitation");
+    let event = this.calEvent();
+    // Is this action reversible? If so, need to check timestamp.
```
NeilRashbrook, line 51: It's reversible in that if you were removed from a meeting then you can be added back, generating a new incoming invitation for you to accept, so you should check the cancellation DTSTAMP against the invitation DTSTAMP. Only two problems with that: a) we don't have DTSTAMP for events yet b) we don't handle rerequests correctly yet.

app/logic/Calendar/Participant.ts
```diff
@@ -4,12 +4,23 @@ import { notifyChangedProperty } from "../util/Observable";
...
-    this.response = response;
+    this._response = response;
+  }
+
+  get response(): InvitationResponse {
+    return this._response;
+  }
+  set response(val: InvitationResponse) {
+    this._response = val;
+    this.lastUpdateTime = new Date();
```
NeilRashbrook: This doesn't seem particularly useful; if you're setting this from an incoming update then it will have its own timestamp anyway, while if you're setting your own response, you need to update the event's timestamp instead.

app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -35,25 +40,40 @@ export class IncomingInvitation {
...
+    let organizer = event.participants.find(participant => participant.response == InvitationResponse.Organizer);
+    if (organizer) {
+      organizer.response = InvitationResponse.Decline;
+      await event.save();
+    }
+  }
+
+  /** ParticpantReply: an invitee replied to your outgoing invitation */
+  async updateParticipantReply() {
```
NeilRashbrook, line 61:
```suggestion
  /** ParticpantReply: an invitee replied to your outgoing invitation */
  protected async updateParticipantReply() {
```
(If this is going to be public API then the subclasses need to override it.)

app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -35,25 +40,40 @@ export class IncomingInvitation {
...
+    assert(this.invitationMessage && this.invitationMessage != InvitationMessage.Invitation, "Can't update from an invitation");
+    let event = this.calEvent();
+    let invitee = this.event.participants.find(participant => participant.response != InvitationResponse.Organizer);
+    let participant = event.participants.find(participant => participant.emailAddress == invitee.emailAddress);
+    let timestamp = this.message.sent; // TODO Use DTSTAMP from ICS
```
NeilRashbrook, line 67: We don't currently have an event field for that, but once we do, it's quite easy to parse from ICS, as it's always in UTC, but it's in iCal compact format, so you might as well send it through `parseDate`.
```diff
   }
+  if (vevent.entries.dtstamp) {
+    [event.timestamp] = parseDate(vevent.entries.dtstamp[0]);
+  }
   if (vevent.entries.dtstart) {
```

app/logic/Calendar/ICal/ICalEMailProcessor.ts
```diff
@@ -24,6 +24,10 @@ export class ICalEMailProcessor extends EMailProcessor {
       event.rawHTMLDangerous = email.rawHTMLDangerous;
     }
     email.event = event;
+
+    // TODO Call IncomingInvitation.update()
+    // let incomingInvitation = selectedCalendar.getIncomingInvitationFor(message);
+    // incomingInvitation.updateFromOtherInvitationMessage();
```
NeilRashbrook:
```suggestion
    if (email.invitationMessage > InvitationMessage.Invitation) {
      let foundCalendars = email.getUpdateCalendars();
      if (foundCalendars.length == 1) { // TODO what if we find multiple calendars?
        let selectedCalendar = foundCalendars.getAt(0);
        let incomingInvitation = selectedCalendar.getIncomingInvitationFor(message);
        await incomingInvitation.updateFromOtherInvitationMessage();
      }
    }
```

app/logic/Calendar/Participant.ts
```diff
@@ -4,12 +4,23 @@ import { notifyChangedProperty } from "../util/Observable";
 
 export class Participant extends PersonUID {
   @notifyChangedProperty
-  response: InvitationResponse;
+  _response: InvitationResponse;
+  /** DTSTAMP when the status was sent that is captured in this object.
+   * Used during auto-update to avoid overwriting with older info. */
+  lastUpdateTime: Date | null;
```
NeilRashbrook, line 12: I have a vague memory of someone complaining because my names didn't match the RFC, which calls this a timestamp (the RFC does also have a last modified time, which is a separate thing).

app/frontend/Calendar/DisplayEvent/Invitation.svelte
```diff
@@ -76,7 +76,13 @@
         </ButtonMenu>
       {/if}
     {:else if $message.invitationMessage && incomingInvitation}
-      <Button label={$t`Update calendar`} disabled={updateDisabled} onClick={onUpdate} classes="font-normal" />
+      {#await onUpdate()}
```
NeilRashbrook, line 79: `onUpdate` will fail if we didn't find a calendar. Also, what happens if we find multiple calendars?

app/logic/Calendar/Invitation/IncomingInvitation.ts
NeilRashbrook: (And then ICalGenerator will need to use the event's timestamp instead of the current date, and we need to update the timestamp if `changesNeedToNotify()` is true.)

app/logic/Calendar/ICal/ICalEMailProcessor.ts
benbucksch: Applied

app/logic/Calendar/Invitation/IncomingInvitation.ts
benbucksch: This is the public API.
Subclasses fixed.

app/logic/Calendar/Invitation/IncomingInvitation.ts
benbucksch: I've added the property `Event.lastUpdateTime` and the check to updateCancelled()

app/logic/Calendar/Invitation/IncomingInvitation.ts
benbucksch: both done

app/logic/Calendar/Participant.ts
benbucksch: Good point! Change reverted.

app/logic/Abstract/PersonUID.ts
benbucksch: Neither am I. You wrote it :)

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: yes

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: fixed, I think

NeilRashbrook:
This commit made my `tsc` output 141 new errors...

app/logic/Calendar/Invitation/IncomingInvitation.ts
NeilRashbrook: Except this line of code never got changed to use `this.event.lastUpdateTime`...

benbucksch: Make a PR, please?

benbucksch: I've just done it

PR #701, opened 2025-07-02
Start and stop editing recurring master so that its invitations work

NeilRashbrook:


benbucksch (approved):
Can you please explain why this was needed and why `unedited` affects invitations?

benbucksch:
@NeilRashbrook I still don't understand this change.

BTW: Why is the master.startEditing()/finishEditing() necessary? It doesn't make sense to me.

a) `unedited` is there only to allow Undo of the modification. Given that master is not being edited in the UI, it should be unnecessary.
b) master is new, so there should be no exceptions to clear.

Issue #707, opened 2025-07-03
Invitation: Incoming: Failed to find unique identity for meeting


jermy-c:
### Reproduction

**In the third party app:**

1. Create an event
2. Add your second email address as participant
3. [Send]

**In Mustang:**

4. Accept the invitation
5. Go to Calendar
6. Change the status to `Tentative` or `Decline`

### Actual Result

- Yellow banner
- Status not sent

### Expected Result

- No yellow banner
- Status sent

### Stack

```
error.ts:6 Error: Failed to find unique identity for meeting
    at assert (util.ts:4:11)
    at _Event.participantMe (Event.ts:388:5)
    at _Event.respondToInvitation (Event.ts:638:44)
    at respond (InvitationResponseButtons.svelte:18:17)
    at onTentative (InvitationResponseButtons.svelte:24:11)
    at HTMLButtonElement.myOnClick (Button.svelte:86:13)
```

jermy-c:
I'm also getting this error the Invitation UI above the email body in Mail.

NeilRashbrook:
When you accept the invitation from the Inbox, Mustang only checks that email account for identities. However, there is currently no way to associate an identity with a calendar, so when you try to accept the invitation from the calendar, it checks all of your accounts, and if it finds multiple potentially matching identities, it will throw this error.

I thought there was an issue regarding associating an identity with a calendar, but I can't find it right now.

NeilRashbrook:
Actually, there's a second bug, which I just discovered, so maybe you're actually seeing that instead.

NeilRashbrook:
Sorry, that was just me misreading TypeScript.

NeilRashbrook:
Forget my previous comments, I found a new error which I think is what you actually ran into. I'll open a PR.

jermy-c:
### Reproduction

1. Click create event
2. Change Calendar to Local calendar
3. Discard changes
4. Click create event
5. Change Calendar to EWS calendar
6. Click Invite

### Actual Result

- Yellow Banner with `No suitable email identity to organise this meeting`
- Cannot invite guests

### Expected Result

- No yellow banner
- Can invite guests

benbucksch:
@jermy-c My commit was after your last comment. Can you check again, please, to see whether my change fixed this bug?

PR #709, opened 2025-07-03
Fix accepting invations from local calendar #707

NeilRashbrook:
... I've no idea how that happened...

### DIFF
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -375,7 +375,7 @@ export class Event extends Observable {
 
   participantMe(mailAccount?: MailAccount): { identity: MailIdentity, myParticipant: Participant } {
     let results = [];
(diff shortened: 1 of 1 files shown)


jermy-c:
I'm getting a `To save an event, it needs to be in a calendar first` error. I'm on ActiveSync for Mail and Calendar.

jermy-c:
The invitation email gets deleted also.

NeilRashbrook:
If the accept API succeeds then ActiveSync will delete the email.

jermy-c:
> If the accept API succeeds then ActiveSync will delete the email.

But the `Accept` response was not sent. However, I do see the event afterwards in the calendar but without an organizer.

NeilRashbrook:
The ActiveSync error is a regression from #550. I should be able to fix it and file a PR.

NeilRashbrook:
The accept API does not send a response. The error arises when we try to send the response.

PR #710, opened 2025-07-03
Allow invitations to be deleted from the calendar

NeilRashbrook:
(The diff looks much nicer if you hide whitespace.)

In the case of a recurring master, I show the menu with only the "Delete entire series" option, as that's the only one that makes sense. Otherwise, I just show the usual delete button.

(Which makes me think... how do we handle the possibility of the user wanting to accept an entire series? They can (only!) do that from the initial message but not from the calendar.)

I also discovered that there was a bug in the delete code as it wants to send a cancellation notification to the organiser but the `respondToInvitation` function assumes you're saving rather than deleting and so the invitation reappears when you restart.

### DIFF
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -13,19 +13,19 @@
             iconSize="16px"
             />
         {/if}
-        {#if !event.isIncomingMeeting}
-          {#if $event.recurrenceCase == RecurrenceCase.Instance}
-            <ButtonMenu bind:isMenuOpen={isDeleteSeriesOpen}>
-              <RoundButton
-                slot="control"
-                label={$t`Delete event`}
-                icon={DeleteIcon}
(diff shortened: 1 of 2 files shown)


NeilRashbrook:
The delete code change I made is the wrong fix because other callers need to be able to send a response on an unsaved event.

benbucksch:
> how do we handle the possibility of the user wanting to accept an entire series? They can (only!) do that from the initial message but not from the calendar.)

The user wants to accept from both places, both from invitation and from calendar.

benbucksch (approved):

NeilRashbrook:
> > how do we handle the possibility of the user wanting to accept an entire series? They can (only!) do that from the initial message but not from the calendar.)
> 
> The user wants to accept from both places, both from invitation and from calendar.

Sure, but how are they supposed to achieve that?

NeilRashbrook:
(Also I should have put this into draft, sigh...)

NeilRashbrook:
That would simplify matters; we could put a check for the recurrenceCase of Instance and if so use the master event instead.

PR #711, opened 2025-07-03
Fix 56 TypeScript errors

NeilRashbrook:
If you don't like this approach I'll look for an alternative but it will be uglier.


app/logic/Calendar/ActiveSync/ActiveSyncIncomingInvitation.ts
```diff
@@ -15,15 +16,16 @@ export class ActiveSyncIncomingInvitation {
...
   myParticipation: InvitationResponse;
 
   constructor(calendar: ActiveSyncCalendar, message: ActiveSyncEMail) {
     this.calendar = calendar;
     this.message = message;
     this.invitationMessage = message.invitationMessage;
     this.event = message.event;
-    let event = calendar.events.find(event => event.calUID == this.event.calUID);
-    this.myParticipation = event?.myParticipation || InvitationResponse.NoResponseReceived;
+    this.calEvent = calendar.events.find(event => event.calUID == this.event.calUID);
```
benbucksch, line 27: Why this?

benbucksch (changes requested):

benbucksch:
The problem was that `OWAIncomingInvitation` does *not* inherit from `IncomingInvitation`, although it clearly is a more specific implementation of the same API.

You can use `declare` to type a class property more specifically than the superclass.

Both implemented in git commit 6ee24df7 . Please take a look.

Does that fix all issues you tried to fix here in this PR?

PR #712, opened 2025-07-03
Don't show meeting response buttons for cancelled events

NeilRashbrook:


app/logic/Calendar/Event.ts
```diff
@@ -452,6 +452,11 @@ export class Event extends Observable {
     return true;
   }
 
+  /** Only for incoming meetings */
+  get isCancelled(): boolean {
+    return !this.participants.some(participant => participant.response == InvitationResponse.Organizer);
```
benbucksch, line 457: Can we please make this cancelled flag more explicit? Capture the info when it reached us.

benbucksch (changes requested):

benbucksch (approved):

app/logic/Calendar/Event.ts
benbucksch: #718

PR #713, opened 2025-07-03
Allow responding to deleted and pseudo-invitations

NeilRashbrook:
In both cases we were trying to save something that shouldn't be saved.


benbucksch:
Can you explain more why we run i to this, please?

This is a bad API, because it assumes an order that is not necessarily given. The caller might set my status first and the call send(). I can find a solution, but I first need to understand the problem.

PR #714, opened 2025-07-03
Update the local calendar to the latest accepted invitation

NeilRashbrook:


app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -27,10 +27,9 @@ export class IncomingInvitation {
 
   async respondToInvitation(response: InvitationResponseInMessage) {
     assert(this.invitationMessage == InvitationMessage.Invitation, "Only invitations can be responded to");
-    let event = this.calEvent();
-    if (!event) {
-      event = this.calendar.newEvent();
-      event.copyFrom(this.event);
+    let event = this.calEvent() || this.calendar.newEvent();
+    event.copyFrom(this.event);
+    if (!this.calendar.events.contains(event)) {
```
benbucksch, line 32: This re-searching is unnecessary, if you separate the 2 cases.

app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -27,10 +27,9 @@ export class IncomingInvitation {
 
   async respondToInvitation(response: InvitationResponseInMessage) {
     assert(this.invitationMessage == InvitationMessage.Invitation, "Only invitations can be responded to");
-    let event = this.calEvent();
-    if (!event) {
-      event = this.calendar.newEvent();
-      event.copyFrom(this.event);
+    let event = this.calEvent() || this.calendar.newEvent();
+    event.copyFrom(this.event);
```
benbucksch, line 31: The copying should be conditional. The user might click [Accept] on an older invitation, although there has already been a newer update, e.g. a description change or participants change. We must not overwrite old data with newer data. So, I've added a timestamp check.

Fixed in git commit 5e3f4641

benbucksch (changes requested):

app/logic/Calendar/Invitation/IncomingInvitation.ts
NeilRashbrook: Actually we should probably show special UI to show that this invitation is out of date in that case.

NeilRashbrook: (But we should keep your change anyway, just to ensure that we don't copy an older event.)

benbucksch: > Actually we should probably show special UI to show that this invitation is out of date in that case.

Nice idea! Can you make a very simple POC, and I'll make the UI?

NeilRashbrook: You probably just have to add
```
      {:else if event && message.event.lastUpdateTime < event.lastUpdateTime}
        {$t`This invitation is out of date`}
```
to `Invitation.svelte`.

PR #717, opened 2025-07-06
Invitation: Incoming: Don't save after deleting and sending Declined response

benbucksch:
Responding to deleted and pseudo-invitations failed, because it they `respondToInvitation()` tried to `save()`.

Neil wrote: Both callers were (before https://github.com/mustang-im/mustang/pull/710) calling respondToInvitation on an incoming invitation, but they didn't want the event to be saved, one because it's a deleted event, the other because it's a pseudo-event. In https://github.com/mustang-im/mustang/pull/710 I wrote some custom code for deleting events but then I discovered

Alternative fix for #713

### DIFF
--- a/app/frontend/Calendar/DisplayEvent/Invitation.svelte
+++ b/app/frontend/Calendar/DisplayEvent/Invitation.svelte
@@ -109,10 +109,9 @@
   export let message: EMail;
   let calendars: Collection<Calendar>;
   let selectedCalendar: Calendar | undefined;
+  let myParticipation: InvitationResponse = InvitationResponse.NoResponseReceived;
   let incomingInvitation: IncomingInvitation;
   let event: Event | undefined;
-
-  $: myParticipation = incomingInvitation?.myParticipation ?? InvitationResponse.NoResponseReceived;
 
   $: if (message.event) {
     loadCalendars();
@@ -135,10 +134,12 @@
     }
     event = selectedCalendar.events.find(event => event.calUID == message.event.calUID);
     incomingInvitation = selectedCalendar.getIncomingInvitationFor(message);
+    myParticipation = incomingInvitation.myParticipation;
   }
 
   async function respond(response: InvitationResponseInMessage) {
     await incomingInvitation.respondToInvitation(response);
+    myParticipation = response;
     event ??= selectedCalendar.events.find(event => event.calUID == message.event.calUID);
   }
   async function onAccept() {


benbucksch:
I am not sure that this change is correct, because I don't have reproduction steps, so I cannot test it.
@NeilRashbrook Could you please describe the reproduction steps, and review the code, and test?

Merging it for now, because there's a bug and I hope this fixes it.

app/logic/Calendar/ActiveSync/ActiveSyncIncomingInvitation.ts
```diff
@@ -28,6 +28,7 @@ export class ActiveSyncIncomingInvitation extends IncomingInvitation {
     await this.calendar.account.callEAS("MeetingResponse", request);
     this.event.myParticipation = response;
     await this.event.respondToInvitation(response); // needs 16.x to do this automatically
+    await this.event.save();
```
NeilRashbrook, line 31: The whole point is that we don't actually want to save this event, because it's the pseudo-event.

benbucksch: Does this pseudo Event have to be an ActiveSyncEvent? If not, how about declaring a subclass InvitedEvent for clarity, and this class overrides save to assert()?

NeilRashbrook: It never has been an ActiveSyncEvent, so I could easily declare a subclass for you.

NeilRashbrook:
## Deleted Event
### Reproduction

1. IMAP account
2. Receive invitation
3. Accept from Inbox
4. Delete from calendar
5. Restart app

### Actual result

- Meeting reappears

## Pseudo event
### Reproduction

1. ActiveSync account
2. Receive incoming invitation by email
3. Click [Accept]

### Actual result

- Error "To save an event, it needs to be in a calendar first"
- Stack:
  - Event.saveLocally
  - Event.respondToInvitation
  - ActiveSyncIncomingInvitation.respondToInvitation
  - respond
  - onAccept

benbucksch:
@NeilRashbrook Ah, thanks. That helps. I can reproduce it, on master.
Can you please send a PR to fix it? Remove the `save()` calls that shouldn't exist, and add a code comment why we are *not* saving.

NeilRashbrook:
Actually, no it wasn't, since a `save` was since added to `respondToInvitationFromCalEvent`...

PR #722, opened 2025-07-07
Invitation: Incoming: Participation reactive

benbucksch:


app/frontend/Calendar/DisplayEvent/Invitation.svelte
```diff
@@ -134,13 +135,11 @@
...
     event = selectedCalendar.events.find(event => event.calUID == message.event.calUID);
     incomingInvitation = selectedCalendar.getIncomingInvitationFor(message);
-    myParticipation = incomingInvitation.myParticipation;
   }
 
   async function respond(response: InvitationResponseInMessage) {
     await incomingInvitation.respondToInvitation(response);
-    myParticipation = response;
-    event ??= selectedCalendar.events.find(event => event.calUID == message.event.calUID);
+    event ??= selectedCalendar?.events.find(event => event.calUID == message.event.calUID);
```
NeilRashbrook, line 142: What's this change for?

benbucksch:
> I don't think I understand Svelte reactivity to understand what this change actually achieves.

This change did not fix a bug, but ensures that the property always has the right value (thanks to `$` and observers) and removes the need for *other* code to set it correctly on every change.

app/frontend/Calendar/DisplayEvent/Invitation.svelte
benbucksch: While testing, I ran into sitations where `selectedCalendar` was null, so this code threw.

NeilRashbrook: Sure, but `incomingInvitation` is `selectedCalendar.getIncomingInvitationFor(message);`, so if that's not null...

Better still, would `event ??= incomingInvitation.calEvent();` work?

NeilRashbrook (commented):
You say observers but `IncomingInvitation` isn't an `Observable` which is confusing me...

PR #725, opened 2025-07-09
Calendar: Edit: Recurrence

benbucksch:
The old code was saving the changes in local variables in the UI, and then pushed it to the event on save. There are at least 2 problems with that:
1. It's against our general idea to a) apply changes directly on the object and b) that changes show up immediately in the rest of the app, whenever possible
2. More importantly, the `let foo =` in RepeatBox.svelte meant that the calendar edit dialog in the sidebar would not refresh and show the values of the previously selected event. This is because `let foo =` is run only when the component is instantiated. Later changes to the properties do *not* re-run that code, and therefore, the variables were not updated to the new data. To fix that, we need to use `$:`, which re-runs the line whenever one of the values changes (as far as Svelte knows, i.e. properties and local vars with assignment, i.e. anywhere on the left side of `=`,  including `foo` in `foo.bar = ...`, and observables with `$`).

However, once I changed to `$:`, the values could not be updated properly by simply writing to them using `bind:value={...}`. Probably, because they were re-calculated from the `event.recurrenceRule` before the latter could be changed.

So, instead of using `bind:`, this is now using `onFooChanged()` user input handlers, which is uglier and more code, and I'm not happy with it, but at least creates a well-defined code path.

This also immediately creates a new `RecurrenceRule` when the user makes changes, and applies it on the event, which means that the change shows in the user's calendar immediately. When he clicks [Revert], we use `unedited` on the master as well to restore the original state.

(Unfortunately, this "immediately see in your calendar" doesn't work for new events yet, simply because they are not yet added to the calendar. This should be fixable.)

This also moves some logic from the UI into the `Event` implementation.

### DIFF
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -1,6 +1,6 @@
 import type { Calendar } from "./Calendar";
 import type { Participant } from "./Participant";
-import { RecurrenceRule } from "./RecurrenceRule";
+import { RecurrenceRule, type RecurrenceInit, Frequency } from "./RecurrenceRule";
 import OutgoingInvitation from "./Invitation/OutgoingInvitation";
 import { InvitationResponse, type InvitationResponseInMessage } from "./Invitation/InvitationStatus";
 import type { MailAccount } from "../Mail/MailAccount";
@@ -310,6 +310,21 @@ export class Event extends Observable {
       this.clearExceptions();
     }
     this.generateRecurringInstances();
+  }
+
+  newRecurrenceRule(frequency: Frequency): void {
+    let init: RecurrenceInit = {
+      masterDuration: this.duration,
+      seriesStartTime: this.startTime,
+      frequency,
+      interval: 1
+    };
+    if (frequency == Frequency.Weekly) {
+      init.weekdays = [this.startTime.getDay()];
+    } else if (frequency == Frequency.Monthly || frequency == Frequency.Yearly) {
+      init.week = 0;
+    }
+    this.recurrenceRule = new RecurrenceRule(init);
   }
 
   /** Create a new instance of the same event.
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -323,11 +323,11 @@ export class Event extends Observable {
       interval,
     };
     if (frequency == Frequency.Weekly) {
-      init.weekdays = weekdays ?? [this.startTime.getDay()];
+      init.weekdays = weekdays ?? [this.startTime.getDay()]; // e.g. Monday and Thursday
     } else if (frequency == Frequency.Monthly || frequency == Frequency.Yearly) {
       init.week = week;
       if (week) {
-        init.weekdays = [this.startTime.getDay()];
+        init.weekdays = [this.startTime.getDay()]; // e.g. 3rd Wednesday of month
       }
     }
     this.recurrenceRule = new RecurrenceRule(init);
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -292,9 +292,9 @@ export class Event extends Observable {
    */
   protected clearRecurrenceRule() {
     if (this._recurrenceRule) {
-      this.clearExceptions();
-      this._recurrenceRule = null;
-      this.recurrenceCase = RecurrenceCase.Normal; // notifies
+      // this.clearExceptions(); in finishEditing()
+      this._recurrenceRule = null; // notifies
+      this.recurrenceCase = RecurrenceCase.Normal;
       this.instances.replaceAll([this]);
     }
   }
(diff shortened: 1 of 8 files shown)


benbucksch:
@jermy-c This is a massive change to how recurring events work. Could you please test this carefully? Also test saving recurring events to EWS, OWA, ActiveSync etc., and loading them in a new profile.

@NeilRashbrook Could you please code-review? Also, please test it, based on problematic cases that you know about.

benbucksch:
@NeilRashbrook

One thing that was very odd, and which took me many hours to debug, was the `if (!this.weekdays)` condition in the `Monthly` and `Yearly` cases in `RecurrenceRule.fillOccurrences()`. Why that condition? `this.weekdays` is not even used in the algo in these cases, and I don't see the inherent relevance, either.

app/logic/Calendar/RecurrenceRule.ts
```diff
@@ -270,7 +271,7 @@ export class RecurrenceRule implements Readonly<RecurrenceInit> {
         };
         break;
       case Frequency.Monthly:
-        if (!this.weekdays) {
+        if (!this.weekdays) { // TODO Why weekdays?
```
benbucksch: @NeilRashbrook See [comment](https://github.com/mustang-im/mustang/pull/725#issuecomment-3050881612)

NeilRashbrook: Sorry, I don't remember. Possibly originally copied from the Weekly code?

app/logic/Calendar/Event.ts
```diff
@@ -290,28 +291,48 @@ export class Event extends Observable {
...
-    this.recurrenceCase = RecurrenceCase.Master; // notifies
-    if (!timesMatch) {
-      this.clearExceptions();
-    }
+    this._muteObservers = true;
+    this.recurrenceCase = RecurrenceCase.Master;
+    this._muteObservers = false;
+    this._recurrenceRule = rule; // notifies
+    // clearExceptions() as necessary in finishEditing()
     this.generateRecurringInstances();
```
NeilRashbrook, line 315: You need to generate the recurring instances after you clear the exceptions, otherwise you might not generate an instance of an exception that gets cleared.

For example, if you have an event that repeats daily, with exceptions in 6 and 7 days, and change it to repeat weekly, the exceptions need to be cleared (because the repeat changed incompatibly) but this needs to happen before you generate the instances, so that you can generate an instance for next week.

I'm not sure how you want that to integrate with your dynamic update; for instance, you might want to keep the exceptions in `unedited` and restore them if you cancel editing.

(Also what was the point in switching the order of the assignments, then adding on a notification, then taking one away again?)

app/logic/Calendar/Event.ts
```diff
@@ -539,12 +572,14 @@ export class Event extends Observable {
...
     this.calUID ??= crypto.randomUUID();
     await this.calendar.storage.saveEvent(this);
+
     if (this.recurrenceCase == RecurrenceCase.Master) {
       this.generateRecurringInstances();
     } else if (this.recurrenceCase == RecurrenceCase.Instance) {
       this.recurrenceCase = RecurrenceCase.Exception;
       this.parentEvent.instances.remove(this);
       this.parentEvent.exceptions.add(this);
+      await this.parentEvent.save();
```
NeilRashbrook, line 582: This seems wrong; `saveLocally` is called by server implementations to sync events from the server.
```suggestion
      await this.parentEvent.saveLocally();
```

app/logic/Calendar/RecurrenceRule.ts
```diff
@@ -237,7 +240,8 @@ export class RecurrenceRule implements Readonly<RecurrenceInit> {
     if (this.occurrences.length < this.count && this.occurrences.at(-1)! < seriesEndTime) {
       this.fillOccurrences(this.count, seriesEndTime);
     }
-    return this.occurrences.filter(date => date >= seriesStartTime && date <= seriesEndTime);
+    return this.occurrences;
+    //return this.occurrences.filter(date => date >= seriesStartTime && date <= seriesEndTime);
```
NeilRashbrook, line 244: ... the whole point of this function was only to return that exact range of occurrences... Mind you all of those callers are now gone so...

app/logic/Calendar/Event.ts
```diff
@@ -433,8 +454,20 @@ export class Event extends Observable {
...
+        this.clearExceptions();
+      }
+    }
     this.unedited = null;
   }
+  cancelEditing() {
+    if (this.unedited) {
+      this.copyFrom(this.unedited);
+    }
+    this.finishEditing();
```
NeilRashbrook, line 469: Is this just to set `unedited` to `null` or does it have some greater purpose?

NeilRashbrook (commented):
> changes show up immediately in the rest of the app, whenever possible

At the time this was implemented, that was of course not possible; in particular there was no easy solution for the problem of exceptions and exclusions. The `unedited` event does make things a lot easier these days!

app/logic/Calendar/Event.ts
NeilRashbrook: In fact that's still wrong, because editing an instance doesn't actually affect the master event data in the database. The line should just be removed.

app/logic/Calendar/RecurrenceRule.ts
benbucksch: Yes, I saw that this is pointless work now.

app/logic/Calendar/Event.ts
benbucksch: "this" = ? Which part do you refer to?

This merely moved code from the UI into logic. This is called on revert button.

The `if` is there, because I often ran into situations where `unedited` was somehow lost (bug?), so every attempt to cancel the dialog failed, forcing me to restart the app.

app/logic/Calendar/Event.ts
benbucksch: OK, I will remove it

app/logic/Calendar/Event.ts
NeilRashbrook: Sorry, I forgot that Github pretends that you're commenting on a single line but you end up commenting on four; I was referring to the call to `finishEditing`, but I see you what you mean, you just moved `onCancel` here. Although there's then the call to `finishEditing` in `onClose`... gotta ensure that edit really finished :wink:

app/logic/Calendar/RecurrenceRule.ts
NeilRashbrook: I found a caller!

PR #726, opened 2025-07-09
Settings: About: Download and install app updates

benbucksch:
* In Settings | About, add a check for updates and an [Install update] button

### TODO:
* Check for update fails. No idea why.
* The updater API is rather wierd, with the events. We need to unsubscribe the events.

### DIFF
--- a/app/frontend/Settings/About/Update.svelte
+++ b/app/frontend/Settings/About/Update.svelte
@@ -2,12 +2,19 @@
   {#await checkForUpdate()}
     <div class="status">{$t`Checking for updates…`}</div>
   {:then}
-    {#if haveUpdate}
+    {#if haveUpdate === true}
       <div class="status">{$t`You can update`}</div>
       <Button label={$t`Install update`} onClick={installUpdate} />
-    {:else}
+    {:else if haveUpdate === false}
       <div class="status">{$t`This is the latest version`}</div>
       <Button label={$t`Check for update`} onClick={checkForUpdate} errorCallback={showError} />
+    {:else if haveUpdate === undefined}
+      <!--
+      <div class="status">
+        {$t`Check for updates failed.`}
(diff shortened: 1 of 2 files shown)


benbucksch:
> you mean it's always remains in the Checking for updates…

Yes

> you need to build the app for the autoUpdater to work

You mean `yarn build` or `yarn build:release:mac` ?

jermy-c:
> You mean yarn build or yarn build:release:mac ?

I normally just do yarn `build:mac`

benbucksch:
@jermy-c Can you please finish this PR? Make it work?

(I remember you specifically asked for this a long time ago.)

jermy-c (approved):
I tested this and checking the status for updates works. But I haven't tested the install update button.

jermy-c:
I forgot to mention it works with: https://github.com/mustang-im/mustang/pull/726/commits/c499b35aa061717760ce0f9e3ab9d2e61a2993b0

benbucksch:
Look at the code that you removed:
```
//return result?.isUpdateAvailable; TODO always false
```
So, we already tried that, and it didn't work. At least not on all OSes.

benbucksch:
> I haven't tested the install update button.

Were you able to test this?

You don't need to build an update yourself. You just need to set the local version to be *lower* than the release version.

jermy-c:
It became `true` for me on MacOS. Here's how I tested it:

1. Merge the changes from [jeremy/update-test](https://github.com/mustang-im/mustang/tree/jeremy/update-test), to test without building
2. `(cd app; bash build/version.sh 0.9.0)` or a higher version
3. open a terminal for `app/` and `yarn dev`
4. open another terminal for `e2/` and `yarn dev`

Could you please test it also?

benbucksch:
`update` is null:

image

Issue #668, opened 2025-06-26
iOS: node.js


benbucksch:
Simple prototype that adds the library for node.js to the iOS app directly in Xcode, making node.js start and stop together with the app:
https://github.com/mustang-im/ios-nodejs

jermy-c:
> On iOS, WASM is unsupported because it requires Just-in-time (JIT) interpretation, which is forbidden by Apple's App Store guidelines.

This causes some errors because the Node.js `fetch()` API relies on WASM.

PR #733, opened 2025-07-10
iOS: Frontend and backend working with Node.js v18 (#668)

jermy-c:
- Working frontend and backend for iOS with Node.js v18
- `fetch()` is not supported on Node.js for iOS, used `node-fetch` for that and disabled `fetch`
- WASM is not supported on iOS


mobile/docs/structure.md
```diff
@@ -6,8 +6,29 @@ The app config directory and files directory should be in `/data/data/im.mustang
 
 [link](https://android.stackexchange.com/a/47951)
 
+
+### iOS
+
+#### Node Native
+
+Node native modules should not be included in the Frameworks section of `App/Target` setting in Xcode.
```
benbucksch, line 14: Thanks for this warning.
Can you please add an explanation why?

mobile/docs/structure.md
```diff
@@ -6,8 +6,29 @@ The app config directory and files directory should be in `/data/data/im.mustang
...
+
+#### Debugging
+
+`console.log` are printed to the debug console in Xcode therefore you must you Xcode to see it. Even the frontend `console.log` are printed to the debug console in Xcode.
```
benbucksch, line 18: Thanks for documenting that. This is very important.

Can you please check whether there are other ways to see the console.log() ?

mobile/ios/App/App/NodeRunner.swift
```diff
@@ -0,0 +1,55 @@
+//
+//  NodeRunner.swift
+//  App
+//
+//  Created by Juergen Pona on 06.05.2025.
```
benbucksch:
```suggestion
//  (c) 2025 Mustang code
```

mobile/ios/App/App/NodeRunner.swift
```diff
@@ -0,0 +1,55 @@
...
+            // Prepare argv: ["node", "<full-path-to-index.mjs>"]
+            let args = ["node", "--no-experimental-fetch", scriptPath]
+            let argc = Int32(args.count)
+            let argv = UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>.allocate(capacity: Int(argc))
+            for (i, arg) in args.enumerated() {
+                argv[i] = strdup(arg)
+            }
+
+            // Launch the embedded Node.js runtime
+            node_start(argc, argv)
```
benbucksch, line 52: Can we link libnode as dynamic library into the main app executable, like we do on Android, instead of starting node as separate process here?

(Should happen in a new PR)

mobile/ios/App/App/Info.plist
```diff
@@ -22,6 +22,19 @@
...
+		<dict>
+			<key>localhost</key>
+			<dict>
+				<key>NSExceptionAllowsInsecureHTTPLoads</key>
+				<true/>
+			</dict>
+		</dict>
+	</dict>
+	<key>NSLocalNetworkUsageDescription</key>
+	<string>This app needs to communicate with a local backend via WebSocket for mail synchronization.</string>
```
benbucksch:
```suggestion
	<string>The app frontend needs to communicate with a local backend via WebSocket for mail synchronization, storage and other tasks. We want the app to run entirely local without any cloud service, that's why the service runs locally on the device.</string>
```

benbucksch (approved):
Nice! I like your minimal changes.

benbucksch:
@jermy-c This is great. I am very happy that you managed to get this working! Well done, congratulations!

:partying_face: :tada: :confetti_ball: :champagne: :100:

benbucksch:
I'll merge it as-is. @jermy-c Can you please fix the review comments in followup PRs?

mobile/docs/structure.md
jermy-c: Ok, will do. It was because it causes a memory error when running the app. And if you check the debug log you'll see `Release/[node native] not found` something like that.

mobile/docs/structure.md
jermy-c: There might a to debug using the command line: https://stackoverflow.com/a/10112466

mobile/ios/App/App/NodeRunner.swift
jermy-c: I'll look for if we can do that because the libnode for iOS is a static library.

However, even for Android we are still running Node.js on a background thread because it blocks the UI. 
https://github.com/hampoelz/Capacitor-NodeJS/blob/f32ceab7221bcc74934a463ae96ac967a8b027a1/android/src/main/java/net/hampoelz/capacitor/nodejs/CapacitorNodeJS.java#L117

jermy-c: We have to run Node.js in a background thread.

> No, Node.js uses a libuv event loop at its core, which is different than the event loop in the WebView. Having the node runtime run in its own thread also prevents Node.js tasks for interfering with the UI, which might cause responsiveness issues.

benbucksch: A thread is still a lot better than a different process, because the lifetime of the thread is automatically tied to the main process, whereas subprocesses are normally not. But that seems to be not an issue on iOS, because iOS will kill any subprocesses together with the main app, as I'm reading.

So, we can keep this solution for now, if it works well

PR #735, opened 2025-07-11
Create InvitationEvent subclass for events used for iMip

NeilRashbrook:
As requested in #717.

(I guess the other approach would be to have the `Event` be a subclass of `InvitationEvent`, but that would mean moving the saving and deletion functions.)

### DIFF
--- a/app/logic/Calendar/Invitation/InvitationEvent.ts
+++ b/app/logic/Calendar/Invitation/InvitationEvent.ts
@@ -1,0 +1,25 @@
+import { Event } from "../Event";
+
+export class InvitationEvent extends Event {
+  constructor() {
+    super();
+  }
+  async save(): Promise<never> {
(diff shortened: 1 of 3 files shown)


app/logic/Calendar/Invitation/InvitationEvent.ts
```diff
@@ -0,0 +1,25 @@
+import { Event } from "../Event";
+
+export class InvitationEvent extends Event {
+  constructor() {
+    super();
+  }
+  async save(): Promise<never> {
+    throw new Error("InvitationEvent holds a temporary copy of event data for incoming and outgoing invitations only");
```
benbucksch:
```suggestion
    throw new NotReached("InvitationEvent holds a temporary copy of event data for incoming and outgoing invitations only");
```

benbucksch:
Removed from master

PR #737, opened 2025-07-13
Invitation: Incoming: Create InvitationEvent subclass for events used for iMip

benbucksch:
As discussed in #717. Re-try of PR #735, which caused a runtime startup error due to circular imports.

### DIFF
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -2,6 +2,7 @@ import type { Calendar } from "./Calendar";
 import type { Participant } from "./Participant";
 import { RecurrenceRule, type RecurrenceInit, Frequency } from "./RecurrenceRule";
 import OutgoingInvitation from "./Invitation/OutgoingInvitation";
+import { InvitationEvent } from "./Invitation/InvitationEvent";
 import { InvitationResponse, type InvitationResponseInMessage } from "./Invitation/InvitationStatus";
 import type { MailAccount } from "../Mail/MailAccount";
 import type { MailIdentity } from "../Mail/MailIdentity";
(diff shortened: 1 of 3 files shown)


app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -2,7 +2,7 @@ import type { Calendar } from "../Calendar";
 import type { Event } from "../Event";
 import { InvitationMessage, InvitationResponse, type InvitationResponseInMessage } from "../Invitation/InvitationStatus";
 import type { EMail } from "../../Mail/EMail";
-import { assert } from "../../util/util";
+import { AbstractFunction, assert } from "../../util/util";
 
 export class IncomingInvitation {
```
NeilRashbrook, line 7:
```suggestion
export abstract class IncomingInvitation {
```

app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -26,23 +26,7 @@ export class IncomingInvitation {
...
-    let { myParticipant } = event.participantMe(this.message.folder.account);
-    /* else add participant? */
-    let hasChanged = event.myParticipation != response;
-    event.myParticipation = this.myParticipation = myParticipant.response = response;
-    await event.save();
-    if (hasChanged) {
-      await event.sendInvitationResponse(myParticipant, this.message.folder.account);
-    }
+    throw new AbstractFunction();
   }
```
NeilRashbrook, line 30:
```suggestion
  abstract respondToInvitation(response: InvitationResponseInMessage): Promise<void>;
```

benbucksch:
> What do you mean when you say "iCal invitation"?

An invitation to a meeting, sent from a personal calendar or CalDAV calendar, via an IMAP or POP3 mail account, in the iCal format.

NeilRashbrook:
> in the iCal format.

... all invitations are sent by iMip in the iCal format, that's the point...

So basically, what you want is:
- An abstract base Invitation class (which just contains the constructor and the `calEvent` function; everything else is abstract)
- An Invitation class for each Exchange calendar type
- An Invitation class for local or CalDAV calendars

I just think that the latter class could be better named :wink:

PR #738, opened 2025-07-13
Invitation: Incoming: Save isCancelled as explicit property on the calendar event

benbucksch:


app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -71,14 +71,20 @@ export class ActiveSyncEvent extends Event {
...
     this.alarm = wbxmljs.Reminder ? new Date(this.startTime.getTime() - k1MinuteMS * sanitize.integer(wbxmljs.Reminder)) : null;
     this.location = sanitize.nonemptystring(wbxmljs.Location, "");
+    this.isCancelled = (sanitize.integer(wbxmljs.MeetingStatus, 0) & 4) !== 0;
     let attendees = ensureArray(wbxmljs.Attendees?.Attendee);
     if (wbxmljs.OrganizerEmail) {
-      let status = wbxmljs.MeetingStatus & 4 ? InvitationResponse.Decline : InvitationResponse.Organizer;
-      let organizer = attendees.find(attendee => attendee.Email == wbxmljs.OrganizerEmail);
+      let status = this.isCancelled ? InvitationResponse.Decline : InvitationResponse.Organizer;
+      let organizerEmail = sanitize.emailAddress(wbxmljs.OrganizerEmail);
+      let organizer = attendees.find(attendee => attendee.Email == organizerEmail);
```
NeilRashbrook, line 79: `sanitize` will lowercase the email address, which could cause the attendee lookup to fail, as Exchange preserves case.

benbucksch: `attendee.Email?.toLowerCase()`, then

FWIW, comparing email addresses case sensitive is almost always a bug, because some people write them capitalized, and other people write the *same* email address as lower case, and they are the same address.

benbucksch:
@NeilRashbrook I didn't really understand. Can you please make a PR?

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
NeilRashbrook: I checked the other implementations, and there I'd made the reverse error of only sanitising the attendee email and not the organiser... :sweat: I'll fix that as part of the fix for #747.

benbucksch: Done

Issue #736, opened 2025-07-13
Calendar: OWA: Sets GMT +00:00 (Reykjavic) as timezone for recurring events


benbucksch:
### Reproduction
1. Set up neil@beone with EWS as account "Neil EWS", in yellow
2. Set up ben@beone with OWA as account "Ben OWA", in green
3. In calendar, create a new event
4. Select the "Ben OWA" account
5. Invite neil@beone
6. Make the event recurring (required for this bug)
7. Refresh calendars
8. Click on the event

### Actual result
* "Neil EWS" has the event in the calendar (correct)
* The event shows as normal event, although it's not yet accepted (other bug)
* The time field shows the time you set (correct)
* The timezone "Reykjavik GMT +00:00" shows, and additionally you see the time in that timezone.

### Expected result
* No timezone shown, given that you created the event, and you view it, i.e. both sender and recipient are in the same timezone.

### Other situations
* Sending the event in EWS and receiving it in OWA (i.e. the opposite direction) does *not* show the but.

### Likely cause
* In OWA, do we send timezone Z (GMT without DST)?
* Please check ActiveSync as well

jermy-c:
The timezone is fixed for OWA but it still adds yourself as a organizer event if you don't click invite.

benbucksch:
Can you file a new bug, please?

PR #745, opened 2025-07-14
Really set the time zone for Office 365 calendar requests #736

NeilRashbrook:
I don't know why Office 365 is special but this does at least work...

### DIFF
--- a/app/logic/Calendar/OWA/OWAEvent.ts
+++ b/app/logic/Calendar/OWA/OWAEvent.ts
@@ -214,6 +214,13 @@ export class OWAEvent extends Event {
     if (this.calendar.account.isOffice365()) {
       request.addField("CalendarItem", "StartTimeZoneId", timezone, "calendar:StartTimeZoneId");
       request.addField("CalendarItem", "EndTimeZoneId", timezone, "calendar:EndTimeZoneId");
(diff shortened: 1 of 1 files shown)


app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -214,6 +214,13 @@ export class OWAEvent extends Event {
     if (this.calendar.account.isOffice365()) {
       request.addField("CalendarItem", "StartTimeZoneId", timezone, "calendar:StartTimeZoneId");
       request.addField("CalendarItem", "EndTimeZoneId", timezone, "calendar:EndTimeZoneId");
+      (request.Header as any).TimeZoneContext = {
```
benbucksch, line 217: Why is the cast necessary?

benbucksch (approved):

app/logic/Calendar/OWA/OWAEvent.ts
NeilRashbrook: Because Header doesn't normally have a `TimeZoneContext` field. (Several different objects are involved, and TypeScript won't know which of them are relevant, so I would have to add it to each one separately with an `any` type and `undefined` value.)

Issue #739, opened 2025-07-14
Invitation: Incoming: Change to participation within calendar event does not stick


benbucksch:
### Reproduction
1. Set up neil@be as EWS account and ben@be as OWA account
2. As neil@be, create meeting
3. Invite ben@be
4. Send
5. As ben@be, [Accept] the invitation.
6. Change it to [Decline]
7. Click on another event
8. Click on this event again

### Actual result
* After step 5:
  * You see the event as [Confirmed] (correct)
  * neil@be gets a notification about it (correct)
* After step 6:
  * You see the event as [Declined] (correct)
  * neil@be gets a notification about it (correct)
* After step 8:
  * You see the event as [Confirmed] (Bug)
  * neil@be gets no notification about it (correct)

### Expected result
* Same as Actual result, but after Step 8, the event shows as [Declined], not [Confirmed].

benbucksch:
FIXED

PR #748, opened 2025-07-14
Exchange: Sync event after changing its response #739

NeilRashbrook:
This will ensure that the event is up-to-date on the client.

I think there's a setting as to whether to delete invitations that you decline, but it might only be available for Office 365. As such I'm not sure what the best approach for OWA is. It's possible that more information is available from the response API.


app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -378,6 +378,15 @@ export class OWAEvent extends Event {
       Id: this.itemID,
     });
     await this.calendar.account.callOWA(request);
+    let events = new ArrayColl<OWAEvent>();
+    try {
+      await this.calendar.getEvents([this.itemID], events);
+    } catch (ex) {
+      console.error(ex);
```
benbucksch, line 386: Why the console.log()? We don't do this, ever. Use either
* `this.calendar.errorCallback(ex)` (adds to server error log - used for errors in individual items) or
* `backgroundError(ex)` (log only, don't show to user, but also logs to error log server), depending on the severity of the error, or
* if the error is expected, we handle or ignore that **specific** error only.

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -378,6 +378,15 @@ export class OWAEvent extends Event {
...
+    }
+    if (events.isEmpty) { // OWA deleted the event from the server
```
benbucksch, line 388: Putting this outside the try/catch will delete the event in case of unexpected server error, e.g. an server protocol change that we didn't adapt to, or throttling, or anything else.

If you ran into a specific server error that you want to treat as if the element was deleted, then catch that specific error only, based on its error ID, and threat it as if the element was deleted. All *other* server errors should not cause us to delete the event.

app/logic/Calendar/OWA/OWAEvent.ts
NeilRashbrook: You're right, I meant to check for the specific error. I must have been too tired...

app/logic/Calendar/OWA/OWAEvent.ts
NeilRashbrook: (moving the code makes no difference; if there is no error then the array will never be empty)

app/logic/Calendar/OWA/OWAEvent.ts
benbucksch: Done

benbucksch:
Done in 300ba14c

Issue #729, opened 2025-07-09
Calendar: Edit: Recurrence: Changing the remainder of series turns it into a single event


jermy-c:
### Reproduction

1. Create a event
2. Add title and start time
3. Add a weekly recurrence
4. Save event
5. Open the second recurrence of the series
6. Edit the title
7. Click save
8. Select `Change remainder of series`

### Actual Result

- The remainder of the series of after the selected event is deleted
- Recurrence of the selected event is gone
- The first recurrence is still there and it's still a recurring event

### Expected Result

- Recurrence still remains on the selected event
- The remainder of the series after the selected event is not deleted

NeilRashbrook:
Regression from #725.

PR #749, opened 2025-07-14
Fix bugs handling remainder of series #725 #729

NeilRashbrook:
Prior to #725, a recurring master's rule didn't get updated until you clicked save. This meant that `onChangeRemainder` would automatically pick up the new rule as it saved the new series, while `Event.seriesStatus` and `Event.truncateRecurrence` were both relying on the master's rule being unedited. `onChangeRemainder` now has to clone the rule, if it wasn't turned off, `Event.seriesStatus` has to now check the unedited state, while for `Event.truncateRecurrence` I just cancel editing to simplify matters. Furthermore, `Event.seriesStatus` suffered the double whammy of an unedited instance not being able to find itself in its master's instance list, presumably due to the recent more dynamic instance generation changes.

### DIFF
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -239,8 +239,14 @@
     master.startEditing();
     master.copyEditableFieldsFrom(event);
     master.calUID = null;
+    if (event.parentEvent.recurrenceRule) {
+      let { frequency, interval, week, weekdays } = event.parentEvent.recurrenceRule;
+      master.newRecurrenceRule(frequency, interval, week, weekdays);
+    }
     await saveEvent(master);
     master.finishEditing();
+    event.parentEvent.cancelEditing();
+    event.cancelEditing();
     await event.truncateRecurrence();
     onClose();
   }
@@ -272,6 +278,8 @@
   }
 
   async function onDeleteRemainder() {
+    event.parentEvent.cancelEditing();
+    event.cancelEditing();
     await event.truncateRecurrence();
     $selectedEvent = null;
     onClose();
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -526,16 +526,17 @@ export class Event extends Observable {
     // Normally the parent of an instance would always have a
     // recurrence rule, but this might get removed during saving,
     // and would cause Svelte to crash if we didn't handle it.
-    let rule = this.parentEvent?.recurrenceRule;
+    let master = this.parentEvent?.unedited ?? this.parentEvent;
+    let rule = master?.recurrenceRule;
     if (!rule) {
       return "none";
     }
-    if (rule.countIs(this.parentEvent.exclusions.length + 1)) {
+    if (rule.countIs(master.exclusions.length + 1)) {
       return "only";
     }
     // Not supporting "last" any more; fortunately nobody needs it
     // (it's prohibitively expensive to compute these days)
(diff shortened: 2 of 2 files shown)


app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -239,8 +239,14 @@
     master.startEditing();
     master.copyEditableFieldsFrom(event);
     master.calUID = null;
+    if (event.parentEvent.recurrenceRule) {
+      let { frequency, interval, week, weekdays } = event.parentEvent.recurrenceRule;
+      master.newRecurrenceRule(frequency, interval, week, weekdays);
+    }
     await saveEvent(master);
     master.finishEditing();
```
benbucksch, line 247: BTW: Why is the master.startEditing()/finishEditing() necessary? It doesn't make sense to me.
a) `unedited` is there only to allow Undo of the modification. Given that master is not being edited in the UI, it should be unnecessary.
b) master is new, so there should be no exceptions to clear.

benbucksch, line 247: FWIW, this whole block (from master = newEvent() until master.finishEditing()) should probably move into the Event class as `cloneSeriesStartingAt(start: Date)`

app/logic/Calendar/Event.ts
```diff
@@ -526,16 +526,17 @@ export class Event extends Observable {
     // Normally the parent of an instance would always have a
     // recurrence rule, but this might get removed during saving,
     // and would cause Svelte to crash if we didn't handle it.
-    let rule = this.parentEvent?.recurrenceRule;
+    let master = this.parentEvent?.unedited ?? this.parentEvent;
```
benbucksch, line 529: Why the unedited?

app/frontend/Calendar/EditEvent/DialogHeader.svelte
benbucksch: Moved to #701

benbucksch: I started this as #750. Continue there.

app/logic/Calendar/Event.ts
NeilRashbrook: Because otherwise editing the event can change what it thinks its series status is.

app/logic/Calendar/Event.ts
```diff
@@ -526,16 +526,17 @@ export class Event extends Observable {
...
       return "none";
     }
-    if (rule.countIs(this.parentEvent.exclusions.length + 1)) {
+    if (rule.countIs(master.exclusions.length + 1)) {
       return "only";
     }
     // Not supporting "last" any more; fortunately nobody needs it
     // (it's prohibitively expensive to compute these days)
-    return this.isNew ? this == this.parentEvent.instances.first ? "first" : "middle" : "exception";
+    return this.isNew ? this.recurrenceStartTime.getDate() == master.instances.first.recurrenceStartTime.getDate() ? "first" : "middle" : "exception";
```
NeilRashbrook, line 539: Bah, just noticed I typo'd this...
```suggestion
   return this.isNew ? this.recurrenceStartTime.getTime() == master.instances.first.recurrenceStartTime.getTime() ? "first" : "middle" : "exception";
```

benbucksch: Put on master in git commit 122c33f

NeilRashbrook: Thanks!

PR #750, opened 2025-07-15
Calendar: Edit: Recurrence: Move changeRemainder() from UI to logic - try 1

benbucksch:


benbucksch:
We need a new property and 2 new functions:
* `Event.setSeriesEndTime(end: Date)`
* `Event.seriesEndTime: Date` with getter and setter (calls `setSeriesEndTime()`)
* `RecurrenceRule.getOccurenceOnOrAfter(time: Date)` returns the occurrence exactly on that time, or the next after that, or null (if there is none after that time)

NeilRashbrook:
So, one big problem with your current code is that it doesn't actually delete the event you selected...

benbucksch:
> one big problem with your current code is that it doesn't actually delete the event you selected...

Can you please post the reproduction steps?
And which code function you mean, what you think it should do where?

benbucksch:
This code still doesn't work. When "changing remainder", it deletes all instances *after* this one, and restores this one to the original instance.

benbucksch:
Alternative #763 has almost the same problem.

app/logic/Calendar/Event.ts
```diff
@@ -743,19 +743,23 @@ export class Event extends Observable {
...
     this.recurrenceRule = new RecurrenceRule({ masterDuration, seriesStartTime, count, frequency, interval, weekdays, week, first });
     await this.trimExceptionsAndExclusions(this.recurrenceRule.getOccurrenceByIndex(count - 1));
   }
 
   /**
    * Changes the current recurrence to be limited by end time.
-   */
+   *
+   * Removes all instances, exceptions and exclusions
+   * on or after that `end` time. */
```
NeilRashbrook, line 760: As I mentioned in #760, this comment is false.

PR #751, opened 2025-07-15
Calendar: Edit: Move calendar - after introducing unedited

benbucksch:
### Reproduction

1. Click create an event
2. Select a different calendar than the initially selected
3. Click to add any of the following: Repeat, Reminder, Invite, Location, Online meeting, Description

### Actual Result

- Selected calendar reverts back to initially selected calendar

### Expected Result

- Selected calendar stays the same

### DIFF
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -166,7 +166,7 @@
 
   export let event: Event;
 
-  let newCalendar: Calendar;
+  let newCalendar = event.calendar;
   $: event.startEditing(); // not `$event`
   $: hasMinimalProps = event && $event.title && $event.startTime && $event.endTime &&
       event.startTime.getTime() <= event.endTime.getTime();
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -166,8 +166,8 @@
 
   export let event: Event;
 
-  let newCalendar = event.calendar;
   $: event.startEditing(); // not `$event`
+  let newCalendar = event.calendar; // Don't listen to $event #730
   $: hasMinimalProps = event && $event.title && $event.startTime && $event.endTime &&
       event.startTime.getTime() <= event.endTime.getTime();
   $: hasMinimalPropsChanged = hasMinimalProps && $event.hasChanged();
@@ -180,7 +180,6 @@
   $: participants = event.participants;
   $: willSend = $participants.hasItems && !$event.isIncomingMeeting;
   $: isFullWindow = $selectedApp instanceof CalendarEventMustangApp;
-
   let isSaveSeriesOpen = false;
   let isDeleteSeriesOpen = false;
 
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -168,6 +168,11 @@
 
   $: event.startEditing(); // not `$event`
   let newCalendar = event.calendar; // Don't listen to $event #730
+  let prevEventID = event.id;
+  $: if (event.id != prevEventID) {
+    newCalendar = event.calendar;
+    prevEventID = event.id;
+  };
   $: hasMinimalProps = event && $event.title && $event.startTime && $event.endTime &&
       event.startTime.getTime() <= event.endTime.getTime();
   $: hasMinimalPropsChanged = hasMinimalProps && $event.hasChanged();
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -166,13 +166,8 @@
 
   export let event: Event;
 
+  let newCalendar: Calendar;
   $: event.startEditing(); // not `$event`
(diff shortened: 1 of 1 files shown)


app/logic/Calendar/Event.ts
```diff
@@ -639,10 +639,12 @@ export class Event extends Observable {
...
    * @returns the new (or same) Event object */
   async moveToCalendar(newCalendar: Calendar): Promise<Event> {
-    if (this.calendar == newCalendar || !newCalendar) {
+    if (this.unedited?.calendar == newCalendar ||
+      (this.calendar == newCalendar && !this.unedited)
+      || !newCalendar) {
       return this;
     }
-    this.calendar?.events.remove(this);
+    (this.unedited.calendar || this.calendar)?.events.remove(this);
```
jermy-c: This is where I'm really concerned because we could have 3 possibilities of calendars:

1. this.uneditied.calendar - this might undefined or we might remove from here
2. this.calendar - we might add or remove in this calendar or both
3. newCalendar - we add in this calendar

I'm a bit confused at this point when making changes to the calendars.

I think it would be clearer if it was only 2 possibilities:

1. this.unedited.calendar - we always remove from this one
2. this.calendar - we always add to this one

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -211,7 +211,7 @@
   }
 
   async function saveEvent(event: Event) {
-    if (event.calendar != newCalendar && event.calendar && newCalendar) {
+    if (event.unedited.calendar != newCalendar && event.calendar && newCalendar) {
       // `moveToCalendar()` does the save and delete as well
       event = await event.moveToCalendar(newCalendar);
```
jermy-c: This is where it would not work. Because event.calendar and newCalendar are the same calendar. But if you do event.unedited.moveToCalendar() then it would copy the unedited event details over so if a user made changes and wants to move it to the newCalendar it would only move it.

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -167,7 +167,7 @@
   export let event: Event;
 
   $: event.startEditing(); // not `$event`
-  let newCalendar = event.calendar; // Don't listen to $event #730
+  $: newCalendar = event.calendar; // Don't listen to $event #730
```
benbucksch: That won't work. Please see the comment!

As soon as you change `event`, e.g. `event.prop = something`, svelte notices that `event` changed, because it's on the right of an `=` assignment, and triggers the `$:` statement, re-runs it, and re-sets `newCalendar` back to its original value, overwriting what the user changed.

jermy-c: This works because here I'm changing the event.calendar directly which means newCalendar is that:

https://github.com/mustang-im/mustang/blob/cd20ee572527b2d47945c45ced664ac53b038e81/app/frontend/Calendar/EditEvent/DialogHeader.svelte#L281

But this was the reason why I said it would mess up the logic.

app/logic/Calendar/Event.ts
benbucksch: > this.unedited.calendar - we always remove from this one
> this.calendar - we always add to this one

Makes sense. But then the data model is incoherent, because the event is not in the collection of the calendar, i.e. `event.calendar.events.find(ev => ev == event)` is false

jermy-c:
I made a change so it only updates if the selected events has changed. I don't think there will be any issues if the event has changed calendars while editing because events are deleted and created.

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -168,6 +168,11 @@
 
   $: event.startEditing(); // not `$event`
   let newCalendar = event.calendar; // Don't listen to $event #730
+  let prevEventID = event.id;
+  $: if (event.id != prevEventID) {
```
benbucksch: Same problem as before. As soon as you have `$: newCalendar = ... event...` on a line, it triggers on `event.foo = bla`

jermy-c: That's why we don't use that anymore because that assigns newCalendar on every update. This will trigger on every update but it won't assign it on every update.

benbucksch: Oh, right. Sorry, I got confused.

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -168,6 +168,11 @@
...
+    newCalendar = event.calendar;
+    prevEventID = event.id;
+  };
```
benbucksch: Write multiple lines as function, like so:

```suggestion
  $: updateEvent(event);
  let prevEventID = event.id;
  function updateEvent(event) {
    if (event.id == prevEventID) {
      return;
    }
    prevEventID = event.id;

    newCalendar = event.calendar;
  };
```

benbucksch:
I cannot reproduce the original bug, on master (without these changes), with the reproduction steps above.

* Do you use full screen event edit, or sidebar?
* Do you change the calendar of a new event (before ever saving it), or of an existing event?

app/frontend/Calendar/EditEvent/DialogHeader.svelte
benbucksch: done

PR #752, opened 2025-07-15
Exchange: Only enforce organiser for meetings #747

### LINKED ISSUE #747: Calendar: OWA: Sets an organizer even when you don't invite
### Reproduction

1. Setup an OWA calendar
2. Click to create an event
3. Add a title but don't invite anyone
4. Save
5. Wait 15 seconds for it be saved on the server
6. Click refresh
7. Open the event again

### Actual Result

- The server adds yourself as an organizer

### Expected Result

- Don't add any organizer if `Invite` was not clicked

NeilRashbrook:
Fixes #747 - Exchange adds an organizer even to single events without participants.

Also fixed the attendee email comparisons which after #738 is now wrong for all Exchange protocols.

### DIFF
--- a/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
@@ -72,7 +72,7 @@ export class ActiveSyncEvent extends Event {
     this.location = sanitize.nonemptystring(wbxmljs.Location, "");
     this.isCancelled = (sanitize.integer(wbxmljs.MeetingStatus, 0) & 4) !== 0;
     let attendees = ensureArray(wbxmljs.Attendees?.Attendee);
-    if (wbxmljs.OrganizerEmail && attendees) {
+    if (wbxmljs.OrganizerEmail && attendees.length) {
       for (let attendee of attendees) {
         attendee.Email = sanitize.emailAddress(attendee.Email);
       }


app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -72,7 +72,10 @@ export class ActiveSyncEvent extends Event {
     this.location = sanitize.nonemptystring(wbxmljs.Location, "");
     this.isCancelled = (sanitize.integer(wbxmljs.MeetingStatus, 0) & 4) !== 0;
     let attendees = ensureArray(wbxmljs.Attendees?.Attendee);
-    if (wbxmljs.OrganizerEmail) {
+    if (wbxmljs.OrganizerEmail && attendees) {
```
benbucksch: `&& attendees` is always going to be `true`, because you just ensured that it exists, on the line before. If you need the check, then say `&& attendees.length`.

benbucksch:
```suggestion
    if (wbxmljs.OrganizerEmail && attendees.length) {
```

benbucksch (changes requested):

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
NeilRashbrook: Oops, thanks! (I could also have written `wbxmljs.Attendees?.Attendee` like I did for the other classes; I've no idea why I didn't here, but your way is probably better anyway.)

PR #755, opened 2025-07-15
OWA: Handle specific error from declining an invitation #748

NeilRashbrook:


app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -378,15 +378,13 @@ export class OWAEvent extends Event {
...
     } catch (ex) {
-      // TODO catch only a specific error, throw all others
-      console.error(ex);
-    }
-    if (events.isEmpty) { // OWA deleted the event from the server
-      await this.deleteLocally();
+      if (ex.type != "ErrorItemNotFound") {
+        throw ex;
+      }
+      await this.deleteLocally(); // OWA deleted the event from the server
```
benbucksch:
```suggestion
      if (ex.type == "ErrorItemNotFound") { // expected
        await this.deleteLocally(); // OWA deleted the event from the server
      } else {
        throw ex;
      }
```

benbucksch (changes requested):

PR #759, opened 2025-07-15
New recurrence rule APIs for finding nearby occurrences #750

NeilRashbrook:
The tests uncovered a bug in `fillOccurrences` so I tweaked it and made it easier to use. At that point, it made creating all of the variations trivial, as `fillOccurrences` now guarantees to fill to the next occurrence if there is one. (Note that this breaks `JSONEvent` even more than it is already.)

### DIFF
--- a/app/logic/Calendar/RecurrenceRule.ts
+++ b/app/logic/Calendar/RecurrenceRule.ts
@@ -235,13 +235,28 @@ export class RecurrenceRule implements Readonly<RecurrenceInit> {
     return this.occurrences[count - 1] != null && this.occurrences[count] == null;
   }
 
-  getOccurrencesByDate(seriesEndTime: Date, seriesStartTime: Date = this.seriesStartTime): Date[] {
-    if (this.seriesEndTime && this.seriesEndTime < seriesEndTime) {
-      seriesEndTime = this.seriesEndTime;
-    }
-    if (this.occurrences.length < this.count && this.occurrences.at(-1)! < seriesEndTime) {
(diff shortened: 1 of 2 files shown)


benbucksch:
> Note that this breaks JSONEvent even more than it is already.

Please explain in detail.

app/logic/Calendar/RecurrenceRule.ts
```diff
@@ -235,13 +235,28 @@ export class RecurrenceRule implements Readonly<RecurrenceInit> {
...
+
+  getOccurrenceNotBefore(date: Date): Date | undefined {
+    this.fillOccurrences(this.count, date);
+    return this.occurrences.find(occurrence => occurrence >= date);
+  }
+
+  getOccurrencesByDate(seriesEndTime: Date/*, seriesStartTime: Date = this.seriesStartTime*/): Date[] {
+    this.fillOccurrences(this.count, seriesEndTime);
     return this.occurrences;
     //return this.occurrences.filter(date => date >= seriesStartTime && date <= seriesEndTime);
```
benbucksch:
```suggestion
```

Might as well remove that bit by now

NeilRashbrook: Should have fixed line 258 while you were there.

NeilRashbrook:
> > Note that this breaks JSONEvent even more than it is already.
> 
> Please explain in detail.

JSONEvent was never updated from the old "instances actually also contains exceptions and exclusions" model. So the further we drift away from that model, the more broken it becomes.

Issue #754, opened 2025-07-15
Invitation: Incoming: Deleted events come back after restart


benbucksch:
### Reproduction

1. IMAP account
2. Receive invitation
3. Accept from Inbox
4. Delete from calendar
5. Restart app

### Actual result

- Meeting reappears

### Reproduction

1. ActiveSync account
2. Receive incoming invitation by email
3. Click [Accept]

### Actual result

- Error "To save an event, it needs to be in a calendar first"
- Stack:
  - Event.saveLocally
  - Event.respondToInvitation
  - ActiveSyncIncomingInvitation.respondToInvitation
  - respond
  - onAccept

_Originally posted by @NeilRashbrook in https://github.com/mustang-im/mustang/issues/717#issuecomment-3070159676_

benbucksch:
@NeilRashbrook Can you please send a PR to fix it? Remove the save() calls that shouldn't exist, and add a code comment why we are not saving.

NeilRashbrook:
Note that the ActiveSync case was fixed by #732.

NeilRashbrook:
Actually, no it wasn't, since a `save` was since added to `respondToInvitationFromCalEvent`...

benbucksch:
@NeilRashbrook @jermy-c Can you please validate that this is fixed?

@NeilRashbrook Can you please review 09da68d?

benbucksch:
Thanks for finding this.

I would like to keep the API as-is, as it was created by 09da. Can you please find a fix that doesn't undo it, and instead fixes specifically those 2 issues that you found?

NeilRashbrook:
That was my original approach, but it failed, because there was no way of fixing case 2.

PR #761, opened 2025-07-15
Allow invitation responses on deleted and pseudo events #754

NeilRashbrook:


benbucksch:
* I didn't like the names here
* so I kept your original idea where `respondToInvitation()` does *not* save, and the caller does it instead.
* I've added a code comment in the places where the save was, to ensure we don't regress this.
* I've changed the API names to be a little clearer. It was very very confusing for me that we had a `respondToInvitation()` both on `IncomingInvitation` and on `Event`, so I renamed the former to `respondToInvitationFromMail()`.

Landed in git commit [09da68d](https://github.com/mustang-im/mustang/commit/09da68de82bca8d6daa63fdb849c8ff2e1b44f7f)

PR #763, opened 2025-07-16
Calendar: Edit: Recurrence: Move changeRemainder() from UI to logic - try 2

benbucksch:


app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -235,19 +235,12 @@
...
-    master.copyEditableFieldsFrom(event);
-    master.calUID = null;
-    if (event.parentEvent.recurrenceRule) {
-      let { frequency, interval, week, weekdays } = event.parentEvent.recurrenceRule;
-      master.newRecurrenceRule(frequency, interval, week, weekdays);
-    }
-    await saveEvent(master);
-    master.finishEditing();
+    let newMaster = event.cloneSeriesFromThisInstance();
+    await newMaster.save();
```
NeilRashbrook, line 239: `saveEvent` used to add the event to the calendar.
```suggestion
    let newMaster = event.cloneSeriesFromThisInstance();
    await newMaster.save();
    newMaster.calendar.events.add(newMaster);
```

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -279,9 +272,10 @@
   }
 
   async function onDeleteRemainder() {
+    let end = new Date(event.startTime);
     event.parentEvent.cancelEditing();
     event.cancelEditing();
-    await event.truncateRecurrence();
+    await event.setRecurrenceEndTime(end);
```
NeilRashbrook, line 278: `event` isn't the master here, and you probably want to delete the current instance too:
```suggestion
    event.parentEvent.cancelEditing();
    event.cancelEditing();
    await event.truncateRecurrencesFromHere();
```

app/logic/Calendar/Event.ts
```diff
@@ -729,6 +729,7 @@ export class Event extends Observable {
    * Used when a master recurrence rule is removed or changed incompatibly.
    */
   protected clearExceptions() {
+    assert(this.recurrenceCase == RecurrenceCase.Master, "Must be a recurring master");
```
NeilRashbrook, line 732: `this.recurrenceCase` can't be relied upon at this point, because we might have just cleared the recurrence rule.

app/frontend/Calendar/DisplayEvent/DialogHeader.svelte
```diff
@@ -132,7 +132,8 @@
   async function onDeleteRemainder() {
     event.parentEvent.cancelEditing();
     event.cancelEditing();
-    await event.truncateRecurrence(event.startTime);
+    event.truncateRecurrencesFromHere();
+    await event.save();
```
NeilRashbrook, line 136: You don't want to save the event here as that tries to turn it from an instance into an exception. (The old `truncateRecurrence` had `await master.save()`, which is why it was `async`.)
```suggestion
    event.truncateRecurrencesFromHere();
    await event.parentEvent.save();
```

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -235,19 +235,12 @@
...
-    }
-    await saveEvent(master);
-    master.finishEditing();
+    let newMaster = event.cloneSeriesFromThisInstance();
+    await newMaster.save();
     event.parentEvent.cancelEditing();
     event.cancelEditing();
-    await event.truncateRecurrence();
+    event.truncateRecurrencesFromHere();
+    await event.save();
```
NeilRashbrook, line 243: Ditto.
```suggestion
    event.truncateRecurrencesFromHere();
    await event.parentEvent.save();
```

app/logic/Calendar/Event.ts
```diff
@@ -772,33 +773,36 @@ export class Event extends Observable {
...
-    }
-    await master.save();
+    let index = master.recurrenceRule.getIndexOfOccurrence(this.recurrenceStartTime);
+    master.setRecurrenceCount(index);
+  }
+
+  /** Creates a new series with the same properties and recurrence,
+   * but starting at this instance.
+   *
+   * It will also truncate the old series at that start time.
```
NeilRashbrook, line 789: Not any more it doesn't.

app/logic/Calendar/Event.ts
```diff
@@ -772,33 +773,36 @@ export class Event extends Observable {
...
+  cloneSeriesFromThisInstance(): Event {
+    assert((this.recurrenceCase == RecurrenceCase.Instance || this.recurrenceCase == RecurrenceCase.Exception) &&
+      this.parentEvent?.recurrenceRule, "Must be a recurring instance");
+    let oldMaster = this.parentEvent;
+    let newMaster = this.calendar.newEvent();
+    newMaster.startEditing();
+    newMaster.copyEditableFieldsFrom(this);
+    newMaster.calUID = null;
+    let { frequency, interval, week, weekdays } = oldMaster.recurrenceRule;
+    newMaster.newRecurrenceRule(frequency, interval, week, weekdays);
```
NeilRashbrook, line 803: `oldMaster.recurrenceRule` could be null if you changed the frequency in the UI first. (This would replace the remainder of the series with a single event.)

NeilRashbrook (changes requested):
> When "changing remainder", it deletes all instances from here on, _including_ this one.

Isn't that the desired behaviour? The new series starts from this instance. (Well, it does if you remember to add it to the calendar...)

benbucksch:
>> When "changing remainder", it deletes all instances from here on, including this one.

> Isn't that the desired behaviour? The new series starts from this instance.

Problem is that the new series is not created

> (Well, it does if you remember to add it to the calendar...)

Oh, duh! Is that the reason why I work on this since 2 days and nothing works? *sigh*

Issue #718, opened 2025-07-06
Invitation: Incoming: Make isCancelled explicit


benbucksch:
#712 is not working reliably, because there may be other reasons why the Organizer is not in the participants list.

`Event.isCancelled` should be an explicitly flag, and saved to disk.

benbucksch:
For iCal, we need to set the new explicit property when we get a cancellation message.

NeilRashbrook:
I can't actually find an iCal property for a cancelled event. (When you cancel an event, you send out an iTip message with a METHOD of CANCEL, but that's not actually part of the VEVENT itself.)

benbucksch:
> iTip message with a METHOD of CANCEL

Good enough. We should register that in the calendar event as `isCancelled = true` and save it to DB when we process the CANCEL message.

benbucksch:
I've implemented this in git commit efc35a91 , or at least I thought I did, but cancelled events do not show up as cancelled for me in the calendar.

### Reproduction
* Send an invitation from Neil EWS to Ben OWA, and vise-versa, with recurring events
* Accept both invitations, at the receiver side
* At the sender, delete the event
* At the receiver, check mail, and refresh the calendar

### Actual result
* In calendar, you see title "Canceled: <old title>"
* In calendar Month View and Week View, you see the event with full opacity

### Actual result
* In calendar Month View and Week View, you see the event with 30% opacity, and the title and time are strike-through.

benbucksch:
@NeilRashbrook Could you please fix the above?

benbucksch:
@NeilRashbrook wrote in #738 
"
(isCancelled will be incorrect for any existing cancelled invitations, except for OWA, which will always resync, or if you change the alarm on a cancelled EWS or ActiveSync invitation (Exchange allows this!) which will trigger a resync of that item.)

For new CalDAV invitations, isCancelled will appear to be correct, but if you sync from another device, isCancelled will always be false, although you could add code to the CalDAV event type to synthesise isCancelled from the lack of an organiser.
"

Also,
```
      let organizer = attendees.find(attendee => attendee.Email?.toLowerCase() == organizerEmail);
```

PR #768, opened 2025-07-16
CalDAV: Support STATUS:CANCELLED #718

NeilRashbrook:


benbucksch:
Where in the spec do you see that these events are cancelled?

Note that we should already be processing CANCELLED invitations and record that.

What I found in the RFCs:

From RFC 6638, Section 3.1:

    A calendar object resource is considered to be a valid organizer scheduling object resource if it contains at least one ATTENDEE property and exactly one ORGANIZER property

RFC 5545, Section 3.8.4.3 (ORGANIZER property) states:

    "The property MUST be specified in an iCalendar object that specifies a group scheduled calendar entity."
    This means any event with ATTENDEE lines (invites) must have exactly one ORGANIZER

.

benbucksch:
Nice, thanks!

@NeilRashbrook Is there any place where we an event with attendees, but no organizer? (As this would be malformed per spec.)

NeilRashbrook:
I think there's one major place, but maybe some other places too.

benbucksch:
> I think there's one major place, but maybe some other places too.

Can you fix them, please?

PR #772, opened 2025-07-16
Try harder to detect outgoing invitations

NeilRashbrook:
if the event isn't in your calendar, look at the participants to see whether you are the organiser.

Or should I look at the organiser to see whether that's you? Well, we already have code that does the first approach; I just had to move it somewhere where I could call it.


benbucksch:
Filed bug #773 to track the bug that you're trying to fix

benbucksch (approved):

Issue #771, opened 2025-07-16
EWS, OWA: Refresh calendars and addressbooks


benbucksch:
We're checking for new mail anyways. When we get change notifications, update the calendar and address books as well.

PR #778, opened 2025-07-17
Calendar: Separate CalendarAccount from Calendar

benbucksch:
* Instead of `Calendar` being an `Account`, separate them into 2 objects.
* A `CalendarAccount` is created only when it has its own login, e.g. CalDAV.
* Separate them also in the DB. Ditto, don't save calendars in accounts.db.
* CalDAV: Allow to have several calendars per CalDAV account


app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
```diff
@@ -15,18 +15,15 @@ const kHalfHour = 30 * 60 * 1000; // milliseconds
...
   readonly folderClass = "Calendar";
   protected readonly requestLock = new Lock();
 
-  get account(): ActiveSyncAccount {
-    return this.mainAccount as ActiveSyncAccount;
-  }
-
   get serverID() {
-    return new URL(this.url).searchParams.get("serverID");
+    return new URL(this.account.url).searchParams.get("serverID");
```
NeilRashbrook: This change is wrong; each ActiveSyncCalendar has its own serverID. You'll need to port 77ffe7f and update all of the places where the url gets used to use the new serverID instead. (Or I could try to do that separately after I've finished #771.)

app/logic/Calendar/CalDAV/CalDAVCalendar.ts
```diff
@@ -1,100 +1,48 @@
...
-import { NotReached, assert, type URLString } from "../../util/util";
-import { gt } from "../../../l10n/l10n";
-import { ArrayColl, Collection } from "svelte-collections";
-import type { DAVClient, DAVCalendar, DAVObject } from "tsdav";
+import type { URLString } from "../../util/util";
+import type { ArrayColl } from "svelte-collections";
+import type { DAVCalendar, DAVObject } from "tsdav";
 
 export class CalDAVCalendar extends Calendar {
   readonly protocol: string = "caldav";
```
NeilRashbrook, line 12: Surely only the account needs a protocol?

NeilRashbrook: Oh, this is because you still need to store each calendar in a database. Ignore me.

app/logic/Mail/ActiveSync/ActiveSyncError.ts
```diff
@@ -1,10 +1,12 @@
 import type { Account } from "../../Abstract/Account";
+import type { Calendar } from "../../Calendar/Calendar";
+import type { Addressbook } from "../../Contacts/Addressbook";
 import { gt } from "../../../l10n/l10n";
 
 export class ActiveSyncError extends Error {
   type: string;
   code: string;
-  constructor(aCommand: string, aStatus: string, account: Account) {
+  constructor(aCommand: string, aStatus: string, account: Account | Calendar | Addressbook) {
```
NeilRashbrook, line 9: Line 231 of `ActiveSyncEvent.ts` passes `this.calendar.account` but you probably want to change that to `this.calendar`.

benbucksch: Done

app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
benbucksch: I've merged 77ffe7f in.
But I don't understand the relation between `folderID`, `serverID` and URL. Could you make the necessary change, please? Same in `ActiveSyncAddressbook`.

benbucksch: Fixed in #808

Issue #309, opened 2024-12-12
Setup: Mail: Manual: Password in manual setup not used


benbucksch:
### Reproduction
1. Setup mail
2. Enter correct email address
3. Enter wrong password
4. Let autoconfig run
5. Go to Manual Config
6. Change the password to the correct one
7. Continue to verification

### Actual result
* Login fail

### Expected result
* Login succeeds

castaway:
Is this one difficult to fix? Its preventing me trying Parula out with all my accounts

benbucksch:
@castaway What's your scenario why it prevents you from setting up the account? It should be sufficient to enter the correct password from the beginning.

castaway:
@benbucksch maybe  I read this too quickly?  I'm trying to setup a manual account, and it keeps failing, I need different passwords for IMAP and for SMTP, so I enter one initially, then have to change it on the SMTP account afterwards.

Then it says this: btw I can't tell which server the error is about, is there any way to tell?

![Image](https://github.com/user-attachments/assets/b3bfc4e9-92a5-4055-821e-94398d8dc2a1)

benbucksch:
Ah, I see. If you need different passwords for IMAP and SMTP, then indeed, you may have run into this bug.

You're right that the error message is unclear whether it's about IMAP and SMTP. In your case, it's about SMTP, because "573 5.7.0" are SMTP error codes, but no normal user can possibly know that, so we need to make this clearer.

Thanks for your report. I will look into this soon, to help you out.

castaway:
@benbucksch any progress?

benbucksch:
Fix in #787

@jermy-c  Can you please validate the fix?

jermy-c:
> @jermy-c Can you please validate the fix?

Yes, it's fixed.

benbucksch:
@jermy-c Thanks for finding the cause!

PR #779, opened 2025-07-17
MailSetup: Assign emailAddress and password from Manual Config #309

jermy-c:
- In SetupMail we were using `emailAddress` and `password` instead of `config.emailAddress` and `config.password` and `<CheckConfig />` uses `emailAddress` and `password`

https://github.com/mustang-im/mustang/blob/ba202c17bee1c70ac0c3dd4f5835416aea9b14f8/app/frontend/Setup/Mail/SetupMail.svelte#L22

- So I just set it to that `emailAddress` to `config.emailAddress` since everywhere else seems to be using that.

### DIFF
--- a/app/frontend/Setup/Mail/SetupMail.svelte
+++ b/app/frontend/Setup/Mail/SetupMail.svelte
@@ -182,7 +182,6 @@
       if (!await manualConfigEl.onContinue()) {
         return;
       }
-      fillConfig(config, emailAddress, password);
       errorMessage = null;
       step = Step.CheckConfig;
     } else if (step == Step.CheckConfig) {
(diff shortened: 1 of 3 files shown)


app/frontend/Setup/Mail/SetupMail.svelte
```diff
@@ -182,7 +182,8 @@
       if (!await manualConfigEl.onContinue()) {
         return;
       }
-      fillConfig(config, emailAddress, password);
```
jermy-c, line 185: I removed this because the step after this is `checkConfig` which takes the `emailAddress` and `password` and does `fillConfig()` and checks it.

jermy-c: Don't remove this because `fillConfig()` might still be needed for more complex configs.

jermy-c:
This doesn't work, if the incoming and outgoing configs have different email address or password. Because `fillConfig()` fills both and also the UI only uses one email address and password variables.

PR #781, opened 2025-07-17
Prevent slow functions from being reentered

NeilRashbrook:
In particular, `listContacts` and `listEvents` are relatively slow on OWA, so we don't want to accidentally run them when they're already running.


benbucksch:
Can we make the sync functions faster? E.g. getting only the latest changes?

NeilRashbrook:
I don't know of a reliable way of doing it, since OWA has no synchronisation primitives.

PR #785, opened 2025-07-17
Improve support of invitations of single occurrences

NeilRashbrook:
In particular, this means that if an attendee changes their attendance of a single instance of a recurring meeting to which they were invited, this will now correctly update a local calendar.


benbucksch:
What do you mean with "single occurrences" in this context?

benbucksch (approved):
Ah, I see now. You're finding the corresponding instance in the calendar that matches the recurrence instance that the invitation is about.

Nice.

PR #786, opened 2025-07-17
EWS, OWA, ActiveSync: Accept an instance of a recurring meeting

NeilRashbrook:
This used to fail because instances are locally generated and so don't have a server or item id.

I'm going to have to look up how to support this in ActiveSync 14.0 as that doesn't support the instance id.

When I tested, I couldn't get the user's calendar to show the new status, although Exchange had sent the email and the organiser's calendar showed their status, but maybe that was just me.

### DIFF
--- a/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
@@ -247,10 +247,16 @@ export class ActiveSyncEvent extends Event {
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
     assert(this.isIncomingMeeting, "Only invitations can be responded to");
     let request = {
-      Request: {
+      // TODO support ActiveSync 14.0
+      Request: this.serverID ? {
         UserResponse: ActiveSyncResponse[response],
         CollectionId: this.calendar.serverID,
         RequestId: this.serverID,
+      } : {
+        UserResponse: ActiveSyncResponse[response],
+        CollectionId: this.calendar.serverID,
+        RequestId: this.parentEvent.serverID,
+        InstanceId: this.recurrenceStartTime.toISOString(),
       },
     };
     await this.calendar.account.callEAS("MeetingResponse", request);
--- a/app/logic/Calendar/EWS/EWSEvent.ts
+++ b/app/logic/Calendar/EWS/EWSEvent.ts
@@ -400,8 +400,28 @@ export class EWSEvent extends Event {
 
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
     assert(this.isIncomingMeeting, "Only invitations can be responded to");
+    let itemID = this.itemID;
+    // Unfortunately this API won't take an OccurrenceItemId directly.
+    if (!itemID) {
+      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance");
+      let request = {
+        m$GetItem: {
+          m$ItemShape: {
+            t$BaseShape: "IdOnly",
+          },
+          m$ItemIds: {
+            t$OccurrenceItemId: {
+              RecurringMasterId: this.parentEvent.itemID,
+              InstanceIndex: this.parentEvent.recurrenceRule.getIndexOfOccurrence(this.recurrenceStartTime) + 1,
+            },
+          },
+        },
+      };
+      let response = await this.calendar.account.callEWS(request);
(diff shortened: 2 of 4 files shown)


NeilRashbrook:
(I noticed that the email sent by Exchange has both a RECURRENCE-iD and an RRULE; I'm not sure whether the combination will confuse any of our iCal code.)

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -247,10 +247,16 @@ export class ActiveSyncEvent extends Event {
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
     assert(this.isIncomingMeeting, "Only invitations can be responded to");
     let request = {
-      Request: {
+      // TODO support ActiveSync 14.0
```
benbucksch, line 250: 1. What *are* we supporting?
2. What are we currently doing for 14.0?

app/logic/Calendar/EWS/EWSEvent.ts
```diff
@@ -400,8 +400,28 @@ export class EWSEvent extends Event {
 
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
     assert(this.isIncomingMeeting, "Only invitations can be responded to");
+    let itemID = this.itemID;
+    // Unfortunately this API won't take an OccurrenceItemId directly.
+    if (!itemID) {
+      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance");
```
benbucksch:
```suggestion
      // In case the invitation is for a single instance of a recurring meeting
      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance");
```

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -372,14 +372,22 @@ export class OWAEvent extends Event {
 
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
     assert(this.isIncomingMeeting, "Only invitations can be responded to");
+    let itemID = this.itemID;
+    // Unfortunately this API won't take an OccurrenceItemId directly.
+    if (!itemID) {
+      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance");
```
benbucksch:
```suggestion
      // In case the invitation is for a single instance of a recurring meeting
      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance");
```

benbucksch (changes requested):

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -372,14 +372,23 @@ export class OWAEvent extends Event {
 
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
     assert(this.isIncomingMeeting, "Only invitations can be responded to");
+    let itemID = this.itemID;
+    // Unfortunately this API won't take an OccurrenceItemId directly.
+    if (!itemID) {
+      // In case the invitation is for a single instance of a recurring meeting
+      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance");
```
benbucksch:
```suggestion
      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance, or have an itemID");
```

app/logic/Calendar/EWS/EWSEvent.ts
```diff
@@ -400,8 +400,29 @@ export class EWSEvent extends Event {
 
   async respondToInvitation(response: InvitationResponseInMessage): Promise<void> {
     assert(this.isIncomingMeeting, "Only invitations can be responded to");
+    let itemID = this.itemID;
+    // Unfortunately this API won't take an OccurrenceItemId directly.
+    if (!itemID) {
+      // In case the invitation is for a single instance of a recurring meeting
+      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance");
```
benbucksch:
```suggestion
      assert(this.recurrenceCase == RecurrenceCase.Instance, "must be an instance, or have an itemID");
```

benbucksch (approved):

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
NeilRashbrook: 1. 14.1
2. it still doesn't work.

PR #787, opened 2025-07-18
Setup: Mail: Manual: Avoid to overwrite the password #309

benbucksch:
The user has manually configured this, so don't overwrite the values that the user has set.


benbucksch:
(Works for me)

Issue #777, opened 2025-07-17
Settings: Mail: Password deleted when opening "Server" settings for any IMAP account


benbucksch:
### Reproduction
0. For easier debugging, in `PasswordChange.svelte`, after `<input />`, add `{password}`. Or, alternatively, check out branch `ben/password-change-2`
1. Set up multiple email accounts. Some IMAP accounts, some EWS accounts.
2. Restart app
3. Check mail
4. -> Works
5. Go to Settings | Mail | Some EWS account | Server
6. Don't change anything, don't save, only look
7. Check mail
8. Go to Settings | Mail | Some IMAP account | Server
9. Don't change anything, don't save, only look
10. Check mail
11. Go to Settings | Mail | Some EWS account | Server
12. Check mail

### Actual result
* Step 7: You see the password. All accounts are logged in. Check mail works.
* Step 10: Password is empty. Some accounts are logged out. Check mail fails with "password wrong".
* Step 12: Password is empty. Some accounts are logged out. Check mail fails with "password wrong".
* It appears that all passwords get deleted
* It seems to happen only if you go to "Server" tab of an IMAP account.
* It does not seem to happen if you go to "Server" tab of EWS accounts.

### Expected result
* Password stays, unless I go into the password field and type

### Cause
* I thought that the `<PasswordChanged>` component is the problem, but branch `ben/password-change-2` shows that this component is not at fault.

benbucksch:
Might be same cause as #309

jermy-c:
Removing the code from the above comment https://github.com/mustang-im/mustang/issues/777#issuecomment-3084710029 seems to fix it but we don't change `config.password` to `""` or `null` so I'm wondering what else could be setting `config.password`.

jermy-c:
Sorry, I think I wasn't reproducing the bug properly. I think you have to do the steps as soon as you start the app also.

jermy-c:
Just a note: Not sure if adding `$` to  `mailAccount` would help here.

https://github.com/mustang-im/mustang/blob/4a8e88517c1fa93b3597a3d467ef5b7b45ce37fe/app/frontend/Settings/Mail/Account/Server.svelte#L5

PR #789, opened 2025-07-18
Settings: Debounce selectedCategory #777

jermy-c:
- When we mount components with `<svelte:components>` there seems to be some issues when the component passed is changed too quick because the component doesn't complete it's life cycle and may set any binded variables to `undefined`.
- Also there seems to be some in-between updates that aren't necessary so could skip those.
- I've debounced the `$selectedCategory` to fix _Settings: Mail: Password deleted when opening "Server" settings for any IMAP account_ and so far in my testing it seems to work
- We have to explicitly set the `delay` to `1ms` because it defaults to `100ms` which is too long and feels slow. `1ms` was enough for me. I think we just need to skip that in-between update.

### DIFF
--- a/app/frontend/Settings/Window/SettingsApp.svelte
+++ b/app/frontend/Settings/Window/SettingsApp.svelte
@@ -14,13 +14,14 @@
       </hbox>
     </hbox>
     <vbox flex class="right-page">
-      <MainContent category={$selectedCategory} />
+      <MainContent category={showCategory} />
     </vbox>
   </Scroll>
 </Splitter>
 
 <script lang="ts">
   import { settingsCategories } from "../SettingsCategories";
+  import type { SettingsCategory } from "./SettingsCategory";
   import { globalSearchTerm, openApp } from "../../AppsBar/selectedApp";
   import { selectedCategory } from "./selected";
   import { mailMustangApp } from "../../Mail/MailMustangApp";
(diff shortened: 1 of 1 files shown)


benbucksch:
Can you please explain more what you mean with "in-between updates"? What exactly happens?

jermy-c:
> Can you please explain more what you mean with "in-between updates" and "changed too quick because the component doesn't complete"? What **exactly** happens?

**In-between updates**

I mean that the binded values are updated on Load when the settingsCategories class loads the component and when the component mounts.

So you get the following updates:

1. Load component <- Updates here with all values being `undefined`
2. Mount component <- updates here again but values are defined
3. Destroy component <- Updates here with all values being `undefined`
4. Repeats 1-3 again before the component is displayed, this is the _**in-between update**_

**changed too quick because the component doesn't complete**

Sorry, I missed doesn't complete the entire component lifecycle. So, I mean that the child components don't complete the entire component lifecycle.

So when you click Account | Server it updates 2 times quickly with no delay one for Account and one for Server and the component lifecycle doesn't complete but the values are still binded which sets them to `undefined`.

So the component would look like:

1. Load component
2. Mount Component <- values are binded for child comonents
3. Destroy Component <- child components haven't unbind values yet
4. Load component <- child components still have binded values which are `undefined` now
5. Mounts component <- child and parent component values are binded

benbucksch:
Thanks for the details, Jeremy.

I've moved the code to `<MainContent>`, otherwise kept it more or less as-is.
FIXED in git commit d94a0efb.

Issue #645, opened 2025-05-30
ActiveSync: Calendar: Malformed request


jermy-c:
### Reproduction

1. Create event on an ActiveSync v14.0 Calendar
2. Add an recurrence
3. Save event
4. Edit event
5. Save edited event

### Actual Result

1. Yellow banner with `Calendar: Malformed request`
2. Edits not saved

### Expected Result

1. Edit saved remotely and locally
2. No yellow banner

### Stack

```
error.ts:6 ActiveSyncError: Calendar: Malformed request
    at ActiveSyncEvent.saveFields (ActiveSyncEvent.ts:164:17)
    at async ActiveSyncEvent.saveToServer (ActiveSyncEvent.ts:130:7)
    at async onChangeAll (DialogHeader.svelte:227:40)
    at async HTMLButtonElement.myOnClick (RoundButton.svelte:63:19)
```

jermy-c:
Also getting the error `this.parentEvent.toFields is not a function` but it seems like to more likely to happen when moving the local event to a ActiveSync calendar.

jermy-c:
It seems to because of the recurrence. If I don't add any repeats then there's no issues.

benbucksch:
The problem is that the `parentEvent` should be an `ActiveSyncEvent`, but it's not. It may be that the moving from personal calendar to ActiveSync account doesn't re-create the event propertly. See `Event.moveToCalendar()`. Maybe it should change `parentEvent` as well.

jermy-c:
I tried to do `newEvent.parentEvent ??= await newEvent.parentEvent?.moveToCalendar(newCalendar);` in `moveToCalendar()` but it's not working. But that might not be correct place to do it.

> The problem is that the parentEvent should be an ActiveSyncEvent, but it's not. It may be that the moving from personal calendar to ActiveSync account doesn't re-create the event propertly. See Event.moveToCalendar(). Maybe it should change parentEvent as well.

I do see it pointing back to the master event and it only happens when editing from a child event and then click on `changeAll`.

jermy-c:
I've just updated the reproduction, to reproduce create an recurring event in an ActiveSync v14.0 calendar and save.

PR #800, opened 2025-07-22
ActiveSync v14.0: Don't include FirstDayOfWeek field #645

jermy-c:
- On ActiveSync v14.0 the `FirstDayOfWeek` field is not supported, therefore, it causes an `Malformed` request when creating a recurring event
- see https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-ascal/438e1602-1042-4d6d-bed9-77e2b0855da9

### DIFF
--- a/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
@@ -127,7 +127,7 @@ export class ActiveSyncEvent extends Event {
         MonthOfYear: this.recurrenceRule.frequency == Frequency.Yearly ? String(this.recurrenceRule.seriesStartTime.getMonth() + 1) : [],
         Until: toCompact(this.recurrenceRule.seriesEndTime) || [],
         DayOfMonth: [Frequency.Monthly, Frequency.Yearly].includes(this.recurrenceRule.frequency) && !this.recurrenceRule.week ? String(this.recurrenceRule.seriesStartTime.getDate()) : [],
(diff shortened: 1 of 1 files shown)


app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -127,7 +127,7 @@ export class ActiveSyncEvent extends Event {
         MonthOfYear: this.recurrenceRule.frequency == Frequency.Yearly ? String(this.recurrenceRule.seriesStartTime.getMonth() + 1) : [],
         Until: toCompact(this.recurrenceRule.seriesEndTime) || [],
         DayOfMonth: [Frequency.Monthly, Frequency.Yearly].includes(this.recurrenceRule.frequency) && !this.recurrenceRule.week ? String(this.recurrenceRule.seriesStartTime.getDate()) : [],
-        FirstDayOfWeek: this.recurrenceRule.frequency == Frequency.Weekly ? String(this.recurrenceRule.first) : [],
+        FirstDayOfWeek: this.recurrenceRule.frequency == Frequency.Weekly && this.calendar.account.protocolVersion != "14.0" ? String(this.recurrenceRule.first) : [],
```
NeilRashbrook: This is logically correct but I don't like the condition ordering; I would prefer the version test to be first, or even to be a separate ternary, like this:
```suggestion
        FirstDayOfWeek: this.calendar.account.protocolVersion == "14.0" ? [] : this.recurrenceRule.frequency == Frequency.Weekly ? String(this.recurrenceRule.first) : [],
```

NeilRashbrook (approved):

benbucksch (approved):

PR #801, opened 2025-07-28
Don't send invitations/cancellations to empty participant list

NeilRashbrook:
I'm not exactly sure how I triggered this but I was editing a meeting and I got an error saying that my invitation message had no recipient.

Since i was there I thought there was no point sending a cancellation to a declined attendee.

(Strictly speaking this is a regression from eef7737 as the previous code sent separate messages for each participants therefore guaranteeing that if there were no relevant participants then no messages were sent.)

### DIFF
--- a/app/logic/Calendar/Invitation/OutgoingInvitation.ts
+++ b/app/logic/Calendar/Invitation/OutgoingInvitation.ts
@@ -31,15 +31,19 @@ export class OutgoingInvitation {
 
   async sendInvitationsTo(participants: Collection<Participant>) {
     let sendTo = participants.filterOnce(participant => participant.response != InvitationResponse.Organizer);
-    for (let participant of sendTo) {
-      participant.response ||= InvitationResponse.NoResponseReceived;
+    if (sendTo.hasItems) {
+      for (let participant of sendTo) {
+        participant.response ||= InvitationResponse.NoResponseReceived;
+      }
+      await this.send("REQUEST", sendTo);
     }
-    await this.send("REQUEST", sendTo);
   }
 
   async sendCancellationsTo(participants: Collection<Participant>) {
-    let sendTo = participants.filterOnce(participant => participant.response > InvitationResponse.Organizer);
-    await this.send("CANCEL", sendTo);
+    let sendTo = participants.filterOnce(participant => [InvitationResponse.Accept, InvitationResponse.Decline, InvitationResponse.NoResponseReceived].includes(participant.response));
+    if (sendTo.hasItems) {
+      await this.send("CANCEL", sendTo);
+    }
   }
 
   async sendInvitations() {
--- a/app/logic/Calendar/Invitation/OutgoingInvitation.ts
+++ b/app/logic/Calendar/Invitation/OutgoingInvitation.ts
@@ -31,12 +31,13 @@ export class OutgoingInvitation {
 
   async sendInvitationsTo(participants: Collection<Participant>) {
     let sendTo = participants.filterOnce(participant => participant.response != InvitationResponse.Organizer);
-    if (sendTo.hasItems) {
-      for (let participant of sendTo) {
-        participant.response ||= InvitationResponse.NoResponseReceived;
-      }
-      await this.send("REQUEST", sendTo);
+    if (sendTo.isEmpty) {
+      return;
     }
+    for (let participant of sendTo) {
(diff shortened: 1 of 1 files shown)


app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -31,15 +31,19 @@ export class OutgoingInvitation {
...
+      }
+      await this.send("REQUEST", sendTo);
     }
-    await this.send("REQUEST", sendTo);
   }
 
   async sendCancellationsTo(participants: Collection<Participant>) {
-    let sendTo = participants.filterOnce(participant => participant.response > InvitationResponse.Organizer);
-    await this.send("CANCEL", sendTo);
+    let sendTo = participants.filterOnce(participant => [InvitationResponse.Accept, InvitationResponse.Decline, InvitationResponse.NoResponseReceived].includes(participant.response));
```
```suggestion
    const participantResponses = [ InvitationResponse.Accept, InvitationResponse.Decline,
      InvitationResponse.Tentative, InvitationResponse.NoResponseReceived];
    let sendTo = participants.filterOnce(participant => participantResponses.includes(participant.response));
```

Add Tentative?

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -31,15 +31,19 @@ export class OutgoingInvitation {
...
+    if (sendTo.hasItems) {
+      await this.send("CANCEL", sendTo);
+    }
```
```suggestion
    if (sendTo.isEmpty) {
      return;
    }
    await this.send("CANCEL", sendTo);
```

Normal cases straight, exceptions as early exit.

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -31,15 +31,19 @@ export class OutgoingInvitation {
...
   async sendInvitationsTo(participants: Collection<Participant>) {
     let sendTo = participants.filterOnce(participant => participant.response != InvitationResponse.Organizer);
-    for (let participant of sendTo) {
-      participant.response ||= InvitationResponse.NoResponseReceived;
+    if (sendTo.hasItems) {
+      for (let participant of sendTo) {
+        participant.response ||= InvitationResponse.NoResponseReceived;
+      }
+      await this.send("REQUEST", sendTo);
     }
```
```suggestion
    if (sendTo.isEmpty) {
      return;
    }
    for (let participant of sendTo) {
      participant.response ||= InvitationResponse.NoResponseReceived;
    }
    await this.send("REQUEST", sendTo);
```

Early exit. See also below.

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -31,15 +31,19 @@ export class OutgoingInvitation {
 
   async sendInvitationsTo(participants: Collection<Participant>) {
     let sendTo = participants.filterOnce(participant => participant.response != InvitationResponse.Organizer);
-    for (let participant of sendTo) {
-      participant.response ||= InvitationResponse.NoResponseReceived;
+    if (sendTo.hasItems) {
+      for (let participant of sendTo) {
+        participant.response ||= InvitationResponse.NoResponseReceived;
```
benbucksch, line 36: Why is the response == 0?

benbucksch (changes requested):

benbucksch:
Added `Participant.isInvitee` and `kInviteeResponses` in git commit eaf6fe3bd

benbucksch:
> there was no point sending a cancellation to a declined attendee.

* You mean an invitee who sent a Decline response to us? Actually, that is interesting information for me to know that a meeting that I cannot join will not happen at all.

* Why you set participants with `InvitationResponse.Unknown` in `sendInvitationsTo()`?
* Why do you ignore them in `sendCancellationsTo()`?
* See also #805

NeilRashbrook:
> > there was no point sending a cancellation to a declined attendee.
> 
>   * You mean an invitee who sent a Decline response to us? Actually, that is interesting information for me to know that a meeting that I cannot join will not happen at all.

Well, that's what I meant... I double checked the list of response types and still used the wrong one... oh well.

>   * Why are there participants with `InvitationResponse.Unknown` at the time when we call `sendInvitationsTo()`?

These are newly added participants.

>   * Why do you ignore them in `sendCancellationsTo()`?

Because we never got around to inviting them. (These days, this should never happen, but I think it might have been a possibility if you added a participant and then changed your mind and deleted the meeting.)

PR #803, opened 2025-07-28
Properly handle acceptance of occurrence of series

NeilRashbrook:
When you invite a participant to a recurring meeting but they have a custom response to a single occurrence, this does not currently get processed properly. There are two problems:
1. The instance turns into an exception, but this means it needs to be added explicitly to its parent calendar's event list.
2. The instance has a shallow copy of the list of participants, which allows it to show the correct master response (if the participant previously replied to the entire series), but we need to deep clone the list before modifying it. (This is not normally a problem because we don't normally modify participants in-place.)

I don't know why `saveLocally` was saving the parent event; creating an exception has no effect on the parent.

### DIFF
--- a/app/logic/Calendar/Invitation/IncomingInvitation.ts
+++ b/app/logic/Calendar/Invitation/IncomingInvitation.ts
@@ -1,5 +1,6 @@
 import type { Calendar } from "../Calendar";
-import type { Event } from "../Event";
+import { RecurrenceCase, type Event } from "../Event";
+import { Participant } from "../Participant";
 import { InvitationMessage, InvitationResponse, type InvitationResponseInMessage } from "../Invitation/InvitationStatus";
 import type { EMail } from "../../Mail/EMail";
 import { AbstractFunction, NotReached, assert } from "../../util/util";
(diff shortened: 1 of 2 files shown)


app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -62,6 +63,9 @@ export class IncomingInvitation {
     assert(this.invitationMessage == InvitationMessage.ParticipantReply, "Not a reply");
     let event = this.calEvent();
     assert(event, "Cannot process invitation update: The event was not found in your calendar");
+    if (event.recurrenceCase == RecurrenceCase.Instance) {
+      event.participants.replaceAll(event.participants.contents.map(participant => new Participant(participant.emailAddress, participant.name, participant.response)));
```
benbucksch:
```suggestion
      event.participants.replaceAll(event.participants.contents.map(p => new Participant(p.emailAddress, p.name, p.response)));
```

benbucksch (approved):

app/logic/Calendar/Invitation/IncomingInvitation.ts
```diff
@@ -62,6 +63,9 @@ export class IncomingInvitation {
     assert(this.invitationMessage == InvitationMessage.ParticipantReply, "Not a reply");
     let event = this.calEvent();
     assert(event, "Cannot process invitation update: The event was not found in your calendar");
+    if (event.recurrenceCase == RecurrenceCase.Instance) {
+      event.participants.replaceAll(event.participants.contents.map(p => new Participant(p.emailAddress, p.name, p.response)));
```
benbucksch:
```suggestion
      event.participants.replaceAll(event.participants.contents.map(p =>
        new Participant(p.emailAddress, p.name, p.response)));
```

PR #811, opened 2025-08-05
Mac: Install update after the last window is closed - fixes #810

### LINKED ISSUE #810: Mac: Auto-updates are not installed
### Reproduction
1. Download an old version of Parula
2. Start Parula
3. Wait 10 seconds
4. Quit Parula

### Actual result
* When running on Mac, I see an OS notification (top right of screen) which says: 'An update to Parula 0.9.2 has been downloaded and will be installed when you exit the application.'
* When I click on it and see older messages, I see that same message about 40 (!) times, the whole right side of the screen is full from the same message.
* The update is **not installed**. Parula stays on version 0.9.1, even after I quit it in the "Parula" application menu on the top right and start it again.

### Expected result
* When I close the last window, and there is an update ready to install, the application quits and installs the update
* When I explicitly manually quit Parula using the application menu on the top right, the update is installed.

benbucksch:
Fixes #810

### DIFF
--- a/app/frontend/Calendar/EditEvent/DateInput.svelte
+++ b/app/frontend/Calendar/EditEvent/DateInput.svelte
@@ -2,12 +2,15 @@
   required
   min={toHTMLString(min)} max={toHTMLString(max)}
   bind:value
-  on:change={onChange}
+  on:change={onChangeDebounced}
+  on:blur={onChange}
   on:change
   {disabled}
   />
 
 <script lang="ts">
+  import debounce from "lodash/debounce";
+
   export let date: Date;
   export let min: Date | null = null;
   export let max: Date | null = null;
@@ -34,7 +37,11 @@
     return date.getFullYear() + "-" + String(date.getMonth() + 1).padStart(2, "0") + "-" + String(date.getDate()).padStart(2, "0");
   }
 
+  const onChangeDebounced = debounce(() => onChange(), 1000);
   function onChange() {
+    if (!value) {
+      return;
+    }
     let [fullYear, month, day] = value.split("-");
     date.setFullYear(parseInt(fullYear), parseInt(month) - 1, parseInt(day) - deltaInDays);
     date = new Date(date);
--- a/app/frontend/Calendar/EditEvent/DurationSelector.svelte
+++ b/app/frontend/Calendar/EditEvent/DurationSelector.svelte
@@ -1,17 +1,50 @@
-<select bind:value={duration}>
-  {#each [10, 15, 20, 30, 45] as min}
-    <option value={min}>{min} {$plural(min, { one: 'minute', other: 'minutes' })}</option>
-  {/each}
-  {#each [1, 1.5, 2, 4, 8] as hour}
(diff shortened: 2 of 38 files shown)


benbucksch:
@jermy-c Let me know whether this works, please.

e2/src/main/index.ts
```diff
@@ -2,7 +2,7 @@ import { setMainWindow, startupBackend, shutdownBackend, startupArgs } from '../
 import { app, shell, BrowserWindow, session } from 'electron'
 import { ipcMain } from 'electron/main';
 import { join } from 'path'
-import electronUpdater from 'electron-updater';
+import electronUpdater, { UpdateCheckResult } from 'electron-updater';
```
jermy-c:
```suggestion
import electronUpdater, { type UpdateCheckResult } from 'electron-updater';
```

e2/src/main/index.ts
```diff
@@ -126,11 +128,15 @@ async function whenReady() {
   })
 
   try {
-    await autoUpdater.checkForUpdatesAndNotify();
+    update = await autoUpdater.checkForUpdatesAndNotify();
     setInterval(async () => {
       try {
+        if (update?.updateInfo.version) {
```
jermy-c:
```suggestion
        if (update?.isUpdateAvailable) {
```

Why not use `isUpdateAvailable` ? Is there still a problem on Linux?

e2/src/main/index.ts
```diff
@@ -144,10 +150,21 @@ async function whenReady() {
...
+  if (process.platform == 'darwin') {
+    updateAndRestartNowIfNeeded();
+  } else {
+    app.quit();
   }
-})
+});
+
+async function updateAndRestartNowIfNeeded() {
+  if (update?.updateInfo.version) {
```
jermy-c:
```suggestion
  if (await update?.downloadPromise) {
```

Maybe we should check if the update is downloaded before installing?

jermy-c (approved):
Seems good to me!

e2/src/main/index.ts
benbucksch:
```suggestion
  await update?.downloadPromise;
  if (update?.updateInfo.version) {
```

e2/src/main/index.ts
benbucksch: There definitely was a problem on Linux. I don't remember whether I tried this specific code.

benbucksch:
> it takes 10 seconds for the app to quit.

What's the UI during the wait?
* Does the main window disappear, but the app stays running without window (as usual on Mac)? That's fine.
* Does the main window just stick around there doing nothing? In that case, we need to show a dialog that says "Installing update".

jermy-c:
> * Does the main window just stick around there doing nothing? In that case, we need to show a dialog that says "Installing update".

Yes, this is what's happening because `quitAndInstall()` is not a promise.

jermy-c:
@benbucksch I have fixed the merge conflicts and tested changes again.

Issue #802, opened 2025-07-28
Changing date and time of series is ignored unless you fiddle with the repeat options


NeilRashbrook:
Steps to reproduce problem:
1. Create a weekly recurring event
2. Edit the first occurrence of the event, and change it to start on a different day of the week
3. Click Save - Change entire series

Expected result: Series changes to different day

Actual result: Weekly recurring UI previews changed recurrence but change does not stick

Additional information: Changing the recurrence (even if only temporarily) sometimes works, but it doesn't always fix it once you tried to do it just by changing the date and time and you may need to try more than once.

benbucksch:
Please define "it"?

I just noticed that we regressed `hasChanged()` in other ways: I was able to save a meeting without title.

benbucksch:
We're editing only the start time of the instance.
We need to update the master startTime and the starttime of the recurrence rule, based on the instance.
Yes, do the change in `onChangeAll()`

NeilRashbrook:
> Why does that same code not work in `onChangeAll()`?

`updateDateUI` updates the `weekdays` variable to make sense with the new date, but we don't have access to that from `onChangeAll`.

NeilRashbrook:
> That's why we avoid `$: ... {` and instead call a function, which is defined directly below.

So more like this?

```diff
   $: $event.startTime, updateDateUI();
+  // Must update the date UI first, as this fixes the week and weekday variables as necessary
+  $: frequency != Frequency.None && ($event.startTime != master.startTime || event.duration != master.duration) && updateRecurrenceDate();
+  function updateRecurrenceDate() {
+    master.startEditing();
+    master.startTime = event.startTime;
+    master.duration = event.duration;
+    master.newRecurrenceRule(frequency, interval, week, weekdays);
+  }
   function updateDateUI() {
```

benbucksch:
Yes, code is fine now.

But I want to keep code like this out of the UI. All code that ensures **internal consistency of the data** has to be in logic, not in the UI. Is there really no way to put this into into logic/?

benbucksch:
Can the weekday be inferred from the startTime? If the start day is different between this and unedited, then remove the old weekday (unedited.startTime day) and add the current day (startTime day), and leave the other weekdays untouched. Isn't that the logic how we set the initial weekday in the first place, i.e. based on startTime?

NeilRashbrook:
[...]

PR #813, opened 2025-08-07
Move recurrence rule validation into `newRecurrenceRule` #802

NeilRashbrook:
- Also update the rule if the user changes the event's duration
- After updating the rule, use the new rule to update the UI, rather than validating the rule in the UI


benbucksch (approved):

benbucksch (approved):

benbucksch:
Thanks that you found a clean way to move the logic code into logic/

Issue #809, opened 2025-07-30
OWA: Auth: Try to use OAuth2 with .default scope and our client ID


benbucksch:
Goal: Get rid of the custom OWA authentication, and use the standard OAuth2 login sequence that we use for other protocols and providers.

1. In Owl with EWS, with the Owl client ID, change the scope to `https://outlook.office.com/.default` instead of `EWS`, see whether that works.
2. Same as above, but use the Mustang client ID.
3. In Mustang with EWS, use the Owl client ID, change the scope to `https://outlook.office.com/.default` instead of `EWS`, see whether that works.
4. Same as above, but use the Mustang client ID.
5. If this works, try to use our normal OAuth2 authentication logic to authenticate, then use the auth token from that to make OWA JSON requests.
6. If this fails, try to see whether we use our OAuth2 auth logic, but make it mimic what OWA does.
7. If this fails, make a subclass of OAuth2Tab and try to shue-horn the actual OWA login sequence and URL into that, and change the "done" logic to detect login similar to what we do in Owl.

benbucksch:
> > 6. try to see whether we use our OAuth2 auth logic, but make it mimic what OWA does.

> I was able to achieve this but only by using OWA's own client ID and redirect URL.
> Owl's OAuth2 object was able to log in and obtain an access token to access OWA apis.

* Can you please make a patch?
* Do you see *any* other way, other than using the OWA client ID?

> > 7. make a subclass of OAuth2Tab and try to shue-horn the actual OWA login sequence and URL into that, and change the "done" logic to detect login similar to what we do in Owl.

Can you please try that as well?

NeilRashbrook:
> > I was able to achieve this but only by using OWA's own client ID and redirect URL.
> > Owl's OAuth2 object was able to log in and obtain an access token to access OWA apis.
> 
> * Can you please make a patch?

For Owl or Mustang?

> * Do you see _any_ other way, other than using the OWA client ID?

We don't get any sort of information (the HTTP 401 response has no body) so there's no way to tell why it doesn't work with our client ID, so no, sorry.

> > > 7. make a subclass of OAuth2Tab and try to shue-horn the actual OWA login sequence and URL into that, and change the "done" logic to detect login similar to what we do in Owl.
> 
> * Can you please try that as well?

Sure; I just ran out of time yesterday.

NeilRashbrook:
[...]

Issue #814, opened 2025-08-11
Auth: Automatic password fill for in-app authentication in a tab


benbucksch:
In Owl, and in our OWA authentication window, we support password fill, for Office365 and for Exchange on-premise. Please implement the same for the in-app browser tab.

Currently, OAuth2EmbeddedBrowser.svelte uses Browser.svelte, which has an onPageChange function. Implement a generic function which can be called for this event, and which accepts a DOM, and fills in the password. You likely need the function to be stateful, so you can implement it as class `PasswordFill`, where the properties are your state, and the `onPageChange(dom)` function is called when the `<Browser>` event `page-change` fired.

PR #815, opened 2025-08-11
Shoehorn OWA login into OAuth2 logic #809

NeilRashbrook:
I may have misunderstood and this is actually option 6 of #809 rather than option 7. Anyway:
- Disable OAuth2 browser choice for OWA login
- Add OWA autofill to in-app OAuth2 browser
- Unprotected all OAuth2 methods because TypeScript won't duck type otherwise
- Make `isAuthDoneURL` async so that OWA can check by running a request
- Force OWA accounts to have the magic `OWAAuth` object instead of a regular `OAuth2` object
- Switch OWA accounts to use `persist:` rather than `persist:login:`

Unfortunately the code in `OWAAuth` and `OAuth2` is so different that I can't really use inheritance. If you prefer I could make a base interface for all of the calls that the various UI object classes make back to the `OAuth2` object.

### DIFF
--- a/app/frontend/Shared/Auth/OAuth2EmbeddedBrowser.svelte
+++ b/app/frontend/Shared/Auth/OAuth2EmbeddedBrowser.svelte
@@ -1,6 +1,7 @@
 <Browser
   title={$t`Authentication`}
   url={startURL}
+  autofill={owaAutoFillLoginPage(dialog.oAuth2.account.username, dialog.oAuth2.account.password)}
   on:page-change={onPageChange}
   on:close={onClose}
   sessionID={dialog.oAuth2.account?.webSessionID}
@@ -15,6 +16,7 @@
   import type { OAuth2Tab } from "../../../logic/Auth/UI/OAuth2Tab";
   import { OAuth2Embed } from "../../../logic/Auth/UI/OAuth2Embed";
   import Browser from "../Browser.svelte";
+  import { owaAutoFillLoginPage } from "../../../logic/Mail/OWA/Login/OWALoginAutoFill";
   import { UserCancelled, UserError, type URLString, sleep } from "../../../logic/util/util";
   import { onMount } from "svelte";
   import { t } from "../../../l10n/l10n";
@@ -26,7 +28,7 @@
 
   async function onPageChange(event: CustomEvent<URLString>) {
     let url = event.detail;
-    dialog.urlChanged(url);
+    await dialog.urlChanged(url);
   }
 
   function onClose() {
--- a/app/logic/Auth/OAuth2.ts
+++ b/app/logic/Auth/OAuth2.ts
@@ -41,15 +41,15 @@ export class OAuth2 extends Observable {
   @notifyChangedProperty
   accessToken?: string;
   @notifyChangedProperty
-  protected refreshToken?: string;
+  refreshToken?: string;
   @notifyChangedProperty
   idToken: string;
   verificationToken: string; /** `state` URL param of authURL/doneURL */
   uiMethod: OAuth2UIMethod = OAuth2UIMethod.Window;
-  protected ui: OAuth2UI | null = null;
+  ui: OAuth2UI | null = null;
 
   expiresAt: Date | null = null;
-  protected expiryTimout: NodeJS.Timeout;
+  expiryTimout: NodeJS.Timeout;
   refreshErrorCallback = (ex: Error) => console.error(ex);
   idTokenCallback: (idToken: string, oAuth2: OAuth2) => void;
 
@@ -196,7 +196,7 @@ export class OAuth2 extends Observable {
    * @returns accessToken
    * @throws OAuth2Error
    */
-  protected async getAccessTokenFromParams(params: any, additionalHeaders?: any, tokenURL: string = this.tokenURL): Promise<string> {
+  async getAccessTokenFromParams(params: any, additionalHeaders?: any, tokenURL: string = this.tokenURL): Promise<string> {
     params.scope = this.scope;
     params.client_id = this.clientID;
     if (this.clientSecret) {
@@ -274,7 +274,7 @@ export class OAuth2 extends Observable {
   }
 
   /** Helper for auth Done URL */
-  isAuthDoneURL(url: URLString): boolean {
+  async isAuthDoneURL(url: URLString): Promise<boolean> {
     let urlParams = Object.fromEntries(new URL(url).searchParams);
     console.log("OAuth2 page change to", url, "doneURL is", this.authDoneURL, "matches", this.authDoneURL == url,
       "is auth done", url.startsWith(this.authDoneURL) && this.verificationToken && urlParams.state == this.verificationToken);
@@ -357,17 +357,17 @@ export class OAuth2 extends Observable {
   /**
    * @returns refreshToken (or null, if not available)
    */
-  protected async getRefreshTokenFromStorage(): Promise<string | null> {
+  async getRefreshTokenFromStorage(): Promise<string | null> {
     return await getPassword(this.storageKey);
   }
-  protected async deleteRefreshTokenFromStorage(): Promise<void> {
+  async deleteRefreshTokenFromStorage(): Promise<void> {
     await deletePassword(this.storageKey);
   }
-  protected async storeRefreshToken(refreshToken: string): Promise<void> {
+  async storeRefreshToken(refreshToken: string): Promise<void> {
     assert(refreshToken, "Nothing to store");
     await setPassword(this.storageKey, refreshToken);
   }
-  protected get storageKey(): string {
+  get storageKey(): string {
     let host = new URL(this.tokenURL).host.replaceAll(".", "-");
     let username = this.account.username.replace(/@/, "-").replaceAll(".", "-");
     return `oauth2.refreshToken.${host}.${username}`;
--- a/app/logic/Auth/OWAAuth.ts
+++ b/app/logic/Auth/OWAAuth.ts
@@ -1,0 +1,114 @@
+import type { OAuth2UIMethod } from "./UI/OAuth2UIMethod";
+import type { OAuth2 } from "./OAuth2";
+import { OAuth2LoginNeeded } from "./OAuth2Error";
+import { OAuth2Tab } from "./UI/OAuth2Tab";
+import type { OWAAccount} from "../Mail/OWA/OWAAccount";
+import { appGlobal } from "../app";
+import { Observable } from "../util/Observable";
+import { assert, type URLString } from "../util/util";
+
+/// Class to perform fake OAuth2 to log into OWA web interface via browser
+export class OWAAuth extends Observable implements OAuth2 {
+  ui: OAuth2Tab | null = null;
+  isLoggedIn = false;
+
+  constructor(public account: OWAAccount) {
+    super();
+  }
+
+  // Called from `TBProfile.readMailAccount()`
+  setTokenURLPasswordAuth(url: string | null | undefined) {
+  }
+
+  // Called from `OWAAccount.loginCommon()`
+  async login(interactive: boolean): Promise<string> {
+    assert(this.account, "Need to set account first");
+    if (this.isLoggedIn) {
+      return;
+    }
+    if (!interactive) {
+      throw new OAuth2LoginNeeded();
+    }
+    return this.loginWithUI();
+  }
+
+  // Called from above to match behaviour of OAuth2 class
+  async loginWithUI(): Promise<string> {
+    this.ui = new OAuth2Tab(this);
+    await this.ui.login();
+    this.ui = null;
+    return "";
+  }
+
+  // Called from `prepareLogin()` during account setup
+  // ui is probably not actually set at this point, but whatever
+  abort() {
+    this.ui?.abort();
+  }
+
+  // Called from `OWAAccount.logout()`
+  async logout() {
+    this.isLoggedIn = false;
+    return appGlobal.remoteApp.OWA.clearStorageData(this.account.partition);
+  }
+
+  // Called from `startLogin()` during account setup
+  async getAccessTokenFromAuthCode(authCode: string): Promise<string> {
+    await appGlobal.remoteApp.OWA.fetchSessionData(this.account.partition, this.account.url, false);
+    return "";
+  }
+
+  // Called from `OAuth2Embed.login()` and `OAuth2Tab.login()`
+  async getAuthURL(): Promise<URLString> {
+    return this.account.url;
+  }
+
+  // Called from `OAuth2Embed.urlChanged()` and `OAuth2Tab.urlChanged()`
+  async isAuthDoneURL(url: URLString): Promise<boolean> {
+    try {
+      await appGlobal.remoteApp.OWA.fetchSessionData(this.account.partition, this.account.url, false);
+      this.isLoggedIn = true;
+    } catch (ex) {
+      this.isLoggedIn = false;
+    }
+    return this.isLoggedIn;
+  }
+
+  // Called from `OAuth2Embed.success()` and `OAuth2Tab.success()`
+  getAuthCodeFromDoneURL(url: URLString): string {
+    return "";
+  }
+
+  // Called from `MailAccount.toConfigJSON()`
+  toConfigJSON() {
+  }
+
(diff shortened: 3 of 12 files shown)


app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -70,7 +73,7 @@ export class OWAAccount extends MailAccount {
   }
 
   async verifyLogin(): Promise<void> {
-    await this.loginCommon(true);
+    await this.loginCommon(false);
```
benbucksch: This has to be interactive, because it's the very first time we log in, and the user is there, and it *is* in fact interactive.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -61,7 +64,7 @@ export class OWAAccount extends MailAccount {
    * multiple OWA accounts for the same host.
    */
   get partition(): string {
-    return 'persist:login:' + this.id;
+    return 'persist:' + this.id;
```
```suggestion
    return 'persist:' + this.webSessionID;
```

app/logic/Auth/OWAAuth.ts
```diff
@@ -0,0 +1,114 @@
...
+  getAuthCodeFromDoneURL(url: URLString): string {
+    return "";
+  }
+
+  // Called from `MailAccount.toConfigJSON()`
+  toConfigJSON() {
+  }
+
+  // Unused stuff but needed to fake out TypeScript for implements OAuth2
+  // Can't actually derive from OAuth2 because all its assertions would fail
```
benbucksch: Which assertions would fail? 2 other ideas:
* Make the assertions pass, e.g. by passing in a tokenURL and authURL.
* Override the function that asserts.
* Make a base class for `OAuth2` and `OWAAuth`, which has all the functions that the UI and other parts of the app need.

Use whatever is cleaner.

benbucksch (changes requested):

app/logic/Auth/OWAAuth.ts
NeilRashbrook: The function that asserts is the constructor, which is the one function you can't override.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: The very first time we log in is when setup does `startLogin`, which is interactive.

benbucksch: `startLogin()` is in the UI and irrelevant here. Whatever the UI does is always - by definition - irrelevant to logic. Logic works independent of the UI.

`verifyLogin()` is inherently interactive.

app/logic/Auth/OWAAuth.ts
benbucksch: That eliminates option 2 (unless the ctor is changed), but not 1 and 3.

NeilRashbrook: I wasn't claiming that it doesn't eliminate the other options.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: It used to break setup, but it seems that somewhere along the line the code stopped trying to open a login tab during setup, so I guess I can revert this after all.

app/frontend/Shared/Auth/OAuth2EmbeddedBrowser.svelte
```diff
@@ -15,6 +16,7 @@
   import type { OAuth2Tab } from "../../../logic/Auth/UI/OAuth2Tab";
   import { OAuth2Embed } from "../../../logic/Auth/UI/OAuth2Embed";
   import Browser from "../Browser.svelte";
+  import { owaAutoFillLoginPage } from "../../../logic/Mail/OWA/Login/OWALoginAutoFill";
```
benbucksch, line 19: We want this to work for all account types. Definitely for all Exchange types.
After this PR, we also want auto-fill for Google, Yahoo etc.

The code here is generic and shouldn't be specific to the account type. Obviously, the password fill script itself might need some provider-specific code.

app/logic/Auth/OAuth2.ts
```diff
@@ -41,25 +41,24 @@ export class OAuth2 extends Observable {
   @notifyChangedProperty
   accessToken?: string;
   @notifyChangedProperty
-  protected refreshToken?: string;
+  refreshToken?: string;
```
benbucksch: Can you keep this `protected`, please?

Same with all the other properties below.

app/logic/Auth/OWAAuth.ts
```diff
@@ -0,0 +1,96 @@
+import { WebAuth } from "./WebAuth";
+import type { OAuth2UIMethod } from "./UI/OAuth2UIMethod";
+import type { OAuth2 } from "./OAuth2";
+import { OAuth2LoginNeeded } from "./OAuth2Error";
+import { OAuth2Tab } from "./UI/OAuth2Tab";
+import type { OWAAccount } from "../Mail/OWA/OWAAccount";
+import { appGlobal } from "../app";
+import { assert, NotReached, type URLString } from "../util/util";
+
+/// Class to perform fake OAuth2 to log into OWA web interface via browser
```
benbucksch:
```suggestion
/** Log into OWA web interface via browser. Mimicks the `OAuth2` login API. */
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -79,9 +81,11 @@ export class OWAAccount extends MailAccount {
    */
   protected async loginCommon(interactive: boolean): Promise<void> {
     if (this.authMethod == AuthMethod.OAuth2) {
-      // The backend has the logic for posing the login page
-      // using the correct cookie jar and auto-filling it.
-      await this.listFolders(interactive);
+      if (!this.oAuth2) {
+        this.oAuth2 = new OWAAuth(this);
+      }
```
benbucksch, line 86:
```suggestion
      this.oAuth2 = new OWAAuth(this);
```

This is nice.

Can we then please remove all the other places that set `Account .oAuth2` and special-case OWA?

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -255,16 +259,12 @@ export class OWAAccount extends MailAccount {
...
-    if (interactive) {
-      autofillJS = owaAutoFillLoginPage(this.username, this.password);
-    }
+  async listFolders(): Promise<void> {
     await this.throttle.throttle();
     let lock = await this.semaphore.lock();
     let sessionData: any;
     try {
-      sessionData = await appGlobal.remoteApp.OWA.fetchSessionData(this.partition, this.url, interactive, autofillJS);
+      sessionData = await appGlobal.remoteApp.OWA.fetchSessionData(this.partition, this.url, false);
```
benbucksch, line 271: In a followup PR, please change this to a normal get folders call, like we did in Owl 1.4.1.1.

app/logic/Mail/AutoConfig/readConfig.ts
```diff
@@ -163,7 +166,10 @@ function readServer(xml: any, displayName: string, fullXML: any, source: ConfigS
   return account;
 }
 
-function getOAuth2Config(account: Account, autoConfigXML: any): OAuth2 {
+function getWebAuthConfig(account: Account, autoConfigXML: any): WebAuth {
+  if (account instanceof OWAAccount) {
+    return new OWAAuth(account);
+  }
```
benbucksch, line 172: You can remove this change (see below)

app/logic/Mail/Import/Thunderbird/TBProfile.ts
```diff
@@ -105,7 +106,7 @@ export class ThunderbirdProfile {
         }
         let url = OAuth2URLs.find(url => url.hostnames.includes(hostname));
         assert(url, `${acc.name}: Need OAuth2 config for host ${hostname}`);
-        acc.oAuth2 = new OAuth2(
+        acc.oAuth2 = acc instanceof OWAAccount ? new OWAAuth(acc) : new OAuth2(
```
benbucksch, line 109: You can remove this change (see below)

app/logic/Auth/WebAuth.ts
```diff
@@ -0,0 +1,27 @@
+import type { Account } from "../Abstract/Account";
+import { Observable } from "../util/Observable";
+import type { URLString } from "../util/util";
+
+// Interface for authentication methods that use web pages, i.e. OWA and oAuth2.
+export abstract class WebAuth extends Observable {
```
benbucksch:
```suggestion
export abstract class WebBasedAuth extends Observable {
```

`WebAuth` is too close to `WebAuthn`, so better to rename.

app/logic/Auth/OWAAuth.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  // Called from `OAuth2Embed.login()` and `OAuth2Tab.login()`
+  async getAuthURL(): Promise<URLString> {
+    return this.account.url;
+  }
+
+  // Called from `OAuth2Embed.urlChanged()` and `OAuth2Tab.urlChanged()`
+  async isAuthDoneURL(url: URLString): Promise<boolean> {
+    try {
+      await appGlobal.remoteApp.OWA.fetchSessionData(this.account.partition, this.account.url, false);
```
benbucksch, line 80: Please first make a cheaper test, whether we have certain data, without network request, like we do in Owl.
The network request should happen only as last double-check.

In a followup PR, please change this to a normal server call, like in Owl. See comment below.

app/frontend/Shared/Auth/OAuth2EmbeddedBrowser.svelte
NeilRashbrook: I forget how generic the password fill script is, but it does at least work for EWS/ActiveSync for Office 365.

app/logic/Auth/OAuth2.ts
NeilRashbrook: Sorry, I forgot to undo this when I switched to using a base class.

app/logic/Auth/OWAAuth.ts
NeilRashbrook: It's not worth making a cheaper test if I'm immediately going to obsolete it by switching to a server call for FindFolders. (And when I say immediately I mean I already have a draft branch for this, but I can't create a PR yet due to merge conflicts.)

app/logic/Auth/WebAuth.ts
NeilRashbrook: Never heard of `WebAuthn`; thanks for tipping me off.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: Sorry, this doesn't work. As I mentioned earlier, the first login actually comes from `startLogin`, and we need `oAuth2` to be set to `OWAAuth` at this point, so that we can fake out setup into letting us use its `Browser`.

app/frontend/Shared/Auth/OAuth2EmbeddedBrowser.svelte
NeilRashbrook: (I can also write a follow-up PR to move the script into a more suitable location and simplify its parameters.)

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: > the first login actually comes from startLogin

Who initiates that? The initiation should come from either verifyLogin() or login(), which then triggers the UI dialog.

Or is there a case in during setup that I'm missing? Can you tell me the case you mean, both from a user perspective (how did he get there) and a code perspective (call stack, and what function triggered the UI), please?

app/frontend/Shared/Auth/OAuth2EmbeddedBrowser.svelte
benbucksch: OK, thanks, here or followup, both is fine with me.

app/logic/Auth/OWAAuth.ts
benbucksch: OK with me. Even if you change to FindFolders, given that this runs on every page change, I do also want a cheaper test that is fully local and best case even sync, without network request, like our `this.OAuthData` test in Owl.

But it's fine for me to do that in a followup PR. Makes sense.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -148,7 +152,7 @@ export class OWAAccount extends MailAccount {
 
   async logout(): Promise<void> {
     this.hasLoggedIn = false;
-    return appGlobal.remoteApp.OWA.clearStorageData(this.partition);
+    return this.oAuth2 ? this.oAuth2.logout() : appGlobal.remoteApp.OWA.clearStorageData(this.partition);
```
benbucksch: We can remove the `return` and make this a normal `if` block.

benbucksch (approved):

app/logic/Auth/OWAAuth.ts
benbucksch: You chose the base class. That's probably cleanest, even if not convenient. Thank you.

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: (This is the only question that blocks the landing.)

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: But be sure to add the `await`s

app/logic/Auth/OWAAuth.ts
NeilRashbrook: FYI the server call (as opposed to `fetchSessionData`) does do a cheap test before the network request. (Well, cheap in that it still has to be done in `owa.ts` via JPC, but it won't hit the network if the test fails.)

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: > verifyLogin() ... which then triggers the UI dialog.

I know you claimed it's interactive, but you really don't want to show any UI during `verifyLogin`, because that destroys the setup page. (Existing OWA uses a pop-up window which doesn't count.)

> Can you tell me the case you mean, both from a user perspective (how did he get there) and a code perspective (call stack, and what function triggered the UI), please?

From a user perspective, this happens after you choose the account type.

From a code perspective, `SetupMail` goes through several steps:
- In `Step.FindConfig`, it creates all of the alternative configuations and applies web-based auth where applicable
- In `Step.FoundConfig` the user chooses the config they wish to use
- `SetupMail` then checks whether that config uses web-based auth
- If so, `Step.Login` logs in to web-based auth using an embedded UI
  - `LoginPage` has provision to ask for the password, but `SetupMail` doesn't use this
  - `OAuth2Login` calls `startLogin` with an `OAuth2Embed` object which then calls `login` triggering the web-based auth in its `OAuth2EmbeddedBrowser`.
- `Step.CheckConfig` then calls `verifyLogin` with no provision for UI.

I would recommend that further questions be directed at the author of `SetupMail`.

benbucksch: Thanks for the explanation. That makes sense.

I didn't remember that anymore.

> I would recommend that further questions be directed at the author of SetupMail.

I tried, but he didn't know from memory, and he's extremely busy with tons of other stuff, so he couldn't investigate.

Thanks that you looked this up. Appreciated.

app/frontend/Shared/Auth/OAuth2EmbeddedBrowser.svelte
NeilRashbrook: PR #814 moved the script into a more suitable location and simplified its parameters.

app/logic/Auth/OWAAuth.ts
NeilRashbrook: PR #818 covers the removal of `fetchSessionData`.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: PR #818 covers the removal of `fetchSessionData`.

PR #816, opened 2025-08-12
Only use `void` as a return type

NeilRashbrook:
`void` has a special meaning as a return type, in that you're not expected to examine the return value at all. We should use `undefined` instead everywhere else, including where you do examine the return value because it's a union type.

Optional chaining works on a variable declared as `string | undefined` but not on one defined as `string | void`.

### DIFF
--- a/app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
@@ -177,7 +177,7 @@ export class ActiveSyncCalendar extends Calendar implements ActiveSyncPingable {
     this.account.addPingable(this);
   }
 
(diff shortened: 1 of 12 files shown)


benbucksch:
If we have no result, the result should be `null` (certain that there's nothing), not `undefined` (don't know).

NeilRashbrook:
Sure, now go back in time and change the API for `Array.find`...

NeilRashbrook:
(It turns out that my editor's syntax highlighter doesn't like a class method declared as returning `void`; it would rather that the type be inferred. It's OK with `Promise<void>` though.)

benbucksch:
> Sure, now go back in time and change the API for `Array.find`...

Yes. Quite a number of the `Array` function signatures are not well considered.

`?? null` would have done it, but fair enough.

benbucksch (approved):

Issue #807, opened 2025-07-29
OWA: Calendar: Check lastModifiedTime before re-fetching the event


benbucksch:
We don't want to fetch all events with all details on every sync.

PR #817, opened 2025-08-12
OWA: Check `lastMod` before fetching the whole event #807

NeilRashbrook:


app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -78,7 +78,17 @@ export class OWACalendar extends Calendar {
         break;
       }
       request.Body.Paging.Offset = sanitize.integer(result.RootFolder.IndexedPagingOffset);
-      await this.getEvents(result.RootFolder.Items.map(item => item.ItemId.Id), events);
+      let eventIDs: string[] = [];
+      for (let item of result.RootFolder.Items) {
+        let event = this.getEventByItemID(item.ItemId.Id);
+        if (event?.lastMod.getTime() == sanitize.date(item.LastModifiedTime).getTime()) {
```
benbucksch: You need defaults, in case they are not set, otherwise it throws:
```suggestion
        if (event?.lastMod?.getTime() == sanitize.date(item.LastModifiedTime, null)?.getTime()) {
```

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -78,7 +78,17 @@ export class OWACalendar extends Calendar {
...
+          events.add(event);
```
benbucksch, line 86: Also please add a code comment for each case, what you intend to do here. It's not immediately clear from the code.

```suggestion
          // Already up-to-date, no need to fetch from server, directly add to results
          events.add(event);
```

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -78,7 +78,17 @@ export class OWACalendar extends Calendar {
...
+          events.addAll(event.exceptions);
+        } else {
+          eventIDs.push(item.ItemId.Id);
```
benbucksch, line 90:
```suggestion
          // Fetch it from server
          eventIDs.push(item.ItemId.Id);
```

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -78,7 +78,17 @@ export class OWACalendar extends Calendar {
         break;
       }
       request.Body.Paging.Offset = sanitize.integer(result.RootFolder.IndexedPagingOffset);
-      await this.getEvents(result.RootFolder.Items.map(item => item.ItemId.Id), events);
+      let eventIDs: string[] = [];
+      for (let item of result.RootFolder.Items) {
+        let event = this.getEventByItemID(item.ItemId.Id);
+        if (event?.lastMod.getTime() == sanitize.date(item.LastModifiedTime).getTime()) {
+          events.add(event);
+          events.addAll(event.exceptions);
```
benbucksch, line 87: I thought `recurrenceColl()` does the expansion, but it does it only for instances, not for exceptions, requiring this special handling here and everywhere. Strange...

app/logic/Calendar/OWA/OWACalendar.ts
NeilRashbrook: Strictly speaking, if `LastModifiedTime` is not set on a new event, we'll just skip that entirely now, rather than throwing. (Also `lastMod` is never nullish.)

app/logic/Calendar/OWA/OWACalendar.ts
NeilRashbrook: Instances aren't saved on the server, while exceptions are, which probably explains the difference.

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -78,7 +78,20 @@ export class OWACalendar extends Calendar {
         break;
       }
       request.Body.Paging.Offset = sanitize.integer(result.RootFolder.IndexedPagingOffset);
-      await this.getEvents(result.RootFolder.Items.map(item => item.ItemId.Id), events);
+      let eventIDs: string[] = [];
+      for (let item of result.RootFolder.Items) {
+        let event = this.getEventByItemID(item.ItemId.Id);
+        if (event?.lastMod.getTime() == sanitize.date(item.LastModifiedTime, null)?.getTime()) {
+          // Our local event is already up-to-date,
+          // so we just add it directly to our results.
```
benbucksch:
```suggestion
          // Our local event is already up-to-date, so we just add it directly to our results.
```

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -78,7 +78,19 @@ export class OWACalendar extends Calendar {
         break;
       }
       request.Body.Paging.Offset = sanitize.integer(result.RootFolder.IndexedPagingOffset);
-      await this.getEvents(result.RootFolder.Items.map(item => item.ItemId.Id), events);
+      let eventIDs: string[] = [];
+      for (let item of result.RootFolder.Items) {
+        let event = this.getEventByItemID(item.ItemId.Id);
+        if (event?.lastMod.getTime() == sanitize.date(item.LastModifiedTime, null)?.getTime()) {
+          // Our local event is already up-to-date, so we just add it directly to our results.
```
benbucksch:
```suggestion
          // Our local event is up-to-date, so just add it directly to the results.
```

benbucksch (approved):

app/logic/Calendar/OWA/OWACalendar.ts
NeilRashbrook: > (Also `lastMod` is never nullish.)

Except nobody told `fromExtraJSON`, sigh...

benbucksch: So, can we add the `?` here, please?
And **also** fix `fromExtraJSON()` to have the expected default value.
I'd like us to fix it in both places.

benbucksch: fromExtraJSON() fixed in #836 
`?` here still TODO

NeilRashbrook: I'm worried about the case where neither the event nor the item have a time, at which point they will now compare equal.

Also, I didn't realise that we don't actually have type checking for null and undefined enabled. I tried enabling it and got 892 more TypeScript errors...

benbucksch: > I'm worried about the case where neither the event nor the item have a time, at which point they will now compare equal.

If both are null, we'll just compare null with null, which is true, which is what we want, no?

> I didn't realise that we don't actually have type checking for null and undefined enabled.

That is deliberate.

It's also deliberate that we use `==` and not `===`.

NeilRashbrook: Is it though? I don't know.

I don't like the way you can write `let foo: Date = null;` and TypeScript doesn't complain because we don't check for that.

I'm not taking about using `==`; this isn't about comparing `null` to `undefined`, this is about writing `null` (or `undefined`) to a variable that's not supposed to be `null` (or `undefined`).

benbucksch: > I don't like the way you can write let foo: Date = null; and TypeScript doesn't complain because we don't check for that.

I don't like the syntax `foo: Date | null = null`, because it's redundant. Generally, if a value is initalized with `null`, then obviously it may be null. The initalized value is also always a valid value, by definition (pun intended). I wish that TypeScript was smarter in this case.

Also, I consider `undefined` to be equal to `null`, in most cases. There is a difference (unknown vs. known to be empty), but so small that most languages do not have that distinction.

OTOH, if a property is initialized with a concrete value, the assumption is that the variable may never be null.

PR #818, opened 2025-08-12
OWA: Stop using sessiondata.ashx

NeilRashbrook:
As requested in PR #815.

Note that `callOWA` uses `fetchJSON` which checks for the `X-OWA-CANARY` cookie before hitting the network.

### DIFF
--- a/app/logic/Auth/OWAAuth.ts
+++ b/app/logic/Auth/OWAAuth.ts
@@ -4,6 +4,7 @@ import type { OAuth2 } from "./OAuth2";
 import { OAuth2LoginNeeded } from "./OAuth2Error";
 import { OAuth2Tab } from "./UI/OAuth2Tab";
 import type { OWAAccount } from "../Mail/OWA/OWAAccount";
+import { owaFindFoldersRequest } from "../Mail/OWA/Request/OWAFolderRequests";
 import { appGlobal } from "../app";
 import { assert, NotReached, type URLString } from "../util/util";
 
@@ -65,7 +66,7 @@ export class OWAAuth extends WebBasedAuth {
 
   // Called from `startLogin()` during account setup
   async getAccessTokenFromAuthCode(authCode: string): Promise<string> {
-    await appGlobal.remoteApp.OWA.fetchSessionData(this.account.partition, this.account.url, false);
+    await this.account.callOWA(owaFindFoldersRequest(), true);
     return "";
   }
 
@@ -77,7 +78,7 @@ export class OWAAuth extends WebBasedAuth {
   // Called from `OAuth2Embed.urlChanged()` and `OAuth2Tab.urlChanged()`
   async isAuthDoneURL(url: URLString): Promise<boolean> {
     try {
-      await appGlobal.remoteApp.OWA.fetchSessionData(this.account.partition, this.account.url, false);
+      await this.account.callOWA(owaFindFoldersRequest(), true);
       this.isLoggedIn = true;
     } catch (ex) {
       this.isLoggedIn = false;
--- a/app/logic/Auth/OWAAuth.ts
+++ b/app/logic/Auth/OWAAuth.ts
@@ -66,7 +66,9 @@ export class OWAAuth extends WebBasedAuth {
 
   // Called from `startLogin()` during account setup
   async getAccessTokenFromAuthCode(authCode: string): Promise<string> {
-    await this.account.callOWA(owaFindFoldersRequest(), true);
+    if (!await this.account.testLoggedIn()) {
+      throw new OAuth2LoginNeeded();
+    }
     return "";
   }
 
@@ -77,13 +79,7 @@ export class OWAAuth extends WebBasedAuth {
 
   // Called from `OAuth2Embed.urlChanged()` and `OAuth2Tab.urlChanged()`
   async isAuthDoneURL(url: URLString): Promise<boolean> {
-    try {
-      await this.account.callOWA(owaFindFoldersRequest(), true);
-      this.isLoggedIn = true;
-    } catch (ex) {
-      this.isLoggedIn = false;
-    }
-    return this.isLoggedIn;
+    return this.isLoggedIn = await this.account.testLoggedIn();
   }
 
   // Called from `OAuth2Embed.success()` and `OAuth2Tab.success()`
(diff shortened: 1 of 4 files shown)


app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -214,8 +214,8 @@ export class OWAAccount extends MailAccount {
     await this.callOWA(request);
   }
 
-  async callOWA(aRequest: any): Promise<any> {
-    if (!this.hasLoggedIn) {
+  async callOWA(aRequest: any, forLogin?: true): Promise<any> {
```
benbucksch: Please remove this parameter. In general, please avoid such special-case modality flags.

app/logic/Auth/OWAAuth.ts
```diff
@@ -77,7 +78,7 @@ export class OWAAuth extends WebBasedAuth {
   // Called from `OAuth2Embed.urlChanged()` and `OAuth2Tab.urlChanged()`
   async isAuthDoneURL(url: URLString): Promise<boolean> {
     try {
-      await appGlobal.remoteApp.OWA.fetchSessionData(this.account.partition, this.account.url, false);
+      await this.account.callOWA(owaFindFoldersRequest(), true);
```
benbucksch: Please make an explicit check whether we have login data *before* issuing the server call.

I saw your comment, but we don't want to depend on implementation details of how exactly `callOWA()` is implemented and assume that it does something in a certain way.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -263,26 +263,9 @@ export class OWAAccount extends MailAccount {
     return result;
   }
 
-  async listFolders(): Promise<void> {
+  async listFolders(forLogin?: true): Promise<void> {
```
benbucksch: Please remove this parameter. In general, please avoid such special-case modality flags.

You can make a new function called `testLoggedIn()`, which does minimal work (and its implementation can be changed at will), instead of calling `listFolders()` for this purpose.

benbucksch: I guess I don't understand why you'd call `listFolders(false)` with a shallow query.

NeilRashbrook: > I guess I don't understand why you'd call `listFolders(false)` with a shallow query.

That's good, because I didn't.

benbucksch: Are there no calls to `listFolders()` (without arguments, meaning false)? Why do you have that flag for a shallow query, then?

NeilRashbrook: `listFolders` is an abstract function in `MailAccount` even though nobody ever calls it (each account type only calls it internally, and `AllAccounts` has an unused implementation) so I have to allow it to have no arguments.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -71,6 +71,21 @@ export class OWAAccount extends MailAccount {
     return this.hasLoggedIn;
   }
 
+  async testLoggedIn(): Promise<boolean> {
+    assert(!this.hasLoggedIn, "Only for use during login");
+    let url = this.url + 'service.svc';
+    let options = { headers: { Action: "FindFolder" } };
+    let bodyJSON = Object.assign({}, owaFindFoldersRequest());
```
benbucksch: What's the point of the `Object.assign({}, ...)`?

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -85,28 +100,19 @@ export class OWAAccount extends MailAccount {
...
+      }
+      let response = await OWALoginBackground.submitLoginForm(this.username, this.password, this.partition, elements);
+      let formURL = new URL(elements.url);
+      let responseURL = new URL(response.url);
+      if (response.status == 401 || responseURL.origin == formURL.origin && responseURL.pathname == formURL.pathname && responseURL.searchParams.get("reason") == "2") {
+        throw new LoginError(null, gt`Password incorrect`);
+      }
+      if (!response.ok || !await this.testLoggedIn()) {
+        throw new Error(`HTTP ${response.status} ${response.statusText}`);
       }
```
benbucksch: This needs to be a different error message: "Error HTTP 200 OK" doesn't make any sense.

```suggestion
      if (!response.ok) {
        throw new Error(`HTTP ${response.status} ${response.statusText}`);
      }
      if (!await this.testLoggedIn()) {
        throw new LoginError(`Login check failed`);
      }
```

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: The OWA requests are implemented as (subclasses of) OWARequest, which provides the `__type` and `Header` properties for you while you provide the `Body` property (or the subclass does). However, we want to strip away the class before sending the request though JPC, as we're only going to pass it to `JSON.stringify` anyway.

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: So, why this modality? Why would you do a shallow listing, ever? Would it make sense to remove the shallow listing option?

app/logic/Mail/OWA/OWAAccount.ts
benbucksch:
```suggestion
    let bodyJSON = Object.assign({}, owaFindFoldersRequest(false)); // Remove class before JPC, not needed for JSON
```

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: Oh, I see you already removed it. Thank you. It's great as-is.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: This is actually a TypeScript error.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: It turns out that `LoginError` takes two parameters...

PR #823, opened 2025-08-13
Password lost when changing account type in manual setup

NeilRashbrook:
During setup, the password on the initial setup screen does not get copied to the manual setup UI. (Whether that is in itself a bug is outside the scope of this PR.)

However, the manual setup UI shows `••••••••` as a placeholder, making it look as if the password has in fact been copied.

I suggest that the placeholder only be set if there is in fact a password.

I've also suggested a different placeholder but if you don't like that idea then that's less important.


NeilRashbrook:
If you decide that the missing password is the true bug then we can just make the above change, not the placeholder.

PR #824, opened 2025-08-13
Allow manual setup of OWA accounts with full page login again

NeilRashbrook:
PR #815 changed the way OWA full page login works, which broke manual setup.

Manual setup also didn't work properly with EWS accounts, and this also improves that.

The changes in `ManualConfigURL` were ported from `ManualConfigHostServer`, while the change in `SetupMail` was ported from line 173 as I assume manual setup has no instructions.


benbucksch (approved):

benbucksch:
> Make a PR, please?

Done

PR #826, opened 2025-08-14
UI: Person Autocomplete: Allow to create Participant objects

NeilRashbrook:
This fixes a bug "introduced" by eaf6fe3.

Feel free to suggest a better name for my variable. (I'm so bad at naming things, I tried calling it `constructor`...)

### DIFF
--- a/app/frontend/Contacts/PersonAutocomplete/PersonsAutocomplete.svelte
+++ b/app/frontend/Contacts/PersonAutocomplete/PersonsAutocomplete.svelte
@@ -41,11 +41,23 @@
   export let tabindex = null;
   export let autofocus = false;
   export let disabled = false;
-  export let onAddPerson: (person: PersonUID) => void | Promise<void> = function(person) { persons.add(person); };
-  export let onRemovePerson: (person: PersonUID) => void | Promise<void> = function(person) { persons.remove(person); };
+  export let onAddPerson: (person: PersonUID) => void | Promise<void> = onAddPersonDefault;
+  export let onRemovePerson: (person: PersonUID) => void | Promise<void> = onRemovePersonDefault;
 
   //$: console.log("persons", persons.contents);
 
+  function onAddPersonDefault(person: PersonUID) {
+    if (!person || persons.contains(person)) {
+      return;
+    }
+    persons.add(person);
+  }
+  function onRemovePersonDefault(person: PersonUID) {
+    if (!person) {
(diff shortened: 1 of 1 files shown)


app/frontend/Contacts/PersonAutocomplete/PersonsAutocomplete.svelte
```diff
@@ -43,6 +43,7 @@
   export let tabindex = null;
   export let autofocus = false;
   export let disabled = false;
+  export let type: new (string, string) => PersonUID;
```
benbucksch:
```suggestion
  export let PersonType: new (string, string) => PersonUID;
```

benbucksch: (Old review comment that I forgot to submit)

app/frontend/Contacts/PersonAutocomplete/PersonsAutocomplete.svelte
```diff
@@ -43,21 +41,11 @@
   export let tabindex = null;
   export let autofocus = false;
   export let disabled = false;
+  export let onAddPerson: (person: PersonUID) => void | Promise<void> = function(person) { persons.add(person); };
```
benbucksch:
```suggestion
  export let onAddPerson: (person: PersonUID) => void | Promise<void> = onAddPersonDefault;
  ...
  function onAddPersonDefault(person: PersonUID) {
    if (!person || persons.contains(person)) {
      return;
    }
    persons.add(person);
  }
```

ditto for remove

app/frontend/Contacts/PersonAutocomplete/PersonsAutocomplete.svelte
NeilRashbrook: Very old; that code no longer exists since the rewrite.

app/frontend/Contacts/PersonAutocomplete/PersonsAutocomplete.svelte
NeilRashbrook: (I had originally put the default methods in `PersonEntry` and `PersonAutocomplete` and when I later moved them back I didn't realise I could put back the default functions.)

benbucksch (approved):
I'm not super happy with changing how the event handlers work.
But I see why you did it, so OK.

PR #828, opened 2025-08-19
Android build with Nodejs v22

jermy-c:
- Updated currently with mustang forks
- Node.js v22.9.0

TODO:
- [ ] Dev environment requires the code signing environment variables to be set. Should allow dev build without signing.
- [x] Bundles libnode for all architectures, should bundle only ARM 64bit
- [ ] Update the APK version and app name from package.json or capacitor.config


.github/workflows/installer-android.yml
```diff
@@ -0,0 +1,62 @@
...
+        gradle-home-cache-includes: |
+          mobile/android/.gradle
+
+    - run: cd app/build; sh mustang-brand.sh
+    - run: cd app; yarn install
+    - run: cd lib; yarn install
+    - run: cd backend; yarn install
+    - run: cd mobile; yarn install
+    - run: cd mobile/backend; yarn install
+    - run: cd lib; yarn add bufferutil # explicitly include bufferutil otherwise it might not be included
```
benbucksch: That's odd. Why does the install fail?

benbucksch: Can you remove this here and add this to the `run: cd lib; yarn install` line above, like so:
```
run: cd lib; yarn install; yarn add bufferutil # explicitly include bufferutil otherwise it might not be included
```

or (preferred, if it works):

```
run: cd lib; yarn add bufferutil; yarn install # explicitly include bufferutil otherwise it might not be included
```

If these don't work, that gives you a hint at what the problem is.

mobile/backend/index.ts
```diff
@@ -2,6 +2,7 @@ import { startupBackend, shutdownBackend } from './backend';
 
 async function startup(): Promise<void> {
   try {
+    console.log(`Node.js version: ${process.version}`);
     console.log("backend starting");
```
benbucksch:
```suggestion
    console.log(`Backend starting with node.js ${process.version}`);
```

mobile/backend/prebuild.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
 cd node_modules;
-(cd better-sqlite3 && npx prebuild-for-nodejs-mobile $MOBILE_ARCH)
-(cd bufferutil && npx prebuild-for-nodejs-mobile $MOBILE_ARCH)
+(cd better-sqlite3 && npx prebuild-for-nodejs-mobile $MOBILE_ARCH $ANDROID_SDK)
+(cd bufferutil && npx prebuild-for-nodejs-mobile $MOBILE_ARCH $ANDROID_SDK)
```
benbucksch, line 7: Will this still work when we build for iOS?

mobile/build.sh
```diff
@@ -2,6 +2,10 @@
 
 mkdir -p ../mobile/dist
 
+export NODE_OPTIONS="--max-old-space-size=32768"
+
 (cd ../app && npm run build && cp -R ./dist/* ../mobile/dist) &&
 (cd backend && npm run build) &&
-npx cap sync
+npx cap copy
```
benbucksch, line 9: Why this change?

mobile/capacitor.config.ts
```diff
@@ -0,0 +1,19 @@
+import type { CapacitorConfig } from '@capacitor/cli';
+
+const config: CapacitorConfig = {
+  appId: "im.mustang.capa",
```
benbucksch, line 4: Can we please also build Parula?

mobile/package.json
```diff
@@ -5,22 +5,25 @@
...
-    "@capacitor/core": "latest",
-    "@capacitor/ios": "^7.2.0",
-    "@capacitor/splash-screen": "latest",
-    "capacitor-nodejs": "https://github.com/hampoelz/capacitor-nodejs/releases/download/v1.0.0-beta.7/capacitor-nodejs.tgz"
+    "@capacitor/android": "^7.0.0",
+    "@capacitor/camera": "^7.0.0",
+    "@capacitor/core": "^7.0.0",
+    "@capacitor/ios": "^7.0.0",
+    "@capacitor/splash-screen": "^7.0.0",
+    "capacitor-nodejs": "https://github.com/mustang-im/Capacitor-NodeJS/releases/download/1.0.0-beta.10/capacitor-nodejs-v1.0.0-beta.10.tgz"
```
benbucksch:
```suggestion
    "capacitor-nodejs": "https://github.com/mustang-im/capacitor-nodejs/releases/download/1.0.0-beta.10/capacitor-nodejs-v1.0.0-beta.10.tgz"
```

mobile/docs/build.md
```diff
@@ -0,0 +1,18 @@
...
+4. `cd mobile; yarn install`
+5. `cd mobile/backend; yarn install`
+6. `cd lib; yarn add bufferutil` explicitly include bufferutil otherwise it might not be included
+7. `mkdir ~/.gyp && echo "{'variables':{'android_ndk_path':''}}" > ~/.gyp/include.gypi`
+8. `export NODE_OPTIONS="--max-old-space-size=32768"`
+9. `export MOBILE_ARCH=android-arm64`
+10. `cd mobile; yarn build`
+11. `npx cap open android` in `/mobile`
+12. Wait for the project to be fully loaded and Gradle Sync to finish running
+13. Go to `Build > Generate Signed App Bundle or APK` and follow the steps
```
benbucksch, line 18: Thank you for the docs!

mobile/capacitor.config.ts
```diff
@@ -0,0 +1,19 @@
...
+  appName: "Mustang",
```
benbucksch, line 5: Can you please also set the app icon correctly?

benbucksch (approved):
Great patch! Thanks for the awesome work.

Just a few small nits.

.github/workflows/installer-android.yml
jermy-c: Because bufferutil is an optional dependency for `ws` it doesn't always install. Or we can include in our `lib/package.json` and commit it. And the package is not included in the build which causes an error on android builds because it still tries to find it.

benbucksch: Yes, please simply add bufferutil to package.json, assuming it doesn't break any desktop OS.

benbucksch: Yes, please simply add bufferutil to package.json, assuming it doesn't break any desktop OS.

mobile/backend/prebuild.sh
benbucksch: If the env var is empty, it will simply be a space, so no problem.

mobile/build.sh
benbucksch: Jeremy says: sync does copy and updates the capacitory addons, and we don't need the latter.

mobile/android/app/build.gradle
```diff
@@ -16,11 +26,38 @@ android {
             ignoreAssetsPattern '!.svn:!.git:!.ds_store:!*.scc:.*:!CVS:!thumbs.db:!picasa.ini:!*~'
         }
     }
+    signingConfigs {
+        release {
+            keyAlias keystoreProperties['keyAlias']
+            keyPassword keystoreProperties['keyPassword']
+            storeFile file(keystoreProperties['storeFile'])
```
benbucksch: Move `keystoreProperties.load(new FileInputStream(keystorePropertiesFile))` above this line.
Then, try to put this block inside an `if (buildType == 'release') {`

.github/workflows/installer-android.yml
jermy-c: On mobile it's not included by default for reason but on the electron build it's included. Maybe it's because it's an optional dependency: https://github.com/websockets/ws?tab=readme-ov-file#opt-in-for-performance

mobile/capacitor.config.ts
benbucksch: Do this in followup PR

mobile/capacitor.config.ts
benbucksch: Do this in followup PR

benbucksch (approved):

benbucksch:
Very nice work, Jeremy! Well done.

PR #829, opened 2025-08-20
Fix TypeScript errors from review of #818

NeilRashbrook:
I'm also getting a weird error from typechecking `app/frontend/Meet/web/vite.config.ts` so I'd like to exclude it.


app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -75,7 +75,7 @@ export class OWAAccount extends MailAccount {
     assert(!this.hasLoggedIn, "Only for use during login");
     let url = this.url + 'service.svc';
     let options = { headers: { Action: "FindFolder" } };
-    let bodyJSON = Object.assign({}, owaFindFoldersRequest(false)); // Remove class before JPC, not needed for JSON
+    let bodyJSON = Object.assign({}, owaFindFoldersRequest()); // Remove class before JPC, not needed for JSON
```
benbucksch: The intent here was to make it **explicit** that we want want only a shallow listing in this particular case. A default param is too subtle.

Instead, change the definition of the function to `owaFindFoldersRequest(deep: boolean)`, please.

benbucksch (changes requested):

benbucksch (approved):
Thank you

PR #830, opened 2025-08-20
EWS: Support additional user-created calendars and addressbooks

NeilRashbrook:
I'm not sure what to name the user calendar and contact folders. EWS (and OWA) just used to name the default folder after the account, while EAS uses the name of the folder and doesn't include the account.

(If you ever want to support tasks, you could still create a dedicated `EWSCalendar` object with the `FolderID` of a Tasks folder, and it would be able to read those tasks.)

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -556,10 +556,10 @@ export class EWSAccount extends MailAccount {
     let haveAddressbook = appGlobal.addressbooks.some(addressbook => addressbook.mainAccount == this);
     if (!haveAddressbook) {
       for (let folder of ensureArray(result.RootFolder.Folders.ContactsFolder)) {
-        // EWS has some internal contacts folders.
-        // Fortunately they all have distinguished folder IDs,
-        // so we are only interested in the primary contacts folder
-        // or a user contact folder with no distinguished ID.
+        /* EWS has some internal contacts folders that we don't want to display.
+           Fortunately, they all have distinguished folder IDs,
+           so we are only interested in the primary user address book ("contacts")
+           or a additional user-created address books (which have no distinguished ID). */
         if (folder.FolderClass == "IPF.Contact" && [undefined, "contacts"].includes(folder.DistinguishedFolderId)) {
           let addressbook = newAddressbookForProtocol("addressbook-ews") as EWSAddressbook;
           addressbook.initFromMainAccount(this);
--- a/app/logic/Calendar/EWS/EWSCalendar.ts
+++ b/app/logic/Calendar/EWS/EWSCalendar.ts
@@ -12,10 +12,10 @@ import type { ArrayColl } from "svelte-collections";
 export class EWSCalendar extends Calendar {
   readonly protocol: string = "calendar-ews";
   declare readonly events: ArrayColl<EWSEvent>;
-  /** Exchange's calendar can only accept incoming invitations from its inbox */
-  readonly canAcceptAnyInvitation = false;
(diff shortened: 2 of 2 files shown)


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -535,51 +535,42 @@ export class EWSAccount extends MailAccount {
...
     for (let folder of ensureArray(result.RootFolder.Folders.CalendarFolder)) {
-      // Link (until #155) or create the default calendar.
-      // TODO: Support user-added calendars. Compare FolderId.
-      // N.B. Only default calendar can handle meeting requests and responses
-      // FolderClass will be IPF.Appointment except for Birthdays calendar
-      // which is IPF.Appointment.Birthdays
-      // Holidays calendar is read-only but it's hard to tell from the API...
-      if (folder.DistinguishedFolderId == "calendar") {
-        let calendar = appGlobal.calendars.find(calendar => calendar.mainAccount == this) as EWSCalendar | null;
+      if (["IPF.Appointment", "IPF.Appointment.Birthday"].includes(folder.FolderClass)) {
```
```suggestion
      if (folder.FolderClass == "IPF.Appointment") {
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -535,51 +535,42 @@ export class EWSAccount extends MailAccount {
...
         addressbook.icon ??= this.icon; // Migration, remove later
         addressbook.color ??= this.color;
-        addressbook.folderID ??= folder.FolderId.Id;
         await addressbook.save();
       }
     }
     for (let folder of ensureArray(result.RootFolder.Folders.CalendarFolder)) {
-      // Link (until #155) or create the default calendar.
-      // TODO: Support user-added calendars. Compare FolderId.
-      // N.B. Only default calendar can handle meeting requests and responses
```
benbucksch, line 578: We need to capture that somehow, and check that in the invitation code.

(Same for OWA)

NeilRashbrook: I could add a boolean flag to the default calendar. I can set the flag here on every startup, or would you prefer it to be in the extra JSON?

NeilRashbrook: I've just set the flag on every startup for now.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -553,45 +553,40 @@ export class EWSAccount extends MailAccount {
...
-      // TODO: Support user-added address books. Compare FolderId.
-      // FolderClass will be IPF.Contacts but some internal folders exist
-      let isMainAddressbook = folder.DistinguishedFolderId == "contacts";
-      if (isMainAddressbook) {
-        let haveAddressbook = appGlobal.addressbooks.find(addressbook => addressbook.mainAccount == this);
-        if (!haveAddressbook) {
+    let haveAddressbook = appGlobal.addressbooks.some(addressbook => addressbook.mainAccount == this);
+    if (!haveAddressbook) {
+      for (let folder of ensureArray(result.RootFolder.Folders.ContactsFolder)) {
+        if (folder.FolderClass == "IPF.Contact" && [undefined, "contacts"].includes(folder.DistinguishedFolderId)) {
```
benbucksch, line 563: [undefined, "contacts"].includes(folder.DistinguishedFolderId) needs a code comment why you do that. What you do filter out here?

NeilRashbrook:
I've fixed the bug whereby contacts and events were always being created in the user's default folder.

benbucksch (approved):

PR #838, opened 2025-08-22
OWA: Support additional user-created calendars and addressbooks

NeilRashbrook:
I'm not sure what to name the user calendar and contact folders. OWA (and EWS) just used to name the default folder after the account, while EAS uses the name of the folder and doesn't include the account.

(If you ever want to support tasks, you could still create a dedicated OWACalendar object with the FolderID of a Tasks folder, and it would be able to read those tasks.)

### DIFF
--- a/app/logic/Mail/OWA/OWAAccount.ts
+++ b/app/logic/Mail/OWA/OWAAccount.ts
@@ -159,15 +159,15 @@ export class OWAAccount extends MailAccount {
     let haveAddressbook = appGlobal.addressbooks.some(addressbook => addressbook.mainAccount == this);
     if (!haveAddressbook) {
       let response = await this.callOWA(new OWAGetPeopleFiltersRequest());
-      for (let filter of response) {
+      for (let ab of response) {
         // Exclude internal contacts folders.
-        if (filter.IsReadOnly || !filter.FolderId?.Id) {
+        if (ab.IsReadOnly || !ab.FolderId?.Id) {
           continue;
         }
         let addressbook = newAddressbookForProtocol("addressbook-owa") as OWAAddressbook;
         addressbook.initFromMainAccount(this);
-        addressbook.name = `${this.name} ${filter.DisplayName}`;
-        addressbook.folderID = filter.FolderId.Id;
+        addressbook.name = `${this.name} ${ab.DisplayName}`;
+        addressbook.folderID = ab.FolderId.Id;
         appGlobal.addressbooks.add(addressbook);
         addressbook.save();
       }
@@ -352,7 +352,7 @@ export class OWAAccount extends MailAccount {
         }
         owaFolder.fromJSON(folder);
         this.folderMap.set(folder.FolderId.Id, owaFolder);
-      } else if (!haveCalendar && folder.FolderClass == "IPF.Appointment") {
+      } else if (folder.FolderClass == "IPF.Appointment" && !haveCalendar) {
         let calendar = newCalendarForProtocol("calendar-owa") as OWACalendar;
         calendar.initFromMainAccount(this);
         if (folder.DistinguishedFolderId == "calendar") {


app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -293,30 +299,24 @@ export class OWAAccount extends MailAccount {
...
-            calendar.icon = this.icon;
-            calendar.color = this.color;
-            calendar.mainAccount = this;
-            appGlobal.calendars.add(calendar);
-          }
-          calendar.icon ??= this.icon; // Migration, remove later
-          calendar.color ??= this.color;
-          calendar.folderID ??= folder.FolderId.Id;
-          await calendar.save();
+      } else if (["IPF.Appointment", "IPF.Appointment.Birthday"].includes(folder.FolderClass)) {
```
```suggestion
      } else if (folder.FolderClass == "IPF.Appointment") {
```

NeilRashbrook:
Still hoping you'll comment one way or another on my addressbook/contacts display naming convention...

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -150,32 +124,36 @@ export class OWAAccount extends MailAccount {
...
-      await addressbook.save();
-      appGlobal.addressbooks.add(addressbook);
-    }
-
-    for (let addressbook of appGlobal.addressbooks) {
-      if (addressbook.mainAccount == this) {
-        addressbook.listContacts()
-          .catch(this.errorCallback);
+    let response = await this.callOWA(new OWAGetPeopleFiltersRequest());
+    for (let filter of response) {
```
```suggestion
    for (let ab of response) {
```

Please name variables after what they contain

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -150,32 +124,36 @@ export class OWAAccount extends MailAccount {
...
+      if (filter.IsReadOnly) {
+        continue;
       }
+      let addressbook = appGlobal.addressbooks.find((addressbook: OWAAddressbook) => addressbook.mainAccount == this && addressbook.folderID == filter.FolderId.Id) as OWAAddressbook | undefined;
+      if (!addressbook) {
```
benbucksch: What we want to do here (for now, until we have proper UI for it): If there are no address books or calendars for this account at all, then add them. If we have any, don't add any more, i.e. respect when the user deleted it.

Long-term, we want a function that lists all address book and another that lists all calendars. And a UI in setup and settings that allows the user to pick which ones he wants to see in Mustang. Only the latter will then be created within Mustang. This function and UI would be generic and work the same for all Exchange protocols and also for JMAP.

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -13,11 +13,13 @@ import { ArrayColl } from "svelte-collections";
...
-  /** Exchange FolderID for this addressbook. Not DistinguishedFolderId */
-  folderID: string;
-  declare readonly events: ArrayColl<OWAEvent>;
+  readonly events: ArrayColl<OWAEvent>;
   /** Exchange's calendar can only accept incoming invitations from its inbox */
   readonly canAcceptAnyInvitation = false;
+  /** Exchange FolderID for this calendar. Not DistinguishedFolderId */
+  folderID: string;
+  /** Is this the default calendar that handles incoming invitations */
+  isInvitationCalendar: boolean = false;
```
benbucksch:
```suggestion
  useForInvitations: boolean = false;
```

(not "used")

app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -13,11 +13,13 @@ import { ArrayColl } from "svelte-collections";
 
 export class OWACalendar extends Calendar {
   readonly protocol: string = "calendar-owa";
-  /** Exchange FolderID for this addressbook. Not DistinguishedFolderId */
-  folderID: string;
-  declare readonly events: ArrayColl<OWAEvent>;
+  readonly events: ArrayColl<OWAEvent>;
   /** Exchange's calendar can only accept incoming invitations from its inbox */
   readonly canAcceptAnyInvitation = false;
```
benbucksch: move this below `folderID` and above `useForInvitations`, as it logically belongs to the latter.

app/logic/Contacts/OWA/OWAAddressbook.ts
```diff
@@ -32,28 +32,14 @@ export class OWAAddressbook extends Addressbook {
   }
 
   async listContactsSlow() {
-    let response = await this.account.callOWA(new OWAGetPeopleFiltersRequest());
-    let contacts = response.find(filter => !filter.IsReadOnly);
-    if (!contacts) {
-      this.persons.clear();
-      this.groups.clear();
```
benbucksch, line 39: The old code assumes that it starts with an empty address book. Did you verify that this is not a problem in the new code?

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: OWA calls the request that it uses to enumerate addressbooks "GetPeopleFilters". I have no idea why. So I named the variable after the request.

app/logic/Calendar/OWA/OWACalendar.ts
NeilRashbrook: Depends on your logic. By my logic, these are all inherited from the base class, so they belong together.

app/logic/Calendar/OWA/OWACalendar.ts
NeilRashbrook: This the calendar that is used for invitations...

app/logic/Contacts/OWA/OWAAddressbook.ts
NeilRashbrook: No, it doesn't. This code just clears the contacts if it can't find an address book.

app/logic/Calendar/OWA/OWACalendar.ts
NeilRashbrook: What I'm trying to say is that `useForInvitations` contains an action, so it sounds like the name of a method rather than a boolean property.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -156,14 +156,21 @@ export class OWAAccount extends MailAccount {
...
-      appGlobal.addressbooks.add(addressbook);
+      let response = await this.callOWA(new OWAGetPeopleFiltersRequest());
+      for (let filter of response) {
+        // Exclude internal contacts folders.
+        if (filter.IsReadOnly || !filter.FolderId?.Id) {
+          continue;
+        }
+        let addressbook = newAddressbookForProtocol("addressbook-owa") as OWAAddressbook;
+        addressbook.initFromMainAccount(this);
+        addressbook.name = `${this.name} ${filter.DisplayName}`;
```
NeilRashbrook: Previously we were assuming that the first viable response "filter" was the default addressbook. Should I continue to do that, so that the name is only set on subsequent responses?

benbucksch: I don't understand how "the name is only set on subsequent responses" relates to "the default addressbook"

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -156,14 +156,21 @@ export class OWAAccount extends MailAccount {
 
     // `listFolders()` will subscribe to new user-added calendars
 
-    // Link (until #155) or create the default address book.
-    // TODO: Support user-added address books. Compare addressbook ID.
-    let haveAddressbook = appGlobal.addressbooks.find(addressbook => addressbook.mainAccount == this);
+    let haveAddressbook = appGlobal.addressbooks.some(addressbook => addressbook.mainAccount == this);
```
benbucksch, line 159: I don't understand why this code is here, whereas the calendar code is in the listFolders() function. Is OWA listing calendars as folders, but not listing addressbooks as folders, whereas EWS lists both as folders?

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: I fixed it

app/logic/Calendar/OWA/OWACalendar.ts
benbucksch: "Should I use this for invitations?" "Yes, use this for invitations"

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: Correct, OWA does list calendars as folders but it a special snowflake when it comes to addressbooks.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: The previous code used to only add the first address book, using the account name.
The current code sets all addressbooks to `{accountname} {abname}`, but I was thinking you might want the first one to be just the same as the account name?

benbucksch: That is correct. The name of the primary personal addressbook should match the name of the mail account.

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: OK, thanks for clarifying that for me

Issue #835, opened 2025-08-21
Android: Build: 16 KB alignment


benbucksch:
<img width="423" height="233" alt="Image" src="https://github.com/user-attachments/assets/5fe69e40-b35b-457e-aa8e-eafc3a83eb4d" />

PR #839, opened 2025-08-23
Android build fixes

jermy-c:
- Updated some dependency versions
- Moved Android envs from workflow file to shell scripts for local build
- Fix UI not set to mobile
- Fix `.aab` build naming
- Fix debug build not working
- Enforce ndkVersion 28 for the project, fixes some parts #835 for the node native modules
- Enforce the same CMake version for the project
- Clear out remains of previous build in the build script

### DIFF
--- a/mobile/android/build.gradle
+++ b/mobile/android/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-
+    
     repositories {
         google()
         mavenCentral()


mobile/android/variables.gradle
```diff
@@ -1,16 +1,18 @@
 ext {
-    minSdkVersion = 23
+    ndkVersion = '28.2.13676358'
```
benbucksch, line 2: You need to state where these values come from. Are these the what I have installed locally, or are these the versions that the node.js-mobile prebuilds have been compiled with? If the latter, please state so clearly, and also tell where in node.js-mobile *exactly* to get the correct versions from.

.github/workflows/installer-android.yml
```diff
@@ -35,17 +38,15 @@ jobs:
...
       env:
-        ANDROID_LIBNODE: "https://github.com/mustang-im/nodejs-mobile/releases/download/v22.9.0/nodejs-mobile-v22.9.0-android.zip"
-        ANDROID_SDK: --sdk35
         MOBILE_ARCH: 'android-arm64'
 
     - run: |
-        mkdir -p ${{ github.workspace }}//private_keys/
-        echo '${{ secrets.KEYSTORE }}' | base64 --decode > ${{ github.workspace }}/private_keys/key.jks
+        mkdir -p ~/private_keys/
+        echo '${{ secrets.KEYSTORE }}' | base64 --decode > ~/private_keys/key.jks
```
benbucksch, line 45: Why this change?

* Is this change still needed, given that the signing works?
* Why not use github workspace? This is the github action script, so it should be set, no?

jermy-c: > Is this change still needed, given that the signing works?

No, it's not needed. It was just for the script to look nicer and possibly for the workflow to be more secure because github workspace is the repo folder which might get checked into git.

> Why not use github workspace? This is the github action script, so it should be set, no?

Because github workspace is the repo folder which might get checked into git.

mobile/android/variables.gradle
jermy-c: - Prebuilds use the NDK set by the ANDROID_NDK_HOME variable to compile the NDK
- The nodejs-mobile is compiled already by the NDK version cannot be changed by our build
- build.gradle would download the version specified if it's not present and this where the version mismatch might happen since it doesn't use the ANDROID_NDK_HOME variable
- Possible solution: Set ANDROID_NDK_HOME to this variable, but we still need to be aware of the nodejs-mobile NDK version it was compiled with there's no way to tell from the libnode.so binary

benbucksch: I guess I don't understand why these are hardcoded here.

PR #840, opened 2025-08-23
Mobile: Set to mobile UI before build

jermy-c:
- Set to mobile UI before building


mobile/package.json
```diff
@@ -7,11 +7,12 @@
     "start": "vite",
     "start:android": "cap run android",
     "start:ios": "cap run ios",
-    "build": "./build.sh",
+    "build": "npm run setup:ui && ./build.sh",
```
benbucksch: Is there a reason to not move this into build.sh?

mobile/package.json
```diff
@@ -7,11 +7,12 @@
...
     "build:android": "npm run build:android:apk && npm run build:android:aab",
     "build:android:apk": "cap sync android && cd android && ./gradlew --configuration-cache assembleRelease",
     "build:android:aab": "cap sync android && cd android && ./gradlew --configuration-cache :app:bundleRelease",
     "preview": "vite preview",
+    "setup:ui": "vite-node ./hooks/common/mobile-ui.ts",
```
benbucksch: ditto

mobile/package.json
jermy-c: No, there isn't. I'll fix that.

benbucksch: Also, isn't it easier to do this as python or perl script, like we do the package.json changes? See app/build/release.sh

mobile/.gitignore
```diff
@@ -9,4 +9,5 @@ dist/
 !build.sh
 NodeMobile.xcframework
 !setup-ios.sh
+!hooks/common/mobile-ui.sh
```
benbucksch: Is this a leftover?

benbucksch:
```suggestion
```

jermy-c: Yes, it was. Sorry.

benbucksch (approved):

Issue #842, opened 2025-08-25
Calendar: Slow to load


benbucksch:
### Reproduction
* You have a few hundred events in multiple calendars, including recurring event instances. (For testing, you can use local calendars.)
* Switch from mail (or chat or any other app) to calendar

### Actual result
* Before the calendar UI appears, you see a delay/freeze of ca. 500ms or more, depending on the speed of your computer and the number of events.

### Expected result
* We are very snappy. 20 to 50 ms.

### Solution
* First, locate which function is slow
* Fix it
* You may trade a little memory for a lot of speed. (Note the tradeoffs "little" and "a lot".)
* If we are doing stupid busy work somewhere, eliminate it.

benbucksch:
Feel free to do the same for other apps, if you find performance bottlenecks elsewhere.

benbucksch:
I think we fixed that

PR #846, opened 2025-08-25
Pre-filter event collections in the day and month views #842

NeilRashbrook:
As mentioned in #842, if I have 11870 events (recurring instances) then that results in 2991240 subscriptions by `filterObservable` and the Calendar first time open is three seconds; with these extra filters, this reduces to 110972 subscriptions and the Calendar first time open time reduces to under half a second.

### DIFF
--- a/app/frontend/Calendar/DayView/EventsBlockCell.svelte
+++ b/app/frontend/Calendar/DayView/EventsBlockCell.svelte
@@ -1,9 +1,7 @@
 <vbox flex class="events" on:dblclick={addEvent}>
   {#if $displayEvents && !$displayEvents.isEmpty}
     {#each $displayEvents.each as event (event.id)}
-      {#if event.startTime && event.endTime}
-        <EventBlock {event} {start} {end} conflicts={displayEvents} />
-      {/if}
+      <EventBlock {event} {start} {end} conflicts={displayEvents} />
     {/each}
   {/if}
 </vbox>
--- a/app/frontend/Calendar/DayView/EventsBlockCell.svelte
+++ b/app/frontend/Calendar/DayView/EventsBlockCell.svelte
@@ -28,7 +28,7 @@
   function setEnd() {
     end = new Date(start);
     end.setHours(end.getHours() + intervalInHours);
-    displayEvents = events.filterObservable(ev => ev.startTime < end && ev.endTime > start && !ev.allDay);
+    displayEvents = events.filterObservable(ev => ev.startTime < end && ev.endTime > start);
   }
 
   function addEvent() {
--- a/app/frontend/Calendar/DayView/EventsBlockCell.svelte
+++ b/app/frontend/Calendar/DayView/EventsBlockCell.svelte
@@ -26,7 +26,7 @@
   function setEnd() {
     end = new Date(start);
     end.setHours(end.getHours() + intervalInHours);
-    displayEvents = events.filterObservable(ev => ev.startTime < end && ev.endTime > start);
+    displayEvents = events.filterObservable(ev => ev.startTime < end && ev.endTime > start && !ev.allDay);
   }
 
   function addEvent() {
--- a/app/frontend/Calendar/DayView/WeekView.svelte
+++ b/app/frontend/Calendar/DayView/WeekView.svelte
@@ -72,9 +72,7 @@
   $: start, setStartTimes();
   function setStartTimes() {
     let startTime = new Date(start);
-    startTime.setMinutes(0);
-    startTime.setSeconds(0);
-    startTime.setMilliseconds(0);
+    startTime.setMinutes(0, 0, 0);
     startTimes = [];
     for (let i = startHour; i < endHour; i += intervalHour) {
       startTime.setHours(i);
@@ -91,10 +89,7 @@
     if (showDays > 3) {
       startTime.setDate(startTime.getDate() - 1);
     }
-    startTime.setHours(startHour);
-    startTime.setMinutes(0);
-    startTime.setSeconds(0);
-    startTime.setMilliseconds(0);
+    startTime.setHours(startHour, 0, 0, 0);
     let filterStart = new Date(startTime);
     days = [];
     for (let i = 0; i < showDays; i++) {
--- a/app/frontend/Calendar/DayView/WeekView.svelte
+++ b/app/frontend/Calendar/DayView/WeekView.svelte
@@ -26,7 +26,7 @@
         {/each}
         {#each startTimes as start}
           <TimeLabel time={start} />
-          <TimeDayRow {days} time={start} {events} />
+          <TimeDayRow {days} time={start} events={visibleEvents} />
         {/each}
       </grid>
     </Scroll>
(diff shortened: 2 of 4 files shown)


app/frontend/Calendar/DayView/WeekView.svelte
```diff
@@ -83,28 +83,24 @@
...
     days = [];
     for (let i = 0; i < showDays; i++) {
       days.push(new Date(startTime));
-      startTime.setDate(startTime.getDate() + 1)
+      startTime.setDate(startTime.getDate() + 1);
     }
-  }
-
-  let allDayEvents: Collection<Event>;
-  $: start, $events, setAllDayEvents();
```
benbucksch, line 104: The trigger conditions for `setDays()` and `setAllDayEvents()` are different.
* Why would you re-calculate `days` when `$events` change?
* Smaller functions are better than larger functions.

app/frontend/Calendar/DayView/WeekView.svelte
```diff
@@ -83,28 +83,24 @@
...
   function setDays() {
     let startTime = new Date(start);
     if (showDays > 3) {
       startTime.setDate(startTime.getDate() - 1);
     }
-    startTime.setHours(startHour);
-    startTime.setMinutes(0);
-    startTime.setSeconds(0);
-    startTime.setMilliseconds(0);
+    let filterStart = startTime.setHours(startHour, 0, 0, 0);
```
benbucksch:
```suggestion
    startTime.setHours(startHour, 0, 0, 0);
    let filterStart = startTime.clone();
```

Do not combine 2 unrelated actions into the same line.
Readability over shortness.

clone(): See comment below

app/frontend/Calendar/MonthView/MonthView.svelte
```diff
@@ -52,11 +53,14 @@
       }
     }
     let startTime = weekDays[0]; // Always start with Monday
+    let filterStart = startTime.getTime();
```
benbucksch:
```suggestion
    let filterStart = startTime.clone();
```

Otherwise you compare timestamp with `Date` object below. While that works with coalasing, it's uh.

benbucksch (changes requested):

app/frontend/Calendar/DayView/WeekView.svelte
NeilRashbrook: I don't actually recalculate anything because `$events` can't actually change.

app/frontend/Calendar/MonthView/MonthView.svelte
NeilRashbrook: Oh, I see, TypeScript will let you compare `date1 < date2` or `date1.getTime() < date2.getTime()` but not a mixture. I didn't know that.

app/frontend/Calendar/DayView/WeekView.svelte
NeilRashbrook: The existing code is actually buggy here; it gets the all day events for a different range. Merging the functions fixes that.

benbucksch: > $events can't actually change

How so? It would help me if you could explain why.

app/frontend/Calendar/DayView/WeekView.svelte
```diff
@@ -93,18 +95,16 @@
     startTime.setMinutes(0);
     startTime.setSeconds(0);
     startTime.setMilliseconds(0);
```
benbucksch: startTime.setHours(startHour, 0, 0, 0);

(as you did elsewhere)

ditto above startTime.setMinutes(0, 0, 0);

app/frontend/Calendar/DayView/WeekView.svelte
benbucksch: > gets the all day events for a different range

Which other range, and where is the bug in the code? And can it easily (!) be fixed in the existing function?

app/frontend/Calendar/MonthView/MonthView.svelte
benbucksch: >  TypeScript will let you compare date1 < date2 or date1.getTime() < date2.getTime() but not a mixture.

Even if TypeScript allowed it, it wouldn't be a good idea :)

app/frontend/Calendar/DayView/WeekView.svelte
```diff
@@ -91,13 +91,17 @@
     if (showDays > 3) {
       startTime.setDate(startTime.getDate() - 1);
     }
-    let filterStart = startTime.setHours(startHour, 0, 0, 0);
+    startTime.setHours(startHour);
+    startTime.setMinutes(0);
+    startTime.setSeconds(0);
+    startTime.setMilliseconds(0);
```
benbucksch:
```suggestion
    startTime.setHours(startHour, 0, 0, 0);
```

Why did you revert that? That was great.

app/frontend/Calendar/DayView/WeekView.svelte
NeilRashbrook: > > $events can't actually change
> 
> How so? It would help me if you could explain why.

`$events` is always `appGlobal.calendarEvents`. Even before then, it was set once in `CalendarApp.svelte` and just inherited down everywhere. (Its contents can change, but `filterObservable` already expects that of course.)

> > gets the all day events for a different range
> 
> Which other range, and where is the bug in the code? And can it easily (!) be fixed in the existing function?

`setAllDayEvents` uses the range of `start` to `start` plus the given number of days. but it doesn't adjust `start` to  midnight, nor does it subtract one day when showing a whole week, so it could be incorrect by up to 2 days.

benbucksch: > $events is always appGlobal.calendarEvents

Yes, but the `$` subscribes to changes, so whenever calendar events are added or removed, this triggers and the `setDays()` function would now run, which is unnecessary. Given that `days` is *not* dependent on events, and changing `days` triggers a refresh of the entire UI, it should not change every time an event is added or removed.

benbucksch: That's why these 2 functions were separate. Could you please keep them separate?

If not too difficult, re-calculate the correct start and end time in `setAllDayEvents()`, fixing the bug that you mentioned, but without merging these 2 functions, please.

benbucksch: Oh! I was assuming that you need to add `$events` as trigger for `setDays()`, as it was for `setAllDayEvents()`, to re-calculate the `visibleEvents` etc.

However, `visibleEvents` is created by `filterObservable()`, which already observes the source `events` collection and refreshes by itself. I didn't consider that. So, `setAllDayEvents()` doesn't need to re-run.

Sorry for my confusion.

app/frontend/Calendar/DayView/EventsBlockCell.svelte
```diff
@@ -28,7 +26,7 @@
   function setEnd() {
     end = new Date(start);
     end.setHours(end.getHours() + intervalInHours);
-    displayEvents = events.filterObservable(ev => ev.startTime < end && ev.endTime > start && !ev.allDay);
+    displayEvents = events.filterObservable(ev => ev.startTime < end && ev.endTime > start);
```
benbucksch: I don't understand this change.

Esp. given that we use `displayEvents` as `conflicts`, and `conflicts` assumes that it really contains only those events that are actually overlapping.

An all-day event is *not* conflicting with a time event. E.g. "Father's day" is not a conflict with "14:00 Go fishing with my son".

```suggestion
    displayEvents = events.filterObservable(ev => ev.startTime < end && ev.endTime > start && !ev.allDay);
```

If you are sure that this change is correct, please create a new PR with it, and explain why.

benbucksch (approved):

app/frontend/Calendar/DayView/WeekView.svelte
NeilRashbrook: If there was a way to tell the collection to re-test all of the events against the filter function we could optimise `visibleEvents` further; the filter function would check against the widget's "globals" and `setDays` would just adjust those and ask the collection to do a re-test, thus avoiding having to resubscribe to everything.

app/frontend/Calendar/DayView/EventsBlockCell.svelte
NeilRashbrook: The events we're displaying in an `EventsBlockCell` already aren't all-day events...

benbucksch: OK, but you're assuming what the caller does. Remember that each component does not trust the others (unless explicitly documented, but only as last resort).

Please assume that callers *will* change. This is important to create a robust system that can later be modified without constant regressions. It is very likely that a lot of this code will change, so such assumptions will kill.

Given that the check is very cheap, can you please leave it in?

benbucksch: I see that this was already fixed.

benbucksch: Compare PR #855

Issue #569, opened 2025-04-21
Calendar: Event: Missing startTime or endTime


benbucksch:
All `Event`s must have `startTime` and `endTime`. Yet, one of them is undefined during `save()`.

See this report:

- ‼️ **TypeError: Cannot read properties of undefined (reading 'toISOString')** `SQLEvent.save(../../../../app/logic/Calendar/SQ...` [View Issue](https://mustang-jq.sentry.io/issues/37935807/?referrer=github-pr-bot)

_Originally posted by @sentry-io[bot] in https://github.com/mustang-im/mustang/issues/560#issuecomment-2817829454_

```
../../../../app/logic/Calendar/SQL/SQLEvent.ts in SQLEvent.save at line 28:63
In App
            calUID, responseToOrganizer, pID, calendarID,
            recurrenceRule, recurrenceMasterEventID,
            recurrenceStartTime, recurrenceIsException
          ) VALUES (
            ${event.title}, ${event.descriptionText}, ${event.descriptionHTML},
-->         ${event.startTime.toISOString()}, ${event.endTime.toISOString()}, ${event.allDay ? 1 : 0},
            ${event.calUID}, ${event.response}, ${event.pID}, ${event.calendar?.dbID},
            ${event.recurrenceRule?.getCalString(event.allDay)}, ${event.parentEvent?.dbID},
            ${event.recurrenceStartTime?.toISOString()},
            ${+!!event.parentEvent?.dbID}
              )`);

../../../../app/logic/Calendar/SQL/SQLCalendarStorage.ts in SQLCalendarStorage.saveEvent at line 15:11
In App ../../../../app/logic/Calendar/Event.ts in Event$1.Event.save at line 237:5
In App ../../../../app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts in responseFunc at line 145:17
In App ../../../../app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts in ActiveSyncCalendar.makeSyncRequest at line 97:15
In App ../../../../app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts in ActiveSyncCalendar.listEvents at line 120:5
In App ../../../../app/logic/Mail/ActiveSync/ActiveSyncAccount.ts in ActiveSyncAccount.login at line 118:9
In App ../../../../app/logic/startup.ts in emailAccountLogin at line 65:9
In App
```

benbucksch:
[...]

PR #848, opened 2025-08-25
Optimise event conflict display #842

NeilRashbrook:
Unless I'm missing something, `EventBlock.conflicts` is just the same as `EventsBlockCell.displayEvents`, isn't it?

### DIFF
--- a/app/frontend/Calendar/DayView/EventsBlockCell.svelte
+++ b/app/frontend/Calendar/DayView/EventsBlockCell.svelte
@@ -1,9 +1,7 @@
 <vbox flex class="events" on:dblclick={addEvent}>
   {#if $displayEvents && !$displayEvents.isEmpty}
     {#each $displayEvents.each as event (event.id)}
-      {#if event.startTime && event.endTime}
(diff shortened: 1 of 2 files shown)


app/frontend/Calendar/DayView/EventsBlockCell.svelte
```diff
@@ -1,9 +1,7 @@
 <vbox flex class="events" on:dblclick={addEvent}>
   {#if $displayEvents && !$displayEvents.isEmpty}
     {#each $displayEvents.each as event (event.id)}
-      {#if event.startTime && event.endTime}
```
benbucksch, line 4: This was a workaround for #569

NeilRashbrook: I moved the check into `WeekView` in #846 but I couldn't remove this line there due to the merge conflict.

benbucksch: > I moved the check into WeekView in https://github.com/mustang-im/mustang/pull/846

I don't see the check there. Make another PR for that, if appropriate, please?

PR #849, opened 2025-08-25
Optimise recurrence rule editing #842

NeilRashbrook:
- Changing the frequency to None wasn't collapsing the repeat box
- Even when it does, sometimes `updateDateUI` runs first; in this case we don't want the update, as the rule would be invalid
- Even when `updateDateUI` is supposed to run, the rule may already be valid, so don't update it, as that will regenerate the instances
- When first editing the event, don't bother generating the instances on the unedited copy, instead just shallow clone them

### DIFF
--- a/app/frontend/Calendar/EditEvent/RepeatBox.svelte
+++ b/app/frontend/Calendar/EditEvent/RepeatBox.svelte
@@ -109,7 +109,8 @@
 
   $: $event.startTime, $event.duration, updateDateUI();
   function updateDateUI() {
-    // This should close the repeat box, but this update might run first.
+    // Setting the frequency to None should close the repeat box,
+    // but Svelte might decide to run this function again first.
     if (frequency == Frequency.None) {
       return;
     }
--- a/app/frontend/Calendar/EditEvent/RepeatBox.svelte
+++ b/app/frontend/Calendar/EditEvent/RepeatBox.svelte
@@ -109,6 +109,11 @@
 
   $: $event.startTime, $event.duration, updateDateUI();
   function updateDateUI() {
+    // This should close the repeat box, but this update might run first.
+    if (frequency == Frequency.None) {
+      return;
+    }
+
     master.startEditing();
     master.startTime = event.startTime;
     master.duration = event.duration;
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -334,7 +334,10 @@ export class Event extends Observable {
         init.weekdays = [this.startTime.getDay()]; // e.g. 3rd Wednesday of month
       }
     }
-    this.recurrenceRule = new RecurrenceRule(init);
+    let recurrenceRule = new RecurrenceRule(init);
+    if (!recurrenceRule.timesMatch(this.recurrenceRule)) {
+      this.recurrenceRule = recurrenceRule;
+    }
   }
 
   /** Create a new instance of the same event.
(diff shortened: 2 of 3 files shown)


app/frontend/Calendar/EditEvent/RepeatBox.svelte
```diff
@@ -109,6 +109,11 @@
 
   $: $event.startTime, $event.duration, updateDateUI();
   function updateDateUI() {
+    // This should close the repeat box, but this update might run first.
```
benbucksch: "This" "this"?

Unclear what the first and second "this" refer to.

app/logic/Calendar/Event.ts
```diff
@@ -442,8 +445,11 @@ export class Event extends Observable {
       return;
     }
     this.timezone ||= myTimezone();
-    this.unedited = this.calendar.newEvent();
+    // Use a raw event to avoid automatic instance generation.
```
benbucksch, line 448: Why does a raw `Event` not generate instances? It should!

Can you find another way which is more explicit?

app/logic/Calendar/Event.ts
```diff
@@ -442,8 +445,11 @@ export class Event extends Observable {
...
+    this.unedited = new Event();
     this.unedited.copyFrom(this);
+    // Instead, just shallow clone our instances (needed for `seriesStatus`)
```
benbucksch, line 451: Unclear how that interacts with all our other recurrence editing and generating and saving code.

benbucksch:
@NeilRashbrook ping

benbucksch (changes requested):
Duh, I didn't send off the review.

app/frontend/Calendar/EditEvent/RepeatBox.svelte
NeilRashbrook: Sorry; for the first "this" I was referring to the frequency being None, while the second "this" refers to this function.

app/logic/Calendar/Event.ts
NeilRashbrook: A raw event doesn't have a calendar, so we don't bother generating instances. The other way is to explicitly set the calendar to null.

app/logic/Calendar/Event.ts
NeilRashbrook: Well, we shouldn't be saving the unedited version...

app/frontend/Calendar/EditEvent/RepeatBox.svelte
benbucksch: Thanks. Can you please fix the comment?

app/logic/Calendar/Event.ts
benbucksch: As mentioned, a raw `Event` *should* generate instances. We may need that for various circumstances in the future. At the very least, please assume that they do.

In general, never rely on assumptions about how other classes, components or other parts of the code are implemented. This leads to fragile code and is a big no-no in Mustang. Each class and component must depend only on the public API of other components, in form of functions and properties, and not on their specific and current implemention behavior. (And this here is nothing you can just document in a code comment, because it will likely change.)

This PR cannot land with this assumption.

app/logic/Calendar/Event.ts
benbucksch: Not sure what "Instead" refers to? What's the alternative that we intentionally don't do?

benbucksch (changes requested):

app/logic/Calendar/Event.ts
NeilRashbrook: > other classes

Then it's a good thing I'm making assumptions about how this class is implemented. Also, we already have other events that don't generate instances.

PR #850, opened 2025-08-25
Android build: Clear cached web assets

jermy-c:
- Remains of the previous builds were found in the Android build because we didn't clear out the `/dist` folder
- The build script now also deletes existing libnodes so always get the one in URL param for the release builds only

### DIFF
--- a/mobile/hooks/common/build.sh
+++ b/mobile/hooks/common/build.sh
@@ -1,23 +1,6 @@
 #!/usr/bin/env bash
 
-mkdir -p ../mobile/dist
-
 export NODE_OPTIONS="--max-old-space-size=32768"
-
-# Clear LIBNODE cache for release build
-is_release=false
-
-for arg in "$@"; do
-  if [[ "$arg" == "--release" ]]; then
-    is_release=true
-    break
-  fi
-done
-
-if "$is_release"; then
-  rm -rf backend/node_modules/prebuild-for-nodejs-mobile/android/libnode
-  rm -rf node_modules/capacitor-nodejs/android/libnode
-fi
 
 # Setup mobile UI
 perl -p -i \
@@ -26,7 +9,7 @@ perl -p -i \
 
 # Build and copy assets
 (cd ../app && npm run build) &&
-(cd ./dist && rm -rf * .[!.]* ..?*) &&
+(rm -rf ./dist; mkdir -p ./dist) &&
 (cd ../app && cp -R ./dist/* ../mobile/dist) &&
 (cd backend && npm run build) &&
 npx cap copy
--- a/mobile/hooks/common/build.sh
+++ b/mobile/hooks/common/build.sh
@@ -4,13 +4,31 @@ mkdir -p ../mobile/dist
 
 export NODE_OPTIONS="--max-old-space-size=32768"
 
+# Clear LIBNODE cache for release build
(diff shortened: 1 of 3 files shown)


mobile/hooks/common/build.sh
```diff
@@ -4,12 +4,30 @@ mkdir -p ../mobile/dist
...
+
 # Setup mobile UI
 perl -p -i \
   -e "s|isMobile = false;$|isMobile = true;|;" \
   ../app/logic/build.ts
 
-(cd ../app && npm run build && cp -R ./dist/* ../mobile/dist) &&
+# Build and copy assets
+(cd ../app && npm run build) &&
+(cd ./dist && rm -rf * .[!.]* ..?*) &&
```
benbucksch: As mentioned in the other PR: `rm -rf .[!.]* ..?*)` is too dangerous. Please be more specific about what you want to delete, esp. for dotfiles.

mobile/package.json
```diff
@@ -9,8 +9,8 @@
     "start:ios": "cap run ios",
     "build": "./hooks/common/build.sh",
     "build:android": "npm run build:android:apk && npm run build:android:aab",
-    "build:android:apk": "export MOBILE_ARCH=android-arm64; npm run build && cap sync android && cd android && ./gradlew --configuration-cache assembleRelease",
-    "build:android:aab": "export MOBILE_ARCH=android-arm64; npm run build && cap sync android && cd android && ./gradlew --configuration-cache :app:bundleRelease",
+    "build:android:apk": "export MOBILE_ARCH=android-arm64;npm run build -- --release && cap sync android && cd android && ./gradlew --configuration-cache assembleRelease",
```
benbucksch: I don't think we should delete the libnode cache on every run.

mobile/hooks/common/build.sh
```diff
@@ -4,12 +4,30 @@ mkdir -p ../mobile/dist
...
+for arg in "$@"; do
+  if [[ "$arg" == "--release" ]]; then
+    is_release=true
+    break
+  fi
+done
+
+if "$is_release"; then
+  rm -rf backend/node_modules/prebuild-for-nodejs-mobile/android/libnode
+  rm -rf node_modules/capacitor-nodejs/android/libnode
```
benbucksch: What is re-downloading them?

I'm not sure we need this, given that we build releases on CI, and CI starts clean every time.

The libnode changes only when upstream made a new release, right? Shouldn't npm already handle this?

benbucksch: Also, I think this should be a separate script. (If we need it at all).

jermy-c: > What is re-downloading them?

The modified `prebuild-for-nodejs-mobile` and `capacitor-nodejs` download them whenever the folder is empty.

> The libnode changes only when upstream made a new release, right?

The libnode changes whenever, you change the `ANDROID_LIBNODE` env for prebuilds and `androidLibNode` for capacitor-nodejs. But we could remove it.

> Shouldn't npm already handle this?

No, npm doesn't handle it. It's the hooks that I placed in the packages that handle downloading the libnode. Before they were included in the package which forced you to use that specific version on CI because you can't manually download the libnode on CI.

benbucksch (approved):
Much better, thanks.

PR #852, opened 2025-08-26
Android: Icons for Mustang and Parula

jermy-c:
- Removed the template icon and splash screen
- Place the Mustang icons as the default icon for the project
- Added a script to change the app name to Parula and change the icon
- There's a color border around the icon

image
image


benbucksch:
> There's a color border around the icon

Can you please remove that border?

benbucksch:
@jermy-c The Parula bird seems a little too much on the left. Otherwise, it's OK.

If Elena has better logos, we can improve them later.

elenaux:
p light
p dark
m light
m dark
dark version is #1B1A1E

benbucksch:
elenaux Thanks

jermy-c:
> @jermy-c The Parula bird seems a little too much on the left. It should be moved a few pixels to the right.

Sorry, I couldn't find a way to move the image using the Image Asset Studio. There's only options for Trim and Resize. I'll try the new Images provided.

jermy-c:
**Updated icons:**

image
image

benbucksch:
@jermy-c The margin of the Mustang icon is too large. The [previous one](https://github.com/mustang-im/mustang/pull/852#issuecomment-3225107443) and [those from Elena](https://github.com/mustang-im/mustang/pull/852#issuecomment-3225527487) are better.

Parula is good now.

jermy-c:
I've reverted the icon for Mustang.

image

benbucksch (approved):

jermy-c:
Templates for creating the icons directly without the need for Image Asset studio:

- https://developer.android.com/distribute/google-play/resources/icon-design-specifications#download_design_templates_resources
- https://www.figma.com/community/file/1131374111452281708/android-app-icons

Issue #841, opened 2025-08-24
More type checking for fewer bugs?


NeilRashbrook:
It's been pointed out to me that we have no type checking for `null` and `undefined`, so as I was bored this weekend I thought I'd have a look to see what type checking we might be missing out on. (If I get bored another weekend I might even write PRs for some of these that aren't strictly bugs.)
### allowUnreachableCode
Turning this on generates 8 "errors" but they are due to conditional compilation or pending features, so not useful.
### allowUnusedLabels
We have no unused labels. (Or more probably no labels at all.)
### exactOptionalPropertyTypes
This changes the way `T?` works as a field; normally it means the same as `T | undefined`, which is the behaviour we want, so we don't want this. (I think this needs `strictNullChecks` anyway.)
### noFallthroughCasesInSwitch
We have no fallthrough cases in switch.
### noImplicitAny
Turning this on generates 296 extra errors (with 125 lines of context).
- 69 of these (with 41 lines of context) are within JPC.
- 102 of these are due to parameters whose types have been (accidentally?) omitted.
- 46 of these are due to objects not being typed as Record.
- 11 of these are due to `Observable`'s decorators using dynamic indexing.
- 4 of these are for functions that return `null`. Unfortunately, because we don't type check for `null`, that's not a valid return type, so it defaults to `any`.
- 4 of these are for objects with a property initialised to `null`. Again, because we don't type check for `null`, that's not a valid type, so it defaults to `any`.
- 4 of these are for methods where TypeScript gave up trying to work out the return type. (It didn't even have to try very hard...)
- 2 of these are because `ActiveSync.verifyLogin` wasn't properly implemented (trying to return a value from an override of a void function).
- 3 of these are for indexing into an enum using its name, because its index signature is limited to the actual values.
- 2 of these are for indexing into an array using the stringified index.
- 4 of these are for modules with no declaration files.
- 3 of these are for global variables with no type annotation.
- 5 of these are for uses of the above variables.
- 6 of these are for indexing into a `StreetAddress` which has no index signature.
- 3 of these are for "indexing" into an EWS request which has no index signature.
- `GraphFolder.processChangedMessages` wants to "index" into `messageJSON["@removed"]`.
- 2 of these are for `JMAPFolder.fromJMAP` wants to index into `this.myRights[right]` and `json.myRights[right]`. Stupid TypeScript not allowing you to set the type of a `for...in` loop.
[...]

PR #863, opened 2025-08-27
Turn on strictFunctionTypes TypeScript type checking #841

NeilRashbrook:
Note: needs a patch to `svelte-collections` for proper type checking of `ArrayColl.find` and `Collection.filterOnce`.

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -448,7 +448,7 @@ export class ActiveSyncAccount extends MailAccount {
             break;
           case FolderType.Calendar:
           case FolderType.UserCalendar:
-            let calendar = appGlobal.calendars.find((calendar: ActiveSyncCalendar) => calendar.mainAccount == this && (calendar.url == url.toString() || calendar.serverID == change.ServerId)) as ActiveSyncCalendar | null;
+            let calendar = appGlobal.calendars.filterOnce((calendar): calendar is ActiveSyncCalendar => calendar.mainAccount == this).find(calendar => (calendar.url == url.toString() || calendar.serverID == change.ServerId));
             console.log("found the ActS cal again", calendar?.name);
             if (calendar) {
               calendar.name = change.DisplayName;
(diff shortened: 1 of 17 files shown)


benbucksch:
Thanks for the `find()` change. It's much better now. That part is OK for me.

Could you please also find a solution for CSSMediaRule which doesn't change runtime and doesn't massively reduce the readability of the code? Casting is perfectly fine with me.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -452,41 +442,53 @@ export class ActiveSyncAccount extends MailAccount {
...
-            if (!haveCalendar) {
-              let calendar = newCalendarForProtocol("calendar-activesync") as ActiveSyncCalendar;
-              calendar.initFromMainAccount(this);
-              if (!isMainCalendar) {
-                calendar.name = sanitize.nonemptylabel(change.DisplayName, calendar.name);
-              }
-              calendar.url = sanitize.url(folderURL);
-              calendar.serverID = sanitize.nonemptystring(change.ServerId);
-              await calendar.save();
+            let calendar = appGlobal.calendars.find((calendar): calendar is ActiveSyncCalendar => calendar.mainAccount == this && (calendar.url == url.toString() || (calendar as ActiveSyncCalendar).serverID == change.ServerId));
```
benbucksch, line 451:
```suggestion
            let calendar = appGlobal.calendars.find((calendar: ActiveSyncCalendar): calendar is ActiveSyncCalendar => calendar.mainAccount == this && (calendar.url == url.toString() || calendar.serverID == change.ServerId));
```

Does this also work?
And: Is there not a shorter (less redundant) way to express the same?

NeilRashbrook: No, `find` wants a function that matches the signature `Calendar => boolean`, not `ActiveSyncCalendar => boolean`. It then returns `Calendar | undefined`, *unless* the signature is written as `Calendar => is ActiveSyncCalendar`, in which case it returns `ActiveSyncCalendar | undefined`.

PR #864, opened 2025-08-27
Add missing Record type declarations #841

NeilRashbrook:
Found by `tsc --noImplicitAny`. It also doesn't like it when you index an enum with a string, so I threw in a few type casts, but maybe that should have been a separate PR.

### DIFF
--- a/app/logic/Contacts/EWS/EWSPerson.ts
+++ b/app/logic/Contacts/EWS/EWSPerson.ts
@@ -138,14 +138,18 @@ export class EWSPerson extends Person {
   }
 }
 
-const PhysicalAddressElements = {
+const PhysicalAddressElements: Recrod<string, string> = {
   street: "Street",
   city: "City",
   postalCode: "PostalCode",
   state: "State",
   country: "CountryOrRegion",
 };
-const PhysicalAddressPurposes = { Business: "work", Home: "home", Other: "other" };
+const PhysicalAddressPurposes: Record<string, string> = {
+  Business: "work",
+  Home: "home",
+  Other: "other",
+};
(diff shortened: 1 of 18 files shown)


benbucksch (commented):
I am not sure that this PR adds value. It does definitely add unless noise.

app/logic/Contacts/EWS/EWSPerson.ts
```diff
@@ -138,14 +138,18 @@ export class EWSPerson extends Person {
   }
 }
 
-const PhysicalAddressElements = {
+const PhysicalAddressElements: Recrod<string, string> = {
```
benbucksch:
```suggestion
const PhysicalAddressElements: Record<string, string> = {
```

You didn't actually compile this, did you?

app/logic/Contacts/EWS/EWSPerson.ts
```diff
@@ -138,14 +138,18 @@ export class EWSPerson extends Person {
...
   street: "Street",
   city: "City",
   postalCode: "PostalCode",
   state: "State",
   country: "CountryOrRegion",
 };
-const PhysicalAddressPurposes = { Business: "work", Home: "home", Other: "other" };
+const PhysicalAddressPurposes: Record<string, string> = {
```
benbucksch, line 148: Bleh :vomiting_face:

app/logic/Contacts/EWS/EWSPerson.ts
NeilRashbrook: Sorry, it was getting late and I was rushing. I've pushed a fix.

PR #865, opened 2025-08-27
Android: Parula brand in the Java code and file path

jermy-c:
- There were still some traces of the "im.mustang.capa" in the Java code which caused the app to crash
- The file path also needed to be changed also
- `App.getInfo().id` is the preferred way to get appID but some of the code was not getting bundled which caused more errors
- We now run the update-project.ts script with node.js instead of vite-node

### DIFF
--- a/mobile/backend/backend.ts
+++ b/mobile/backend/backend.ts
@@ -438,7 +438,7 @@ function directory(type: string): string {
 }
 
 // workaround: @capacitor/app App.getInfo().id has some code not getting bundled
-const appID = "com.beonex.parula";
+const appID = "im.mustang.capa";
 let configDir: string;
 /**
  * Get the user config directory on disk.
--- a/mobile/backend/backend.ts
+++ b/mobile/backend/backend.ts
@@ -437,6 +437,7 @@ function directory(type: string): string {
   return os.homedir();
 }
 
+const appID = "im.mustang.capa";
 let configDir: string;
 /**
  * Get the user config directory on disk.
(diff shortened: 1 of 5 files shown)


mobile/hooks/common/parula-brand.sh
```diff
@@ -13,3 +13,8 @@ perl -p -i \
   -e "s|im.mustang.capa|com.beonex.parula|;" \
   -e "s|"Mustang"|"Parula"|;" \
   ./capacitor.config.ts
+
+# Relace all appID in /mobile/backend
+perl -p -i \
+  -e "s|im.mustang.capa|com.beonex.parula|;" \
+  ./backend/backend.ts
```
benbucksch, line 20: :heavy_check_mark:

mobile/backend/backend.ts
```diff
@@ -437,6 +437,8 @@ function directory(type: string): string {
   return os.homedir();
 }
 
+// workaround: @capacitor/app App.getInfo().id has some code not getting bundled
+const appID = "im.mustang.capa";
```
benbucksch, line 441: :heavy_check_mark:

jermy-c, line 441: The update would have missed this update since it runs after `yarn build`.

benbucksch:
OK, that's pretty bad. It seems that the "android-template.tar.gz" is not actually a template, but generated specifically for our project, including the IDs. I wonder: Why do they have to be checked in as .tar.gz file, then?

I'm OK with this PR in this case.

jermy-c:
> Please do change the filenames and comments to be more specific than "update foo", but rather describe what you change in there (application IDs) and why (branding).

I've updated the comments. Which filenames should be changed?

benbucksch:
> I've updated the comments. Which filenames should be changed?

It's OK as-is

jermy-c:
> I made a CI build of this branch and installed it, and cleared all data, and it still crashes.

Ok, I'll look into it. Maybe something else needs to be updated.

benbucksch:
I tested with Samsung A16 with Android 15.

benbucksch (approved):

PR #871, opened 2025-08-28
Android: Use directories by Capacitor

jermy-c:
- We should use the directories pointed by Capacitor for mobile
- For the FilesDir we need to use "Documents" otherwise wouldn't be able to open them
- "/{appID}" is used because it would place the files directly as  in the "Documents" folder


benbucksch:
> However, placing files in the Documents folder means that the files would be deleted when the app is uninstalled

Did you mean "would *not* be deleted"?
We want them to be deleted.

> `/{appID}` is used

Use `{appName}`.

But: We will likely not take this appraoch. I don't want to clutter user directories.

jermy-c:
> Did you mean "would not be deleted"?

 Yes, that's what I meant.





> Use `{appName}`.
> 
> But: We will likely not take this appraoch. I don't want to clutter user directories.

Yes, I agree. I just realized this wouldn't work for iOS since the Documents directory is sandboxed Documents directory only for the app which would make sense to have another `/{appName}` directory within that directory.

app/logic/util/backend-wrapper.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * The files in here are useful mostly to the app itself, in internal data formats,
+ * and not really useful without the app.
+ * Consequently, this is a directory that is usually *not* shown to the user,
+ * but still accessible, if needed by technical people or for backups.
+ *
+ * - Linux: /home/USER/.mustang/
+ * - Windows: C:\Users\USER\AppData\Mustang\
+ * - Mac OS: /Users/USER/Library/Application Support/Mustang
+ * - Android: /data/user/USER/im.mustang.capa/files/
```
benbucksch:
```suggestion
 * - Android: /data/user/0/im.mustang.capa/files/
```

benbucksch: Is this the "current directory"? If so, we can just use relative paths for the SQL database file and avoid all this.

app/logic/util/backend-wrapper.ts
```diff
@@ -0,0 +1,66 @@
+import { appGlobal } from "../app";
```
benbucksch, line 1: The concept here makes sense. A lot of the capacitor functions are running in the frontend. That means essentially all the OS integration functions on mobile will need to be in the frontend. On Electron, they are in the node.js backend process.

I just don't like the name and location of this file. I don't know a better one, either. We need a clearer concept of what functions this file will contain. E.g. "OS Integration" would make sense. But SQLLite functions would also be here.

benbucksch (approved):

Issue #867, opened 2025-08-28
Android: Open URL in system browser


benbucksch:
### Reproduction
* Create a meeting with an online meeting URL, and manually set the URL to `https://mustang.im/test`
* Click on the [Globe] icon after the URL

### Actual result
* Nothing happens

### Expected result
* The phone's preferred browser opens.
* Note: The browser that the user chose, not the manufacturer's original preference, in case the user changed it.

### Code
* mobile/backend/backend.ts , function `openExternalURL()` needs to be implemented
* My failed attempt to use `@capacitor/browser` is in git commit 354f44c3d7

### Other situations
* If you click on an http:// link in an email, then the system browser opens. This is a different code path and doesn't call `openExternalURL()`.

jermy-c:
Seems like there's another error. Whenever we do `parseMIME()` we clear the collection and we add `new Attachment()` objects without the `filepathLocal`.

https://github.com/mustang-im/mustang/blob/77297fbf13d3b82ef264e9e4ac40e77cbd8c4091/app/logic/Mail/EMail.ts#L275-L278

When we `loadMIME()` we may trigger it also.

https://github.com/mustang-im/mustang/blob/77297fbf13d3b82ef264e9e4ac40e77cbd8c4091/app/logic/Mail/EMail.ts#L352

PR #872, opened 2025-09-01
Android: Open URL in system browser (#867)

jermy-c:
- Open URLs in the user's default browser.


benbucksch:
VSCode should automatically adapt the imports, but please double check. Please also test all instances in the UI where this is called.

jermy-c:
I checked the imports again and tested the UI also.

app/logic/util/backend-wrapper.ts
```diff
@@ -0,0 +1,13 @@
+// #if [MOBILE]
```
benbucksch: I'm sorry,   Jeremy! My comment must have gotten lost in the github UI. I wrote:

Could you please rename this file backend-wrapper to `os-integration.ts` ?

Sorry again...

jermy-c: Ok, no problem. Done.

PR #875, opened 2025-09-03
Fix ActiveSyncAccount.verifyLogin #841

NeilRashbrook:
The code for `verifyLogin` appears to be based on similar code for `callEAS` but the difference there is that `callEAS` does actually return a value while `verifyLogin` does not.

Also fixes some cases where TypeScript has trouble working out the type. (These all produce the same TypeScript error.)

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -189,9 +189,8 @@ export class ActiveSyncAccount extends MailAccount {
     if (response.status == 401) {
       const repeat = async () => {
         this.retries++;
-        let result = await this.verifyLogin(); // repeat the call
+        await this.verifyLogin(); // repeat the call
         this.retries = 0;
-        return result;
       }
       if (this.retries) {
         let ex = Error(`HTTP ${response.status} ${response.statusText}`);
@@ -199,7 +198,7 @@ export class ActiveSyncAccount extends MailAccount {
       } else if (this.oAuth2) {
         this.oAuth2.reset();
         await this.oAuth2.login(false); // will throw error, if interactive login is needed
-        return repeat();
+        await repeat();
       } else if (!/\bBasic\b/.test(response.WWWAuthenticate)) {
         throw this.fatalError = new ConnectError(null,
           "Unsupported authentication protocol(s): " + response.WWWAuthenticate);
--- a/app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
@@ -35,7 +35,7 @@ export class ActiveSyncCalendar extends Calendar implements ActiveSyncPingable {
     return new ActiveSyncEvent(this, parentEvent);
   }
 
-  getIncomingInvitationForEMail(message: ActiveSyncEMail) {
+  getIncomingInvitationForEMail(message: ActiveSyncEMail): ActiveSyncIncomingInvitation {
     return new ActiveSyncIncomingInvitation(this, message);
   }
 
--- a/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
@@ -135,7 +135,7 @@ export class ActiveSyncEvent extends Event {
     };
   }
 
-  get outgoingInvitation() {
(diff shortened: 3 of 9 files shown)


benbucksch (changes requested):
1. This mixes 2 kinds of changes: Trivial addition of obvious return type, and changes to verifyLogin(). Please make 2 separate PRs.
2. The explicit return types are very obvious. Why does TypeScript have issues? Is it due to how you defined these types?
3. The verifyLogin chnage looks wrong to me. You remove a number of `return` statements. That means the code now continues. Isn't there a `throw` at the end, or do I misread it? In any case, please keep the old logic and keep the returns. Also, please explicitly define the return type as void, because it's not pbvious *here*.
4. (Not this change, but the old code) In general, please do `return foo();` only if you really want to return the return value of `foo()`. If foo() returns undefined, void, Promise of void or similar null types, never write `return foo();` , but instead write `foo(); return;` (in 2 lines). It's far less confusing and much easier to read.

NeilRashbrook:
> 1. This mixes 2 kinds of changes: Trivial addition of obvious return type, and changes to verifyLogin(). Please make 2 separate PRs.

Sorry, I put them together because they were the same TypeScript error code. The return types aren't important enough for me to make a separate PR, so I'll just remove them from here.

> 2. The explicit return types are very obvious. Why does TypeScript have issues? Is it due to how you defined these types?

(Those objects were your idea.) It's because TypeScript can't validate the construction of the return value because it depends on `this` which it's in the middle of doing type checking for but it needs the type of the return value to complete that type checking.

> 3. The verifyLogin chnage looks wrong to me. You remove a number of `return` statements. That means the code now continues. Isn't there a `throw` at the end, or do I misread it? In any case, please keep the old logic and keep the returns. Also, please explicitly define the return type as void, because it's not obvious _here_.

You're quite right, there should be a `return` there. Thanks for spotting my error. (Also the return type is of course `Promise<void>`, since the function is async.)

> 4. (Not this change, but the old code) In general, please do `return foo();` only if you really want to return the return value of `foo()`. If foo() returns undefined, void, Promise of void or similar null types, never write `return foo();` , but instead write `foo(); return;` (in 2 lines). It's far less confusing and much easier to read, because I don't have to look up the return type of `foo()` .

I'll let the author of 07e37d0 know... oh, that was you!

(Two relevant tseslint rules; `return-await` would flag the `return repeat();` as needing an `await`; `no-confusing-void-expression` rule would in theory disallow using the awaited `void` value, although the problem here is that TypeScript couldn't figure out that the function had no return value, so I have no idea if that would work.)

benbucksch:
> I'll let the author of 07e37d0 know... oh, that was you!

Message received ;-) I normally don't do that. Not sure why I did that here. My bad. Sorry.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -189,17 +189,17 @@ export class ActiveSyncAccount extends MailAccount {
     if (response.status == 401) {
       const repeat = async () => {
         this.retries++;
-        let result = await this.verifyLogin(); // repeat the call
+        await this.verifyLogin(); // repeat the call
         this.retries = 0;
-        return result;
```
benbucksch, line 194: Still removing the return?

NeilRashbrook: This is a different return and there's nothing to return and we're at the end of the function.

benbucksch: There's ton of other code in the 'else' blocks,  and some of them have 'return'. Pleass keep the explicit 'return' here.

benbucksch (changes requested):

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: There is no else block here; you're looking at the wrong place.

benbucksch: I see that you remove the "return;" on line 194, there is an "if" on line 196, and an "else if" on lines 199 and 203, another "else" on line 206, and a "throw" on line 211.
At least that's what GitHub shows me here.
If you remove the "return", you fall into one of these throws.

Unless you are the very end of the function, could you please just keep the "return" in there?

benbucksch: Oh, this is an inline function created with const =. I couldn't see that correctly here on the small screen.

Issue #874, opened 2025-09-02
Android: SQLite database is not saved to disk


benbucksch:
### Reproduction 1
1. Read an email
2. Click on the author name
3. In the menu, select [Add to personal address book]
4. Force-quit Mustang
5. Start Mustang again
6. Go to address books

### Reproduction 2
1. Go to address books
2. Select Personal address book
3. Tap [+] to add a contact
4. Tap on the selected empty contact
5. Add a name

7. Force-quit Mustang (Nabvar, [|||] button to see the list of apps, the swipe the app away. Alternatively, long-tap the app icon on the Android start screen, Edit, Force-Quit)
8. Start Mustang again
9. Go to address books

### Actual result
* The new contact is not there

### Expected result
* The new contact is there

PR #880, opened 2025-09-04
Mail: Save recipient to disk and remotely

jermy-c:
- We were only adding the person to the collection and not saving to disk or remotely
- This fixes **Reproduction 1** of #874

### DIFF
--- a/app/frontend/Mail/Message/Recipient.svelte
+++ b/app/frontend/Mail/Message/Recipient.svelte
@@ -76,7 +76,6 @@
 
   async function addToAddressbook(addressbook: Addressbook) {
     let person = recipient.createPerson(addressbook);
-    addressbook.persons.add(person);
     await person.save();
     recipient = recipient;
   }
--- a/app/frontend/Mail/Message/Recipient.svelte
+++ b/app/frontend/Mail/Message/Recipient.svelte
@@ -76,6 +76,7 @@
 
   async function addToAddressbook(addressbook: Addressbook) {
     let person = recipient.createPerson(addressbook);
+    addressbook.persons.add(person);
     await person.save();
     recipient = recipient;
   }
--- a/app/frontend/Mail/Message/Recipient.svelte
+++ b/app/frontend/Mail/Message/Recipient.svelte
@@ -31,7 +31,7 @@
   {#each appGlobal.addressbooks.each as addressbook}
     <MenuItem
       label={$t`Add to ${addressbook.name}`}
-      onClick={() => addToAddressbook(addressbook)} />
+      onClick={() => catchErrors(() => addToAddressbook(addressbook))} />
   {/each}
 
   <MenuLabel label={$t`Other actions`} />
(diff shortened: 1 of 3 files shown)


app/frontend/Mail/Message/Recipient.svelte
```diff
@@ -31,7 +31,7 @@
   {#each appGlobal.addressbooks.each as addressbook}
     <MenuItem
       label={$t`Add to ${addressbook.name}`}
-      onClick={() => addToAddressbook(addressbook)} />
+      onClick={() => catchErrors(() => addToAddressbook(addressbook))} />
```
benbucksch: This change is unnecessary, because our `<MenuItem>` `onClick` does the catch.
https://github.com/mustang-im/mustang/blob/1d3199900a17a522bf3b4ce3207014c81936a8fa/app/frontend/Shared/Menu/MenuItem.svelte#L62

We need to catch errors in `on:click`, but `onClick` does it for us.

benbucksch:
> This fixes Reproduction 1 of Android: SQLite database is not saved to disk #874

I confirmed that this is the cause of the Repro 1.

app/frontend/Mail/Message/Recipient.svelte
jermy-c: Ok, thanks!

app/frontend/Mail/Message/Recipient.svelte
```diff
@@ -74,9 +74,11 @@
     isPopupOpen = true;
   }
 
-  function addToAddressbook(addressbook: Addressbook) {
+  async function addToAddressbook(addressbook: Addressbook) {
     let person = recipient.createPerson();
     addressbook.persons.add(person);
+    person.addressbook = addressbook;
+    await person.save();
```
NeilRashbrook: This will only work for local address books. Maybe something like this? (untested)
```suggestion
    await person.moveToAddressbook(addressbook);
```

jermy-c: Ok, thanks. It does work. I thought there would be a problem since `moveToAddress()` expects it to be already saved on disk but turns out it doesn't cause any problems.

benbucksch:
Neil is right about the problem,  but the problem is earlier. I've committed a suggestion what I mean. I typed this on my cellphone and couldn't test it. Can you please test both callers? The outer is in the mail composer when you add a new email address,  in the pop-up. Please also check that i caught all callers of createPerson().

app/logic/Abstract/PersonUID.ts
```diff
@@ -1,5 +1,6 @@
 import type { PersonOrGroup } from "../../frontend/Contacts/Person/PersonOrGroup";
 import { Group } from "./Group";
+import { Addressbook } from "../Contacts/Addressbook";
```
NeilRashbrook: `import type`?

jermy-c: Thanks!

PR #881, opened 2025-09-08
Fix duplicate delete strategy enum value

NeilRashbrook:
I assume this was an unintentional typo. This probably means that the `Flag` delete strategy has never been tested...


benbucksch:
Thanks

Issue #404, opened 2025-01-16
EWS, OWA, ActS: Shared folders, calendars, address books - outgoing


benbucksch:
Our user wants to share their mail folders, calendar or address book with another user from his organization, on the same server.

In Settings, there should be a tab "Share" as part of the mail account, calendar and address book, and the user can give address to others there. Elena has UX mockups for how the UI will look like.

There are multiple sharing levels:
1. Free/Busy (other person can only query specific times)
2. Availability (see all times, but only whether free or busy)
3. Limited read (can see titles of meetings, but not details)
4. Full read (can see entire calendar)
5. Edit (can modify entire calendar).

- [ ] EWS
   - [ ] Mail folders
   - [ ] Calendar
   - [ ] Address book
- [ ] OWA
   - [ ] Mail folders
   - [ ] Calendar
   - [ ] Address book
- [ ] ActiveSync
   - [ ] Mail folders
   - [ ] Calendar
   - [ ] Address book

benbucksch:
https://github.com/microsoftgraph/microsoft-graph-docs-contrib/blob/main/concepts/outlook-share-or-delegate-calendar.md

PR #885, opened 2025-09-10
IMAP, EWS, OWA: Allow others to access my mail folders #404

NeilRashbrook:
Although #404 claimed that UX mockups exist it doesn't actually help because I don't know where they are, so I wrote some basic UI.

- When viewing an EWS account, it shows you the Exchange Delegation feature where you can assign delegates for your email, contatcs and calendar. You also get to assign basic permissions to your delegates. I can't remember whether this allows them to send mail as or on behalf of you or not.
- When viewing an EWS folder, it shows you the Exchange permissions for that folder. It uses the easy permissions model, so you get to choose from 9 permission levels rather than 288 possible permission combinations.
- When viewing an EWS calendar, it shows you the Exchange permissions for that calendar, again using the easy permissions model, although now there are 11 permission levels to choose from.

Permissions are entered using the autocomplete UI. You can add users using their email address and click on a user to edit their permissions. In the case of delegated users, each user is edited and saved separately using a button inside the popup, while in the case of folder and calendar permissions, these are all saved in one go using a separate button.

So that the easy permission UI can be slightly more easily understood, some of their effective permissions are shown as a group of checkboxes, but these are display only.

### DIFF
--- a/app/logic/Mail/EWS/EWSFolder.ts
+++ b/app/logic/Mail/EWS/EWSFolder.ts
@@ -540,7 +540,7 @@ export class EWSFolder extends Folder {
     return "You cannot change Exchange special folders.";
   }
 
-  async getPermissions(): Promise<ArrayColl<ExchangePermission>> {
+  async getSharedPersons(): Promise<ArrayColl<PersonUID>> {
     let request = {
       m$GetFolder: {
         m$FolderShape: {
@@ -559,10 +559,32 @@ export class EWSFolder extends Folder {
       },
     };
     let result = await this.account.callEWS(request);
-    return new ArrayColl(result.Folders.Folder.PermissionSet.Permissions.Permission.map(permission => ExchangePermission.fromExchange(permission, this.account.emailAddress)));
+    return getSharedPersons(result.Folders.Folder.PermissionSet.Permissions.Permission);
   }
 
-  async setPermissions(permissions: ArrayColl<ExchangePermission>) {
+  async getPermissions(): Promise<ExchangePermission[]> {
+    let request = {
+      m$GetFolder: {
+        m$FolderShape: {
+          t$BaseShape: "IdOnly",
+          t$AdditionalProperties: {
+            t$FieldURI: {
+              FieldURI: "folder:PermissionSet",
+            },
+          },
+        },
+        m$FolderIds: {
+          t$FolderId: {
+            Id: this.id,
+          },
+        },
+      },
+    };
+    let result = await this.account.callEWS(request);
+    return result.Folders.Folder.PermissionSet.Permissions.Permission.map(permission => new ExchangePermission(permission));
+  }
+
+  async setPermissions(permissions: ExchangePermission[]) {
     let request = {
       m$UpdateFolder: {
         m$FolderChanges: {
@@ -578,7 +600,7 @@ export class EWSFolder extends Folder {
                 t$Folder: {
                   t$PermissionSet: {
                     t$Permissions: {
-                      t$Permission: permissions.contents.map(permission => permission.toEWSFolderPermission()),
+                      t$Permission: permissions.map(permission => permission.toEWSFolderPermission()),
                     },
                   },
                 },
@@ -622,34 +644,17 @@ export class ExchangePermissions {
   }
 }
 
-export class ExchangePermission extends PersonUID {
+type ExchangeUser = { UserId: { DisplayName: string; DistinguishedUser: string; PrimarySmtpAddress: string; } };
+
+export class ExchangePermission {
   exchangePermissions: ExchangePermissions;
   distinguishedUser?: string;
+  emailAddress?: string;
 
-  constructor(emailAddress: string, name: string, permissions: Partial<ExchangePermissions> = {}, distinguishedUser?: "Anonymous" | "Default") {
-    super(emailAddress, name);
+  constructor(permissions: Partial<ExchangePermissions> & ExchangeUser) {
     this.exchangePermissions = new ExchangePermissions(permissions);
-    this.distinguishedUser = distinguishedUser;
-  }
-
-  static fromExchange(permissions: any, emailAddress: string) {
-    let name;
-    let distinguishedUser = permissions.UserId.DistinguishedUser as "Anonymous" | "Default" | undefined;
-    switch (distinguishedUser) {
-    case "Anonymous":
-      emailAddress = "*@*";
-      name = gt`Anonymous`;
-      break;
-    case "Default":
-      emailAddress = emailAddress.replace(/.*@/, "*@");
-      name = gt`All other users`;
-      break;
-    default:
-      emailAddress = sanitize.emailAddress(permissions.UserId.PrimarySmtpAddress, null);
-      name = sanitize.string(permissions.UserId.DisplayName, null);
-      break;
-    }
-    return new ExchangePermission(emailAddress, name, permissions, distinguishedUser);
+    this.distinguishedUser = permissions.UserId.DistinguishedUser;
+    this.emailAddress = permissions.UserId.PrimarySmtpAddress;
   }
 
   toEWSFolderPermission() {
@@ -687,21 +692,25 @@ export class ExchangePermission extends PersonUID {
   }
 }
 
-export async function deleteExchangePermissions(target: { getPermissions(): Promise<ArrayColl<ExchangePermission>>, setPermissions(permission: ArrayColl<ExchangePermission>): Promise<void> }, otherPerson: PersonUID) {
+export function getSharedPersons(permissions: ExchangeUser[]): ArrayColl<PersonUID> {
+  return new ArrayColl(permissions.filter(permission => !permission.UserId.DistinguishedUser).map(permission => new PersonUID(permission.UserId.PrimarySmtpAddress, permission.UserId.DisplayName)));
+}
+
+export async function deleteExchangePermissions(target: { getPermissions(): Promise<ExchangePermission[]>, setPermissions(permission: ExchangePermission[]): Promise<void> }, otherPerson: PersonUID) {
   let targetPermissions = await target.getPermissions();
-  let personPermission = targetPermissions.find(permission => permission.emailAddress == otherPerson.emailAddress);
-  if (personPermission) {
-    targetPermissions.remove(personPermission);
+  let personPermission = targetPermissions.findIndex(permission => permission.emailAddress == otherPerson.emailAddress);
+  if (personPermission >= 0) {
+    targetPermissions.splice(personPermission, 1);
     await target.setPermissions(targetPermissions);
   }
 }
 
-export async function setExchangePermissions(target: { getPermissions(): Promise<ArrayColl<ExchangePermission>>, setPermissions(permission: ArrayColl<ExchangePermission>): Promise<void> }, person: PersonUID, access: string, ...permissions: MailShareIndividualPermissions[]) {
+export async function setExchangePermissions(target: { getPermissions(): Promise<ExchangePermission[]>, setPermissions(permission: ExchangePermission[]): Promise<void> }, person: PersonUID, access: string, ...permissions: MailShareIndividualPermissions[]) {
   let targetPermissions = await target.getPermissions();
   let personPermission = targetPermissions.find(permission => permission.emailAddress == person.emailAddress);
   if (!personPermission) {
-    personPermission = new ExchangePermission(person.emailAddress, person.name, { IsFolderVisible: true });
-    targetPermissions.add(personPermission);
+    personPermission = new ExchangePermission({ IsFolderVisible: true, UserId: { DisplayName: undefined, DistinguishedUser: undefined, PrimarySmtpAddress: person.emailAddress } });
+    targetPermissions.push(personPermission);
   }
   let permission = personPermission.exchangePermissions;
   switch (access) {
--- a/app/logic/Calendar/EWS/EWSCalendar.ts
+++ b/app/logic/Calendar/EWS/EWSCalendar.ts
@@ -4,12 +4,12 @@ import type { PersonUID } from "../../Abstract/PersonUID";
 import { EWSEvent } from "./EWSEvent";
 import { EWSIncomingInvitation } from "./EWSIncomingInvitation";
 import type { EWSAccount } from "../../Mail/EWS/EWSAccount";
-import { ExchangePermission, deleteExchangePermissions, setExchangePermissions } from "../../Mail/EWS/EWSFolder";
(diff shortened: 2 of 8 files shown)


app/frontend/Calendar/EditEvent/ParticipantsBox.svelte
```diff
@@ -14,16 +14,16 @@
...
+      {person.emailAddress}
+    </hbox>
     <hbox slot="person-pill-before-avatar" let:person={participant}>
       <ParticipantConfirmIcon {participant} />
     </hbox>
     <hbox slot="person-popup-bottom" let:person={participant}>
       <ParticipantConfirmText {participant} />
     </hbox>
-    <hbox slot="result-bottom-row" let:person>
-      <PersonAvailability {person} />
```
benbucksch, line 25: Could you please explain why you change this? It removes the Availability display.

app/frontend/Contacts/PersonAutocomplete/PersonAutocomplete.svelte
```diff
@@ -42,13 +42,13 @@
...
   import { assert } from "../../../logic/util/util";
   import { t } from "../../../l10n/l10n";
-  const dispatch = createEventDispatcher<{ addPerson: PersonUID }>();
 
   export let skipPersons: Collection<PersonUID> = new ArrayColl<PersonUID>();
   export let placeholder = $t`Add person`;
   export let tabindex = null;
   export let autofocus = false;
   export let typedText: string = ""; /* in/out */
+  export let onAddPerson: (person: PersonUID) => void | Promise<void>;
```
benbucksch, line 51: Please make these Autocomplete event handlers changes a separate PR with its own justification.

app/frontend/Settings/AccountExchangeCalendar.svelte
```diff
@@ -0,0 +1,70 @@
...
+  <hbox slot="header">Exchange - {$t`Advanced`}</hbox>
+
+  {#if permissions}
+    <hbox class="align-start">
+      <label for="permissions">{$t`Calendar permissions`}</label>
+      <PersonsAutocomplete persons={permissions} placeholder={$t`Add permisson`} {onAddPerson} {onRemovePerson}>
+        <hbox slot="result-bottom-row" class="recipient-email-address font-small" let:person>
+          {person.emailAddress}
+        </hbox>
+        <CalendarExchangePermissions slot="person-popup-bottom" let:person {person}/>
```
benbucksch: UI should be independent of Exchange. We need the same feature and UI in JMAP and even in IMAP accounts. IMAP supports this, and we have a specific request to implement that for Dovecot.

app/logic/build.ts
```diff
@@ -4,6 +4,6 @@ export const production = false;
 export const appVersion: string = '0.9.5-dev';
 export const appName: string = 'Mustang';
 export const siteRoot: string = 'https://mustang.im';
-export const isMobile = true;
+export const isMobile = false;
```
benbucksch, line 7: 🙈

benbucksch: Sorry.
Could you please make a separate PR for that?

app/frontend/Calendar/EditEvent/ParticipantsBox.svelte
NeilRashbrook: You're right, this doesn't belong in this PR; I was just putting in `{person.emailAddress}` everywhere, since it really frustrates me to see 10 contacts named "Ben" without knowing which one is the email address I want.

As it happens, `<PersonAvailability>` is just a stub. Instead Availability is shown further down using `<AvailabilityGrid>`.

app/frontend/Contacts/PersonAutocomplete/PersonAutocomplete.svelte
NeilRashbrook: Indeed, I think this is a better solution for #826, so I'll switch that PR over to this method. (But it would need to land before I could rebase the changes away.)

app/frontend/Settings/AccountExchangeCalendar.svelte
NeilRashbrook: Actually this section is for account permissions i.e. who has access to your Exchange Calendar "account" or Addressbook "account". The folder permissions are in`app/frontend/Settings/Mail/Account/` but I'd have to look at how folder permissions work in IMAP and JMAP to see whether I could unify the UI at all.

app/logic/build.ts
NeilRashbrook: Filed #892.

app/frontend/Settings/AccountExchangeCalendar.svelte
NeilRashbrook: IMAP rights work differently to Exchange folder permissions, so I can't really unify them. (They work more like Exchange delegate permissions, in that for instance the "Remove" action is immediate.) I tried to find out how JMAP mailbox permissions work, but I failed.

app/frontend/Calendar/EditEvent/ParticipantsBox.svelte
benbucksch: > it really frustrates me to see 10 contacts named "Ben" without knowing which one is the email address I want.

Yes, I discussed that with @elenaux and we plan to add the email address to the results.

app/logic/Mail/IMAP/IMAPFolder.ts
```diff
@@ -645,4 +646,63 @@ export class IMAPFolder extends Folder {
...
+  async getPermissions(): Promise<ArrayColl<IMAPPermission> | undefined> {
+    if (!await this.account.hasCapability("ACL")) {
+      return undefined;
+    }
+    let conn = await this.account.connection();
+    let attributes: Array<{ type: string, value: string }>;
+    let permissions = new ArrayColl<IMAPPermission>();
+    await conn.exec('GETACL', [{ type: 'ATOM', value: this.path }], { untagged: { async ACL(untagged) { attributes = untagged.attributes; } } });
+    for (let i = 0; i < attributes.length; i += 2) {
+      let name = attributes[i].value;
```
benbucksch: `sanitize()` everything

app/logic/Mail/IMAP/IMAPFolder.ts
```diff
@@ -645,4 +646,63 @@ export class IMAPFolder extends Folder {
...
+  w = "write",
+  i = "create",
+  k = "subfolder",
+  x = "remove",
+  t = "delete",
+  e = "expunge",
+  a = "owner",
+}
+
+export class IMAPPermission extends PersonUID {
```
benbucksch: An `IMAPPermission` should *have* a person, not *be* a person.

```suggestion
export class IMAPPermission {
  for: PersonUID;
```

app/logic/Mail/IMAP/IMAPFolder.ts
```diff
@@ -645,4 +646,63 @@ export class IMAPFolder extends Folder {
...
+    for (let i = 0; i < attributes.length; i += 2) {
+      let name = attributes[i].value;
+      let rights = attributes[i + 1].value;
+      let emailAddress = name.includes('@') ? name : name + this.account.emailAddress.slice(this.account.emailAddress.indexOf('@'));
+      permissions.add(new IMAPPermission(emailAddress, name, rights));
+    }
+    return permissions;
+  }
+
+  async setPermission(permission: IMAPPermission, right: keyof typeof IMAPACL) {
```
benbucksch: `left:` is missing, _right_?

app/logic/Mail/EWS/EWSFolder.ts
```diff
@@ -518,6 +520,171 @@ export class EWSFolder extends Folder {
...
+  PublishingAuthor = Create | Read | Subfolders | WriteOwn,
+  Editor = Create | Read | WriteAll,
+  PublishingEditor = Create | Read | Subfolders | WriteAll,
+  Owner = Create | Read | Subfolders | WriteOwn | WriteAll,
+  // Calendar permission values
+  FreeBusyTimeOnly = 1 << 5,
+  FreeBusyTimeAndSubjectAndLocation = 1 << 6,
+}
+
+export class ExchangePermission extends PersonUID {
```
benbucksch: See IMAP: `for: PersonUID`. It *is* not a Person.


```suggestion
export class ExchangePermission {
  for: PersonUID;
```

app/logic/Mail/EWS/EWSFolder.ts
```diff
@@ -518,6 +520,171 @@ export class EWSFolder extends Folder {
...
+            },
+          },
+        },
+      },
+    };
+    await this.account.callEWS(request);
+  }
+}
+
+export enum ExchangePermissions {
```
benbucksch: We need to find a way to unify that with IMAPPermissions

app/logic/Mail/IMAP/IMAPFolder.ts
```diff
@@ -645,4 +646,63 @@ export class IMAPFolder extends Folder {
...
+    await conn.exec('SETACL', [{ type: 'ATOM', value: this.path }, { type: 'ATOM', value: permission.name }, { type: 'ATOM', value: permission[IMAPACL[right]] ? "+" + right : "-" + right }]);
+  }
+
+  async removePermission(permission: PersonUID) {
+    let conn = await this.account.connection();
+    await conn.exec('DELETEACL', [{ type: 'ATOM', value: this.path }, { type: 'ATOM', value: permission.name }]);
+  }
+}
+
+export enum IMAPACL {
```
benbucksch: We need to find a way to unify that with ExchangePermissions

app/logic/Mail/IMAP/IMAPFolder.ts
NeilRashbrook: Well, we'll see what what your final UI does, but for now, I'm using PersonsAutocomplete to hold the permitted users, and that needs the permissions to derive from, not aggregate, PersonUID.

app/logic/Mail/EWS/EWSFolder.ts
```diff
@@ -527,6 +529,152 @@ export class EWSFolder extends Folder {
...
+
+  toEWSCalendarPermission() {
+    return Object.assign({
+      t$UserId: this.distinguishedUser
+      ? { t$DistinguishedUser: this.distinguishedUser, }
+      : { t$PrimarySmtpAddress: this.emailAddress, },
+    }, { t$CalendarPermissionLevel: "Custom" }, this.exchangePermissions.toEWS());
+  }
+
+  toOWAFolderPermission() {
```
benbucksch, line 676: OWA code in EWS?


The general direction is a generic API for permissions that works across all protocols.

benbucksch:
Please:
* Merge to master
* Move protocol-specific logic from the UI to the backend. For the high-level permissions and the low-level permissions, you may make 2 backend functions.
* The logic whether of the high-level permissions applies (based on the low level permissioins set) should be in the backend, not the frontend.

NeilRashbrook:
> - Merge to master

Done.

> - Move protocol-specific logic from the UI to the backend. For the high-level permissions and the low-level permissions, you may make 2 backend functions.

Done. I also restored `PersonsAutocomplete`.

> - The logic whether of the high-level permissions applies (based on the low level permissions set) should be in the backend, not the frontend.

I may have satisfied this inadvertently but I'm not actually sure what you mean here sorry.

> I installed 2.3 instead of 2.4

Ubuntu's default repos include 2.3.21; the instructions say use `apt update` and `apt upgrade` to upgrade but in fact `apt update` is also required for a new installation of 2.4, otherwise 2.3.21 gets installed.

app/frontend/Settings/Mail/SharingOutgoing.svelte
```diff
@@ -187,61 +189,50 @@
...
-  async function onAddPerson(person: PersonUID) {
-    assert(account instanceof EWSAccount || account instanceof OWAAccount, "Not supported");
-    addPerson = null;
-    if (sharedFolders.includes("msgfolderroot")) {
-      await account.addSharedFolders(person, "msgfolderroot");
-    } else if (sharedFolders.includes("inbox")) {
-      await account.addSharedFolders(person, "inbox");
-    }
-    if (sharedFolders.includes("contacts")) {
-      await account.addSharedAddressbook(person);
```
benbucksch, line 240: Can you please use the API that I sketched here?

Actually, it should be:
```
await account.addSharedFolders(person, "inbox", permission);
...
await mainAddressbook.addSharedPerson(person, permission);
...
await mainCalendar.addSharedPerson(person, permission);
```

For deletion, the delete on the mail account should delete only the mail sharing, not everything. The calendar and address book should have `deleteSharedPerson(person)` functions on their classes.

benbucksch (changes requested):

benbucksch:
More generally, please don't make the backend API specific to this UI. As mentioned, the UI is likely to change over time, and the backend should stay as-is despite any reasonable UI changes.

benbucksch:
There are also other review comments from 3 weeks still open. E.g. Permissions should not inherit from `PersonUID`, but have one.

We should get rid of `ExchangePermissions` and `IMAPACL` and use the generic permissions, if possible.

app/frontend/Settings/Mail/SharingOutgoing.svelte
NeilRashbrook: This looked as if it was simply copied and pasted from incoming sharing. I had assumed that was an oversight.

Some of the other changes were not done because I was using that code for testing purposes.

benbucksch: > This looked as if it was simply copied and pasted from incoming sharing. I had assumed that was an oversight.

Yes, it was. Indeed, it was an oversight and I was confused. Sorry about that.

Of course the function for incoming sharing and outgoing sharing have to be different functions. It wouldn't make any sense to unify them into one.

What I meant with my comment is that mail, calendar and address book sharing shouldn't all be in the same function. But whatever account I want to grant access to, I should call a function on that account. You do that in your latest revision, so that's nice. The API now looks sane.

However, I am confused how the `addSharedPerson()` now relates to the incoming sharing, which uses the same function.

benbucksch: `addSharedPerson()` makes sense for incoming sharing.
For outgoing, how about `grantAccessToPerson()`?

NeilRashbrook:
- dovecot 2.4: ACL, read works, write fails silently, didn't test delete

app/frontend/Settings/Mail/SharingOutgoing.svelte
NeilRashbrook: For incoming sharing, you're adding the other person's folders/addressbook/calendar to your accounts.

For outgoing sharing, you're adding other persons who can access your folders/addressbook/calendar.

PR #887, opened 2025-09-11
Close in-app OAuth after login

NeilRashbrook:
Steps to reproduce problem:
- Start app
- Click "disconencted" icon for OWA account
- Watch as OWA logs in

Expected result: Mail automatically selected after login completes

Actual result: Must manually click Mail icon to actually start reading mail

### DIFF
--- a/app/frontend/Mail/MailMustangApp.ts
+++ b/app/frontend/Mail/MailMustangApp.ts
@@ -61,6 +61,7 @@ const tabsObserver = {
     for (let tab of tabs) {
       mailMustangApp.subApps.remove(tab.mustangApp);
     }
+    openApp(mailMustangApp, {});
   },
 };
 oAuth2TabsOpen.registerObserver(tabsObserver);


app/frontend/Mail/MailMustangApp.ts
```diff
@@ -61,6 +61,7 @@ const tabsObserver = {
     for (let tab of tabs) {
       mailMustangApp.subApps.remove(tab.mustangApp);
     }
+    openApp(mailMustangApp, {});
```
benbucksch, line 64:
```suggestion
    navigate(-1);
```

Does this also work?

NeilRashbrook: Sort of. If you don't touch anything, you're fine. But if you accidentally click Calendar before logon completes, then figure out you went wrong and click back on the login, then when login does complete, you switch back to Calendar, but Mail stays highlighted in the sidebar.

PR #893, opened 2025-09-17
iOS: Node.js 24

jermy-c:
- Update code to use Node.js mobile v24
- Automatically detects node native modules and signs them and then copies them into the app bundle
- Not tested for Release build yet


mobile/capacitor.config.ts
```diff
@@ -1,22 +1,34 @@
...
+    appName: "Mustang",
+    webDir: "./dist",
+    plugins: {
+      CapacitorNodeJS: {
+        nodeDir: "nodejs",
+        androidLibNode: "https://github.com/mustang-im/nodejs-mobile/releases/download/v22.9.0/nodejs-mobile-v22.9.0-android.zip",
+        androidArchitectures: ["arm64"],
+      },
+      SplashScreen: {
+        "launchAutoHide": true
```
benbucksch: This change looks unnecessary and not ideal.

mobile/backend/package.json
```diff
@@ -9,30 +9,29 @@
     "build": "vite build"
   },
   "dependencies": {
-    "@radically-straightforward/sqlite": "^1.0.0",
-    "@types/nodemailer": "^6.4.14",
-    "adm-zip": "^0.5.12",
-    "better-sqlite3": "^11.10.0",
+    "@radically-straightforward/sqlite": "^1.1.10",
```
benbucksch: Which of these changes are needed?

mobile/capacitor.config.ts
jermy-c: This change is for dynamically setting the settings for the iOS build because `process.env` only works in functions. But this might change because I still don't how to build for production/release.

mobile/backend/package.json
jermy-c: None of them, I'll revert the changes. It was for trying to remove the `punycode` deprecation warning but this doesn't solve it.

mobile/docs/build.md
```diff
@@ -44,3 +44,36 @@ fails when there's no tag for the commit which fails when you manually trigger i
...
+8. `cd mobile; yarn setup:ios`
+9. `cd mobile; npx cap sync ios`
+10. `npx cap open ios` in `/mobile`
+11. Select `Any iOS Device (arm64)` at the top bar under the menu
+12. Click `[Product]` > `[Archive]`
+13. Go to `~//Library/Developer/Xcode/Archives` to find the Archive
+
+#### Troubleshooting
+
+- You may need to wait a little bit after opening the project to select device and build because Xcode needs to index the files
```
benbucksch: Please quantify "a little bit". 10 Seconds? 10 Minutes?

mobile/hooks/ios/setup.sh
```diff
@@ -0,0 +1,16 @@
...
+
+mkdir -p ./resources
+cd ./resources
+
+DOWNLOAD_URL="https://github.com/mustang-im/nodejs-mobile/releases/download/v24.5.0/nodejs-mobile-v24.5.0-ios.zip"
+TMP_ZIP_FILE="tmp.zip"
+
+# Download the file
+echo "Downloading $DOWNLOAD_URL"
+curl -L -o $TMP_ZIP_FILE "$DOWNLOAD_URL"
```
benbucksch, line 17: You set `IOS_LIBNODE` earlier. Why is that not sufficient?

Why do we need to manually download,  if that wasn't necessary for node 18?

benbucksch: If we keep this, please make sure to do that only when the file doesn't exist yet.

mobile/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -255,14 +256,54 @@
...
+			inputPaths = (
+			);
+			name = "Code Sign Node Native Modules";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "find \"${CONFIGURATION_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}\" -name '*.node' -exec codesign --force --sign \"${EXPANDED_CODE_SIGN_IDENTITY}\" {} \\;\n";
```
benbucksch: Why this new shell script? Why as inline code? Can you use built-in functioons of the build system?

mobile/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -255,14 +256,54 @@
...
+			inputPaths = (
+			);
+			name = "Copy Node Native Modules";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "SRC_DIR=\"${SRCROOT}/App/nodejs-project\"\n\nDEST_DIR=\"${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}\"\n\nmkdir -p \"$DEST_DIR\"\n\nfind \"$SRC_DIR\" -name '*.node' -exec cp --parents {} \"$DEST_DIR\" \\;\n";
```
benbucksch: ditto

mobile/hooks/ios/setup.sh
jermy-c: > Why do we need to manually download, if that wasn't necessary for node 18?

We did manually download in node18, it was just that setup.sh was placed in different directory.

> You set IOS_LIBNODE earlier. Why is that not sufficient?

Because IOS_LIBNODE in the build script which runs after this. But we can maybe have the variables in one file better.

mobile/ios/App/App.xcodeproj/project.pbxproj
jermy-c: There's a built-in function for this but it needed be configured with the UI every run.

mobile/ios/App/App.xcodeproj/project.pbxproj
jermy-c: There's a built-in function for code-signing the prebuilds but I didn't know how to configure it properly and it caused a lot of problems.

benbucksch: Can you please at least make the script a file on its own that you call here, instead of writing this inline?

mobile/ios/App/App.xcodeproj/project.pbxproj
benbucksch: OK, thanks for scripting this.

mobile/hooks/ios/setup.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+mkdir -p ./resources
+cd ./resources
+
+DOWNLOAD_URL="https://github.com/mustang-im/nodejs-mobile/releases/download/v24.5.0/nodejs-mobile-v24.5.0-ios.zip"
```
benbucksch: Same URL as `IOS_LIBNODE=` above? Please share them. Put this variable in a separate script file.

mobile/hooks/ios/setup.sh
benbucksch: OK

mobile/hooks/ios/setup.sh
benbucksch: Also, please keep the old code at least as a comment. It looks useful.
```
DOWNLOAD_URL=$(curl -s "https://api.github.com/repos/$OWNER/$REPO/releases/latest" \
  | jq -r '.assets[] | select(.name | test("ios.*\\.zip")) | .browser_download_url')
```

mobile/docs/build.md
benbucksch: ping

jermy-c: 10 minutes

mobile/ios/App/App.xcodeproj/project.pbxproj
jermy-c: Yes, 9031464c46fd6ba20d9728d80eac34efd504c82a

mobile/docs/build.md
jermy-c: 4d75d6bfa524729e1b1bae6fc68eedf456893c2c

mobile/hooks/ios/setup.sh
jermy-c: 90728c1414eedaa3e0910a63898622a55ef69b48

mobile/hooks/ios/variables.sh
```diff
@@ -0,0 +1 @@
+IOS_LIBNODE="https://github.com/mustang-im/nodejs-mobile/releases/download/v24.5.0/nodejs-mobile-v24.5.0-ios.zip"
```
benbucksch, line 1: Please rename file to hooks/ios/variables.sh

In another PR, please rename hooks/ to scripts/ or better - if possible - hooks/android/ to android/scripts/ and hooks/ios/ to ios/scripts/

mobile/hooks/ios/setup.sh
```diff
@@ -0,0 +1,22 @@
...
+
+mkdir -p ./ios/App/resources
+cd ./ios/App/resources
+
+# Gets the latest release download URL from GitHub API
+# DOWNLOAD_URL=$(curl -s "https://api.github.com/repos/$OWNER/$REPO/releases/latest" \
+#   | jq -r '.assets[] | select(.name | test("ios.*\\.zip")) | .browser_download_url')
+
+DOWNLOAD_URL="$IOS_LIBNODE"
+TMP_ZIP_FILE="tmp.zip"
```
benbucksch: Make the file name describe the content.

benbucksch: libnode-download.zip

mobile/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -453,12 +471,8 @@
 				);
 				LIBRARY_SEARCH_PATHS = (
 					"$(inherited)",
-					"$(PROJECT_DIR)/App/nodejs-project/builds/better_sqlite3.node",
-					"$(PROJECT_DIR)/App/nodejs-project/build/bufferutil.node",
-					"$(PROJECT_DIR)/App/nodejs-project/build/better_sqlite3.node",
-					"$(PROJECT_DIR)/App/nodejs-project/build/sdf",
```
benbucksch, line 459: You can extract the downloaded binaries before loading xcode, to avoid that xcode removes them from the project.

jermy-c: I don't think this will be an issue for CI build since the orders is always to download and build the binaries before building the app. But this will something to remember when build locally or local development.

PR #894, opened 2025-09-17
iOS: Load UI with SceneDelegate instead of AppDelegate

jermy-c:
-  Load UI with SceneDelegate instead of AppDelegate to comply with iOS 26 requirements

> In the next major release following iOS 26, UIScene lifecycle will be required when building with the latest SDK; otherwise, your app won’t launch. While supporting multiple scenes is encouraged, only adoption of scene life-cycle is required.

https://developer.apple.com/documentation/technotes/tn3187-migrating-to-the-uikit-scene-based-life-cycle


jermy-c:
Yes, this one is ready to merge.

Issue #403, opened 2025-01-16
EWS, OWA, ActS: Shared folders, calendars, address books - incoming


benbucksch:
When another user shared their folder, calendar or address book with us, we should display that.
It should appear as another account. We can then think about the presentation.

- [ ] EWS
   - [ ] Mail folders
   - [ ] Calendar
   - [ ] Address book
- [ ] OWA
   - [ ] Mail folders
   - [ ] Calendar
   - [ ] Address book
- [ ] ActiveSync
   - [ ] Mail folders
   - [ ] Calendar
   - [ ] Address book

benbucksch:
https://github.com/microsoftgraph/microsoft-graph-docs-contrib/blob/main/concepts/outlook-get-shared-events-calendars.md

benbucksch:
Should also support SendOnBehalf/SendAs , if permitted

PR #898, opened 2025-09-19
EWS: Open other users' shared mailbox, contacts and calendars #403

NeilRashbrook:
Note: This is based on #830, since without that we can't actually access any non-default contacts or calendars. Apparently once that lands, GitHub will automatically turn this into a PR against master.

Currently this only tries to look for the sharing user's default contacts or calendars. In theory if the entire mailbox has been shared it would be possible to search for all contacts or calendars but I don't know whether that's useful.

I can't get the list of existing shared contacts or calendars to update dynamically. (I can't even get `each` to work with the collection; it tries to index it as if it was an array, so I've had to use `.contents` just to stop it crashing.)

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -685,7 +685,7 @@ export class EWSAccount extends MailAccount {
 
   async createToplevelFolder(name: string): Promise<EWSFolder> {
     if (this.sharedFolderRoot == "inbox") {
-      throw new Error(gt`You only have access to the Inbox of these shared folders`);
+      throw new Error(gt`You only have access to the Inbox of this shared account`);
     }
     let request = {
       m$CreateFolder: {
@@ -715,7 +715,7 @@ export class EWSAccount extends MailAccount {
     return folder;
   }
 
-  async findNamedFolders(person: PersonUID, names: string[]): Promise<string[]> {
+  async findSharedFolders(person: PersonUID, DistinguishedIDs: string[]): Promise<string[]> {
     if (this.mainAccount) {
       throw new NotReached();
     }
@@ -730,8 +730,8 @@ export class EWSAccount extends MailAccount {
           },
         },
         m$FolderIds: {
-          t$DistinguishedFolderId: names.map(name => ({
-            Id: name,
+          t$DistinguishedFolderId: DistinguishedIDs.map(DistinguishedID => ({
+            Id: DistinguishedID,
             t$Mailbox: {
               t$EmailAddress: person.emailAddress,
             },
--- a/app/frontend/Settings/AccountGeneral.svelte
+++ b/app/frontend/Settings/AccountGeneral.svelte
@@ -148,7 +148,7 @@
 
   async function checkForShares(person: PersonUID) {
     sharedPerson = person;
-    let shares = await (account as EWSAccount).findNamedFolders(sharedPerson, ["msgfolderroot", "inbox", "contacts", "calendar"]);
+    let shares = await (account as EWSAccount).findSharedFolders(sharedPerson, ["msgfolderroot", "inbox", "contacts", "calendar"]);
     if (!shares.includes("msgfolderroot")) {
       shareAccountDisabled = await $t`The root folder is inaccessible`;
     } else if (account.dependentAccounts().find(other => other.protocol == 'ews' && other.username == sharedPerson.emailAddress)) {


benbucksch:
After restart of Mustang, do the shared address books re-appear, and work, and refresh?
* If not, could you please add that?
* If yes, how does it work - I see addressbook.username = personUID.emailAddress, but I don't know how that works. Does the `folderID` make it work automatically?

> In theory if the entire mailbox has been shared it would be possible to search for all contacts or calendars but I don't know whether that's useful.

Yes, that would be very useful.

NeilRashbrook:
> I think it might be possible but we would have to subscribe to the other individual folders separately.

I double-checked and we can subscribe to other folders separately. I don't know when the best time to do this is, since if we do it in the main account we won't pick up new shared folders but if we do it in each addressbook/calendar then we'll create a separate subscription for each folder which might be excessive. Well, I guess we could combine the two, so that existing shared folders are all subscribed in one go at at startup...

app/logic/Calendar/EWS/EWSCalendar.ts
```diff
@@ -11,13 +11,15 @@ import type { ArrayColl } from "svelte-collections";
...
 export class EWSCalendar extends Calendar {
   readonly protocol: string = "calendar-ews";
-  declare readonly events: ArrayColl<EWSEvent>;
+  readonly events: ArrayColl<EWSEvent>;
   /** Exchange's calendar can only accept incoming invitations from its inbox */
   readonly canAcceptAnyInvitation = false;
   /** Exchange FolderID for this calendar. Not DistinguishedFolderId */
   folderID: string;
   /** Is this the default calendar that handles incoming invitations */
   isInvitationCalendar: boolean = false;
```
benbucksch:
```suggestion
  useForInvitations: boolean = false;
```

app/logic/Calendar/EWS/EWSCalendar.ts
```diff
@@ -11,13 +11,15 @@ import type { ArrayColl } from "svelte-collections";
...
+  /** If this is a shared calendar, should we subscribe to its notifications? */
+  needsStreamingSubscription: boolean = true;
```
benbucksch: Can you write the code without this?

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -33,6 +34,7 @@ export class EWSAccount extends MailAccount {
   readonly folderMap = new Map<string, EWSFolder>;
   throttle = new Throttle(50, 1);
   semaphore = new Semaphore(20);
+  folder: "msgfolderroot" | "inbox" | null;
```
benbucksch: Please document the 3 states here.

benbucksch: Actually, can you add a separate `isSharedFolder: boolean = false`?
When I see folder == "inbox", of course I think it's my inbox.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -553,40 +606,48 @@ export class EWSAccount extends MailAccount {
...
-          if (!isMainAddressbook && folder.DisplayName) {
-            addressbook.name = `${this.name} ${folder.DisplayName}`;
-          }
-          addressbook.folderID = sanitize.nonemptystring(folder.FolderId.Id);
-          await addressbook.save();
+    if (this.folder) {
+      return; // Don't automatically add shared addressbook or calendar.
+    }
+    for (let folder of ensureArray(result.RootFolder.Folders.ContactsFolder)) {
+      if (folder.FolderClass == "IPF.Contact" && [undefined, "contacts"].includes(folder.DistinguishedFolderId)) {
```
benbucksch: [undefined, "contacts"].includes(folder.DistinguishedFolderId) need a code comment why you do that. What you do filter out here?

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -553,40 +606,48 @@ export class EWSAccount extends MailAccount {
...
+        let addressbook = appGlobal.addressbooks.find((addressbook: EWSAddressbook) => addressbook.mainAccount == this && addressbook.folderID == folder.FolderId.Id) as EWSAddressbook | undefined;
+        if (!addressbook) {
+          addressbook = newAddressbookForProtocol("addressbook-ews") as EWSAddressbook;
+          addressbook.name = `${this.name} ${folder.DisplayName}`;
```
benbucksch: if main account, continue to use just the name of the main email account.
Same for calendar.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -597,10 +658,18 @@ export class EWSAccount extends MailAccount {
   }
 
   async createToplevelFolder(name: string): Promise<EWSFolder> {
+    if (this.folder == "inbox") {
+      throw new Error(gt`You only have access to the Inbox of these shared folders`);
```
```suggestion
      throw new Error(gt`You only have access to the Inbox of this shared account`);
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -597,10 +658,18 @@ export class EWSAccount extends MailAccount {
...
+    }
     let request = {
       m$CreateFolder: {
         m$ParentFolderId: {
-          t$DistinguishedFolderId: {
+          t$DistinguishedFolderId: this.folder ? {
```
benbucksch: Can you please put the `? {` and `: {` at the beginning of the line? That makes it clearer where and what the cases are.

Same everywhere.

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Actually, just rename this property to `this.sharedFolder`, and document the 3 states.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -617,6 +686,136 @@ export class EWSAccount extends MailAccount {
     folder.id = sanitize.nonemptystring(result.Folders.Folder.FolderId.Id);
     return folder;
   }
+
+  async findNamedFolders(person: PersonUID, names: string[]): Promise<string[]> {
```
benbucksch: Please make a  JSDoc for the function, explaining what "named folders" are.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -617,6 +686,136 @@ export class EWSAccount extends MailAccount {
...
+    };
+    try {
+      let result = await this.callEWS(request);
+      return result.filter(folder => folder.ResponseClass == "Success").map(folder => getEWSItem(folder.Folders).DistinguishedFolderId);
+    } catch (ex) {
+      return [];
+    }
+  }
+
+  async addSharedFolders(person: PersonUID, folder: "msgfolderroot" | "inbox") {
```
```suggestion
  async addSharedUser(person: PersonUID, folder: "msgfolderroot" | "inbox") {
```

benbucksch: Document that this is called from the UI on setup/settings, not automatically on login.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -617,6 +686,136 @@ export class EWSAccount extends MailAccount {
...
+  }
+
+  fromConfigJSON(json: any) {
+    super.fromConfigJSON(json);
+    this.folder = sanitize.enum(json.folder, ["msgfolderroot", "inbox"], null);
+  }
+
+  toConfigJSON(): any {
+    let json = super.toConfigJSON();
+    json.folder = this.folder;
```
```suggestion
    json.sharedFolder = this.folder;
```

same above

benbucksch (changes requested):

PR #900, opened 2025-09-25
Android: Run Node.js based on native App lifecycle

### LINKED ISSUE #884: Mobile: Manage the backend according to native app lifecyle
We currently have issues where:

1. Frontend is loaded but backend is not
2. Backend is still running even when app is not in view

This is because we're not loading Node.js according to the native app lifecyle but rather just loading Node.js only.

I've attempted to fix this in https://github.com/mustang-im/capacitor-nodejs/pull/2 and https://github.com/mustang-im/capacitor-nodejs/pull/3 but both were unsuccessful.

jermy-c:
- Run Node.js based on native App lifecycle
- Remove Capacitor-nodejs dependency
- Capacitor Bridge is not used anymore since Node.js runs a local server

### DIFF
--- a/.github/workflows/installer-android-parula.yml
+++ b/.github/workflows/installer-android-parula.yml
@@ -36,7 +36,6 @@ jobs:
     - run: cd backend; yarn install
     - run: cd mobile; yarn install
     - run: cd mobile/backend; yarn install
-    - run: cd mobile; yarn build
 
     - run: |
         mkdir -p ${{ github.workspace }}/private_keys/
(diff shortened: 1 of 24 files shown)


benbucksch:
I built it on github CI and installed it on a real Android device (where Mustang Android from master is already running well), and it's crashing on startup.

jermy-c:
Sorry, I haven't written the code to download the libnode to the project. And there's also a linking error I'm seeing whenever the libnode is missing you have to open Android Studio to relink it.

Also this got broken again since 57843e62e4bcd4797894c3db67ab754d7fc974b2 because I was trying to reuse the V8 instance so when the app pauses we can stop node but not reinitialize V8.

jermy-c:
The functions to start and stop Node.js work now but both are too slow.

When the startup is too slow, it causes the frontend to throw an error stating that the backend is not present.

When stopping Node.js is too slow, it causes the app to crash because the old Node.js instance is still present. But this only happens when tap the back button to exit and quickly open the app again.

PR #902, opened 2025-09-29
OWA: Open other users' shared mailbox, contacts and calendars #403

NeilRashbrook:
This is the OWA equivalent of #898, and as such, is based on #838.

The shared addressbooks and calendar won't automatically refresh, because I don't know how to do that for OWA.

### DIFF
--- a/app/frontend/Settings/Mail/SharingIncoming.svelte
+++ b/app/frontend/Settings/Mail/SharingIncoming.svelte
@@ -76,7 +76,6 @@
   import CloseIcon from "lucide-svelte/icons/x";
   import { assert } from "../../../logic/util/util";
   import { gt, t } from "../../../l10n/l10n";
-  import { ArrayColl } from "svelte-collections";
 
   export let account: Account;
   $: sharedWith = getAllAccounts().filterObservable(other => other.protocol == account.protocol && other.mainAccount == account);
@@ -125,6 +124,7 @@
 
   async function onAddPerson(person: PersonUID) {
     assert(account instanceof EWSAccount || account instanceof OWAAccount, "Not supported");
+    sharedPerson = null;
     if (sharedFolders.includes("msgfolderroot")) {
       await account.addSharedFolders(person, "msgfolderroot");
     } else if (sharedFolders.includes("inbox")) {
@@ -136,7 +136,6 @@
     if (sharedFolders.includes("calendar")) {
       await account.addSharedCalendar(person);
     }
-    sharedPerson = null;
   }
 
   function onCloseAddDialog() {
--- a/app/logic/Mail/OWA/OWAAccount.ts
+++ b/app/logic/Mail/OWA/OWAAccount.ts
@@ -288,7 +288,8 @@ export class OWAAccount extends MailAccount {
 
   async callOWA(aRequest: any, { mailbox = null } = {}): Promise<any> {
     if (this.mainAccount) {
-      return (this.mainAccount as OWAAccount).callOWA(aRequest, { mailbox: this.username });
+      let mainAccount = this.mainAccount as OWAAccount;
+      return await mainAccount.callOWA(aRequest, { mailbox: this.username });
     }
     if (!this.hasLoggedIn) {
       throw new LoginError(null, "Please login");
@@ -300,9 +301,9 @@ export class OWAAccount extends MailAccount {
         Action: aRequest.action,
         Authorization: this.authorizationHeader,
         "Content-Type": "application/json",
-        "x-anchormailbox": mailbox || this.emailAddress,
+        "x-anchormailbox": mailbox ?? this.emailAddress,
         "x-customowascenariodata": "MailboxAccess:SharedMailbox,ExplicitLogon",
-        "x-owa-explicitlogonuser": mailbox || this.emailAddress,
+        "x-owa-explicitlogonuser": mailbox ?? this.emailAddress,
       },
       method: "POST",
     };


app/logic/Calendar/OWA/OWACalendar.ts
```diff
@@ -26,6 +26,10 @@ export class OWACalendar extends Calendar {
     return this.mainAccount as OWAAccount;
   }
 
+  get mailbox(): string | undefined {
```
benbucksch: What is this? Why it is needed? Why is it called "mailbox" when it returns the username? Can you please rename it to reflect its meaning? Like e.g.


```suggestion
  /** If this is a calendar of another user, return the username of that other user account.
   * In the normal case, when this is the calendar of our user, returns undefined. */
  get sharedAccountUsername(): string | undefined {
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -48,6 +49,7 @@ export class OWAAccount extends MailAccount {
   protected notifications: OWANotifications;
   throttle = new Throttle(50, 1);
   semaphore = new Semaphore(20);
+  folder: "msgfolderroot" | "inbox" | null;
```
benbucksch: See discussion in EWS. #898

```suggestion
  sharedFolderRoot: "msgfolderroot" | "inbox" | null;
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -230,13 +253,21 @@ export class OWAAccount extends MailAccount {
     await this.callOWA(request);
   }
 
-  async callOWA(aRequest: any): Promise<any> {
+  async callOWA(aRequest: any, aMailbox?: string): Promise<any> {
```
```suggestion
  /**
   * options
   *   aUsername:
   *     If this request is for a shared account: the username of that account.
   *     Otherwise: undefined.
   */
  async callOWA(aRequest: any, options: { aUsername?: string }): Promise<any> {
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -230,13 +253,21 @@ export class OWAAccount extends MailAccount {
...
+    if (this.mainAccount) {
+      return await (this.mainAccount as OWAAccount).callOWA(aRequest, this.username);
```
benbucksch: I don't understand this. If `this.mainAccount` is set, then this is a dependent account. Could be a shared account, or a calendar, or something else. If this is a shared account, why use `this.username` and not `aMailbox` (= `options.aUsername`) ?

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -230,13 +253,21 @@ export class OWAAccount extends MailAccount {
...
       throw new LoginError(null, "Please login");
     }
     let url = this.url + 'service.svc';
     let options = {
-      headers: {
+      headers: aMailbox ? {
+        Action: aRequest.action,
+        "x-anchormailbox": aMailbox,
+        "x-customowascenariodata": "MailboxAccess:SharedMailbox,ExplicitLogon",
+        "x-owa-explicitlogonuser": aMailbox,
```
benbucksch: So both these are the shared user, not my user? explicitlogonuser suggests to me that this is the user I logged in with, i.e. my user, not the shared user.

app/logic/Mail/OWA/OWAFolder.ts
```diff
@@ -191,6 +191,16 @@ export class OWAFolder extends Folder {
     return this.messages.find((m: OWAEMail) => m.itemID == id) as OWAEMail | undefined;
   }
 
+  protected async moveOrCopyMessages(action: "move" | "copy", messages: Collection<EMail>): Promise<boolean> {
+    // We can copy messages to and from shared folders for the main account,
+    // but the messages all have to be from the same account.
+    let sourceAccount = messages.first.folder.account;
+    if ((sourceAccount.mainAccount || sourceAccount) == (this.account.mainAccount || this.account) && messages.contents.every(msg => msg.folder.account = sourceAccount)) {
```
benbucksch:
```suggestion
    if ((sourceAccount.mainAccount ?? sourceAccount) == (this.account.mainAccount ?? this.account) &&
        messages.contents.every(msg => msg.folder.account == sourceAccount)) {
```

You used `=` instead of `==`. Phew, close one!

(Nits: Used `??` instead of `||` and added a linebreak)

Why is this condition different from the one in EWS #898 ? Please check every difference (I found at least 3 differences), and see whether they should be made the same.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -420,6 +459,82 @@ export class OWAAccount extends MailAccount {
       this.errorCallback(ex);
     }
   }
+
+  async findNamedFolders(person: PersonUID, names: string[]): Promise<string[]> {
```
benbucksch: Please apply all review comments from #898 also here in this PR.

In this case:
Please make a JSDoc for the function, explaining what "named folders" are.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -350,7 +386,10 @@ export class OWAAccount extends MailAccount {
   }
 
   async createToplevelFolder(name: string): Promise<OWAFolder> {
-    let result = await this.callOWA(owaCreateNewTopLevelFolderRequest(name));
+    if (this.folder == "inbox") {
+      throw new Error(gt`You only have access to the Inbox of these shared folders`);
```
```suggestion
      throw new Error(gt`You only have access to the Inbox of this shared account`);
```

benbucksch:
@NeilRashbrook Please let me know when this is ready, i.e. merged to master, review comments here done, all review comments from #898 done.

For the UI, please merge to master. The UI on master is completely re-written, so simply throw away the UI here. Adapting the UI on master should be straight-forward: In Sharing*.svelte, search for `EWSAccount` and add `OWAAccount` as valid case. Then, in SettingsCategories.ts, add:
```
accountSettings.add(new AccSetting(OWAAccount, "mail-sharing", gt`Sharing *=> Accessing mail account of team mates`, MailSharing));
```
below the `EWSAccount` line.

app/logic/Calendar/OWA/OWACalendar.ts
NeilRashbrook: This is needed because OWA requests need to provide the mailbox [Exchange term] of the account handling the request. (Hotmail seems to require it on all requests; I have no idea whether it supports sharing. Office 365 requires it on all requests for shared mailboxes, but accepts it on all requests. I don't know what on-premises needs, but it accepts it anyway.)

It would be wrong to call it the username, as it needs to be the user's primary SMTP address, even when they log in with DOMAIN\User. That's my fault for reusing the username property. It should probably be a separate property saved in the JSON.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: I'm going to have to expand on this comment. Or maybe rewrite the code. These are the cases:
- An addressbook or calendar sets this on its requests if it's a shared addressbook or calendar
- Folders do not set this on their requests
- Shared folders call their shared account's callOWA, which then sets this when it forwards to the main account

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: None of the folder (or email) functions set the mailbox in their calls to this function, but if they are shared folders then they actually invoke it on the shared account, which then needs to forward the call to the main account (which is responsible for authentication etc.) but telling the main account to use its mailbox.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: For once, a bad name that isn't my fault!

app/logic/Mail/OWA/OWAFolder.ts
NeilRashbrook: Thanks for spotting the typo.

As the comment says, in this case, all of the messages need to be from the same source account. Both source and target accounts have to have the same main account (or themselves if they are the main account).

app/frontend/Settings/Mail/SharingIncoming.svelte
```diff
@@ -11,14 +11,14 @@
     {#if $sharedWith.isEmpty}
       <hbox class="nothing">{$t`You are not accessing other accounts`}</hbox>
     {:else}
-      {#each $sharedWith.each as otherUser}
+      {#each $sharedWith.each as otherAccount}
         <hbox class="existing-person">
-          <hbox class="name" flex>{otherUser.name}</hbox>
+          <hbox class="name" flex>{otherAccount.identities.first.realname}</hbox>
```
benbucksch, line 16: This in a few `?` for good measure and throw-avoidance?

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: odd

benbucksch (changes requested):

NeilRashbrook:
I've added the "option(s)" argument as requested.

PR #906, opened 2025-10-02
EWS: PoC of searching messages on the server

NeilRashbrook:
This is implemented only for the per-folder search at the bottom of the message list because Exchange does not support deep search traversal.

This is implemented only for EWS, because:
- ActiveSync only has a free-text and/or message date received search, plus it has custom IDs for search results that cannot be mapped onto the original message. Thanks, Microsoft!
- OWA does not use `FindItem` for searching, so as yet I don't even know whether you can specify search terms. Instead it uses its own free-text based search. For Office 365 you can at least search for unread messages (this works by appending ` isRead:false` to the search keyword).

### DIFF
--- a/app/frontend/Mail/LeftPane/FolderFooter.svelte
+++ b/app/frontend/Mail/LeftPane/FolderFooter.svelte
@@ -133,19 +133,7 @@
       searchMessages = null;
       return;
     }
-    searchMessages = folder.messages.filterOnce(msg =>
-      (!isShowStarred || msg.isStarred === true) &&
-      (!isShowUnread || msg.isRead === false) &&
-      (!searchTerm || searchTerm.length > 1) &&
-      (!searchTerm ||
-        msg.subject?.toLowerCase().includes(searchTerm) ||
-        msg.contact?.name?.toLowerCase().includes(searchTerm) ||
-        msg.from?.name?.toLowerCase().includes(searchTerm) ||
-        msg.from?.emailAddress?.toLowerCase().includes(searchTerm) ||
-        msg.to?.some(to => to.name?.toLowerCase().includes(searchTerm) ||
-          to.emailAddress?.toLowerCase().includes(searchTerm)) ||
-        msg.text?.toLowerCase().includes(searchTerm))
-    ) as ArrayColl<EMail>;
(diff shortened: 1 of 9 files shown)


NeilRashbrook:
Looking into searching with OWA, I discovered that it uses a FindItem request to search for deleted contacts. I was eventually able to adapt this request into a request to search messages for starred, unread or keywords.

app/frontend/Mail/LeftPane/FolderFooter.svelte
```diff
@@ -133,19 +133,7 @@
...
-      (!searchTerm ||
-        msg.subject?.toLowerCase().includes(searchTerm) ||
-        msg.contact?.name?.toLowerCase().includes(searchTerm) ||
-        msg.from?.name?.toLowerCase().includes(searchTerm) ||
-        msg.from?.emailAddress?.toLowerCase().includes(searchTerm) ||
-        msg.to?.some(to => to.name?.toLowerCase().includes(searchTerm) ||
-          to.emailAddress?.toLowerCase().includes(searchTerm)) ||
-        msg.text?.toLowerCase().includes(searchTerm))
-    ) as ArrayColl<EMail>;
+    searchMessages = await folder.searchMessages(isShowStarred, isShowUnread, searchTerm);
```
benbucksch, line 136: This is quick search. It's searching only the displayed emails and therefore extremely fast. It matches the Thunderbird quick search.

The API to implement is `SearchEMail`. This is searches the entire DB. Now it should also search the server. Please implement that API.

I understand that Exchange can search only 1 folder at a time. We need to discuss how to deal with that.

benbucksch (changes requested):

app/frontend/Mail/LeftPane/FolderFooter.svelte
NeilRashbrook: I wanted to try it out on a search query that I knew I would be able to code so I implemented it for tags but unfortunately it turns out that the tags pane doesn't work at all for some reason.

NeilRashbrook: The tags bug was actually FastList's fault for not showing me the tagged messages. Filed PR #910.

benbucksch:
Compare #1417

PR #909, opened 2025-10-14
iOS: Upload app to GitHub Release & Test Flight

jermy-c:
- Upload the IPA to GitHub Releases for a reference of what was built (working)
- Upload the IPA to TestFlight for testing the app on a real iOS device by installing it and running it (not working yet)
- Output is still a `App.ipa` with the Product name being `App` also. I tried changing it but it breaks the build and project.

### DIFF
--- a/mobile/ios/App/App/Info.plist
+++ b/mobile/ios/App/App/Info.plist
@@ -20,6 +20,8 @@
 	<string>$(MARKETING_VERSION)</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>ITSAppUsesNonExemptEncryption</key>
+	<false/>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>NSAppTransportSecurity</key>


benbucksch:
How does the capacitor node add-on solve this?

.github/workflows/installer-ios-parula.yml
```diff
@@ -53,9 +53,39 @@ jobs:
...
 
+      - name: Find IPA file and set path
+        run: |
+          IPA_FILE=$(find mobile/output -name "*.ipa" -type f | head -1)
+          if [ -z "$IPA_FILE" ]; then
+            echo "Error: No IPA file found in mobile/output/"
+            exit 1
+          fi
+          echo "Found IPA file: $IPA_FILE"
+          echo "IPA_FILE_PATH=$IPA_FILE" >> $GITHUB_ENV
```
benbucksch, line 64: Why is this line needed? What's the purpose?

What does this line do, technically? Please be detailed.

Also, does it append to github_env, or replace it? Can you please double-check that it does what you intended?

jermy-c: This is because the `softprops/action-gh-release@v2` doesn't take a glob and you need to provide the exact filename to the GitHub action. So finds the `IPA` file and get's the filename and appends it to the `GITHUB_ENV` because `export [ENV]=VALUE` doesn't work.

PR #910, opened 2025-10-15
FastList doesn't update properly when its items change

NeilRashbrook:
The `updateSize` that triggers when `items` changes never does anything because there are no content rows at this point. This means that when the `onScroll` triggers, `hasRows` is greater than `items.length`, resulting in a silly negative `startPos`.

While I was there I noticed that `scrollIntoView` was completely broken so I rewrote it.

Since `onScroll` couldn't rely on `showRows` being no greater than `items.length`, I thought I might as well remove the dependency and always have `showRows` be defined by the visible size only.

### DIFF
--- a/app/frontend/Shared/FastList.svelte
+++ b/app/frontend/Shared/FastList.svelte
@@ -156,7 +156,7 @@
       rowHeight = contentRow.offsetHeight;
       let availableHeight = listE.offsetHeight - headerE.offsetHeight;
 
-      showRows = Math.min(items.length, Math.ceil(availableHeight / rowHeight));
+      showRows = Math.ceil(availableHeight / rowHeight);
       //console.log("size", "contentrow", contentRow.offsetHeight, "list", listE.offsetHeight, "header", headerE.offsetHeight, "rowheight", rowHeight, "available", availableHeight, " showrows", showRows);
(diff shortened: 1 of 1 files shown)


NeilRashbrook:
> With your fix here, did you consider in the code, and test in practice, the following situations?
> 
> * No items

Seems to work fine.

> * Fewer items in list than there is space for (e.g. folder with only 3 emails)

This is actually the case I was trying to fix!

> * Need to listen to list changes and update the list display when items are added or removed.
> 
> 
> There's a special snowflake:
> 
> * Account list should be as small (in height) as possible, and its height needs to grow to the number of accounts, up to the

Thanks, I'll check these, both without and with my patch, to see if there is any difference in behaviour.

NeilRashbrook:
Note: While testing I did find an intermittent bug in my patch...

> * Need to listen to list changes and update the list display when items are added or removed.

I couldn't find any problems adding or removing items. (When I accidentally used the "withoutObserver" versions by mistake, my patch was slightly less broken than the current code, but I have no idea why.)

> * Account list should be as small (in height) as possible, and its height needs to grow to the number of accounts, up to the

I don't know how far the hight needs to grow to, but I don't have enough accounts to trigger it, either with or without the patch.

PR #912, opened 2025-10-15
iOS: Rename the product name and IPA output

jermy-c:
- Rename the IPA for iOS from `App.ipa` to `Mustang_[Version].ipa`
- Rename the App Product Name for iOS

### DIFF
--- a/mobile/ios/build/build.sh
+++ b/mobile/ios/build/build.sh
@@ -21,7 +21,6 @@ xcodebuild -workspace './ios/App/App.xcworkspace' -scheme App -destination gener
 # Export the archive
 xcodebuild archive -archivePath App.xcarchive -exportArchive -exportOptionsPlist ./ios/archive.plist -exportPath output -allowProvisioningUpdates
 
-cd mobile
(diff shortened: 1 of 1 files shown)


mobile/ios/App/App.xcodeproj/project.pbxproj
```diff
@@ -175,7 +175,7 @@
 			);
 			name = App;
 			productName = App;
```
benbucksch, line 177: Should these two change to "Mustang" as well?

benbucksch (changes requested):

mobile/ios/App/App.xcodeproj/project.pbxproj
jermy-c: No, that change would break the project.

Because the Pod file expects a Target name `App`. And I don't what to change the Pod file because it's how Capacitor manages its plugins.

https://github.com/mustang-im/mustang/blob/4145c8c16e9ad601572473ad0e1ccb13fe5a5044/mobile/ios/App/Podfile#L21

benbucksch: OK

PR #914, opened 2025-10-17
Build: Replace icon.svg with icon.png - fixes #913

### LINKED ISSUE #913: Build: build/icon.png?asset not found
### Reproduction
1. Check out a fresh source tree
2. Follow the build instructions in docs/INSTALL.md

### Actual result

App does not start
```
 ERROR  Could not resolve "../../build/icon.png?asset" from "src/main/index.ts"
```

### Expected result
Application runs, without errors.

jermy-c:
- Replace icon.svg with icon.png
- Scaled the icon a little so it is not too big
- Fixes #913, since it was expecting an icon.png instead of an icon.svg

Running on Mac as an iOS app through Xcode
image

Android
image

MacOS
image


benbucksch:
Can you please not delete the SVG? Otherwise good.

benbucksch:
On a Samsung device, the circle around Mustang is now touching the icon border, whereas there was a slight gap before.

The icons in your screenshots look fine and correct and should stay as-is.

PR #917, opened 2025-10-23
iOS: Build Node Natives as Frameworks

jermy-c:
- Build Node Natives as Frameworks

### DIFF
--- a/mobile/ios/build/ios-create-plists-and-dlopen-override.js
+++ b/mobile/ios/build/ios-create-plists-and-dlopen-override.js
@@ -1,4 +1,11 @@
-// From: <https://github.com/nodejs-mobile/nodejs-mobile-cordova/blob/6cd38b0abc86f0acdbc3e899b36937abf906b119/install/helper-scripts/ios-create-plists-and-dlopen-override.js>
+/**
+ * This was imported from
+ * https://github.com/nodejs-mobile/nodejs-mobile-cordova/blob/6cd38b0abc86f0acdbc3e899b36937abf906b119/install/helper-scripts/ios-create-plists-and-dlopen-override.js
+ * and modified
+ *
+ * MIT license
+ * Copyright (c) 2018 Janea Systems, Inc. and Contributors
+ */
 
 import fs from 'fs';
 import path from 'path';
--- a/mobile/ios/build/override-dlopen-paths-preload.js
+++ b/mobile/ios/build/override-dlopen-paths-preload.js
@@ -1,4 +1,11 @@
-// From: <https://github.com/nodejs-mobile/nodejs-mobile-cordova/blob/6cd38b0abc86f0acdbc3e899b36937abf906b119/install/helper-scripts/override-dlopen-paths-preload.js>
+/**
+ * This was imported from
+ * https://github.com/nodejs-mobile/nodejs-mobile-cordova/blob/6cd38b0abc86f0acdbc3e899b36937abf906b119/install/helper-scripts/override-dlopen-paths-preload.js
+ * and modified
+ *
+ * MIT license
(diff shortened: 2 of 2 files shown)


mobile/docs/structure.md
```diff
@@ -14,7 +14,16 @@ The variable for libnode is set in `hooks/common/variables.sh` and used in `hook
 
 #### Node Native
 
-Node native modules should not be included in the Frameworks section of `App/Target` setting in Xcode.
+Node native modules should not be included in the Frameworks section of `App/Target` setting in Xcode. Node native modules need to be converted to frameworks and embedded into the app using the script `ios/build/xcode-embed-node-native.sh` which should be run in Xcode because it needs the environment variables set in Xcode. If the frameworks are not embedded in the app, the app will not be accepted by the App Store or TestFlight. However, when you embed the frameworks the path and the file name changes so you need to override the dlopen calls at runtime to load the correct framework. The `override-dlopen-paths-preload.js` script is used to override the dlopen calls at runtime to load the correct framework.
```
benbucksch:
```suggestion
Node native modules should not be included in the Frameworks section of `App/Target` setting in Xcode.
Node native modules need to be converted to frameworks and embedded into the app using
the script `ios/build/xcode-embed-node-native.sh`, which should be run in Xcode, because it
needs the environment variables set in Xcode. If the frameworks are not embedded in the app,
the app will not be accepted by the App Store or TestFlight. However, once you embed the
frameworks, the path and the file name will change, so you need to override the `dlopen()` calls
at runtime to load the correct framework. The `override-dlopen-paths-preload.js` script is used
for that.
```

a) added linebreaks
b) addec missing commas where needed - it was very hard to understand without them
c) rephrased the "once you embed" sentence
c) removed duplication in last sentence

mobile/ios/App/App/NodeRunner.swift
```diff
@@ -38,10 +38,21 @@ import NodeMobile
...
                 }
+            
+                let overrideDLOpenScript = "override-dlopen-paths-preload.js"
+                // —— attempt to find override-dlopen-paths-preload.js ——
+                guard let overrideDLOpenPath = Bundle.main.path(
+                        forResource: "override-dlopen-paths-preload",
+                        ofType: "js",
+                        inDirectory: "nodejs-project"
+                      ) else {
+                    print("❌ NodeRunner: could not find override-dlopen-paths-preload.js in nodejs-project")
```
benbucksch, line 49: Is there a specific command to print error messages? I think there is.

mobile/ios/build/ios-create-plists-and-dlopen-override.js
```diff
@@ -0,0 +1,140 @@
...
+          }
+        }
+        recursivelyFindFrameworks(currentFilename);
+      }
+    }
+  }
+  recursivelyFindFrameworks(projectPath);
+
+  console.log("Found " + countValidFrameworks + " valid frameworks and " + countInvalidFrameworks + " invalid frameworks after rebuilding the native modules for iOS.");
+  if (foundFrameworks.length<1) {
```
benbucksch: Code formatting:
Add spaces around operators: ` < `
Same below, many places: ...`projectPath, cur`...

mobile/docs/structure.md
benbucksch: @jermy-c ping

mobile/ios/App/App/NodeRunner.swift
jermy-c: There's no `console.error` equivalent in Swift. It's just the normal `print()`.

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/errorhandling/

PR #926, opened 2025-11-05
Mobile: Debug log

jermy-c:
- Add Debug option in Settings for seeing debug logs on Mobile where Dev tools are not available

### DIFF
--- a/app/frontend/Settings/About/Debug.svelte
+++ b/app/frontend/Settings/About/Debug.svelte
@@ -28,7 +28,7 @@
       const seen = new WeakSet();
       return JSON.stringify(obj, (key, value) => {
         if (key && key.startsWith("_")) {
-          return "";
+          return undefined;
         }
         if (value == null) { // or undefined
           return "";


app/logic/util/logHistory.ts
```diff
@@ -0,0 +1,15 @@
+import { ArrayColl } from "svelte-collections";
```
benbucksch, line 1: Move this file to `logic/util/logHistory.ts` (camelCase), please.

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,53 @@
+<vbox>
+  <h1>Debug</h1>
+  <button on:click={() => logHistory.clear()}>Clear log history</button>
```
benbucksch: Move this button to the bottom of the screen, please

Also, please add a button to copy all contents to clipboard (we have a few other places where we copy to clipboard). This "Copy to clipboard" button is above the "Clear log history" button.

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,53 @@
+<vbox>
+  <h1>Debug</h1>
```
benbucksch: Localize the label, please. We might want to enable that in production, maybe hidden behind a another button. Ditto button labels etc.

app/frontend/Settings/SettingsCategories.ts
```diff
@@ -155,3 +157,7 @@ about.subCategories.addAll([
   new SettingsCategory("opensource", gt`Open-Source`, OpenSource),
 ]);
 settingsCategories.add(about);
+
+if (!production) {
+  about.subCategories.add(new SettingsCategory("debug", "Debug", Debug));
```
benbucksch: Localize

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,53 @@
...
+          if (value instanceof Node) {
+            return `[${value.nodeName || "Node"}]`;
+          }
+          if (seen.has(value)) {
+            return "[Circular]";
+          }
+          seen.add(value);
+        }
+        return value;
+      });
```
benbucksch, line 51: And if it's not an object, but a string or number?

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,53 @@
...
+  import { logHistory } from "./LogHistory";
+
+  function safeStringify(obj: any): string {
+    try {
+      const seen = new WeakSet();
+      return JSON.stringify(obj, (key, value) => {
+        if (key && key.startsWith("_")) {
+          return undefined;
+        }
+        if (typeof value === "object" && value !== null) {
```
benbucksch:
```suggestion
        if (value == null) { // or undefined
          return;
        } else if (typeof value === "object") {
```

benbucksch (changes requested):

app/frontend/Settings/About/Debug.svelte
jermy-c: It is returned as-is and they can be converted to strings, it's the objects that have been causing errors.

jermy-c: Correction... There seems to be an error if there's a `undefined` or `null`.

benbucksch: my code suggestion above should fix that

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,76 @@
+<vbox>
+  <h1>{$t`Debug`}</h1>
+  <div class="log-history" bind:this={logHistoryEl}>
```
benbucksch:
```suggestion
  <div class="log-history selectable" bind:this={logHistoryEl}>
```

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,76 @@
...
+    {#each $logHistory.each as log, i}
+      <div class="selectable">{i}: {log.map((v) => typeof v === "object" && v !== null ? safeStringify(v) : String(v)).join(" ")}</div>
```
benbucksch:
```suggestion
      <div>{i}: {log.map((v) => typeof v === "object" && v !== null ? safeStringify(v) : String(v)).join(" ")}</div>
```

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,76 @@
+<vbox>
+  <h1>{$t`Debug`}</h1>
+  <div class="log-history selectable" bind:this={logHistoryEl}>
+    {#each $logHistory.each as log, i}
+      <div>{i}: {log.map((v) => typeof v === "object" && v !== null ? safeStringify(v) : String(v)).join(" ")}</div>
```
benbucksch:
```suggestion
      <div>{i}: {log?.map(v => safeStringify(v)).join(" ")}</div>
```

benbucksch: Please see whether that works now

benbucksch (changes requested):

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,76 @@
...
+  import { logHistory } from "../../../logic/util/logHistory";
+  import { catchErrors } from "../../Util/error";
+  import { t } from "../../../l10n/l10n";
+
+  function safeStringify(obj: any): string {
+    try {
+      const seen = new WeakSet();
+      return JSON.stringify(obj, (key, value) => {
+        if (key && key.startsWith("_")) {
+          return undefined;
```
benbucksch:
```suggestion
          return "";
```

app/frontend/Settings/About/Debug.svelte
```diff
@@ -0,0 +1,76 @@
...
+        }
+        if (value == null) { // or undefined
+          return;
```
benbucksch:
```suggestion
          return "";
```

benbucksch (changes requested):

app/frontend/Settings/About/Debug.svelte
benbucksch: Update: added `.?` in case of `console.log(null)`

app/frontend/Settings/About/Debug.svelte
jermy-c: I tried this earlier and doesn't make the child elements selectable.

app/frontend/Settings/About/Debug.svelte
jermy-c: I change it to `ensureArray(log).map` because you will still get an error if it's a single element. e.g. `console.log(5)`

Because you can't apply the `.map()` method.

jermy-c:
The debugger works now. It was tested with:

```js
  onMount(() => {
    logHistory.addAll([1, 2, 3, "sdf", { a: 1, b: 2, c: 3 }, { d: "sdf" }, null, undefined, { e: undefined }, { f: null }, new Set([1, 2, 3]), new Map([["a", 1], ["b", 2], ["c", 3]]), new Date(), new Error("test"), new Event("test")]);
  });
```

app/frontend/Settings/About/Debug.svelte
jermy-c: This makes the `JSON.stringify()` also print out the original key and an empty string `""` value for those starting with `_`

benbucksch (approved):

NeilRashbrook:
Is there a way to stop this being loaded in my development build? It's annoying to see all of my `console.log` commands being reported as from `logHistory.ts`.

PR #928, opened 2025-11-06
FastList: Change Grid height from min-content to auto

jermy-c:
- On Safari Browsers the FastList items were only taking the space of the screen and the items were not readable
- Changing the grid height to auto makes it take up the space it needs to display the items properly


benbucksch (approved):
Thank you!

PR #933, opened 2025-11-12
Build: Don't mark `-dev` builds as latest

jermy-c:
- If the version number ends with `-dev` lets not mark it as latest, this messes up our auto update for Desktop

### DIFF
--- a/.github/workflows/installer-ios-parula.yml
+++ b/.github/workflows/installer-ios-parula.yml
@@ -86,7 +86,7 @@ jobs:
         with:
           tag_name: ${{ env.APP_VERSION }}
           make_latest: ${{ env.RELEASE }}
-          prerelease: ${{ env.RELEASE != 'true' }}
+          prerelease: ${{ env.RELEASE == 'false' }}
           files: ${{ env.IPA_FILE_PATH }}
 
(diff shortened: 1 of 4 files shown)


.github/workflows/installer-ios-parula.yml
```diff
@@ -75,11 +75,18 @@ jobs:
           cd mobile
           VERSION=$(jq -r '.version' package.json)
           echo "APP_VERSION=v$VERSION" >> $GITHUB_ENV
+          if [[ "$VERSION" == *dev ]]; then
+            echo "RELEASE=false" >> $GITHUB_ENV
+          else
+            echo "RELEASE=true" >> $GITHUB_ENV
+          fi
```
benbucksch, line 82: Do you need the same for the desktop builds?

Otherwise the PR is OK for me.

jermy-c: No, on desktop Electron builder takes care of releasing but if we do decide to build test builds on CI then we will need something similar.

benbucksch: > if we do decide to build test builds on CI then we will need something similar.

We may well do test builds on CI.

PR #935, opened 2025-11-18
iOS: Set unique build numbers

jermy-c:
- Set's the build number to `Major.Minor.Unix time in seconds`
- This for uploading builds to TestFlight where the build number needs to be unique and incremented
- Haven't tested if TestFlight accepts this format yet

### DIFF
--- a/app/build/mustang-brand.sh
+++ b/app/build/mustang-brand.sh
@@ -10,7 +10,7 @@ perl -p -i \
   ../../e2/package.json
 
 MAJOR_MINOR=$(echo "$VERSION" | sed 's/^\([0-9]*\.[0-9]*\).*/\1/')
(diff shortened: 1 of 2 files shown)


app/build/mustang-brand.sh
```diff
@@ -9,8 +9,12 @@ perl -p -i \
   -e "s|\"version\": \".*\"|\"version\": \"$VERSION\"|;" \
   ../../e2/package.json
 
+MAJOR_MINOR=$(echo "$VERSION" | sed 's/^\([0-9]*\.[0-9]*\).*/\1/')
+BUILD_VERSION="${MAJOR_MINOR}.$(date +%Y%m%d%H%M%S)"
```
benbucksch, line 13: smart that you use the first 2 components from our actual version number. Thank you!

PR #941, opened 2025-11-21
OWA: Stream Office 365 notifications in the front end

NeilRashbrook:
`@microsoft/fetch-event-source` is an extension of the regular EventSource that allows for custom `fetch` properties; in our case, headers. However the API to consume the events differs from both our in-house event source and the regular EventSource...

### DIFF
--- a/app/logic/Mail/OWA/Notification/OWAExchangeNotifications.ts
+++ b/app/logic/Mail/OWA/Notification/OWAExchangeNotifications.ts
@@ -12,9 +12,9 @@ export class OWAExchangeNotifications extends OWANotifications {
     // This loop only ends by exception (e.g. logout) or app shutdown.
     while (true) {
       url = this.account.url + "ev.owa2?ns=PendingRequest&ev=PendingNotificationRequest&UA=0&cid=" + cid + "&X-OWA-CANARY=";
-      let stream = await appGlobal.remoteApp.OWA.streamJSON(this.account.partition, url);
+      let stream = await appGlobal.remoteApp.OWA.fetchJSON(this.account.partition, url);
       if (!stream.ok) {
-        throw new Error(`streamJSON failed with HTTP ${stream.status} ${stream.statusText}`);
+        throw new Error(`stream fetch failed with HTTP ${stream.status} ${stream.statusText}`);
       }
       for await (let chunk of stream.body) {
         // Avoid racing with ourselves, if we caused the notification.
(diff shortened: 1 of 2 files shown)


app/logic/Mail/OWA/Notification/OWAOffice365Notifications.ts
```diff
@@ -29,37 +29,43 @@ export class OWAOffice365Notifications extends OWANotifications {
...
         url = this.account.url + URLPart`notificationchannel/connect?transport=serverSentEvents&cid=${cid}&connectionToken=${json.ConnectionToken}`;
-        let stream = await appGlobal.remoteApp.OWA.streamEvents(this.account.partition, url, streamOptions);
+        let stream = await fetch(url, streamOptions);
         //console.log("stream", stream.ok ? "OK" : "Failed", "status", stream.status, "statusText", stream.statusText, "stream obj", stream);
         if (!stream.ok) {
-          throw new Error(`streamEvents failed with HTTP ${stream.status} ${stream.statusText}`);
+          throw new Error(`stream fetch failed with HTTP ${stream.status} ${stream.statusText}`);
         }
-        for await (let chunk of stream.body) {
+        for await (let chunk of stream.body.pipeThrough(new TextDecoderStream()).pipeThrough(new TransformStream(new EventDecoder()))) {
```
NeilRashbrook, line 68: Regarding the "must have a '\[Symbol.asyncIterator]()' method" TypeScript error, this happens if:
- your TypeScript version is lower than 5.5
- your target is lower than ES2018

For instance this means that I will always see that error:
- I can't get TypeScript 5.5.x to compile at all because I get error TS6305
- I can't get TypeScript 5.6-7.x to compile because they additionally no longer support `importsNotUsedAsValues` and `preserveValueImports`
- I can't get TypeScript 5.8.x or later to compile because they don't like a file in the `faker` module for some reason

benbucksch: > * your TypeScript version is lower than 5.5
> * your target is lower than ES2018

Is there and AND or OR between those?

NeilRashbrook: OR I'm afraid; you need TypeScript 5.5 or later targeting ES2018 or later for the code to compile.

benbucksch: So, you need TypeScript 5.5+ AND targeting ES2018+

PR #951, opened 2025-12-09
ActiveSync: Detect user calendars and addressbooks again

NeilRashbrook:
Regression from 535dad46969c21cb86b0064d5c18e5dc3f66ead0 which prevented ActiveSync from detecting any user calendars or addressbooks.

This PR restores the previous behaviour, which is:
- All user folders, calendars and addressbooks are detected on account creation
- Creations and renamings of user folders, calendars and addressbooks are detected on folder sync
- User calendar and addressbook accounts can be deleted from mustang, but will reappear if renamed on the server

Additionally I removed all the remaining references to URLs which was the way that user calendars and addressbooks were detected before the server ID was added to the config JSON.

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -460,8 +460,11 @@ export class ActiveSyncAccount extends MailAccount {
               appGlobal.addressbooks.add(addressbook);
             }
             if (!isMainAddressbook) {
+              // Rename and new address books
               addressbook.name = sanitize.nonemptylabel(change.DisplayName, addressbook.name);
             }
+            // ActiveSync doesn't list all folders, only the changed ones.
+            // So, if we're here, then the address book changed.
             await addressbook.save();
             break;
           case FolderType.Calendar:


app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -454,35 +451,33 @@ export class ActiveSyncAccount extends MailAccount {
...
-              }
-              addressbook.url = sanitize.url(folderURL);
               addressbook.serverID = sanitize.nonemptystring(change.ServerId);
-              await addressbook.save();
-              appGlobal.addressbooks.add(addressbook);
             }
+            if (!isMainAddressbook) {
+              addressbook.name = sanitize.nonemptylabel(change.DisplayName, addressbook.name);
+            }
+            await addressbook.save();
```
benbucksch, line 468: Why do we need to save the addressbook on every login?

Calendars and address books should be saved in the DB, including their name, and re-loaded on every startup. They should not need to be created, not need to be saved (unless something actually changed), and not need to be added to `appGlobal.addressbooks`, because it's already in there since load, long before we get there.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -454,35 +451,33 @@ export class ActiveSyncAccount extends MailAccount {
...
-              let calendar = newCalendarForProtocol("calendar-activesync") as ActiveSyncCalendar;
+            let isMainCalendar = change.Type == FolderType.Calendar;
+            let calendar = appGlobal.calendars.find((calendar: ActiveSyncCalendar) => calendar.mainAccount == this && calendar.serverID == change.ServerId) as ActiveSyncCalendar | null;
+            if (!calendar) {
+              calendar = newCalendarForProtocol("calendar-activesync") as ActiveSyncCalendar;
               calendar.initFromMainAccount(this);
-              if (!isMainCalendar) {
-                calendar.name = sanitize.nonemptylabel(change.DisplayName, calendar.name);
-              }
-              calendar.url = sanitize.url(folderURL);
```
benbucksch, line 484: .url should be set. At least as long as long as they are accounts, until #778 lands.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -454,35 +451,33 @@ export class ActiveSyncAccount extends MailAccount {
...
               calendar.serverID = sanitize.nonemptystring(change.ServerId);
-              await calendar.save();
-              appGlobal.calendars.add(calendar);
             }
+            if (!isMainCalendar) {
+              calendar.name = sanitize.nonemptylabel(change.DisplayName, calendar.name);
```
benbucksch, line 481: Why set this on every login? The user will want to change this. You'd overwrite user changes this way.

benbucksch:
> prevented ActiveSync from detecting any user calendars or addressbooks.

That's strange. I tested it and it did add address books and calendars for me.

Did you try with a new account, or an existing one? Please try with a new account.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: We don't; this only happens whenever FolderSync tells us about an address book. However, you're right about adding to `appGlobal.addressbooks`; I forgot to move that line back into the `if (!addressbook)` block.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: `initFromMainAccount` already sets the URL.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: Again, this doesn't happen on every login, only when ActiveSync notifies us about it.

NeilRashbrook (commented):
> > prevented ActiveSync from detecting any user calendars or addressbooks.
> 
> That's strange. I tested it and it did add address books and calendars for me.
> 
> Did you try with a new account, or an existing one?

Trying with an existing account is pointless because the code doesn't run in this case.

> Please try with a new account.

Without the patch only one of the calendars was added. (I didn't check address books because there was only one to add.)

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -457,35 +451,33 @@ export class ActiveSyncAccount extends MailAccount {
...
-              if (!isMainAddressbook) {
-                addressbook.name = sanitize.nonemptylabel(change.DisplayName, addressbook.name);
-              }
-              addressbook.url = sanitize.url(folderURL);
               addressbook.serverID = sanitize.nonemptystring(change.ServerId);
-              await addressbook.save();
               appGlobal.addressbooks.add(addressbook);
             }
+            if (!isMainAddressbook) {
+              addressbook.name = sanitize.nonemptylabel(change.DisplayName, addressbook.name);
```
benbucksch, line 464:
```suggestion
              // Rename and new address books
              addressbook.name = sanitize.nonemptylabel(change.DisplayName, addressbook.name);
```

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -457,35 +451,33 @@ export class ActiveSyncAccount extends MailAccount {
...
+            }
+            await addressbook.save();
```
benbucksch, line 468:
```suggestion
            // ActiveSync doesn't list all folders, only the changed ones.
            // So, if we're here, then the address book changed.
            await addressbook.save();
```

benbucksch (approved):

PR #952, opened 2025-12-10
UI: Remove SvelteUI and replace with native HTML elements

jermy-c:
- Replaced useDebounce and throttle with functions from lodash
- Use native widgets for Checkbox and ViewSelector
- Radio widget not implement yet because it's unused in `app/frontend/Calendar/DisplayEvent/DisplayEvent_unused.svelte`

### DIFF
--- a/app/frontend/Calendar/ViewSelector.svelte
+++ b/app/frontend/Calendar/ViewSelector.svelte
@@ -5,7 +5,7 @@
     {/each}
   </select>
   <hbox class="icon">
-    <ChevronsUpDownIcon size={16} strokeWidth={4}/>
+    <ChevronsUpDownIcon size={16} strokeWidth={2}/>
   </hbox>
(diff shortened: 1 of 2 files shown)


app/frontend/Mail/Search/SearchPane.svelte
```diff
@@ -59,7 +59,7 @@
   import XIcon from "lucide-svelte/icons/x";
   import { showError } from "../../Util/error";
   import { ArrayColl } from "svelte-collections";
-  import { useDebounce } from '@svelteuidev/composables';
+  import debounce from 'lodash/debounce';
```
benbucksch: Can you please make a separate PR for the removal of debounce?

jermy-c: Yes, PR #954 has the removal of debounce.

PR #953, opened 2025-12-10
Debug: Only enable debug log for mobile development build

jermy-c:
- Only enable debug log for mobile development build

### DIFF
--- a/app/frontend/Settings/SettingsCategories.ts
+++ b/app/frontend/Settings/SettingsCategories.ts
@@ -27,7 +27,7 @@ import ChatNotifications from "./Chat/Notifications.svelte";
 import About from "./About/About.svelte";
 import License from "./License/Page/License.svelte";
 import OpenSource from "./About/OpenSource.svelte";
-// #if [DEV && MOBILE]
+// #if [MOBILE]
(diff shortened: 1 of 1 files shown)


app/frontend/Settings/SettingsCategories.ts
```diff
@@ -27,7 +27,9 @@ import ChatNotifications from "./Chat/Notifications.svelte";
 import About from "./About/About.svelte";
 import License from "./License/Page/License.svelte";
 import OpenSource from "./About/OpenSource.svelte";
+// #if [DEV && MOBILE]
```
benbucksch:
```suggestion
// #if [MOBILE]
```

app/frontend/Settings/SettingsCategories.ts
```diff
@@ -156,8 +157,8 @@ about.subCategories.addAll([
 ]);
 settingsCategories.add(about);
 
-if (!production) {
-  about.subCategories.add(new SettingsCategory("debug", gt`Debug`, Debug));
-}
+// #if [DEV && MOBILE]
```
benbucksch:
```suggestion
// #if [MOBILE]
```

PR #955, opened 2025-12-10
Mobile: Backend: Lazy Load heavy modules

jermy-c:
- Based on the benchmarks this is not faster but it always fixes `JPC: Connection failed to open` on my Android Emulator

### DIFF
--- a/mobile/backend/backend.ts
+++ b/mobile/backend/backend.ts
@@ -8,42 +8,6 @@ import os from "node:os";
 import fs from "node:fs";
 import fsPromises from "node:fs/promises";
 
-let HTTPServer: typeof import("../../backend/HTTPServer").HTTPServer;
-async function importHTTPServer() {
-  if (HTTPServer) return;
-  HTTPServer = (await import("../../backend/HTTPServer")).HTTPServer;
-}
-
-let ImapFlow: typeof import("imapflow").ImapFlow;
-async function importImapFlow() {
-  if (ImapFlow) return;
-  ImapFlow = (await import("imapflow")).ImapFlow;
-}
-
-let ky: typeof import("ky").default;
-async function importKy() {
-  if (ky) return;
-  ky = (await import("ky")).default;
-}
-
-let nodemailer: typeof import("nodemailer");
-async function importNodemailer() {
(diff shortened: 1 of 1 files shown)


mobile/backend/backend.ts
```diff
@@ -1,19 +1,49 @@
...
-import { DAVClient } from "tsdav";
 import { createType1Message, decodeType2Message, createType3Message } from "../../backend/ntlm";
 import path from "node:path";
 import os from "node:os";
 import fs from "node:fs";
 import fsPromises from "node:fs/promises";
 
+let HTTPServer: typeof import("../../backend/HTTPServer").HTTPServer;
+async function importHTTPServer() {
+  if (HTTPServer) return;
```
benbucksch: I had asked you to cache the module in a variable. Thanks that you did that.

Upon looking into it further, it seems that this is unnecessary: The module system already caches modules, based on their URL, and simply returns the module that's already loaded. That's true even for dynamic modules. So, they are already cached and we don't need to cache them again.

mobile/backend/backend.ts
```diff
@@ -123,10 +153,11 @@ async function writeFile(path: string, permissions: number, contents: Uint8Array
  * let json = await ky.get(https://api.example.com/users", { result: "json" });
  * ```
  */
-function kyCreate(defaultOptions) {
+async function kyCreate(defaultOptions) {
   /* `ky` (like axios) is both a function and acts like an object with functions get(), post() etc. as properties,
    * which confuses jpc, so make it only an object. */
   let kyObj = {};
+  await importKy();
```
benbucksch:
```suggestion
  const ky = await import("ky");
```

Same everywhere below

benbucksch (changes requested):

mobile/backend/backend.ts
jermy-c: Ok, thanks. Glad that you found that out, it looks much nicer now.

mobile/backend/backend.ts
jermy-c: Done in 4ab7397be6d20d69006e6538e35d56ba89eb6c8f

PR #961, opened 2025-12-12
Build: Lazy load heavy modules

jermy-c:
**Lazy load:**
- Matrix
- SIP
- Faker
- LiveKit
- XMPP

Class constructors cannot be async so I had to lazy load the entire class.

PR #962, opened 2025-12-13
Build: Upgrade to Node v24 in workflows

jermy-c:
Our Desktop builds are using a Node.js v20 which is causing the builds to fail because the Matrix SDK requires Node.js > v22.

https://github.com/mustang-im/mustang/pull/961#issuecomment-3648723161


jermy-c:
This was not test on CI yet. It based on:

1. The error message thrown by Yarn. `error matrix-js-sdk@39.4.0-rc.0: The engine "node" is incompatible with this module. Expected version ">=22.0.0". Got "20.19.6"
`. https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines

2. I'm using Node.js v24 locally.

PR #967, opened 2025-12-15
EWS: Fast fail when OAuth2 isn't logged in

NeilRashbrook:
When an EWS request is made, there's no actual check that the OAuth2 object is logged in. By comparison with OWA, the request fails immediately if it hasn't been logged in.

This is one way to fix a race condition that exists when you try to log in to EWS via the in-app tab. (The race condition causes the OAuth2 object to log out when the racing request fails, even though it is logging in at the time.)


NeilRashbrook:
I thought I might as well make an ActiveSync port so I added another commit for that.

benbucksch (approved):
I'm not sure that this is the correct fix, or at least not a complete fix.
But I'll let this land.

We should not even getting there in this state in the first place. The Auth panel should not return until it's actually logged in and the login is usable.

PR #968, opened 2025-12-16
Add pID to ContactBase and use it in subclasses

NeilRashbrook:
As of 0a31eab `id` is just a unique per-session identifier used to distinguish between new contacts.

Note: I may have missed some of the `id`s, or fixed them incorrectly, but I was at least able to create an EWS contact on Grommunio.

### DIFF
--- a/app/logic/Abstract/Contact.ts
+++ b/app/logic/Abstract/Contact.ts
@@ -11,6 +11,9 @@ export type Contact = Person | Group;
 export class ContactBase extends Observable {
   id: string;
   dbID: number;
+  /** Protocol-specific ID for this person.
+   * null = new contact that has not been saved to the server yet. */
+  pID: string | null = null;
   addressbook: Addressbook | null;
   @notifyChangedProperty
   name: string;
@@ -24,6 +27,14 @@ export class ContactBase extends Observable {
     super();
     this.id = randomID();
     this.addressbook = addressbook;
+  }
+
+  get isNew(): boolean {
+    return !this.dbID;
+  }
+
+  get existsOnServer(): boolean {
+    return !!this.pID;
   }
 
   async save() {
--- a/app/logic/Abstract/Person.ts
+++ b/app/logic/Abstract/Person.ts
@@ -7,7 +7,6 @@ import { ArrayColl } from 'svelte-collections';
 import { sanitize } from '../../../lib/util/sanitizeDatatypes';
 
 export class Person extends ContactBase {
-  id: string;
   @notifyChangedProperty
   firstName: string | null;
   @notifyChangedProperty
--- a/app/logic/Contacts/SQL/SQLGroup.ts
+++ b/app/logic/Contacts/SQL/SQLGroup.ts
@@ -22,14 +22,14 @@ export class SQLGroup extends Group {
             name, description, pID, addressbookID, json
           ) VALUES (
             ${group.name}, ${group.description},
-            ${group.id}, ${group.addressbook?.dbID}, ${jsonStr}
+            ${group.pID}, ${group.addressbook?.dbID}, ${jsonStr}
           )`);
         group.dbID = insert.lastInsertRowid;
       } else {
         await (await getDatabase()).run(sql`
           UPDATE "group" SET
             name = ${group.name}, description = ${group.description},
-            pID = ${group.id}, addressbookID = ${group.addressbook?.dbID},
+            pID = ${group.pID}, addressbookID = ${group.addressbook?.dbID},
             json = ${jsonStr}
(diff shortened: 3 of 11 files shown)


benbucksch:
> As of https://github.com/mustang-im/mustang/commit/0a31eab4774a5ba3eebed387cd4993e3854f7cf3 id is just a unique per-session identifier used to distinguish between new contacts.

Sorry, the initializer was merely intended to be a default init. You are still free to use it for your protocol-specific ID.

app/logic/Abstract/Contact.ts
```diff
@@ -3,21 +3,21 @@ import type { Person } from './Person';
...
-import { assert, randomID, type URLString } from '../util/util';
+import { Lock } from '../util/Lock';
+import { randomID, type URLString } from '../util/util';
 
 export type Contact = Person | Group;
 
 export class ContactBase extends Observable {
   id: string;
   dbID: number;
+  /** Protocol-specific ID for this person */
```
benbucksch:
```suggestion
  /** Protocol-specific ID for this person.
    * null = new contact that has not been saved on the server yet. */
```

app/logic/Abstract/Person.ts
```diff
@@ -1,13 +1,10 @@
 import { ContactBase } from './Contact';
 import type { Addressbook } from '../Contacts/Addressbook';
-import type { PublicKey } from '../Mail/Encryption/PublicKey';
-import { publicKeyFromJSON } from '../Mail/Encryption/KeyFromJSON';
 import { notifyChangedProperty, Observable } from '../util/Observable';
 import { ArrayColl } from 'svelte-collections';
-import { sanitize } from '../../../lib/util/sanitizeDatatypes';
+import { assert } from '../util/util';
```
benbucksch, line 10: Is really nothing using this, apart from for the protocol-specific ID?

app/logic/Contacts/SQL/SQLGroup.ts
```diff
@@ -13,24 +13,22 @@ import { appGlobal } from "../../app";
...
+            ${group.pID}, ${group.addressbook?.dbID}
           )`);
         group.dbID = insert.lastInsertRowid;
       } else {
         await (await getDatabase()).run(sql`
           UPDATE "group" SET
             name = ${group.name}, description = ${group.description},
-            pID = ${group.id}, addressbookID = ${group.addressbook?.dbID},
-            json = ${jsonStr}
+            pID = ${group.pID}, addressbookID = ${group.addressbook?.dbID}
```
benbucksch: We're already saving it to the DB as `pID`, which suggests that this is the right move here.

benbucksch (approved):

benbucksch:
> I may have missed some of the ids, or fixed them incorrectly, but I was at least able to create an EWS contact on Grommunio.

please double-check.

And un-bitrot.

benbucksch:
Can you please add a
```
get isNew(): boolean {
  return !this.pID;
}
```
getter? (On `ContactBase` or one of its subclasses, as appropriate.) This has multiple advantages:
* It's easier to read `if (foo.isNew)` than `if (!foo.pID)`
* It makes the intention explicit and unambigous. It shows what you want to accomplish, not just the how.
* It's easier to refactor. If we later use a dedicated boolean flag, we just need to change 1 or 2 places, instead of everywhere. The API stays the same.
* Same API as `Event.isNew`
* It keeps the implementation detail (use non-existing ID as signal) isolated within the class. Not spread across the whole app.

NeilRashbrook:
> > I may have missed some of the ids, or fixed them incorrectly, but I was at least able to create an EWS contact on Grommunio.
> 
> please double-check.
> 
> And un-bitrot.

CardDavPerson: Never used itemID in the first place.
GraphChatMessage, GraphhatRoom, MatrixAccount and XMPPChat now use a PersonUID subclass instead of Person.
JMAPGroup/Person: now uses jmapID, which is stored in the extra JSON, for which support was added to Group in a976127. (Not sure what `original` or `uid` are for though...)

Given the changes above, I might be better off adding an itemID/serverID field to the Exchange subclasses which is saved in the JSON, but that would be a separate PR.

benbucksch:
Superceded by #1311

NeilRashbrook:
> > I may have missed some of the ids, or fixed them incorrectly, but I was at least able to create an EWS contact on Grommunio.
> 
> please double-check.
> 
> And un-bitrot.

I've rebased but I don't know whether I should try to change:
- CardDAVPerson seems to use `url` as a protocol-specific id
- JMAPGroup and JMAPPerson seem to use `jmapID` as a protocol-specific id

Issue #948, opened 2025-12-08
IMAP: Shared folders: See emails of other users on the same server


benbucksch:
Our user has a colleage in his team, and that colleage gave access to his inbox or a subfolder (the "shared folder") to our user. We want to automatically detect that and show these folders.

Ideally, these folders can be shown either as part of the user's folder hierarchy, or as separate email account, similar to what we do for EWS.

This is the same as #898, but for IMAP.
#949 allows to give other users access to my mail folders, i.e. the other direction.

benbucksch:
> As for sending, it would depend on your SMTP server whether you would be allowed to send with a different address, but if you were, you would just have to set them up as an identity.

Yes, same thing there. We'd need to know a) which user is sharing (email address and realname), and b) which folder belongs to him, so that we can automatically select the right From address and save the sent email in the right Sent folder.

NeilRashbrook:
> Is there an easy way that we can detect that this is a shared folder

If the server supports the `NAMESPACE` command (RFC 2342), it will tell you which folders are actually other users or public folders. (If your own folders are in their own namespace, it will tell you that too.)

benbucksch:
Great, can you support that, please? We definitely want to know public folders from other users from our own user folders.

NeilRashbrook:
Good news: ImapFlow supports the `namespace` command via `conn.run`.

Bad news: It only ever returns the first namespace, so I need to use `conn.exec` again...

NeilRashbrook:
> You can also fix ImapFlow. https://github.com/mustang-im/imapflow/

Filed https://github.com/mustang-im/imapflow/pull/1 if that's what you meant.

> `listNamespaces()` should return the namespaces, instead of silently assigning the class variable, and the caller assigns them to the class variable.

Sorry, I was just copying the style of `listFolders`.

PR #973, opened 2025-12-18
IMAP: Support namespace command #948

NeilRashbrook:
If the server supports namespaces, this will set the `this.namespaces.personal`, `this.namespaces.otherUser` and `this.namespaces.shared` properties to arrays of namespace lists (as per RFC 2342), and also as a convenience it will map prefixes to namespace types.

I guess I should point out that the result of `conn.run("CAPABILITY")` does get cached by ImapFlow; we can't read the cached variable through JPC but fortunately `conn.run("CAPABILITY")` checks the cache first.

Grommunio: Does not support namespaces
Beonex: Only supports the personal namespace
Kolab Now: Supports all three namespace types

### DIFF
--- a/app/logic/Mail/IMAP/IMAPAccount.ts
+++ b/app/logic/Mail/IMAP/IMAPAccount.ts
@@ -22,12 +22,7 @@ export class IMAPAccount extends MailAccount {
   deleteStrategy: DeleteStrategy = DeleteStrategy.MoveToTrash;
   /**
    * An object representing the IMAP namespaces on this server.
-   * There are three categories, "personal", "other" and "shared".
-   * Each category has an array of namespace values, which are
-   * a pair of { prefix, delimiter } strings.
-   * The prefix allows you to determine whether a path is in this namespace.
-   * The delimiter tells you how you should construct subfolder paths,
-   * becuse it can be different for each namespace for some reason...
+   * See @IMAPNamespace, @IMAPNamespaceRecord
    */
   namespaces = kDefaultNamespaces;
   /** if polling is enabled, how often to poll.
@@ -249,6 +244,8 @@ export class IMAPAccount extends MailAccount {
     let conn = await this.connection();
     // conn.capabilities doesn't work; it's an object property,
     // and JPC doesn't notice direct changes to properties.
+    // Fortuantely `conn.run("CAPABILITY")` has its own cache,
+    // and only makes a network request if absolutely necessary.
     let capabilities = await conn.run('CAPABILITY');
     return await capabilities.has(capa);
   }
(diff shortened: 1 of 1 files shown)


app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -375,6 +428,11 @@ export class IMAPAccount extends MailAccount {
   }
 }
 
+/** The three types of IMAP namespace. */
+type IMAPNamespace = "personal" | "other" | "shared";
```
benbucksch, line 436: What is "other"? Public?

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -234,10 +247,50 @@ export class IMAPAccount extends MailAccount {
 
   async hasCapability(capa: string): Promise<boolean> {
     let conn = await this.connection();
-    let capabilities = await conn.capabilities;
+    // conn.capabilities doesn't work; it's an object property,
+    // and JPC doesn't notice direct changes to properties.
+    let capabilities = await conn.run('CAPABILITY');
```
benbucksch, line 249: We shouldn't re-run the network command every time some function needs the result. Can you fix that the function works and uses the result stored on the backend? Obviously, it should be the result from the last run of `CAPABILITY`, e.g. be different after login.

benbucksch: > but fortunately conn.run("CAPABILITY") checks the cache first.

OK, but that needs at least a comment.

Ideally, we'd fix that in JPC. At worst with a `__refresh()` function or so.

app/logic/Mail/IMAP/IMAPAccount.ts
NeilRashbrook: No, "shared" is for public folders and "other" is for other users.

benbucksch: Could you please document that in the JSDoc?

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -20,6 +20,16 @@ export class IMAPAccount extends MailAccount {
...
   pathDelimiter: string; /** Separator in folder path. E.g. '.' or '/', depending on server */
   deleteStrategy: DeleteStrategy = DeleteStrategy.MoveToTrash;
+  /**
+   * An object representing the IMAP namespaces on this server.
+   * There are three categories, "personal", "other" and "shared".
+   * Each category has an array of namespace values, which are
+   * a pair of { prefix, delimiter } strings.
+   * The prefix allows you to determine whether a path is in this namespace.
+   * The delimiter tells you how you should construct subfolder paths,
+   * becuse it can be different for each namespace for some reason...
```
benbucksch:
```suggestion
   * because it can be different for each namespace for some reason...
```

oh, Great! ...

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -375,6 +429,11 @@ export class IMAPAccount extends MailAccount {
   }
 }
 
+/** The three types of IMAP namespace. */
```
benbucksch:
```suggestion
/** The three types of IMAP namespace:
 * personal = user's own folders
 * other = other users' folders
 * shared = public folders, within the company */
```

app/logic/Mail/IMAP/IMAPAccount.ts
benbucksch: I made a suggestion above.

app/logic/Mail/IMAP/IMAPAccount.ts
benbucksch: This comment would be better in a `type` definition for `{ prefix, delimiter }`, above `kDefaultNamespace`

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -234,10 +247,51 @@ export class IMAPAccount extends MailAccount {
 
   async hasCapability(capa: string): Promise<boolean> {
     let conn = await this.connection();
-    let capabilities = await conn.capabilities;
+    // conn.capabilities doesn't work; it's an object property,
+    // and JPC doesn't notice direct changes to properties.
```
benbucksch, line 246:
```suggestion
    // and JPC doesn't notice direct changes to properties.
    // Fortunately, `conn.run("CAPABILITY")` checks its own cache first, so not network trip.
```

benbucksch (approved):

PR #974, opened 2025-12-18
Android: Run Node.js in a ViewModel Coroutine

jermy-c:
- Moved Node.js into ViewModel
- Redirect C++ std::out and std::out logs to the frontend in Coroutines for debugging errors with SMTP or IMAP
- The C++ Libraries are now loaded in Coroutine also but it is always loaded before running C++ code.
- The Android compileSDK and targetSDK had to be upgraded because `androidx.core:core-ktx:1.17.0` would not work
- Build scripts are updated to download libnode and also update the package name and also the directory structure to match the package name
- The Logger Class logs to the frontend if available but always logs to Logcat
- This should replace #900
- A build from this PR: https://github.com/mustang-im/mustang/releases/download/v0.9.9-dev/Mustang-0.9.9-dev.apk


benbucksch:
I don't know where the kotlin and C++ code comes from?

mobile/android/scripts/download-libnode.sh
```diff
@@ -0,0 +1,18 @@
+# Run from mobile/android
+
+# Check if libnode already exists
+if [ -f "./app/src/main/cpp/libnode/bin/arm64-v8a/libnode.so" ]; then
+    echo "libnode already exists, skipping download."
+    exit 0
+fi
+
+DOWNLOAD_URL="https://github.com/mustang-im/nodejs-mobile/releases/download/v22.9.0/nodejs-mobile-v22.9.0-android.zip"
```
benbucksch, line 9: In general, downloading binary blobs is uncool and dangerous. It will get is thrown out of F-Droid (or rejected from), for instance.
It's also against reproducible builds.

mobile/android/scripts/update-project.js
```diff
@@ -0,0 +1,133 @@
...
+    let content = await fs.readFile(file, 'utf8');
+    content = content.replace(/^package [^;\n]*/m, `package ${APP_ID}`);
+    await fs.writeFile(file, content);
+  }));
+  if (files.length) console.log(`✓ Updated ${files.length} Kotlin files`);
+}
+
+// Task 4: Update C++ Files
+async function updateCppFiles() {
+  const OLD_JNI_PREFIX = "im/mustang/capa";
```
benbucksch, line 76: Does that make any difference which package name the code has?
Can we just leave it as-is?
Does a user see the difference?

I understand that stack traces would be different, but that's rather a disadvantage that an advantage.

benbucksch (changes requested):

mobile/android/scripts/download-libnode.sh
benbucksch: Idea: Put the libs into an npm release.
Then, get them from node_modules/...
Still a binary module, but at least using the normal package installer dependency system, not just fetching it via https.

mobile/android/scripts/update-project.js
jermy-c: I think this can be removed we're no longer calling any method from Java.

Issue #965, opened 2025-12-15
EWS: Office365: Exceeded available subscriptions


benbucksch:
```
streamHTTP failed with HTTP 500 Internal Server Error
callStream	@	EWSAccount.ts:371
await in callStream		
streamNotifications	@	EWSAccount.ts:465
```
and:
```
error.ts:46 EWSItemError: You have exceeded the available subscriptions for your account.  Remove unnecessary subscriptions and try your request again.
    at EWSAccount.checkResponse (EWSAccount.ts:224:13)
    at EWSAccount.callEWS (EWSAccount.ts:309:21)
    at async EWSAccount.streamNotifications (EWSAccount.ts:453:20)
logError	@	error.ts:46
backgroundErrorInAccount	@	startup.ts:142
account.errorCallback	@	startup.ts:70
```

Both repeat multiple times.

This is during dev, where I keep restarting the client.
But we need to support 10+ shared accounts.

PR #976, opened 2025-12-19
EWS: Unsubscribe all notifications on logout #965

NeilRashbrook:


benbucksch:
Shouldn't you also close the connections from our side?

benbucksch:
Make a `MailAccount.disconnect()` function, which we can call on app shutdown.

`logout()` also actively forgets the login tokens, which we don't want to do on app shutdown.

app/logic/Abstract/Account.ts
```diff
@@ -145,6 +152,7 @@ export class Account extends Observable {
    * and likely deletes all local information from this account.
    * Does not delete the account on the server. */
   async deleteIt(): Promise<void> {
+    await this.disconnect();
```
benbucksch: This should be `logout()`

app/logic/Abstract/Account.ts
```diff
@@ -87,6 +87,13 @@ export class Account extends Observable {
   async logout(): Promise<void> {
   }
 
+  /**
+   * Used to request that any long-lived connections be disconnected,
+   * either due to app shutdown or account deletion.
```
benbucksch: `* e.g. due to app shutdown.`

benbucksch: Thanks that you added the function, and documented it

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -365,7 +367,7 @@ export class EWSAccount extends MailAccount {
   }
 
   async callStream(request: Json, responseCallback: (message: Record<string, any>) => Promise<void>) {
-    let lastAttempt;
+    let lastAttempt, controller;
```
benbucksch: As 2 lines, please. And define their types.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -39,6 +39,8 @@ export class EWSAccount extends MailAccount {
   // msgfolderroot: if this is an account shared with us
   // inbox: if this is an inbox shared with us
   sharedFolderRoot: "msgfolderroot" | "inbox" | null;
+  // JPC remoted AbortControllers for all currently streaming notifications
+  controllers = new Set<AbortController>();
```
```
notificationAbort = new Set<AbortController>();
```

benbucksch (approved):
Thank you, Neil!

PR #977, opened 2025-12-19
Android: Don't overwrite our Android Project files with Capacitor template files

jermy-c:
- We were extracting the template files before changing the package name every time
- This will overwrite our changes for our Java files
- The other change is that it will update all Java files instead of just the `MainActivity.java`
- The packageName is not directly visible to the user but I think it is visible in the App details section of the Play Store


benbucksch:
Please double-check whether the java package name has to match the application ID. We want the app ID to change, yes. But *not* the source code package names.

If the main() function package name has to match the app ID, then you can still call the other package from the main() function.

PR #982, opened 2025-12-23
Android: Run Node.js in a view model coroutine

jermy-c:
- The NodeProcess is now a viewModel to have the same lifecycle of the app
- Everything is now run in a IO context to coroutine to minimize blocking the UI thread
- Files are copied in parallel now because we have multiple `.js` files
- Native Methods are explicitly linked on link for performance and early error catching on load to check if the function signature matches.


mobile/android/app/src/main/cpp/node-process.cpp
```diff
@@ -0,0 +1,60 @@
...
+        return JNI_ERR;
+    }
+
+    // Find the NodeProcess class
+    jclass c = env->FindClass("im/mustang/capa/nodeprocess/NodeProcess");
+    if (c == nullptr) return JNI_ERR;
+
+    // Register your class' native methods.
+    static const JNINativeMethod methods[] = {
+            {"startNode", "([Ljava/lang/String;)I", reinterpret_cast<jint*>(startNode)},
```
benbucksch: Is there no shutdown needed?

mobile/android/app/src/main/cpp/node-process.cpp
```diff
@@ -0,0 +1,60 @@
...
+        env->DeleteLocalRef(arg);
+    }
+
+    printf("Starting node in C++ with %d arguments\n", argc);
+    int exitCode = node::Start(argc, argv);
+    return jint(exitCode);
+}
+
+// Called by System.loadLibrary("node-process")
+// Registers native methods in the Kotlin class without letting
```
benbucksch: "without letting" ... what?

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/FileOperations.kt
```diff
@@ -0,0 +1,77 @@
...
+import java.io.FileOutputStream
+
+object FileOperations {
+    private const val TAG = "FileOperations"
+
+    // Asynchronous copying
+    suspend fun copyAssetsDir(assetManager: AssetManager, from: String, to: String): Boolean = coroutineScope {
+        try {
+            // Get the list of files in the assets directory
+            // if there's no files return
```
benbucksch:
```suggestion
            // if there's no files, then just return
```

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/NodeProcess.kt
```diff
@@ -0,0 +1,87 @@
...
+    private val TAG = "NodeProcess"
+    private lateinit var job: Job
+    private val mainJS = "index.mjs"
+    private val nodeDir = "nodejs"
+    // Don't initialize until needed
+    private val assetManager: AssetManager by lazy {
+        context.assets
+    }
+    private val filesDir: File by lazy {
+        context.filesDir
```
benbucksch: Inline this? (and the previous property) You use both only once

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/NodeProcess.kt
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    private external fun startNode(args: Array<String>): Int
+
+    fun start() {
+        if (this::job.isInitialized && job.isActive) return
+        job = viewModelScope.launch(Dispatchers.IO) {
+            try {
+                Log.d(TAG, "Starting to load libraries")
```
benbucksch:
```suggestion
                Log.d(TAG, "Starting to load node.js libraries")
```

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/NodeProcess.kt
```diff
@@ -0,0 +1,87 @@
...
+                loadLibraries()
+                Log.d(TAG, "Libraries loaded")
```
benbucksch:
```suggestion
                Log.d(TAG, "node.js libraries loaded")
```

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/FileOperations.kt
benbucksch: Why not throw an exception? The calling code throws anyways, if this returns false.

mobile/android/app/src/main/cpp/node-process.cpp
```diff
@@ -0,0 +1,60 @@
...
+               jobjectArray args) {
+
+    // Argument count
+    jsize argc = env->GetArrayLength(args);
+
+    char** argv = new char*[argc];
+
+    for (int i = 0; i < argc; i++) {
+        jstring arg = static_cast<jstring>(env->GetObjectArrayElement(args, i));
+        jsize argSize = env->GetStringUTFLength(arg);
```
benbucksch, line 14:
```suggestion
        // Returns the length of the UTF8 byte stream, in bytes.
        // ! `GetStringLength()` returns the Unicode chars and is unsuitable for array allocation
        jsize argSize = env->GetStringUTFLength(arg);
```

mobile/android/app/src/main/cpp/node-process.cpp
```diff
@@ -0,0 +1,60 @@
+#include <jni.h>
+#include <node.h>
+
+using namespace std;
+
+// Copy arguments to one contiguous memory
```
benbucksch:
```suggestion
// Copy arguments from JNI to C++ format
```

benbucksch: You should make this (line 11 to 31) a function.

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/NodeProcess.kt
```diff
@@ -0,0 +1,87 @@
...
+
+
+                // Node.js assets are in the APK archived and cannot be accessed with a path
+                // Because the we have node native modules and multiple files, we need to have
+                // it copied to a physical location for node to find the .node files using relative paths
```
benbucksch:
```suggestion
                // Because we have node native modules and multiple files, we need to have them
                // copied to a physical location, for node to find the `.node` files using relative paths
```

mobile/android/app/capacitor.build.gradle
```diff
@@ -14,8 +14,8 @@ dependencies {
     implementation project(':capacitor-camera')
     implementation project(':capacitor-filesystem')
     implementation project(':capacitor-splash-screen')
+    implementation project(':capacitor-status-bar')
```
benbucksch: Where does this come from?
If this is needed, it should be a separate PR

benbucksch (approved):
I cannot really review this code well, because it's too deep in Android, JNI, node.js. I looked as well as I could, but my review is necessarily only on the surface.

Looks OK to me. Just a few nits.

Only major question is: Do we need to shut down node.js?

mobile/android/app/src/main/cpp/node-process.cpp
jermy-c: It would be good to shutdown Node.js since `job.cancel()` does shutdown the process. Unfortunately, the `node::Start()` does not return an `env` variable required by the `node::Stop()` function. But I've not seen any issues so far even after restarting the app immediately.

mobile/android/app/capacitor.build.gradle
jermy-c: This was from when we added the Capacitor Status Bar plugin and the changes in this file was not committed. I'll remove this. But this will be added anyways when you do `cap sync`.

mobile/android/capacitor.settings.gradle
```diff
@@ -17,8 +17,8 @@ project(':capacitor-filesystem').projectDir = new File('../node_modules/@capacit
 include ':capacitor-splash-screen'
 project(':capacitor-splash-screen').projectDir = new File('../node_modules/@capacitor/splash-screen/android')
 
+include ':capacitor-status-bar'
```
jermy-c: This is also from `cap sync`. This is will still work even if it's not committed, do we still want to commit changes like this in the future? But I'm not sure if it'll still work if the file is not present in the git history but the changes don't need to be committed.

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/FileOperations.kt
jermy-c: Yes, you're right, I've fixed it.

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/NodeProcess.kt
jermy-c: Yes, thanks! I've inlined it.

benbucksch:
@NeilRashbrook Could you review this, please?

mobile/android/app/src/main/cpp/node-process.cpp
```diff
@@ -0,0 +1,74 @@
+#include <jni.h>
+#include <node.h>
+
+using namespace std;
+
+// Copy arguments from JNI to C++ format
+char** argsJavaToC(JNIEnv *env, jsize argc, jobjectArray args) {
+    char** argv = new char*[argc];
```
NeilRashbrook, line 8: I finally found the source to `node::Start`, and fortunately it doesn't require a C-style `argv` which would need to be `nullptr`-terminated.

mobile/android/app/src/main/cpp/node-process.cpp
NeilRashbrook: From the [source](https://github.com/nodejs/node/blob/8e561a3d0672f7538e647019fe9e5aa0aed20d2d/src/node_main_instance.cc#L110), `node::Start` runs the script to completion i.e. it spins the event loop until it would naturally shut down, and doesn't actually return until then. I'm not sure how you could get to explicitly cancel the job in that case.

If I'm wrong and you do get to cancel the job then it might be possible to use `Environment* env = Environment::GetCurrent(Isolate::GetCurrent());` or some such.

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/NodeProcess.kt
```diff
@@ -0,0 +1,87 @@
...
+                }
+
+                val mainJS = File(to, mainJS)
+                if (!mainJS.exists()) {
+                    throw Exception("Main JS file not found")
+                }
+
+                val args = arrayOf("node", mainJS.absolutePath)
+                Log.d(TAG, "Starting node with arguments: \"${args.joinToString(" ")}\"")
+                startNode(args)
```
NeilRashbrook, line 74: As I mentioned in my other comment, it looks as if this is going to block the current thread. Is that acceptable behaviour?

benbucksch:
Test build https://github.com/mustang-im/mustang/actions/runs/21985774694

I'm running this on my phone, for testing

mobile/android/app/src/main/java/im/mustang/capa/nodeprocess/NodeProcess.kt
benbucksch: No

benbucksch:
I see symptoms that the backend stays running, even if I force-kill the app (= swipe it away in the app switcher). I get errors about too many IMAP connections being open, an error that I know from the desktop during dev when the frontend reloads and the backend stays alive.

The objective here is to *ensure* that the lifetime of the backend is identical to the frontend, i.e. the backend must be killed in *all* situations when the frontend is killed, be it manually by the user or automatically by the OS.

PR #983, opened 2025-12-23
Auth: Always set config.oAuth2 for OAuth2 accounts

benbucksch:
There were cases where we did not fill in an OAuth2 config. This is a revision of PR #981.

The automatic account setup failed to fill in the OAuth2 config. Before PR #969, this appeared to work, because the login step would always show the separate window UI, but then the account would be created with the popup window as the default auth UI.

### DIFF
--- a/app/frontend/Setup/Mail/SetupMail.svelte
+++ b/app/frontend/Setup/Mail/SetupMail.svelte
@@ -188,7 +188,6 @@
       }
       step = Step.FoundConfig;
       await onContinue();
-      return;
     } else if (step == Step.CheckConfig) {
       step = Step.FinalizeConfig;
     } else if (step == Step.FinalizeConfig) {
--- a/app/logic/Auth/OAuth2Util.ts
+++ b/app/logic/Auth/OAuth2Util.ts
@@ -11,10 +11,10 @@ export function getOAuth2BuiltIn(config: Account): WebBasedAuth | undefined {
     return new OWAAuth(config as OWAAccount);
   }
   let hostname = (config as TCPAccount).hostname ??
-    config.url ? new URL(config.url).hostname : null;
+    (config.url ? new URL(config.url).hostname : null);
   let o = OAuth2URLs.find(o => o.hostnames.some(h => h == hostname));
   if (!o) {
-    return undefined;
+    return null;
   }
   let oAuth2 = new OAuth2(config, o.tokenURL, o.authURL, o.authDoneURL, o.scope, o.clientID, o.clientSecret, o.doPKCE);
   oAuth2.setTokenURLPasswordAuth(o.tokenURLPasswordAuth);
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -80,6 +80,7 @@ export class ActiveSyncAccount extends MailAccount {
     await super.login(interactive);
     if (this.authMethod == AuthMethod.OAuth2) {
       this.oAuth2 ??= getOAuth2BuiltIn(this);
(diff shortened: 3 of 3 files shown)


app/frontend/Setup/Mail/SetupMail.svelte
```diff
@@ -182,13 +186,9 @@
...
-      fillConfig(config, emailAddress, password);
-      errorMessage = null;
-      if (config.oAuth2) {
-        step = Step.Login;
-      } else {
-        step = Step.CheckConfig;
-      }
+      step = Step.FoundConfig;
+      await onContinue();
+      return;
```
NeilRashbrook: This is inconsistent with line 168 which doesn't explicitly return (because it's the last code in its block).

app/logic/Auth/OAuth2Util.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function getOAuth2BuiltIn(config: Account): WebBasedAuth | undefined {
+  if (config.protocol == "owa") {
+    return new OWAAuth(config as OWAAccount);
+  }
+  let hostname = (config as TCPAccount).hostname ??
+    config.url ? new URL(config.url).hostname : null;
+  let o = OAuth2URLs.find(o => o.hostnames.some(h => h == hostname));
+  if (!o) {
+    return undefined;
```
NeilRashbrook: `config.oAuth2` normally defaults to `null`.

Also, while I'm here, `??` has higher precedence than `?:`, so you'll need to wrap line 14 in `()`s.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -80,12 +79,7 @@ export class ActiveSyncAccount extends MailAccount {
     await ensureLicensed();
     await super.login(interactive);
     if (this.authMethod == AuthMethod.OAuth2) {
-      if (!this.oAuth2) {
-        let urls = OAuth2URLs.find(a => a.hostnames.includes(this.hostname));
-        assert(urls, gt`Could not find OAuth2 config for ${this.hostname}`);
-        this.oAuth2 = new OAuth2(this, urls.tokenURL, urls.authURL, urls.authDoneURL, urls.scope, urls.clientID, urls.clientSecret, urls.doPKCE);
-        this.oAuth2.setTokenURLPasswordAuth(urls.tokenURLPasswordAuth);
-      }
+      this.oAuth2 ??= getOAuth2BuiltIn(this);
```
NeilRashbrook, line 82: This `assert` went missing, although one was kept on line 168.

app/logic/util/util.ts
```diff
@@ -1,5 +1,5 @@
 
-export function assert(test: boolean | Object | null, errorMessage: string): asserts test {
+export function assert(test: boolean | Object | null | undefined, errorMessage: string): asserts test {
```
NeilRashbrook: What's complaining about this?

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
benbucksch: fixed

app/logic/Auth/OAuth2Util.ts
benbucksch: Thanks a lot! It was actually failing for IMAP.

I was somehow wrongly assuming that `??` and `?:` have the same precedence and are executed in order of appearance, so I would have been looking at that problem for quite a while.

Fixed.

app/frontend/Setup/Mail/SetupMail.svelte
benbucksch: return removed

PR #989, opened 2026-01-07
Android: WebView for OAuth2

jermy-c:
## Android: WebView for OAuth2

- WebViews or iframes with embedded User-Agents don't work for OAuth2 on Mobile
- Changing the User-Agent for iframe doesn't seem to work at all
- Takes that size and position as the element to look like an embedded iframe
- Uses the https://github.com/Cap-go/capacitor-inappbrowser/?tab=readme-ov-file#openwebview instead of the Capacitor InAppBrowser because this one lets us inject the autofill JS
- Not working for iOS yet, not sure why the WebView would stop the app


jermy-c:
The example here works for Android only: https://github.com/Cap-go/capacitor-inappbrowser/?tab=readme-ov-file#open-webview-with-custom-dimensions

PR #990, opened 2026-01-09
Update all other dependencies except ImapFlow, Node.js, Electron, Capacitor, Svelte

jermy-c:
- Update all other dependencies except ImapFlow, Node.js, Electron, Capacitor, Svelte
- `vite` for e2 cannot be updated because it will break `electron-vite`
- the resolve conditions need to be explicitly defined otherwise `pkce-challenge` won't resolve with Vite 7+

### DIFF
--- a/mobile/backend/package.json
+++ b/mobile/backend/package.json
@@ -26,7 +26,7 @@
     "ws": "^8.19.0"
   },
   "resolutions": {
-    "**/iconv-lite": "^0.6.3"
+    "**/iconv-lite": "^0.7.2"
   },
   "devDependencies": {
     "@rollup/plugin-esm-shim": "^0.1.8",


app/package.json
```diff
@@ -20,75 +20,75 @@
     "@capacitor/file-viewer": "^2.0.0",
     "@capacitor/filesystem": "^7.1.4",
     "@capacitor/splash-screen": "^7.0.3",
-    "@faker-js/faker": "^9.8.0",
+    "@faker-js/faker": "^10.2.0",
```
benbucksch, line 23: Can you please double-check all files that use faker?

jermy-c: I don't see any problems while using it but these lines have typescript errors with params passed.

https://github.com/mustang-im/mustang/blob/3da43bbefba52ebf3154f67de4059ed32cc5d434/app/logic/testData.ts#L164-L167

```
Expected 0-1 arguments, but got 2.
```

```
Type '"+49-170-### ####"' has no properties in common with type '{ style?: "human" | "national" | "international"; }'.
```

benbucksch: Yes, that's what I mean. Finding and fixing these kind of errors is part of the task here.

Please also check whether other module updates caused other code errors.

jermy-c: Oh ok, sorry, I understand now, I'll fix the errors.

jermy-c:
I've checked and fixed the errors and tested it again.

mobile/backend/package.json
```diff
@@ -9,29 +9,30 @@
...
-    "vite-plugin-static-copy": "^2.3.1",
-    "ws": "^8.16.0"
+    "ky": "^1.14.2",
+    "nodemailer": "^7.0.12",
+    "tsdav": "^2.1.6",
+    "vite-plugin-static-copy": "^3.1.4",
+    "ws": "^8.19.0"
   },
   "resolutions": {
     "**/iconv-lite": "^0.6.3"
```
benbucksch:
```suggestion
    "**/iconv-lite": "^0.7.2"
```

jermy-c:
@benbucksch commit ef76076dedaa09fa214eacad667a73ab5114d001 added duplicate dependency records to `backend/package.json`

benbucksch:
> commit [ef76076](https://github.com/mustang-im/mustang/commit/ef76076dedaa09fa214eacad667a73ab5114d001) added duplicate dependency records to `backend/package.json`

I had to manually merge it with master and probably made a mistake.
I don't see the problem. Can you please fix it?

Issue #937, opened 2025-11-18
Mac: Camera and Microphone permissions not requested


jermy-c:
### Reproduction

1. Settings | Meet | Devices
2. Select your Microphone
3. Select your Camera

### Actual Result

1. No permission popup or no permission requested
2. You don't see yourself

### Expected Result

1. You can see yourself
2. You see a permission request

### Environment

- MacOS

PR #993, opened 2026-01-12
Mac: Request camera and mic access #937

jermy-c:
- Request camera and mic access for MacOS


app/logic/Meet/LocalMediaDeviceStreams.ts
```diff
@@ -37,6 +40,17 @@ export class LocalMediaDeviceStreams extends MediaDeviceStreams {
     await this.setCameraMicOn(this._cameraOn, on, this._cameraDevice, device);
   }
   async setCameraMicOn(cameraOn: boolean, micOn: boolean, cameraDevice: string = this._cameraDevice, micDevice: string = this._micDevice) {
+    // #if [!WEBMAIL && !MOBILE]
```
benbucksch: Can you please check for Mac *before* you call the backend? That saves a useless call on Windows and Linux. And then you won't need the conditional compile, either.

PR #994, opened 2026-01-12
OAuth2: isLoggedIn = false when expired #350

benbucksch:
If the computer was asleep (notebook lid down) and waking up now, and we refresh the IMAP connection, but the access token is expired (because too much time passed since the notebook went to sleep), then `isLoggedIn` should return false, because we *know* for a fact that the access token won't be valid anymore.

In `IMAPAccount.connection()`, we're checking that. If `isLoggedIn` is false, we try to log in again, non-interactively, and wait for that before trying to connect. If we have a refresh token, we should be able to get a new access token non-interactively, and the IMAP connection should then accept it.

This should theoretically fix the #350 Gmail "Invalid credentials" bug.

### DIFF
--- a/app/logic/Auth/OAuth2.ts
+++ b/app/logic/Auth/OAuth2.ts
@@ -169,7 +169,7 @@ export class OAuth2 extends WebBasedAuth {
 
   isExpired(): boolean {
     return this.expiresAt
-      ? this.expiresAt.getTime() + 2000 < Date.now()
+      ? this.expiresAt.getTime() - 2000 < Date.now()
       : false;
   }
 
--- a/app/logic/Mail/IMAP/IMAPFolder.ts
+++ b/app/logic/Mail/IMAP/IMAPFolder.ts
@@ -76,8 +76,7 @@ export class IMAPFolder extends Folder {
         }
       } catch (ex) {
         console.log("Opening IMAP folder failed", ex);
-        if (ex.code == "NoConnection" || // TCP connection dropped
-            ex.message == "Invalid credentials (Failure)") { // Gmail access token expired #350
+        if (ex.code == "NoConnection") {
           conn = await this.account.reconnect(conn);
           if (doLock) {
             lock = await this.account.connectionLock.get(conn).lock();


app/logic/Auth/OAuth2.ts
```diff
@@ -164,7 +164,11 @@ export class OAuth2 extends WebBasedAuth {
   }
 
   get isLoggedIn(): boolean {
-    return !!this.accessToken;
+    return !!this.accessToken && !this.isExpired;
+  }
+
+  isExpired(): boolean {
+    return this.expiresAt.getTime() + 2000 > Date.now();
```
benbucksch:
```suggestion
    return this.expiresAt.getTime() + 2000 < Date.now();
```

app/logic/Auth/OAuth2.ts
```diff
@@ -164,7 +164,11 @@ export class OAuth2 extends WebBasedAuth {
   }
 
   get isLoggedIn(): boolean {
-    return !!this.accessToken;
+    return !!this.accessToken && !this.isExpired;
+  }
+
+  isExpired(): boolean {
+    return this.expiresAt.getTime() + 2000 < Date.now();
```
jermy-c: I think this will break when the OAuth2 token doesn't expire because it will throw an error:

https://github.com/mustang-im/mustang/blob/3da43bbefba52ebf3154f67de4059ed32cc5d434/app/logic/Auth/OAuth2.ts#L51

https://github.com/mustang-im/mustang/blob/3da43bbefba52ebf3154f67de4059ed32cc5d434/app/logic/Auth/OAuth2.ts#L237-L244

benbucksch: Good catch! Thank you.

I've added a null check.

Did you just notice this in the code, or do you actually hit that? If you hit it, that means Gmail doesn't send expiry, which means this fix (PR) cannot help Gmail.

jermy-c: I just noticed it.

benbucksch: Thanks for seeing that in the code.

app/logic/Auth/OAuth2.ts
NeilRashbrook: Doesn't this wait until 2 seconds _after_ the server-side expiry time?

jermy-c:
Tested version 78c482084ab7cf700301bcb8e7d13f088afc2251

The email account state is correct where it show it's logged out but when you click the log in button it throws an error. The `expiresAt` is not updated.

```
LoginError: Check your login, username, and password.
Invalid credentials (Failure)
    at IMAPAccount.connection (IMAPAccount.ts:135:35)
    at async IMAPAccount.login (IMAPAccount.ts:56:5)
    at async login (AccountListItem.svelte:62:11)
    at async HTMLButtonElement.myOnClick (Button.svelte:95:13)
```

app/logic/Mail/IMAP/IMAPFolder.ts
```diff
@@ -76,7 +76,8 @@ export class IMAPFolder extends Folder {
         }
       } catch (ex) {
         console.log("Opening IMAP folder failed", ex);
-        if (ex.code == "NoConnection") {
+        if (ex.code == "NoConnection" || // TCP connection dropped
+            ex.message == "Invalid credentials (Failure)") { // Gmail access token expired #350
```
jermy-c: I simulated the OAuth2 expiry by:

1. Logging sucessfully with OAuth2
2. Revoking the token from my Google Account

The error was still there because no commands are issued here:

https://github.com/mustang-im/mustang/blob/3da43bbefba52ebf3154f67de4059ed32cc5d434/app/logic/Mail/IMAP/IMAPFolder.ts#L67-L76

Which means this condition isn't met.

app/logic/Mail/IMAP/IMAPFolder.ts
jermy-c, line 92: This is where it the error is thrown.

jermy-c:
Commit 3410ee819a753f9a2321bf294991f0ef026e7266

```
Error: Reconnect: Connection unknown
    at assert$2 (index-CjvZQbIG.js:1250:11)
    at IMAPAccount.reconnect (index-CjvZQbIG.js:114721:5)
    at IMAPFolder.runCommand (index-CjvZQbIG.js:114040:35)
    at async IMAPFolder.fetchUIDList (index-CjvZQbIG.js:114204:5)
    at async IMAPFolder.checkDeletedMessages (index-CjvZQbIG.js:114365:22)
    at async IMAPFolder.getNewMessages (index-CjvZQbIG.js:114223:5)
    at async getMail (index-CjvZQbIG.js:173985:7)
    at async catchErrors (index-CjvZQbIG.js:68590:5)
```
```
Error: Reconnect: Connection unknown
    at assert$2 (index-CjvZQbIG.js:1250:11)
    at IMAPAccount.reconnect (index-CjvZQbIG.js:114721:5)
    at IMAPFolder.runCommand (index-CjvZQbIG.js:114040:35)
    at async IMAPFolder.fetchMessageList (index-CjvZQbIG.js:114150:5)
    at async IMAPFolder.listChangedMessages (index-CjvZQbIG.js:114119:27)
    at async IMAPFolder.listMessages (index-CjvZQbIG.js:114069:19)
    at async loadFolder (index-CjvZQbIG.js:281527:25)
```
```
Error: Reconnect: Connection unknown
    at assert$2 (index-CjvZQbIG.js:1250:11)
    at IMAPAccount.reconnect (index-CjvZQbIG.js:114721:5)
    at IMAPFolder.runCommand (index-CjvZQbIG.js:114040:35)
    at async IMAPFolder.fetchMessageList (index-CjvZQbIG.js:114150:5)
    at async IMAPFolder.listChangedMessages (index-CjvZQbIG.js:114119:27)
    at async IMAPFolder.listMessages (index-CjvZQbIG.js:114069:19)
    at async loadFolder (index-CjvZQbIG.js:273570:25)
```

benbucksch:
From what I understood, there are 3 problems - please correct me when I'm wrong:

1. IMAP connection died
When the notebook lid is closed, the TCP connections die, so when we are awaken from suspend mode, we think we have an IMAP connection, but in fact it's gone. We try to use the existing `connection` object, but the call fails with "NoConnection" (and into reconnect), or similar. The "or similar" might be a case that we don't handle yet and need to add code for.

2. Gmail IMAP fails all commands when the access token expired
Gmail IMAP server refuses to answer to IMAP commands when the access token expires, even if the IMAP connection still stands. I think that's the "IMAPCommandError: Invalid credentials (Failure)" cases you mentioned in comment https://github.com/mustang-im/mustang/pull/994#issuecomment-3741701827.

3. OAuth2.isLoggedIn
Our `OAuth2.isLoggedIn` doesn't consider expiry. This (and only this) is what I try to fix in this PR. After a suspend awake, our IMAP connection won't work anymore, because TCP connection dropped. And when we reconnect, `isLoggedIn` returns true, because we still have the access token, but the access token is no longer valid, because a lot of time has passed, but we were not able to refresh it at the expiry time, because the notebook was asleep. My change to OAuth2.ts (and only that) in this PR should theoretically fix this problem 3, if I did the code right. Can you please test whether this fixes this problem 3?

PR #996, opened 2026-01-18
IMAP: Don't run reconnect() twice at the same time #350

benbucksch:
### Problem

This tries to fix https://github.com/mustang-im/mustang/issues/350#issuecomment-3765448337

1.
- an event listener triggers `reconnect()`, the function starts
- `runCommand()` triggers `reconnect()`, the function call starts

2.
- the first call executes up to `this.connections.set(purpose, null);`, this is sets connection purpose to null
- the second call executes up to `let purpose = this.connections.getKeyForValue(connection);assert(purpose, "Connection purpose unknown");`, this gets the connection purpose as null because of the first call

3.
- The first call continues executing and resolves
- the second call throws, the `runCommand()` stays in the failed state

### Fix

This changes the `RunOnce` class to be a wrapper around an async function to prevent it from being called twice at the same time.
The second call will wait for the first call to finish, and then get the
result from the first call, without calling the function a second time.

### DIFF
--- a/app/logic/Mail/IMAP/IMAPAccount.ts
+++ b/app/logic/Mail/IMAP/IMAPAccount.ts
@@ -76,8 +76,7 @@ export class IMAPAccount extends MailAccount {
     if (conn) {
       return conn;
     }
-    let runOnce = this.connectRunOnce.get(purpose);
-    return await runOnce.runOnce(async () => {
+    return await this.connectRunOnce.get(purpose).runOnce(async () => {
       this.fatalError = null;
 
       // Auth method


benbucksch:
Tip: Hide whitespace changes while reviewing.
image

jermy-c:
I ran built and ran the code. Then let my Macbook sleep and then clicked [Get Mail].
```
Opening IMAP folder failed Error: Connection not available
    at message.js:155:16
    at new Promise (<anonymous>)
    at WSCall.makeCall (message.js:150:12)
    at JPCWebSocket.callRemote (protocol.js:184:31)
    at RemoteClass.getMailboxLock (obj.js:240:55)
    at IMAPFolder.runCommand (IMAPFolder.ts:71:36)
    at async IMAPFolder.fetchUIDList (IMAPFolder.ts:276:5)
    at async IMAPFolder.checkDeletedMessages (IMAPFolder.ts:460:22)
    at async IMAPFolder.getNewMessages (IMAPFolder.ts:295:5)
    at async getMail (GetMailButton.svelte:57:16)
    at async catchErrors (error.ts:65:5)
```

```
TypeError: Cannot read properties of undefined (reading 'lock')
    at IMAPFolder.runCommand (IMAPFolder.ts:82:62)
    at async IMAPFolder.fetchUIDList (IMAPFolder.ts:276:5)
    at async IMAPFolder.checkDeletedMessages (IMAPFolder.ts:460:22)
    at async IMAPFolder.getNewMessages (IMAPFolder.ts:295:5)
    at async getMail (GetMailButton.svelte:57:16)
    at async catchErrors (error.ts:65:5)
```

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -214,30 +216,31 @@ export class IMAPAccount extends MailAccount {
   }
 
   async reconnect(connection: ImapFlow): Promise<ImapFlow> {
-    // Note: Do not stop polling
+    await this.reconnectRunOnce.runOnce(async () => {
```
jermy-c: I think this will cause an error if both the `Main` and `Fetch` connection trigger a reconnect at the same time. Only the first executed one will have a renewed connection.

benbucksch: Right, the `RunOnce` lock has to be per `purpose`. Can you please adapt the code to make it work? Until it works.

benbucksch:
Your changes look good to me. Can you test it, please?

jermy-c:
The changes seem to be working so far. I checked:

- OAuth2: expiryAt, the `setTimeout()` seems to be working even when sleeping
- connections, connectRunOnce, reconnectRunOnce properties for the correct ImapFlow connection ID, before the connections with different purposes had the same ID.

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -215,29 +215,32 @@ export class IMAPAccount extends MailAccount {
...
-    this.connectionLock.delete(connection);
 
     let purpose = this.connections.getKeyForValue(connection);
     assert(purpose, "Connection purpose unknown");
-    this.connections.set(purpose, null);
-    this.notifyObservers();
 
-    if (this.authMethod == AuthMethod.OAuth2 && this.oAuth2 &&
+    let runOnce = this.reconnectRunOnce.get(purpose);
+    return await runOnce.runOnce(async () => {
```
benbucksch:
```suggestion
    return await this.reconnectRunOnce.get(purpose).runOnce(async () => {
```

app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -73,9 +76,8 @@ export class IMAPAccount extends MailAccount {
     if (conn) {
       return conn;
     }
-    let lock = await this.connectLock.get(purpose).lock();
-    try {
-
+    let runOnce = this.connectRunOnce.get(purpose);
+    return await runOnce.runOnce(async () => {
```
benbucksch:
```suggestion
    return await this.connectRunOnce.get(purpose).runOnce(async () => {
```

PR #998, opened 2026-01-19
EWS: Remove namespace attributes when converting DOM to JSON

NeilRashbrook:


app/logic/Mail/EWS/XML2JSON.ts
```diff
@@ -1,6 +1,10 @@
 import type { Json } from "../../util/util";
 
 export function XML2JSON(aNode: Element): Json {
+  // This node may have been namespaced. We're not interested in that attribute.
+  // (This doesn't matter for Microsoft Exchange because it declares all of its
+  // namespaces on the result node, but other EWS implementations may vary.)
```
benbucksch:
```suggestion
  // We don't need the `xmlns` attribute, because we have the `prefix` property.
  // This attribute confuses Kopano.
  // MS Exchange declares all of its namespaces on the result node.
```

Does this comment make sense? If it's wrong or incomplete, feel free to adjust.

app/logic/Mail/EWS/XML2JSON.ts
```diff
@@ -1,6 +1,10 @@
 import type { Json } from "../../util/util";
 
 export function XML2JSON(aNode: Element): Json {
+  // We don't need the `xmlns` attribute, because we have the `prefix` property.
+  // This attribute confuses Kopano.
+  // MS Exchange declares all of its namespaces on the result node.
+  aNode.removeAttribute(aNode.prefix ? "xmlns:" + aNode.prefix : "xmlns");
```
benbucksch:
```suggestion
  /* We don't need the `xmlns` attribute, because we don't consider namespaces while reading.
   * Kopano puts the namespace asttributes on individual XML nodes.
   * This confuses our code which detects empty nodes, and that code thinks that the node
   * is not empty.
   * MS Exchange declares all of its namespaces on the result node, so the problem
   * doesn't appear there. */
  aNode.removeAttribute(aNode.prefix ? "xmlns:" + aNode.prefix : "xmlns");
```

benbucksch (changes requested):

Issue #922, opened 2025-11-03
IMAP: Wrong email


benbucksch:
We have a race condition where sometimes the wrong email gets downloaded. The email has the right headers (= sender, recipients, subject, date), but the wrong body (= text content of the email). Sadly, the wrong body is even saved to disk, so that a restart of Mustang doesn't fix it.

It can happen even for emails where the body already shows, but maybe the attachments are not downloaded yet.

* I didn't see this bug between February and July.
* I first saw this bug after commit 9b4d5ca4, which was supposed to fix #648. I reverted it in dc02b764 . But even after the revert, the bug still appeared sometimes, albeit less often.
* I added a workaround in commit ccaaddc4 , which made the bug mostly go away, but I have still seen the bug after that workaround (if I was not mistaken by running the wrong build).

benbucksch:
I haven't seen this bug anymore recently. Maybe the bug appears only when you set up a new account and download a large mailbox.

benbucksch:
> downloading message bodies happens on a different connection to fetching new mail

See `IMAPFolder.downloadMessages()`, which uses the `Fetch` connection - deliberately, so that a download due to an interactive click on an email cannot be blocked by a background email download.

benbucksch:
@NeilRashbrook You did try this while filtering messages, right? I see the bug only then. See my comments above.

jermy-c:
### SQL operations running while transaction in progress

We're reading emails and maybe some other operations while saving emails in a transaction. I think it might cause an issue if the email record is not committed to the database when the email is being queried. Because changes are only committed after the transaction is complete.

**What's happening**
1. Acquire transaction lock in SQLEmail
2. Start transaction
3. Read email operation is triggered in another function, I see this happening when I place a `console.log()`
4. It is allowed since it doesn't need to acquire a transaction lock
5. It is considered part of the transaction, which seems to make transaction slightly slower about 1ms per message compared to running all `email.save()` in `await Promise.all()` without transactions
6. Other operations might still be running
7. Transaction ends with all the other operations considered as part of the transaction

PR #1002, opened 2026-01-21
IMAP: Pass a fallback purpose - fixes #350

### LINKED ISSUE #350: IMAP: Gmail: Invalid credentials (Failure) in some sessions
1. Go to Mail
2. Click to an account
3. Click to Sent Mail

Actual result
Instead of the content there is 'Invalid credential (Failure)'

Expected result
Seeing the content of the sent mail

image

jermy-c:
There's a race condition where:

1. The first call triggered by the listener executes the code before the runOnce and inside the runOnce. Inside the runOnce sets the connection to null in the map
2. The second call triggered by the reconnect in `runCommand()`, tries to get the purpose for connection but the connection no longer exist in the map

**Fix**

1. Allow for passing the purpose as a fallback
2. Getting the purpose by connection will still need to take precedence so we close that connection instead of another one. And it also prevents the case that incase the connection still works and another function is still using has the connection closed mid operation.
3. If the purpose has no connection then the passed fallback purpose will let the `connection()` know to create one for it
4. If the purpose has no connection it is also likely it is already being reconnected but we just need to know which purpose it is to get the correct `runOnce()`
5. The condition also matches the one in `runCommand()`, if there's a connection supplied use that, else use the purpose. Here we have if we can get the purpose from the connection use that else use the fallback.

Might affect #922

### DIFF
--- a/app/logic/Mail/IMAP/IMAPAccount.ts
+++ b/app/logic/Mail/IMAP/IMAPAccount.ts
@@ -217,7 +217,7 @@ export class IMAPAccount extends MailAccount {
     // Note: Do not stop polling
     assert(connection, "Reconnect: Connection unknown");
 
-    purpose ??= this.connections.getKeyForValue(connection);
+    purpose = this.connections.getKeyForValue(connection) ?? purpose;
     assert(purpose, "Connection purpose unknown");
 
     return await this.reconnectRunOnce.get(purpose).runOnce(async () => {


app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -213,11 +213,11 @@ export class IMAPAccount extends MailAccount {
     });
   }
 
-  async reconnect(connection: ImapFlow): Promise<ImapFlow> {
+  async reconnect(connection: ImapFlow, purpose?: ConnectionPurpose): Promise<ImapFlow> {
     // Note: Do not stop polling
     assert(connection, "Reconnect: Connection unknown");
 
-    let purpose = this.connections.getKeyForValue(connection);
+    purpose = this.connections.getKeyForValue(connection) ?? purpose;
```
benbucksch, line 220:
```suggestion
    purpose ??= this.connections.getKeyForValue(connection);
```

benbucksch:
I applied a syntax fix. `??=` is a new operator specifically for this situation.

app/logic/Mail/IMAP/IMAPAccount.ts
jermy-c: I didn't do that way because if the connection has a different purpose than the one passed, we will be closing the wrong.

jermy-c:
> I applied a syntax fix. `??=` is a new operator specifically for this situation.

I didn't do it that way because if the connection has a different purpose than the one passed, we will be closing the wrong connection and creating a new connection for different purpose.

benbucksch:
> I didn't do it that way because if the connection has a different purpose than the one passed, we will be closing the wrong connection and creating a new connection for different purpose.

Can you please explain to me what you think the difference between the 2 versions is? From what I see, they should do *exactly* the same thing, and it's just a shorter syntax. Can you explain your logic, please, what you would like to do differently? Maybe you see something that I don't see.

benbucksch:
Oh, nevermind, I see the difference between the 2 statements. Mine prefers the passed-in value for `purpose`, yours prefers the purpose of the passed-in `connection`. My mistake, sorry.

But can you please still explain your reasoning? You have good point, I just didn't understand it. Can you walk me through "closing the wrong connection" in more detail, code line by line?

jermy-c:
> 2\. It will `reconnect()` for main and return a main connection. And we have to wait for it but if we had used the same key from purpose we could use the same `fetch` connection which has a `runOnce()` running for the connection.

Correction, it will return a main connection regardless in this case because the connection is no longer in the map.

The other case I think we will have this problem(connection purpose not equal to purpose) is if the `reconnect()` in the `runCommand()` is called first then the `reconnect()` in the listener.

benbucksch:
> Oh, nevermind, I see the difference between the 2 statements. Mine prefers the passed-in value for `purpose`, yours prefers the purpose of the passed-in `connection`. My mistake, sorry.

I've reverted my wrong change. Thanks for pointing that out!

I'm now running a build with your original PR here as my primary mail client, to test it regarding #922.

PR #1003, opened 2026-01-26
WebView: Convert string to Blob URL - fixes #986

### LINKED ISSUE #986: Mail: ERR_INVALID_URL when loading email content - keeps previous email content in display
### Reproduction
* Click on email A (any)
* Click on email B (specific email from Elena, subject "...what I sent to...")

### Actual result
* The email header changes to the email B
* The email content does *not* change and stays with the content of email A.
* The following error appears on the error console:

```
node:electron/js2c/isolated_bundle:2 Unexpected error while loading URL Error: Error invoking remote method 'GUEST_VIEW_MANAGER_CALL': Error: ERR_INVALID_URL (-300) loading 'data:text/html;charset=utf-8;base64,PGh0bWwgeG1sbnM9Imh0dHA6Ly93d3cudzMub3J…BVKzFFRjItMUVGRiwgVSsyMDIwLCBVKzIwQTAtMjBBQiwgVSsyMEFELTIwQ0YsIFUrMjExMywg'
```

### Expected result
* The content of email B appears

jermy-c:
- Convert string to Blob URL instead of Data URL because size limitations
- The WebView was throwing errors if it was displaying 2 4K inline images

### DIFF
--- a/app/frontend/Shared/WebView.svelte
+++ b/app/frontend/Shared/WebView.svelte
@@ -1,8 +1,8 @@
 // #if [!WEBMAIL && !MOBILE]
-<webview bind:this={webviewE} src={url ?? dataURL} {title} class:hidden {partition} />
+<webview bind:this={webviewE} src={url ?? blobURL} {title} class:hidden {partition} />
 // #else
 <!-- TODO Security: Test that this <webview> is untrusted and jailed -->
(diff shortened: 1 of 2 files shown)


benbucksch:
Thanks, Jeremy, for your fix here!

I filed #1005 about the inline images mentioned above.

PR #1006, opened 2026-01-28
Mobile: Don't include capacitor on desktop builds - #1004

### LINKED ISSUE #1004: Desktop: Mobile dependencies bundled
Mobile dependencies are being bundled in the desktop build also. It's increasing the app size and for some reason memory also.

There's a condition here, and it should be `false`. It's coming from the Capacitor import in `NavigationM.svelte`.
https://github.com/mustang-im/mustang/blob/a314f84d9ddb0babfcda4e15c5a6bac1c8b19e67/app/frontend/MainWindow/MainWindow.svelte#L24-L33

Image

benbucksch:
Fix #1004 Capacitor should not be included in desktop builds

Use compile-time `#if [MOBILE]` instead of runtime Svelte `{#if isMobile}`.

Completely untested @jermy-c Can you check whether this works on desktop and mobile, and whether it fixes the bug? If not, please adopt the PR and fix it to make it work.

### DIFF
--- a/app/frontend/MainWindow/MainWindow.svelte
+++ b/app/frontend/MainWindow/MainWindow.svelte
@@ -19,7 +19,7 @@
     {/if}
     <vbox flex>
       <NotificationBar notifications={$notifications} />
-      {#if appGlobal.isMobile}
+      // #if [MOBILE]
         <Router primary={false} {history}>
           <SplitterHorizontal name="sidebar" initialBottomRatio={0.7} hasTop={!!sidebar}>
             <vbox flex class="sidebar" slot="top">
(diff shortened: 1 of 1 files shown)


app/frontend/MainWindow/MainWindow.svelte
```diff
@@ -21,17 +21,18 @@
...
+        <SplitterHorizontal name="sidebar" initialBottomRatio={0.7} hasTop={!!sidebar}>
+          <vbox flex class="sidebar" slot="top">
+            <svelte:component this={sidebar} />
+          </vbox>
+          <AppContentRoutes slot="bottom" />
+        </SplitterHorizontal>
+        <NavigationM />
+      </Router>
+      // #else
+      {#if $selectedApp}
```
jermy-c: A runtime conditional within a conditional compile doesn't work properly... It causes it to compile the entire code to plain text `#if`

```svelte
// #else
{#if $selectedApp}
```

jermy-c (approved):
Capacitor is removed from the desktop build now.

app/frontend/MainWindow/MainWindow.svelte
```diff
@@ -21,26 +21,26 @@
...
+      <Router primary={false}>
+        <SplitterHorizontal name="sidebar" initialBottomRatio={0.7} hasTop={!!sidebar}>
+          <vbox flex class="sidebar" slot="top">
+            <svelte:component this={sidebar} />
+          </vbox>
+          <AppContentRoutes slot="bottom" />
+        </SplitterHorizontal>
+        <NavigationM />
+      </Router>
+      // #else
```
benbucksch, line 32: `{#if $selectedApp}` is still needed

benbucksch:
@jermy-c I only see `#if` in the desktop main window. I have no idea why, or what broke.

Issue #1005, opened 2026-01-28
Mail: Attachments: Inline images: Replace data: URLs with blob: URLs


benbucksch:
Compare #986
The problem is that [Chrome doesn't support `data:` URLs larger than 2 MB](https://groups.google.com/a/chromium.org/g/chromium-dev/c/zYYSVvn6a5M/m/WyZUweM4BAAJ) *in the renderer process*.

There's another place where we create `data:` URLs for attachments that are potentially larger: inline images. We embed these into the HTML that we generate. See [`EMail.addCID()`](https://github.com/mustang-im/mustang/blob/a314f84d9ddb0babfcda4e15c5a6bac1c8b19e67/app/logic/Mail/EMail.ts#L548). This is the case b) that I was [talking about above](https://github.com/mustang-im/mustang/pull/1003#issuecomment-3800799280) - sorry for the confusion. We can storing the blob URL on the attachment object. However, there we have the problem that we don't get a notification when the attachment is being freed. Weak references would be a solution, yes.

It seems we can use [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) for that purpose. Something like:
```
class Attachment ... {
  ...
  urlFinalizer = new FinalizationRegistry(url => {
    URL.revokeObjectURL(url);
    this._blobURL = null;
  });

  getBlobURL(attachment): string {
    if (this._blobURL) {
      return this._blobURL;
    }
    this._blobURL = URL.createObjectURL(this.content);
    this.urlFinalizer.register(attachment, this._blobURL);
    return this._blobURL;
  }
}
```

But I'm not sure whether that code makes any sense, particularly:
a) what the "held value" should be, and
b) whether a FinalizationRegistry can be a property of the `target` (here: Attachment object), of whether the FinalizationRegistry itself would keep the target alive.

I also don't know whether this is an actual issue. The problem in Chrome is only when the `data:` URL is sent over the IPC border to the renderer process. If we don't actually render the `attachment`s data: URL, then we might skirt around the problem. That's why this is a minor issue at the moment, unless we know otherwise.

PR #1008, opened 2026-01-29
Mail: Attachments: Inline images: Replace data: URLs with blob: URLs #1005

jermy-c:
- Using `blob` URLs seems to be better for memory and speed. At least with my 15MB inline image. The blobURL HTML is 12KB and the dataURL HTML is 22MB plus the existing the `File` object.

> Blob URLs are similar to [data URLs](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data), because they both allow representing in-memory resources as URLs; the difference is that data URLs embed resources in themselves and have severe size limitations, whereas blob URLs require a backing Blob or MediaSource and can represent larger resources.

- I think it's faster and using less memory because the URL is just pointing to the existing `File` object instead of creating more data or the string

- But we have an existing issue where the once the attachment is loaded it's not cleared after viewing but clearing it will require refetching it from the DB every time displaying it.

### DIFF
--- a/app/frontend/Shared/WebView.svelte
+++ b/app/frontend/Shared/WebView.svelte
@@ -97,7 +97,7 @@
     </style>`;
     let servers = allowServerCalls ? `* 'unsafe-inline'` : `'unsafe-inline'` ;
     const head = `<meta http-equiv="Content-Security-Policy" content="default-src 'none';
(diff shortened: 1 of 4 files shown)


benbucksch:
@jermy-c Why is this a Draft? Is this ready for merge and release to end users?

Did you test this properly?

benbucksch:
OK, as long as it's not regressing, let's land this.

PR #1009, opened 2026-01-29
IMAP: modseq is a bigint

NeilRashbrook:
I accidentally stumbled over this.

ImapFlow uses the `bigint` type for the `modseq` value, because that is a 64-bit integer and therefore not representable as a `number`.

Unfortunately JPC doesn't support the `bigint` type, turning any `bigint` values into `undefined` instead. As such, any `modseq` values that might have been generated by IMAP servers are getting ignored.

Because they are not `number`, I have of course updated the relevant IMAP code to use the correct type. I don't know the database support for `bigint` but I assume saving them as strings works.

<s>I've not tested any of the IMAP changes so I'm drafting this PR for now. (The updated JPC test passes locally.)</s>

### DIFF
--- a/app/logic/Mail/IMAP/IMAPFolder.ts
+++ b/app/logic/Mail/IMAP/IMAPFolder.ts
@@ -39,7 +39,7 @@ export class IMAPFolder extends Folder {
 
   /** Last sequence number seen */
   get lastModSeq(): bigint {
-    return BigInt(this.syncState ?? 0);
+    return BigInt(this.syncState ?? 0n);
   }
   set lastModSeq(val: bigint) {
     assert(typeof (val) == "bigint", "IMAP Folder modseq must be a bigint");
--- a/lib/jpc/protocol.test.js
+++ b/lib/jpc/protocol.test.js
@@ -90,17 +90,21 @@ test('Protocol', async () => {
   let collection = await start.makeCollection();
   expect(collection).toHaveProperty("self", collection);
   await expect(collection.isEmpty).resolves.toBe(true);
+  await collection.add(0);
   await collection.add(0n);
   await collection.add("");
+  await collection.add(null);
   await collection.add(undefined);
   await collection.add(false);
   await expect(collection.isEmpty).resolves.toBe(false);
-  await expect(collection.size).resolves.toBe(4);
+  await expect(collection.size).resolves.toBe(5);
   await expect(collection.contains(false)).resolves.toBe(true);
   await expect(collection.contains(true)).resolves.toBe(false);
-  await collection.setSize(3);
+  await collection.setSize(4);
+  await expect(collection.remove(0)).resolves.toBe(true);
   await expect(collection.remove(0n)).resolves.toBe(true);
   await expect(collection.size).resolves.toBe(2);
+  await expect(collection.remove(0)).resolves.toBe(false);
   await expect(collection.remove(0n)).resolves.toBe(false);
   await collection.clear();
   await expect(collection.isEmpty).resolves.toBe(true);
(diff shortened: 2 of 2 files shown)


lib/jpc/protocol.test.js
```diff
@@ -90,18 +90,18 @@ test('Protocol', async () => {
   let collection = await start.makeCollection();
   expect(collection).toHaveProperty("self", collection);
   await expect(collection.isEmpty).resolves.toBe(true);
-  await collection.add(0);
+  await collection.add(0n);
   await collection.add("");
-  await collection.add(null);
+  await collection.add(undefined);
```
benbucksch, line 97: Why this change? If you want to test `undefined`, then test both, but also `null`, please.

lib/jpc/protocol.test.js
```diff
@@ -90,18 +90,18 @@ test('Protocol', async () => {
   let collection = await start.makeCollection();
   expect(collection).toHaveProperty("self", collection);
   await expect(collection.isEmpty).resolves.toBe(true);
-  await collection.add(0);
+  await collection.add(0n);
```
benbucksch, line 94: Please keep the test for `0`. If necessary, add `0n`, but don't replace `0`.

app/logic/Mail/IMAP/IMAPFolder.ts
```diff
@@ -38,12 +38,12 @@ export class IMAPFolder extends Folder {
   }
 
   /** Last sequence number seen */
-  get lastModSeq(): number {
-    return this.syncState as number ?? 0;
+  get lastModSeq(): bigint {
+    return BigInt(this.syncState ?? 0);
```
```suggestion
    return BigInt(this.syncState ?? 0n);
```

benbucksch (changes requested):

app/logic/Mail/IMAP/IMAPFolder.ts
```diff
@@ -38,12 +38,12 @@ export class IMAPFolder extends Folder {
...
-    return this.syncState as number ?? 0;
+  get lastModSeq(): bigint {
+    return BigInt(this.syncState ?? 0n);
   }
-  set lastModSeq(val: number) {
-    assert(typeof (val) == "number", "IMAP Folder modseq must be a number");
-    this.syncState = sanitize.integer(val);
+  set lastModSeq(val: bigint) {
+    assert(typeof (val) == "bigint", "IMAP Folder modseq must be a bigint");
+    this.syncState = String(val);
```
benbucksch, line 46: Please *always* `sanitize`

NeilRashbrook: Sanitize what? I need this to be a string, because that's what the database is expecting. (Also despite TypeScript's best efforts to protect us, we're already checking the type of the value, plus `sanitizeDatatypes.ts` doesn't support `bigint` anyway.)

lib/jpc/protocol.test.js
NeilRashbrook: The old code wouldn't distinguish `0n` from `undefined`, thus this would detect any regression. But you're right, I should have added both and then fixed up all of the checks. (In particular, you changed the checks for size from `4` to `5` but of course those should be `6`; another check at the end of the file also fails now.)

app/logic/Mail/IMAP/IMAPFolder.ts
NeilRashbrook: Thanks for catching this; I should at the very least have used `"0"` (to match the type of `syncState`), or maybe `this.syncState ? BigInt(this.syncState) : 0n` was the way to go.

app/logic/Mail/IMAP/IMAPFolder.ts
benbucksch: > sanitizeDatatypes.ts doesn't support bigint anyway

Would be worth adding, given that we have integer() and float().

benbucksch: > Sanitize what?

That it's a bigint - as expected - before you save it as string.

lib/jpc/protocol.test.js
benbucksch: Yes, I couldn't get the test to run. Could you please adapt it?

NeilRashbrook: It runs when backported to the JPC repo (which is how I tested it), or do you mean it should run as a test within Mustang?

app/logic/Mail/IMAP/IMAPFolder.ts
NeilRashbrook: > > sanitizeDatatypes.ts doesn't support bigint anyway
> 
> Would be worth adding, given that we have integer() and float().

Filed PR #1023.

PR #1013, opened 2026-02-03
IMAP: Don't lock again if already locked - #1012

### LINKED ISSUE #1012: IMAP: Connection not available
I got these errors while using the app. I don't think it there's a specific reproduction other than setting up an IMAP account.

```
Error: Connection not available
    at message.js:155:16
    at new Promise (<anonymous>)
    at WSCall.makeCall (message.js:150:12)
    at JPCWebSocket.callRemote (protocol.js:184:31)
    at RemoteClass.next (obj.js:240:55)
    at IMAPFolder.ts:231:22
    at async IMAPFolder.runCommand (IMAPFolder.ts:93:14)
    at async IMAPFolder.fetchMessageList (IMAPFolder.ts:221:5)
    at async IMAPFolder.listChangedMessages (IMAPFolder.ts:189:27)
    at async IMAPFolder.listMessages (IMAPFolder.ts:124:19)
    at async loadFolder (MailApp.svelte:43:11)
```

```
Opening IMAP folder failed Error: Connection not available
    at message.js:155:16
    at new Promise (<anonymous>)
    at WSCall.makeCall (message.js:150:12)
    at JPCWebSocket.callRemote (protocol.js:184:31)
    at RemoteClass.getMailboxLock (obj.js:240:55)
    at IMAPFolder.runCommand (IMAPFolder.ts:72:36)
    at async IMAPFolder.fetchMessageList (IMAPFolder.ts:221:5)
    at async IMAPFolder.listChangedMessages (IMAPFolder.ts:189:27)
    at async IMAPFolder.listMessages (IMAPFolder.ts:124:19)
    at async loadFolder (MailApp.svelte:43:11)
```

jermy-c:
- There might have been a deadlock in `runCommand()`

If the `getMailboxLock(this.path)` before it throws. We don't release the lock for that reference. But we create a new lock without releasing the old lock.
https://github.com/mustang-im/mustang/blob/5dd1f847baa8ae13b9b5870b31891efa7f701bfd/app/logic/Mail/IMAP/IMAPFolder.ts#L67-L72

- Might fix #1012

### DIFF
--- a/app/logic/Mail/IMAP/IMAPFolder.ts
+++ b/app/logic/Mail/IMAP/IMAPFolder.ts
@@ -81,7 +81,7 @@ export class IMAPFolder extends Folder {
           conn = await this.account.reconnect(conn, purpose);
           if (doLock) {
             lock ??= await this.account.connectionLock.get(conn).lock();
-            lockMailbox = await conn.getMailboxLock(this.path);
+            lockMailbox ??= await conn.getMailboxLock(this.path);
           } else {
             await conn.mailboxOpen(this.path);
           }


app/logic/Mail/IMAP/IMAPFolder.ts
```diff
@@ -80,7 +80,7 @@ export class IMAPFolder extends Folder {
         if (ex.code == "NoConnection") {
           conn = await this.account.reconnect(conn, purpose);
           if (doLock) {
-            lock = await this.account.connectionLock.get(conn).lock();
+            lock ??= await this.account.connectionLock.get(conn).lock();
             lockMailbox = await conn.getMailboxLock(this.path);
```
```suggestion
            lockMailbox ??= await conn.getMailboxLock(this.path);
```

jermy-c:
> How exactly did you find this? What made you think that this might be the problem?

I found this while trying to fix #1012. I didn't have any idea how or where to find the fix so I just decided to read the code and find out logic flows if there were some issues with concurrency. Also, we have had existing issues with concurrency that's why decided to find out all cases even if it meant that I wouldn't get any fix from it.

app/logic/Mail/IMAP/IMAPFolder.ts
jermy-c: I didn't add this condition here because the `getMailboxLock()`. 1.) selects the mailbox for the given new connection 2.) the gets the lock for the given new connnection 

But if there's any new errors this might be the cause of it.

benbucksch: hm, you're right. Good point. Thank you.
The function name is misleading.
So, we need to first attempt to release the existing lock, then get a new lock.

jermy-c: Yes, the existing lock cannot be released until the entire command operation is complete otherwise we'll get a race condtion.

benbucksch: `reconnect()` now uses `runOnce()`, so another command will wait for the reconnect as well.

PR #1022, opened 2026-02-10
EMailCollection: Read message once

jermy-c:
- `getIndexRange()` was getting called very often and it is necessary to call it often to feel responsive. However, it also mean't that messages were being re-read while there was already a call been made to fetch the message from the DB.

**Call Process**

1. `getIndexRange(0, 10)`, while this is not resolve another called is already made
2. `getIndexRange(3, 10)`, indexes 3-12 are redundant calls in this case and the same will repeat if you scroll further

### Fix

- Wrap readMessage in `runOnce()`

### Addition

- Now that the readMessage in `runOnce()` we can use `this.readMessagesFromDB_parallel()`

### DIFF
--- a/app/logic/Mail/Store/EMailCollection.ts
+++ b/app/logic/Mail/Store/EMailCollection.ts
@@ -1,11 +1,13 @@
 import { ArrayColl, SortedCollection } from 'svelte-collections';
 import { PromiseAllDone } from '../../util/PromiseAllDone';
+import { RunOnce } from '../../util/RunOnce';
 import type { EMail } from '../EMail';
 import type { Folder } from '../Folder';
 
 export class EMailCollection<T extends EMail> extends SortedCollection<T> {
   folder: Folder;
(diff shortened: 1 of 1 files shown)


app/logic/Mail/Store/EMailCollection.ts
```diff
@@ -23,7 +25,7 @@ export class EMailCollection<T extends EMail> extends SortedCollection<T> {
 
     if (emails.length) {
       // async: Return before filtering and reading emails
-      this.readMessagesFromDB(emails)
+      this.readMessagesFromDB_parallel(emails)
```
benbucksch, line 28: Did you test how this performs, sequential vs. parallel? When I tested, the sequential one was faster, but that was without WAL.

jermy-c: It seems to be faster sequential is around 30-40ms and parallel is around 10-20ms.

PR #1023, opened 2026-02-10
Add support for bigint to sanitiser

NeilRashbrook:
As requested in #1009.

I wanted to mirror `sanitize.integer` which accepts things like `0xDEADBEEF` but I can easily change this.

### DIFF
--- a/lib/util/sanitizeDatatypes.ts
+++ b/lib/util/sanitizeDatatypes.ts
@@ -52,10 +52,7 @@ class Sanitize {
     if (typeof (unchecked) == "bigint") {
       return unchecked;
     }
-    if (!/^\s*(\d+|0x[\da-f]+)\s*$/i.test(unchecked)) {
-      return haveError("Not a bigint", unchecked, fallback);
-    }
-    return BigInt(unchecked);
+    return haveError("Not a bigint", unchecked, fallback);
   }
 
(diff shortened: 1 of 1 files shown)


benbucksch:
Thanks

> I wanted to mirror sanitize.integer which accepts things like 0xDEADBEEF but I can easily change this.

Yes, let's be strict about bigint and not accept strings in this case.

lib/util/sanitizeDatatypes.ts
```diff
@@ -48,6 +48,16 @@ class Sanitize {
...
   }
 
+  bigint(unchecked: bigint | string | null | undefined, fallback: bigint | null | Symbol = throwErrors): bigint {
+    if (typeof (unchecked) == "bigint") {
+      return unchecked;
+    }
+    if (!/^\s*(\d+|0x[\da-f]+)\s*$/i.test(unchecked)) {
+      return haveError("Not a bigint", unchecked, fallback);
+    }
+    return BigInt(unchecked);
```
NeilRashbrook: So like this?

```suggestion
    if (typeof (unchecked) == "bigint") {
      return unchecked;
    }
    return haveError("Not a bigint", unchecked, fallback);
```

app/logic/Mail/IMAP/IMAPFolder.ts
```suggestion
    return BigInt(this.syncState, 0n);
```

PR #1024, opened 2026-02-10
Build: Move e2/ to desktop/ and backend/ to desktop/backend/

benbucksch:
Move directories:
* e2/ -> desktop/
* backend/ -> desktop/backend/


benbucksch:
> TypeError: Cannot read properties of undefined (reading '_properties')

That's the `Observable` import in desktop/backend.ts:
import { Observable, notifyChangedProperty } from '../../lib/util/Observable';
but I cannot figure out what's wrong. VSCode says somethign about `rootDir`, but I don't know where that is. It's already in tsconfig.node.json.

PR #1026, opened 2026-02-10
EWS, OWA, ActS, Graph: Auth: Auto re-login

benbucksch:
Before: When auth fails, we showed an error to the user.
Now: When a call calls due to auth, we attempt exactly once to login non-interactively, and if that works, we re-try the call. If it fails, we show an error to the end user.

### DIFF
--- a/app/logic/Mail/OWA/OWAAccount.ts
+++ b/app/logic/Mail/OWA/OWAAccount.ts
@@ -287,7 +287,7 @@ export class OWAAccount extends MailAccount {
     await this.callOWA(request);
   }
 
-  async callOWA(aRequest: any, { mailbox = null, authRepeat = false } = {}): Promise<any> {
+  async callOWA(aRequest: any, { mailbox = null } = {}): Promise<any> {
     if (this.mainAccount) {
       let mainAccount = this.mainAccount as OWAAccount;
       return await mainAccount.callOWA(aRequest, { mailbox: this.username });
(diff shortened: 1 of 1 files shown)


app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -259,7 +259,7 @@ export class ActiveSyncAccount extends MailAccount {
     };
     if (this.oAuth2) {
       if (!this.oAuth2.isLoggedIn) {
-        throw new LoginError(null, "Please login");
+        await this.oAuth2.login(false);
       }
```
NeilRashbrook, line 255: Did you change the other protocols and forget to update this? Or, you could just write the code like this:
```suggestion
      await this.oAuth2.login(false);
```

benbucksch: I don't understand the problem with my code. In all protocols, I do the OAuth2 login only if we're not logged in (or the login expired). So, my code looks correct to me? I don't want to unconditionally log in every time, only when needed.

NeilRashbrook: > I don't want to unconditionally log in

Fair enough. But if `login` is supposed to be unconditional, why then does it start `if (this.isLoggedIn) { return`?

(The other problem is that the change in this file isn't the same as in the other files.)

benbucksch:
> OWA doesn't have the concept of a non-interactive OAuth login, so that code is utterly useless.

I removed the OWA auth re-try code.

PR #1031, opened 2026-02-12
Windows & Linux: Separate output for ARM

jermy-c:
- Change the artifact pattern for Windows and Linux only
- Change the Channel name for Windows ARM only, that is the update file `.yml`

### DIFF
--- a/.github/workflows/installer-windows-parula.yml
+++ b/.github/workflows/installer-windows-parula.yml
@@ -18,8 +18,8 @@ jobs:
       with:
         node-version: 24.x
 
-    - shell: bash
-      run: cd app/build; sh parula-brand.sh
+    - run: cd app/build; bash fix-installer-filenames.sh
+    - run: cd app/build; sh parula-brand.sh
     - run: cd app; yarn install
     - run: cd lib; yarn install
     - run: cd desktop; yarn install
--- a/app/build/mustang-brand.sh
+++ b/app/build/mustang-brand.sh
@@ -9,22 +9,6 @@ perl -p -i \
   -e "s|\"version\": \".*\"|\"version\": \"$VERSION\"|;" \
   ../../desktop/package.json
 
-# Include arm in artifact name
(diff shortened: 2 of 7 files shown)


.github/workflows/installer-windows-parula.yml
```diff
@@ -18,7 +18,8 @@ jobs:
       with:
         node-version: 24.x
 
-    - run: cd app/build; sh parula-brand.sh
+    - shell: bash
+      run: cd app/build; sh parula-brand.sh
```
benbucksch:
```suggestion
    - run: cd app/build; bash parula-brand.sh
```

app/build/mustang-brand.sh
```diff
@@ -9,6 +9,22 @@ perl -p -i \
   -e "s|\"version\": \".*\"|\"version\": \"$VERSION\"|;" \
   ../../desktop/package.json
 
+# Include arm in artifact name
```
benbucksch: This shouldn't be in the brand scripts, because we want this for all brands. This should run before branding.

Did you check whether electron builder allows to set the filename per arch in its own yml file,  avoiding the script based modifications?

benbucksch (changes requested):

app/build/mustang-brand.sh
jermy-c: > Did you check whether electron builder allows to set the filename per arch in its own yml file, avoiding the script based modifications?

Yes, I did. You can't set it per arch.

benbucksch: > You can't set it per arch.

Thanks for having checked that.

> This shouldn't be in the brand scripts, because we want this for all brands. This should run before branding.

This still needs to be fixed, but I need to make a release and land this now.
Can you please fix this in a followup PR?

benbucksch: DONE

.github/workflows/installer-windows-parula.yml
benbucksch: Fixed

PR #1033, opened 2026-02-13
Windows & Linux: Fix artifact and channel name for ARM

jermy-c:
- Linux: AppImage: Use default artifact name pattern which only includes arch if it's not x64
- Windows: Override the the artifact name pattern to include arch for ARM only
- Windows: Override the channel to include OS and ARCH in the updateInfo filename also
- Upgrade Electron Builder and its updater
- Sample Windows ARM run: https://github.com/mustang-im/mustang/actions/runs/22002970056
- Sample Linux ARM run: https://github.com/mustang-im/mustang/actions/runs/21994646168
- Sample Linux x64 run: https://github.com/mustang-im/mustang/actions/runs/21997174651
- Alternative to #1031


desktop/electron-builder.yml
```diff
@@ -47,8 +47,6 @@ linux:
   # <https://specifications.freedesktop.org/menu-spec/latest/additional-category-registry.html>
   category: Network
   icon: build/icon.icns
-appImage:
```
benbucksch, line 50: Why did we explicitly set this, originally?

jermy-c: This was from the `electron-vite` template. Commit https://github.com/mustang-im/mustang/commit/9ecf358656afd051e4484fd63d56410c12bd17af

PR #1036, opened 2026-02-23
Streamline login and startup code

NeilRashbrook:
Not only was `startup.ts` starting up all accounts in parallel at startup, the accounts themselves were also trying to start up their dependent accounts (obviously, only once they had finished logging in).
- Centralised code for each account to start up their dependent accounts once they've logged in.
- Centralised code to allow the user to ask a dependent account to log in and it will ask the main account.
- Centralised code for each type of account to start itself up in its own way.
- Don't try to start up dependent accounts automatically from `startup.ts`.
- EWS is a special snowflake because of its streaming notifications.


app/frontend/MainWindow/MainWindow.svelte
```diff
@@ -119,7 +119,7 @@
     if (appGlobal.emailAccounts.isEmpty && appGlobal.chatAccounts.isEmpty) {
       setup();
     } else {
-      await loginOnStartup(console.error, backgroundError);
+      await loginOnStartup();
```
benbucksch: I deliberately separated
* errors that happen during automatic login at startup (auto, non-interactive) (these should be handled specifically, because we don't want to spew errors on the user during startup, e.g. he might not have Wifi yet)
* errors that happen while the user manually logs in (interactive, must show error to user - these exceptions must bubble the call stack and throw into the button that the user clicked on - that's intentional and by design, so that the UI can decide how to show those - I will work on a nicer error display UI in the new 2 months or so)
* errors that happen during biff / mail check (should go to `Account.errorCallback`)
* errors that happen while parsing invididual mails (this is currently the same error handler as the previous one, but they should actually be separate as well.)

Please keep that separation.

app/logic/Abstract/Account.ts
```diff
@@ -58,8 +58,20 @@ export class Account extends Observable {
     this.color = randomAccountColor();
   }
 
+  /**
+   * Perform one-time post-login actions, such as performing an initial sync
+   * and/or setting up polling or notifications.
+   * By default all we do is to asynchronously notify dependent accounts.
+   * N.B. We can't make this `protected` because EWS needs to call it.
```
benbucksch:
```suggestion
```

Not specific to EWS. In fact, in this very implementation here, you are calling   `dependent.startup()`, which already would violate `protected` in JavaScript (unlike Java).

app/logic/Abstract/Account.ts
```diff
@@ -58,8 +58,20 @@ export class Account extends Observable {
     this.color = randomAccountColor();
   }
 
+  /**
+   * Perform one-time post-login actions, such as performing an initial sync
+   * and/or setting up polling or notifications.
+   * By default all we do is to asynchronously notify dependent accounts.
```
benbucksch:
```suggestion
   * Subclasses should override this and do protocol-specific actions like fetching messages,
   * but also call `super.startup()`.
   * By default, we asynchronously start up all dependent accounts.
```

app/logic/Abstract/Account.ts
```diff
@@ -58,8 +58,20 @@ export class Account extends Observable {
...
+   */
+  async startup() {
+    for (let dependent of this.dependentAccounts()) {
+      dependent.startup().catch(dependent.errorCallback);
+    }
+  }
+
   get isLoggedIn(): boolean {
-    return false;
+    return this.isDependentAccount && this.mainAccount.isLoggedIn;
```
benbucksch: This gives me a little bad feeling that this default generic implementation is for dependent accounts. That feels like opportunistic programming, not software architecture design.

app/logic/Files/FileSharingAccount.ts
```diff
@@ -16,6 +16,11 @@ export class FileSharingAccount extends Account {
     return dir;
   }
 
+  async startup() {
+    await this.sync();
+    await super.startup();
```
benbucksch: Shouldn't we call `super` first? (everywhere)

benbucksch (changes requested):
I really like your idea of the `startup()` function, which separates code out of the `login()` function. It was a mistake of mine to muddle those together. (We already separated them in Owl, and I think even in previous projects.)

That would also simplify the `verifyLogin()` implementation in many cases, hopefully removing `loginBase()`.

I also like how that simplifies `startup.ts`, which has incrementally grown. Really nice. Did you double-check that the new code really is functionally equivalent to the old one?

This is the right direction.

I don't like that the generic `Account` has the default implementation for dependent accounts. That happens to be convenient for the current code, but doesn't sound like the right software architecture design.

app/logic/Abstract/Account.ts
NeilRashbrook: TypeScript only complains about the call in EWS; it allows a method of a class to call any protected method visible in that class on any object of that class or any class derived from it. I believe this follows C++ behaviour; I'm not sure what Java does (it's confusing because it has an extra permission level).

app/logic/Files/FileSharingAccount.ts
NeilRashbrook: I feel that each account ought to be sure that it's been able to start itself up first before trying to start any of its dependent accounts.

app/frontend/MainWindow/MainWindow.svelte
NeilRashbrook: Thanks for reminding me about that use case; I've restored `startupErrorCallback` (`backgroundErrorCallback` was unused).

I can't quite decide what to do about the case where the main account logs in on startup but then the dependent account throws an error; with my current code this still gets logged to the account error callback, not the startup callback.

app/logic/Abstract/Account.ts
benbucksch: Can you give those 2 functions dedicated names, and *not* call them automatically? Specific subclasses can then call them as needed - at a time that is good (see the other comment below).

app/logic/Files/FileSharingAccount.ts
benbucksch: Yes, see my comment above for the Account.startup() function. If you make the dependent account init a separate function, you can first call `super.startup()` (which might not do anything at the moment), and then at the end of the function `this.startupDependentAccounts()` or somesuch. This makes it a) explicit and b) lets the subclass control when or if to init the subaccouts, and c) allows `super.startup()` to later do actually generic init code, at the start, not the end.

benbucksch (changes requested):

app/logic/Abstract/Account.ts
NeilRashbrook: I just removed the `isLoggedIn` changes. I think they were wrong anyway. (Maybe I could have got it to work with `!this.mainAccount || this.mainAccount.isLoggedIn`.)

app/logic/Files/FileSharingAccount.ts
NeilRashbrook: I moved some generic mail account init code into `super.startup()`.

app/logic/Abstract/Account.ts
```diff
@@ -77,6 +97,10 @@ export class Account extends Observable {
    */
   async login(interactive: boolean): Promise<void> {
     this.errors.clear();
+    if (this.isLoggedIn) {
+      return;
+    }
+    await this.mainAccount?.login(interactive);
   }
```
benbucksch: I've changed that to a) make it clearer that it's only for dependentAccount b) check dependent account before `isLoggedIn`, because the former is much cheaper than the latter.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -102,25 +102,29 @@ export class EWSAccount extends MailAccount {
...
+      if (dependent instanceof EWSAddressbook) {
+        dependent.listContacts()
+          .then(() => dependent.username != this.username && this.streamNotifications(dependent.folderID))
+          .catch(dependent.errorCallback);
+      } else if (dependent instanceof EWSCalendar) {
+        dependent.listEvents()
+          .then(() => dependent.username != this.username && this.streamNotifications(dependent.folderID))
+          .catch(dependent.errorCallback);
+      } else {
+        dependent.startup().catch(dependent.errorCallback);
```
benbucksch: I've removed the redundant code a little.

app/logic/WebMail/startup.ts
```diff
@@ -29,26 +29,26 @@ export async function getStartObjects(): Promise<void> {
...
-    await account.inbox.getNewMessages();
+    account.login(false).catch(startupErrorCallback);
   }
 }
 
 function backgroundErrorInAccount(ex: Error, account: Account) {
   account.errors.add(ex);
+  if (ex?.message) {
+    ex.message = `${account.name}: ${ex.message}`;
+  }
```
benbucksch: We don't need the account name in WebMail, because it's only for a single account of the provider running the webmail.

app/logic/WebMail/startup.ts
```diff
@@ -29,26 +29,26 @@ export async function getStartObjects(): Promise<void> {
...
-    await account.login(false);
+    if (!account.isDependentAccount) {
+      account.login(false).catch(startupErrorCallback);
+    }
   }
 
   for (let account of appGlobal.emailAccounts) {
     account.errorCallback = (ex) => backgroundErrorInAccount(ex, account);
-    await account.login(false);
-    assert(account.inbox, "Inbox not found");
```
benbucksch, line 45: The assert got lost. This can happen, if the server returns no folders. We need to show a proper error message to the user in this case, not just "getNewMessages of undefined".

app/logic/startup.ts
```diff
@@ -45,101 +45,27 @@ export async function getStartObjects(): Promise<void> {
...
-
-function errorWithAccountName(account: Account, errorCallback: (ex: Error) => void) {
-  return (ex: Error) => {
-    account.errors.add(ex);
-    if (ex?.message) {
-      ex.message = `${account.name}: ${ex.message}`;
-    }
-    errorCallback(ex);
-  };
-}
```
benbucksch, line 140: This got lost for startupErrorCallback

app/logic/startup.ts
```diff
@@ -45,101 +45,27 @@ export async function getStartObjects(): Promise<void> {
...
-
 function backgroundErrorInAccount(ex: Error, account: Account) {
   account.errors.add(ex);
+  if (ex?.message) {
+    ex.message = `${account.name}: ${ex.message}`;
+  }
```
benbucksch, line 76: This will also change the message in the `account.errors` list, which is not what we want.

But I see that `errorWithAccountName` had the same bug.

Do you have a smart way to fix that? Note that `Object.assign(new Error(), ex);` doesn't work, because .message, .stack, .name, possibly .cause are not enumerable. The new `structuredClone()` might work?

benbucksch (commented):
There is one big unclarity for me: login vs. list folders vs. get new messages

Does login() get new messages? a) At startup? b) When I manually click on get messages and/or login, in various places in the UI?

What from I can see, this PR changes behavior. In some cases, when the user manually clicks "get mail", we'd do 2 `getNewMessages()` in a row: one from login() -> startup() -> getNewMessages(), and another from the button itself, because login() so far didn't do getNewMessages().

This has to be staightened out before this can land.

benbucksch:
My preference would be (maybe not part of this PR, because it's a different refactoring change):
* *Not* let login() call startup(), but let the caller decide.
* *Not* let MailAccount.startup() call getNewMessages(), but let the caller decide. Unless all callers want that.

app/logic/startup.ts
```diff
@@ -45,90 +41,24 @@ export async function getStartObjects(): Promise<void> {
...
-      calendarLogin(account)
-        .catch(errorWithAccountName(account, startupErrorCallback));
+  for (let account of allAccounts) {
+    if (account.loginOnStartup && !account.isDependentAccount) {
+      (async () => {
+        if (!account.isLoggedIn) {
+          await account.login(false);
+        }
+        await account.startup();
+      })().catch(errorWithAccountName(account, startupErrorCallback));
```
NeilRashbrook, line 57: So you do want to separate login and startup? I don't have a problem with that, assuming you go on to remove all of the calls to `this.startup();` from all of the `login` implementations.
```suggestion
      (async () => {
        try {
          if (!account.isLoggedIn) {
            await account.login(false);
          }
          await account.startup();
        } catch (ex) {
          startupErrorInAccount(ex, startupErrorCallback);
        }
      })();
```
(thought it looked clearer to use try/catch; needs implementation of `startupErrorInAccount`)

app/logic/startup.ts
NeilRashbrook: Yes, `structuredClone` should work on `Error` objects.

benbucksch (approved):
Looks good!

NeilRashbrook:
> My preference would be (maybe not part of this PR, because it's a different refactoring change):
> 
>  * _Not_ let login() call startup(), but let the caller decide.

Indeed, main accounts are still getting started up twice at the moment.

PR #1040, opened 2026-02-24
Mac: Build: Include bufferutil and utf-8-validate in x64ArchFiles - fixes #1037

### LINKED ISSUE #1037: Mac: Build fails with "bufferutil.node ... same in both x64 and arm64 builds"
https://github.com/mustang-im/mustang/actions/runs/22345203463/job/64657926916

Detected file "Contents/Resources/app.asar.unpacked/node_modules/bufferutil/prebuilds/darwin-arm64/bufferutil.node" that's the same in both x64 and arm64 builds and not covered by the x64ArchFiles rule: "undefined"  failedTask=build stackTrace=Error: Detected file "Contents/Resources/app.asar.unpacked/node_modules/bufferutil/prebuilds/darwin-arm64/bufferutil.node" that's the same in both x64 and arm64 builds and not covered by the x64ArchFiles rule: "undefined"

jermy-c:
- Include bufferutil and utf-8-validate in x64ArchFiles
- Both of these packages download the prebuild for all OSs and Archs and it's trying to merge the `.node` files
- The `Release/` has a `.node` file also
- This makes it keep them without trying to merge them
- Also it's a fixed based on the error message:
> "Contents/Resources/app.asar.unpacked/node_modules/bufferutil/prebuilds/darwin-arm64/bufferutil.node" that's the same in both x64 and arm64 builds and not covered by the x64ArchFiles rule: "undefined"

### DIFF
--- a/desktop/electron-builder.yml
+++ b/desktop/electron-builder.yml
@@ -35,6 +35,7 @@ mac:
     target: default
     arch:
       - universal
+  x64ArchFiles: '**/{bufferutil,utf-8-validate}.node'
 dmg:
   artifactName: ${name}-${version}.${ext}
 linux:


desktop/electron-builder.yml
```diff
@@ -35,6 +35,7 @@ mac:
     target: default
     arch:
       - universal
+  x64ArchFiles: '**/{bufferutil,utf-8-validate}.node'
```
benbucksch, line 38: Is this a complete list, or additional to the automatically selected files? (should be the latter)

benbucksch (approved):

desktop/electron-builder.yml
jermy-c: I think it's the complete list. I couldn't find what's the default. But bettersqlite3 is a native module but without downloading all prebuilds. It is rebuilt everytime.

That's the complete list: 
- https://github.com/electron-userland/electron-builder/blob/5d4651dbb29d0730d0fb3be96de79d9348c0a99c/packages/app-builder-lib/src/macPackager.ts#L195
- https://www.electron.build/app-builder-lib.interface.macconfiguration#x64archfiles

jermy-c: The binaries are single arch but electron-universal detects them as merged binaries: https://github.com/electron/universal?tab=readme-ov-file#skip-lipo-for-certain-binaries-in-your-universal-app

benbucksch: It's OK as hotfix, but we should have a proper fix.

benbucksch:
I've triggered a test build https://github.com/mustang-im/mustang/actions/runs/22373954633

Issue #1047, opened 2026-02-27
JPC: Invalid string length - breaks CardDAV sync


benbucksch:
### Reproduction
* Setup CardDAV account with 1000 contacts
* Sync

### Actual result
* Error: "Invalid string length"
* The error comes from `JSON.stringify()` in `WSCall.send()`
* The object that we try to send is too big

### Expected result
* JPC works with any data, even large data

#### De-duplicate objects

* The trigger seems to be that the tsdav library includes, for every contact, a reference to the server response object, which contains the entire list of contacts. So, the amount of data multiplies: 1000 contacts each contain 1000 responses = 1000000 responses.
* When marshalling the object, we should keep track (in a map) of which objects we have already included, and if an object repeats, include only a reference to it, i.e. de-duplicating them. The object IDs should be created only on-demand, only if an object actually repeated.
* When unmarshalling, the objects need to be reconstructed. When encountering an object with ID, put it in a map. When encountering a reference, look up the original object in the map, and include the full object in the result.

#### Split up large data objects or arrays

* Despite de-duplication, it can be that the actual data is simply large, even without object duplication. We should not fail in this case.
* Figure out a way to split up large objects, and send them in multiple messages, so that each message stays below the limit for Javascript strings and `JSON.stringify()`.
* Split-up should happen only when necessary, only for really large objects.

Please make 2 separate PRs for the above 2 fixes. In both cases, make sure to not slow down the normal cases that don't need these fixes.

benbucksch:
> De-duplicate objects

We already have `_recursiveObjects`, but that doesn't seem to work in this case.
In this case, it's not recursive, the object is just repeated.

PR #1051, opened 2026-02-27
JPC: Don't send the same plain object twice in one message #1047

NeilRashbrook:
Note that this does not currently apply to arrays, so if you have an array of identical arrays of identical objects the identical objects will be deduplicated but the identical arrays will not; this is why the test actually uses an object that mimics an array.

### DIFF
--- a/lib/jpc/obj.js
+++ b/lib/jpc/obj.js
@@ -46,10 +46,25 @@ class RemoteClass {
   }
 }
 
+class IncomingContext {
+  /** @type Map<number, object> */
+  plainObjects = new Map();
+  /** @type Map<string, object> */
+  stubs = new Map();
+}
+
+class OutgoingContext {
+  count = 0;
+  /** @type Map(object, object) */
+  plainObjects = new Map();
+  /** @type Map(object, string) */
+  localObjects = new Map();
+}
+
 export default class BaseProtocol {
   /**
    * @param method {string}
-   * @paqram [payload] {any}
+   * @param [payload] {any}
    * @returns {Promise<any>}
    */
   callRemote(method, payload) {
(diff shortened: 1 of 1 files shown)


benbucksch:
Here's an idea: If we want to output the id only for actually repeated objects, you could make it a map from the object to the ouput value. If it repeats, you can retroactively add the id, given that we actually serialize only at the end.

benbucksch:
Instead of letting all callers set the repeated, can you make a context object, default null? if mapOutgoingObjects finds its context to be null, it creates it. This way, you localize the context creation in a single point. Make it a context object, not just the bare repeated object, because we could integrate also other state, like _recursiveObjects , and other data later on, so let's make it generic. There is only one context per response.

BTW: Wouldn't it make sense to merge repeated and _recursiveObjects , given that they do almost the same thing?

lib/jpc/obj.js
```diff
@@ -186,24 +201,23 @@ export default class BaseProtocol {
...
 
-  async updateObjectProperties(obj, properties) {
+  updateObjectProperties(obj, properties) {
     for (let propName in properties) {
       Object.defineProperty(obj, propName, {
         configurable: true,
         enumerable: true,
         writable: false,
-        value: await this.mapIncomingObjects(properties[propName]),
+        value: properties[propName],
```
benbucksch, line 220: Can you please explain this change?

NeilRashbrook: The callers now call `mapIncomingObjects` on the `properties` object. This simplifies the code, but also allows deduplication between the properties.

benbucksch: > allows deduplication between the properties

How so? Please elaborate.

NeilRashbrook: Previously the code for handling Svelte notifications at line 561/583 called `mapIncomingObjects` separately for each property. I could have created a context manually to share it between the properties, but I thought it was easier to call `mapIncomingObjects` on the `properties` object at 573/595. Without that, each property would have been serialised individually.

PR #1057, opened 2026-03-03
Mail: Move messages in a generic function for all protocols

benbucksch:
* Create a generic function `Folder.moveOrCopyMessagesHere()`. Move all the logic into that function. That ensures that the logic is the same across all protocols.
* The abstract function `moveOrCopyMessagesOnServer()` is called by it and implemented by all the protocols. In most cases, that's all that a protocol needs to implement.
* A protocol can still override `moveMessagesHere()` or `copyMessagesHere()`, if necessary.

### DIFF
--- a/app/logic/Mail/Folder.ts
+++ b/app/logic/Mail/Folder.ts
@@ -129,26 +129,33 @@ export class Folder extends Observable implements TreeItem<Folder> {
   }
 
   async moveMessagesHere(messages: Collection<EMail>) {
-    throw new AbstractFunction();
+    await this.moveOrCopyMessagesHere("move", messages);
   }
 
   async copyMessagesHere(messages: Collection<EMail>) {
-    throw new AbstractFunction();
+    await this.moveOrCopyMessagesHere("copy", messages);
   }
 
   /**
-   * Helper function for `copyMessagesHere()` and `moveMessagesHere()`
-   * @returns
-   * true = Move has already been done (across accounts)
-   * false = Caller needs to move messages (within the same account) */
-  protected async moveOrCopyMessages(action: "move" | "copy", messages: Collection<EMail>): Promise<boolean> {
+   * Helper function for `copyMessagesHere()` and `moveMessagesHere()`.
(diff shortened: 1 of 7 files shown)


app/logic/Mail/Folder.ts
```diff
@@ -129,37 +129,56 @@ export class Folder extends Observable implements TreeItem<Folder> {
...
+
+    this.countTotal += messages.length;
+    if (action == "move") {
+      sourceFolder.countTotal -= messages.length;
+      for (let sourceMsg of messages) {
+        await sourceMsg.deleteMessageLocally();
+      }
+    }
+
+    await this.moveOrCopyMessagesOnServer(action, messages);
```
NeilRashbrook, line 177: Of course this doesn't handle the case where the move fails, e.g. due to insufficient permissions to delete the messages from the source folder.

app/logic/Mail/Folder.ts
```diff
@@ -207,7 +226,7 @@ export class Folder extends Observable implements TreeItem<Folder> {
 
   async rename(newName: string): Promise<void> {
     let disabled = this.disableRename();
-    assert(!disabled, disabled);
+    assert(!disabled, disabled || "Cannot rename");
```
NeilRashbrook, line 229: (Always nice to see the compile error count go down!)

app/logic/Mail/Folder.ts
NeilRashbrook: Thanks for (mostly) fixing this in df0883a. (I think the messages have still been visibly removed from the folder's message list at this point.)

benbucksch: Yes, I was thinking about what you said, and you are right that this way is safer, for the reasons that you mentioned. I just needed to ponder over this a few days.

And yes, it was important for me that the user gets immediate visual feedback, but, as you say, we provide that by removing the messages from folder.messages immediately. This gives us both: The immediate UI feedback, but if something goes wrong, we still have the emails saved in the folder.

PR #1062, opened 2026-03-04
Build: Allow seeing node.js network traffic including fetch() in Chrome DevTools

benbucksch:
To debug, we need to see the network requests that our app does. To get around cross-origin limitations of Chrome, almost all of them happen in the node.js. So, we need to see those.

Once this works, a developer should be able to do:
* `cd desktop/; yarn debug` (instead of `dev`)
* Start Chrome
* Type `chrome://inspect`
* Enable "[x] Discover network targets"
* Below "Remote target", you should see `localhost:9222`
* Below it, there is a small blue link "Inspect". Click that.
* You should see the normal Chrome DevTools, but now connected to node.js, instead of the webpage
* Open the "Networking" tab
* Check an EWS account, or any other HTTPS-based account
* You should see a network call in the DevTools.

But you don't. In node.js, we use `ky`, which uses node `fetch()`, which uses `undici`, not `http`. Unfortunately, the new network debugging in node.js 22.6.0+ only supports `http` and not `undici`, so the calls don't appear. @jermy-c Can you look into this, please?

### DIFF
--- a/desktop/package.json
+++ b/desktop/package.json
@@ -32,6 +32,7 @@
     "bufferutil": "^4.1.0",
     "electron-updater": "^6.3.3",
     "node-network-devtools": "^1.0.29",
+    "undici": "^7.22.0",
     "utf-8-validate": "^6.0.6",
     "ws": "^8.19.0"
   },
--- a/desktop/package.json
+++ b/desktop/package.json
@@ -31,6 +31,7 @@
     "better-sqlite3": "^12.5.0",
     "bufferutil": "^4.1.0",
     "electron-updater": "^6.3.3",
+    "node-network-devtools": "^1.0.29",
     "utf-8-validate": "^6.0.6",
     "ws": "^8.19.0"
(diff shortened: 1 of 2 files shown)


NeilRashbrook:
OWA for Office 365 already uses `fetch` in the front-end and you can debug it "normally" there (you get lots of CORS requests which the back end fakes out). Would you like me to see whether I can get EWS and ActiveSync to work using front-end `fetch`? (Note that OWA for on-premises Exchange needs cookie separation, which still has to be done in the back end.)

benbucksch:
@jermy-c Can we remove "undici" and the other new dependencies?

jermy-c:
> @jermy-c Can we remove "undici" and the other new dependencies?

I think we can remove "undici" because Electron is updated and the main process Node.js is v24 now.

But we might still need "node-network-devtools" because Node.js doesn't fully support network debugging.

PR #1068, opened 2026-03-05
EWS, OWA: Implement `mayHaveSubAccounts` and `listPossibleSubAccounts()` #1061

NeilRashbrook:
When implementing this for EWS I discovered a better way to skip the unwanted contact folders.

### DIFF
--- a/app/logic/Mail/OWA/Request/OWAFolderRequests.ts
+++ b/app/logic/Mail/OWA/Request/OWAFolderRequests.ts
@@ -1,9 +1,6 @@
 import { OWARequest } from "./OWARequest";
 import type { OWAEMail } from "../OWAEMail";
 import type { ExchangePermission } from "../../EWS/EWSFolder";
-
-// IconIndexPidTag is currently unused because I don't know how to
-// correctly format an OWA request for it, so we can't import it yet.
 // import { IconIndexPidTag } from "../../EWS/EWSFolder";
 
 export function owaFindMsgsInFolderRequest(folderID: string, maxFetchCount: number): OWARequest {


benbucksch:
Isn't this duplicating a lot of the code for listing folders?

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -648,6 +648,10 @@ export class EWSAccount extends MailAccount {
               }, {
                 FieldURI: "folder:DistinguishedFolderId",
               }],
+              t$ExtendedFieldURI: {
+                PropertyTag: "4340",
```
benbucksch: Can you please make this a constant, whose name also expresses what 4340 means?
Use it below as well.

benbucksch: What exactly does this new filter do? Why is it OK to filter this, when it wasn't filtered before?

NeilRashbrook: It was filtered before, but it was a hack - it checked the DistinguishedFolderId was either unset or set to "contacts". This filter is the official way.

I've converted all of the various `PropertyTag` constants and also linked their definitions on learn.microsoft.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -518,6 +497,84 @@ export class OWAAccount extends MailAccount {
...
+    let accounts = [];
+    for (let folder of contacts) {
+      if (this.dependentAccounts().find(account => account.protocol == "addressbook-owa" && (account as OWAAddressbook).folderID == folder.FolderId.Id)) {
+        isMainAddressbook = false;
+        continue;
+      }
+      let addressbook = newAddressbookForProtocol("addressbook-owa") as OWAAddressbook;
+      addressbook.initFromMainAccount(this);
+      if (!isMainAddressbook && folder.DisplayName) {
+        addressbook.name = `${(folder.account || this).name} ${folder.DisplayName}`;
```
benbucksch: What is this change?
a) This doesn't belong here
b) What is the old behavior and what is the new one?
c) `folder` here is the server response, no? The new code will ignore the name that the user gave to the account, won't it?

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Thanks.

I've changed the names to not use Hungarian notation. While Microsoft does that, we don't do that in our code.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -180,22 +180,11 @@ export class OWAAccount extends MailAccount {
 
     let haveAddressbook = appGlobal.addressbooks.some(addressbook => addressbook.mainAccount == this);
     if (!haveAddressbook) {
-      let isMainAB = true;
       let response = await this.callOWA(new OWAGetPeopleFiltersRequest());
-      for (let ab of response) {
-        // Exclude internal contacts folders.
-        if (ab.IsReadOnly || !ab.FolderId?.Id) {
-          continue;
```
benbucksch, line 188: Can readonly folders, or those without ID, appear first in the list?
If so, then the new code won't work (both yours and mine).
In that case, we could use an out parameter, like `createCalendarAccount(folder: any, { isMain = false })`. Not nice, but luckily only in this case.

benbucksch: Also, calendar has a distinguishedfolderid even in OWA, but the main addressbooks does not?
There's other code below that checks it: `if (folder.DistinguishedFolderId == "contacts") {`. Why does the OWA code for addressbooks (but not calendar, and not EWS) use the first addressbook as main, and not the `DistinguishedFolderId == "contacts"` criteria to find main, like everywhere else?

NeilRashbrook: Yes, read-only folders can appear first in the list. I'm not sure why you think the new code doesn't work. I don't think the caller needs to know whether there is a main address book. (Well, not until you rewrote it.)

`GetPeopleFilters` doesn't return distinguished folder IDs. I guess we could make a separate request to find the folder ID of the contacts folder and identify the main addressbook that way...

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: Github isn't showing this as a change, so I'm not 100% sure what you mean, but the new code needs to work with shared addressbooks, for which we set the subaccount for the inbox/mail folders as the account property.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -28,8 +28,8 @@ import { assert, blobToBase64, ensureArray, NotReached, NotSupported, type Json
 import { gt } from "../../../l10n/l10n";
 import { ArrayColl } from "svelte-collections";
 
-// https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxprops/01b52d3c-d194-4a8c-83ee-4ac7506339da
-const PidTagAttributeHidden = "0x10F4";
+// <https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxprops/01b52d3c-d194-4a8c-83ee-4ac7506339da>
+const HiddenPidTag = "0x10F4";
```
NeilRashbrook, line 32: Note that there are `PidTag` instances in OWA code as well; I assume you want them to be changed too?

(Technically there are other `PidTag` values for hidden things, but I doubt we'll use them.)

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -828,59 +830,51 @@ export class EWSAccount extends MailAccount {
         }
       }
     }
-    accounts.addAll(this.createAddressbookAccounts(contacts));
-    accounts.addAll(this.createCalendarAccounts(calendars));
+    accounts.addAll(addressbooks.map(ab => this.createAddressbookAccount(ab)).filter(a => a));
+    accounts.addAll(calendars.map(cal => this.createCalendarAccount(cal)).filter(a => a));
```
NeilRashbrook:
```suggestion
    accounts.addAll(addressbooks.map(ab => this.createAddressbookAccount(ab)).filter(Boolean));
    accounts.addAll(calendars.map(cal => this.createCalendarAccount(cal)).filter(Boolean));
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -846,7 +846,7 @@ export class EWSAccount extends MailAccount {
     addressbook.initFromMainAccount(this);
     let isMainAddressbook = !folder.account && folder.DistinguishedFolderId == "contacts";
     if (!isMainAddressbook && folder.DisplayName) {
-      addressbook.name = `${(folder.account || this).name} ${folder.DisplayName}`;
+      addressbook.name = `${this.name} ${sanitize.nonemptylabel(folder.DisplayName)}`;
```
NeilRashbrook: ... which makes this wrong, sorry.

```suggestion
      addressbook.name = `${(folder.account || this).name} ${sanitize.nonemptylabel(folder.DisplayName)}`;
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -868,7 +868,7 @@ export class EWSAccount extends MailAccount {
     if (isMainCalendar) {
       calendar.useForInvitations = true;
     } else if (folder.DisplayName) {
-      calendar.name = `${(folder.account || this).name} ${folder.DisplayName}`;
+      calendar.name = `${this.name} ${sanitize.nonemptylabel(folder.DisplayName)}`;
```
NeilRashbrook:
```suggestion
      calendar.name = `${(folder.account || this).name} ${sanitize.nonemptylabel(folder.DisplayName)}`;
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -698,14 +698,16 @@ export class EWSAccount extends MailAccount {
     }
     let haveAddressbook = appGlobal.addressbooks.some(addressbook => addressbook.mainAccount == this);
     if (!haveAddressbook) {
-      for (let addressbook of this.createAddressbookAccounts(ensureArray(result.RootFolder.Folders.ContactsFolder))) {
+      for (let folder of ensureArray(result.RootFolder.Folders.ContactsFolder)) {
+        let addressbook = this.createAddressbookAccount(folder);
         await addressbook.save();
```
NeilRashbrook, line 705: `addressbook` is possibly `null`.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -698,14 +698,16 @@ export class EWSAccount extends MailAccount {
...
         appGlobal.addressbooks.add(addressbook);
       }
     }
     let haveCalendar = appGlobal.calendars.some(calendar => calendar.mainAccount == this);
     if (!haveCalendar) {
-      for (let calendar of this.createCalendarAccounts(ensureArray(result.RootFolder.Folders.CalendarFolder))) {
+      for (let folder of ensureArray(result.RootFolder.Folders.CalendarFolder)) {
+        let calendar = this.createCalendarAccount(folder);
```
NeilRashbrook, line 714: `calendar` is possibly `null`.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -537,7 +537,7 @@ export class OWAAccount extends MailAccount {
     let addressbook = newAddressbookForProtocol("addressbook-owa") as OWAAddressbook;
     addressbook.initFromMainAccount(this);
     if (!isMainAddressbook && folder.DisplayName) {
-      addressbook.name = `${(folder.account || this).name} ${folder.DisplayName}`;
+      addressbook.name = `${this.name} ${sanitize.nonemptylabel(folder.DisplayName)}`;
```
NeilRashbrook:
```suggestion
      addressbook.name = `${(folder.account || this).name} ${sanitize.nonemptylabel(folder.DisplayName)}`;
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -556,7 +556,7 @@ export class OWAAccount extends MailAccount {
     if (isMainCalendar) {
       calendar.useForInvitations = true;
     } else if (folder.DisplayName) {
-      calendar.name = `${(folder.account || this).name} ${folder.DisplayName}`;
+      calendar.name = `${this.name} ${sanitize.nonemptylabel(folder.DisplayName)}`;
```
NeilRashbrook:
```suggestion
      calendar.name = `${(folder.account || this).name} ${sanitize.nonemptylabel(folder.DisplayName)}`;
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -516,63 +517,52 @@ export class OWAAccount extends MailAccount {
...
           if (folder.DistinguishedFolderId == "calendar") {
             calendars.push(folder);
           }
         }
       }
     }
-    accounts.addAll(this.createAddressbookAccounts(contacts));
-    accounts.addAll(this.createCalendarAccounts(calendars));
+    accounts.addAll(addressbooks.map(ab => this.createAddressbookAccount(ab, addressbooks[0] == ab)).filter(a => a));
+    accounts.addAll(calendars.map(cal => this.createCalendarAccount(cal)).filter(a => a));
```
NeilRashbrook:
```suggestion
    accounts.addAll(calendars.map(cal => this.createCalendarAccount(cal)).filter(Boolean));
```

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -516,63 +517,52 @@ export class OWAAccount extends MailAccount {
...
           }
           if (folder.DistinguishedFolderId == "calendar") {
             calendars.push(folder);
           }
         }
       }
     }
-    accounts.addAll(this.createAddressbookAccounts(contacts));
-    accounts.addAll(this.createCalendarAccounts(calendars));
+    accounts.addAll(addressbooks.map(ab => this.createAddressbookAccount(ab, addressbooks[0] == ab)).filter(a => a));
```
NeilRashbrook: This is extremely unobvious.
```suggestion
    accounts.addAll(addressbooks.map((ab, i) => this.createAddressbookAccount(ab, i == 0)).filter(Boolean));
```

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Feel free to import these constants (and only that) directly from EWSAccount.ts et al. Better than duplicating them, given that they are identical, due to how Exchange works.

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Note: The old code did only `this.name`.
I asked about that in another comment. Please answer in the other comment where I asked the question.

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Right.

Can you please fix that? (ditto below)

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: Ah, I thought about that, but didn't know that `map()` has an `i` param. Nice.

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/map#parameter

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: Ah, you filter before. I missed that. All good, then.

let contacts = response.filter(ab => !ab.IsReadOnly && ab.FolderId?.Id);

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: > This is extremely unobvious.

Maybe, in this case, we make an explicit `for` loop:
```
```

```suggestion
    let i = 0;
    for (let folder of addressbooks) {
      let isMain = i++ == 0;
      let addressbook = this.createAddressbookAccount(ab, isMain);
      if (!addressbook) {
        continue;
      }
      accounts.add(addressbook);
    }
```

You won't like that either, because you like brevity.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: I did. https://github.com/mustang-im/mustang/pull/1068#discussion_r2942623526

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: OK, that is very confusing. Because `folder` is the server response. Setting an app object as property of an XML object is ... unexpected.

Is there a clearer way to pass the account?

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -691,40 +698,20 @@ export class EWSAccount extends MailAccount {
...
-          addressbook.initFromMainAccount(this);
-          let isMainAddressbook = folder.DistinguishedFolderId == "contacts";
-          if (!isMainAddressbook && folder.DisplayName) {
-            addressbook.name = `${this.name} ${folder.DisplayName}`;
-          }
-          addressbook.folderID = sanitize.nonemptystring(folder.FolderId.Id);
-          await addressbook.save();
-          appGlobal.addressbooks.add(addressbook);
-        }
+      let addressbooks = ensureArray(result.RootFolder.Folders.ContactsFolder).filter(folder => folder.ExtendedProperty.Value != "true");
```
benbucksch: Much nicer. Now just a linebreak.

```suggestion
      let addressbooks = ensureArray(result.RootFolder.Folders.ContactsFolder)
        .filter(folder => folder.ExtendedProperty.Value != "true");
```

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -766,6 +753,126 @@ export class EWSAccount extends MailAccount {
...
+    }
+    if (folder.account) {
+      addressbook.username = folder.account.username;
+    }
+    addressbook.folderID = sanitize.nonemptystring(folder.FolderId.Id);
+    return addressbook;
+  }
+
+  private createCalendarAccount(folder: any): EWSCalendar | null {
+    if (this.dependentAccounts().find(account => account.protocol == "calendar-ews" && (account as EWSCalendar).folderID == folder.FolderId.Id)) {
```
benbucksch, line 864:
```suggestion
    assert(folder.FolderClass == "IPF.Appointment", "Need calendar");
    if (this.dependentAccounts().find(account => account.protocol == "calendar-ews" && (account as EWSCalendar).folderID == folder.FolderId.Id)) {
```

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: I've tried to pass the account more clearly in commit 15c4cb8.

app/logic/Mail/OWA/Request/OWAFolderRequests.ts
```diff
@@ -2,6 +2,10 @@ import { OWARequest } from "./OWARequest";
 import type { OWAEMail } from "../OWAEMail";
 import type { ExchangePermission } from "../../EWS/EWSFolder";
 
+// IconIndexPidTag is currently unused because I don't know how to
+// correctly format an OWA request for it, so we can't import it yet.
```
benbucksch:
```suggestion
```

app/logic/Mail/OWA/Request/OWAFolderRequests.ts
```diff
@@ -23,7 +24,7 @@ export function owaFindMsgsInFolderRequest(folderID: string, maxFetchCount: numb
         /*}, {
           __type: "PropertyUri:#Exchange",
           ExtendedFieldURI: {
-            PropertyTag: "0x1080",
+            PropertyTag: IconIndexPidTag,
```
benbucksch, line 27: If you want to capture why it's commented out, you can put the comment here, instead of at the import.

(There are a number of other places in the code where I disabled code, and I left the import in place, or commented out. But the comment is in the code, not at the import)

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: Thanks, also for the comment of this unusual code bit.

Issue #1058, opened 2026-03-04
iCal: Preserve unknown properties


benbucksch:
Like vCard in CardDAV, we should be able to read, modify, and save an iCal event in CalDAV without destroying properties that we don't know/support.

We already save the original iCal string together with the event in our database. After the user modified one property, and we save the event back to the server, we should read the original iCal, apply the properties from the `Event`, and leave unsupported properties as-is, then save that back to the server. It's OK (and to a degree necessary) that we overwrite properties that we do support. After all, the user needs to be able to delete specific properties. See how the process works for vCard in `CardDAVPerson`. It should work the same in `CalDAVEvent`.

Please do make the function names more intutive and not purely opportunistic. See e.g. how `JSContact` and `JSCalendarEvent` work.

PR #1072, opened 2026-03-10
CalDAV: Preserve unknown properties #1058

NeilRashbrook:
Note: This is quite a large PR; you may want to review each commit individually - 5e949c3 is the big one so to speak; the rest is just boilerplate to simplify the changes.


app/logic/util/VParser.ts
```diff
@@ -1,11 +1,11 @@
-import { gt } from "../../../l10n/l10n";
-import { assert } from "../../util/util";
+import { gt } from "../../l10n/l10n";
```
benbucksch, line 1: Please leave this file in its original directory. `logic/util/` is not a generic dumping ground for any file used in more than 2 directories, but only for code that is potentially needed everywhere.

The new filename is good. So, app/logic/Calendar/ICal/VParser.ts.
It's OK that Contacts/ imports it from Calendar/ . There is no better solution in this case, given that it's specific to iCal and vCard only.

app/logic/Calendar/ICal/EventToICal.ts
```diff
@@ -2,152 +2,137 @@ import type { Event } from "../Event";
...
-      break;
-    default:
-      lines.push(["ATTENDEE", "RSVP", "TRUE", "CN", participant.name, "MAILTO:" + participant.emailAddress]);
-      break;
-    }
-  }
-  lines.push(["END", "VEVENT"]);
-  lines.push(["END", "VCALENDAR"]);
-  return lines.map(line2ical).join("");
+  return `BEGIN:VCALENDAR\r\n${method ? `METHOD:${method}\r\n` : ""}VERSION:2.0\r\nPRODID:-//Beonex//${appName}//EN\r\n${eventToVEvent(event)}END:VCALENDAR\r\n`;
```
benbucksch: Can you write that as array of lines (strings) and `.join("\r\n")`, please? That would be more readable. (End with a "" line in the array, to keep thr trailing newline.)

same below

app/logic/Calendar/ICal/EventToICal.ts
```diff
@@ -2,152 +2,137 @@ import type { Event } from "../Event";
...
+export function vEventToContainer(vevent: VObject): Record<string, string[]> {
+  let container: Record<string, string[]> = Object.create(null);
+  for (let key in vevent.entries) {
+    container[key] = vevent.entries[key].map(entry => entry.line);
+  }
+  return container;
+}
+
+export function updateContainerFromEvent(event: Event, container: Record<string, string[]>) {
+  setValue(container, "dtstamp", datetime2ical(new Date()));
```
benbucksch, line 58: Why not make `container` a proper class? `container.setValue("dtstamp", ...)`

app/logic/Calendar/ICal/EventToICal.ts
```diff
@@ -2,152 +2,137 @@ import type { Event } from "../Event";
...
+  setValue(container, "uid", event.calUID);
+  setValue(container, "summary", event.title);
+  setValue(container, "location", event.location || event.isOnline && event.onlineMeetingURL);
```
benbucksch, line 63: You lost all the code comments. They are critically important information on why we do it. They are an **essential** part of the code.
Why did you remove them??

Without these comments, nobody else will understand the logic, and there will be regressions and bugs later.

Same everywhere below.

If you want my reviews to be fast, I need to trust you that your patch *only* changes what it advertizes in the PR title, and keeps everything else as-is. Otherwise, it makes my job as reviewer exceedingly difficult. I feel like in a mine field.

I will continue reviewing this PR another time.

app/test/logic/Calendar/ICal/ICalEMailProcessor.test.ts
```diff
@@ -29,19 +32,34 @@ const allFiles = await fs.readdir(dataDir);
 const testFiles = allFiles.filter(name => name.endsWith(".ics")).map(name => name.slice(0, -4));
 test.each(testFiles)("Parse %s", async name => {
   const calendar = await fs.readFile(new URL(name + ".ics", dataDir), { encoding: 'utf-8' });
+  let text = calendar;
```
benbucksch:
```suggestion
  let text = calendar; // updated below
```

*Please* make your code readable.

app/test/logic/Calendar/ICal/ICalEMailProcessor.test.ts
```diff
@@ -29,19 +32,34 @@ const allFiles = await fs.readdir(dataDir);
...
   const [invitationMessage, event] = JSON.parse(await fs.readFile(new URL(name + ".json", dataDir), { encoding: 'utf-8' }));
   const processor = new ICalEMailProcessor();
   const email: EMail = {
     attachments: [{
       mimeType: "text/calendar",
       content: {
         text() {
-          return calendar;
+          return text;
```
benbucksch, line 43:
```suggestion
          return text; // variable `text` is updated below, after creation of the object
```

app/test/logic/Calendar/ICal/ICalEMailProcessor.test.ts
```diff
@@ -17,7 +18,7 @@ function toJSON(event: Event) {
     timezone: event.timezone != "UTC" && event.timezone || undefined,
     allDay: event.allDay,
     calUID: event.calUID,
-    location: event.location,
+    location: event.location || undefined,
```
benbucksch: Why is this necessary? is `location` an empty string?

If you mean to convert `null`  to `undefined`, please use `??`, not `||`. Use `||` only when you specifically want to cover an empty string, false, 0 etc.

app/test/logic/Calendar/ICal/ICalEMailProcessor.test.ts
```diff
@@ -17,7 +18,7 @@ function toJSON(event: Event) {
...
     descriptionText: event.descriptionText || undefined,
```
benbucksch: This looks wrong. An empty string here should be valid as description, no?

app/test/logic/Calendar/ICal/ICalEMailProcessor.test.ts
```diff
@@ -53,4 +54,12 @@ test.each(testFiles)("Parse %s", async name => {
   text = getICal(email.event);
   await processor.process(email, null);
   expect(toJSON(email.event as Event)).toEqual(event);
+  let parsed = new VContainer(calendar);
+  text = getUpdatedVEvent(email.event, parsed.objects.vevent[0]);
+  for (let line of calendar.match(/^BEGIN:VEVENT$.*^END:VEVENT$/ims)[0].match(/^\w.+/gm)) {
+    // Ignore lines that we know we will have rewritten
+    if (!/\b(begin|method|end|version|prodid|dtstamp|summary|location|description|styled-description|x-alt-desc|dtstart|dtend|recurrence-id|rrule|status|organizer|attendee)\b/i.test(line)) {
```
benbucksch, line 61: This is going to be difficult to maintain. It will break every time we add another property.
is there another way, other than listing all properties here?

app/logic/Calendar/ICal/EventToICal.ts
```diff
@@ -11,6 +11,11 @@ export function getICal(event: Event, method?: iCalMethod): string {
   return `BEGIN:VCALENDAR\r\n${method ? `METHOD:${method}\r\n` : ""}VERSION:2.0\r\nPRODID:-//Beonex//${appName}//EN\r\n${eventToVEvent(event)}END:VCALENDAR\r\n`;
 }
 
+export function getUpdatedICal(event: Event, original: string, method?: iCalMethod) {
+  let parsed = new VContainer(original);
+  return `BEGIN:VCALENDAR\r\n${method ? `METHOD:${method}\r\n` : ""}VERSION:2.0\r\nPRODID:-//Beonex//${appName}//EN\r\n${getUpdatedVEvent(event, parsed.objects.vevent[0])}END:VCALENDAR\r\n`;
```
benbucksch: I think this at least is the third place where we have the header, including the specific app ID. Can we have a function for that, please?

app/logic/Calendar/ICal/EventToICal.ts
```diff
@@ -2,152 +2,137 @@ import type { Event } from "../Event";
...
 
-function utc2ical(date: Date): string {
-  return date.toISOString().replace(/-|:|\..../g, "");
+function containerToVEvent(container: Record<string, string[]>): string {
+  return "BEGIN:VEVENT\r\n" + Object.values(container).flat().map(line => line.match(/.{1,75}/gu).join("\r\n ")).join("\r\n") + "\r\nEND:VEVENT\r\n";
 }
 
-function date2ical(date: Date): string {
-  return String(date.getFullYear()) + String(date.getMonth() + 1).padStart(2, "0") + String(date.getDate()).padStart(2, "0");
+function setValue(container: Record<string, string[]>, key: string, value: string | null, parameters: Record<string, string> = {}) {
```
benbucksch, line 141: Please document this function, extensively, given how central it is.

benbucksch (changes requested):
I like your renames very much. The new names are much clearer than the old names. I'd really like to take this patch.

Also because it fixes an important bug.

I also like that we no longer write iCal just by pushing lines.

Unfortunately, the PR changes too many things at once, and therefore, I will need a lot more time to review.

app/logic/Calendar/ICal/EventToICal.ts
benbucksch: also, as mentioned below, this should probably be factored out into a function.

NeilRashbrook: Unfortunately it's not that easy, as `method` is optional, and `eventToVEvent` returns a string that already ends in a newline. I could split it into separate lines and use `+`...

app/test/logic/Calendar/ICal/ICalEMailProcessor.test.ts
NeilRashbrook: The JSON check files don't specify a description if the ICS file doesn't have one. Which is all but 13 of them. My guess is that when the test was originally written, if the ICS file didn't have a description, the retrieved value was still undefined.

app/test/logic/Calendar/ICal/ICalEMailProcessor.test.ts
NeilRashbrook: I don't know. I have a similar problem with the vCard test.

app/test/logic/Calendar/ICal/ICalEMailProcessor.test.ts
benbucksch: I am specifically asking about the difference between "no description" and "description is the empty string". What if the VCS file defines "" as description? We should be able to test for that, no?

NeilRashbrook: I might be mistaken, but I seem to recall that an empty description is illegal per the vCard spec.

benbucksch: From what I know, empty string is legal, but please do check. If you're right, then this code is fine.
Otherwise, this should probably be `??` instead of `||`

NeilRashbrook: An empty string is legal, but old versions of Thunderbird used to log a console warning when their parser found an empty string. (I haven't checked current Thunderbird.)

As such there is no difference between "no description" and "description is the empty string".

However the actual issue was that when the test was originally written, we did distinguish between them by making the description undefined when there was no description. This was changed by PR #449 to make it so that the raw text was null when there was no description. I've updated the code accordingly.

I may have been wrong about "location" though; I had an incorrect test file I had left over for some reason. I've removed that change.

benbucksch: Make sure to write the tests after the spec, not adapting the test code to the failures. :-)

Either way is fine for me, for now. Just please be conscious that `||` is different from `??`, and you want `??` most of the time, unless you *specfically* want to treat "" as null. Use `??` otherwise. Sorry that I'm so picky about this specific operator.

app/logic/util/VParser.ts
benbucksch: Still TODO

PR #1073, opened 2026-03-10
Check for various consent errors during OAuth2

NeilRashbrook:
Note: I don't actually know of an account that I can used to test for consent failures; all the accounts I know of already have consent or aren't compatible for some reason.

### DIFF
--- a/app/logic/Auth/OAuth2Error.ts
+++ b/app/logic/Auth/OAuth2Error.ts
@@ -18,17 +18,34 @@ export class OAuth2ServerError extends OAuth2Error {
   codes: number[];
   details: any;
   consentRequired = false;
+  adminPolicy = false;
   constructor(prefix: string, json: any) {
     let msg = prefix + ": " +
       sanitize.nonemptystring(json?.error_description, null)?.split(/[\r\n]/)[0].replace(/^\w+: /, "")
       ?? sanitize.nonemptystring(json?.error, null)?.replace("_", " ")
       ?? "Login failed. Unknown OAuth2 error.";
     super(msg);
+    if (json.error == "access_denied") {
+      this.consentRequired = true;
+    }
+    // Google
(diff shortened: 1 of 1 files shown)


app/logic/Auth/OAuth2Error.ts
```diff
@@ -18,18 +18,35 @@ export class OAuth2ServerError extends OAuth2Error {
...
   consentRequired = false;
+  adminPolicy = false;
   constructor(prefix: string, json: any) {
     let msg = prefix + ": " +
       sanitize.nonemptystring(json?.error_description, null)?.split(/[\r\n]/)[0].replace(/^\w+: /, "")
       ?? sanitize.nonemptystring(json?.error, null)?.replace("_", " ")
       ?? "Login failed. Unknown OAuth2 error.";
     super(msg);
+    if (json.error == "access_denied") {
+      this.consentRequired = true;
```
benbucksch, line 29: This is too broad. `access_denied` is *any* OAuth2 failure, including wrong password.

benbucksch (changes requested):

app/logic/Auth/OAuth2Error.ts
NeilRashbrook: https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#error-codes-for-authorization-endpoint-errors says otherwise.

benbucksch: What does the OAuth2 spec say?

https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1
"access_denied
               The resource owner or authorization server denied the
               request."

Note "or authorization server"

That matches what I remember: `access_denied` is the fallback error code whenever the auth server wants to say "no", and there's no more specific error code.

PR #1074, opened 2026-03-10
Mail: SQL: Turn on WAL mode and minimize duplicates - fixes #1065

### LINKED ISSUE #1065: Mail: Lag when email with event is being loaded or parsed
### Reproduction

1. Start Mustang and make sure everything is just being loaded
2. Select an email with a event that's not loaded yet
3. Select another email with or without an event
4. Repeat

### Actual Result

1. Lags after step 2 with a rainbow spinning wheel on Mac

### Expected Result

1. No lag
2. No rainbow spinning wheel
3. Can continue loading/parsing in the background

### Environment

- MacOS
- Test with EWS and IMAP accounts

**Related to** https://github.com/mustang-im/mustang/issues/1055#issuecomment-4000975841. The lag might not be from `listMessage()` maybe from `loadBody()`, `loadMIME()` or `SQLEmail.read()`.

jermy-c:
- `REPLACE` didn't work because the `emailID` column didn't have a `UNIQUE` constraint
- We didn't delete the email from the content DB when it was delete from the email DB

### DIFF
--- a/app/logic/Mail/SQL/Source/SQLSourceEmailMigrate.ts
+++ b/app/logic/Mail/SQL/Source/SQLSourceEmailMigrate.ts
@@ -1,39 +1,33 @@
 import { mailSourceDatabaseSchema } from "./createSourceDatabase";
-import { RunOnce } from "../../../util/flow/RunOnce";
 import sql, { type Database } from "../../../../../lib/rs-sqlite";
 
-const runOnce = new RunOnce<void>();
-
 export async function migrateToNewSchema(database: Database) {
-  // Run in runOnce because this seems to be called multiple times in parallel
-  await runOnce.runOnce(async () => {
-    if (await isEmailIDUnique(database)) {
-      return;
-    }
-    try {
-      await database.execute(sql`
-        BEGIN TRANSACTION;
+  if (await isEmailIDUnique(database)) {
+    return;
+  }
+  try {
+    await database.execute(sql`
+      BEGIN TRANSACTION;
 
-        -- Delete duplicates from the DB, keep the latest version
-        DELETE FROM emailMIME
-        WHERE id NOT IN
-          (SELECT MAX(id)
-            FROM emailMIME
-            GROUP BY emailID);
+      -- Delete duplicates from the DB, keep the latest version
+      DELETE FROM emailMIME
+      WHERE id NOT IN
+        (SELECT MAX(id)
+          FROM emailMIME
+          GROUP BY emailID);
 
-        ALTER TABLE emailMIME RENAME TO emailMIME_old;
-        $${mailSourceDatabaseSchema}
+      ALTER TABLE emailMIME RENAME TO emailMIME_old;
+      $${mailSourceDatabaseSchema}
 
-        INSERT INTO emailMIME SELECT * FROM emailMIME_old;
-        DROP TABLE emailMIME_old;
+      INSERT INTO emailMIME SELECT * FROM emailMIME_old;
+      DROP TABLE emailMIME_old;
 
-        END TRANSACTION;
-      `);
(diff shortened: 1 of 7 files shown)


benbucksch (changes requested):
Good changes.

Can you please add migration (see issue comment)?

app/logic/Mail/SQL/Source/SQLSourceEmailMigrate.ts
```diff
@@ -0,0 +1,51 @@
...
+
+      END TRANSACTION;
+    `);
+  } catch(ex) {
+    await database.run(sql`ROLLBACK TRANSACTION`);
+    throw ex;
+  }
+}
+
+export async function isEmailIDUnique(database: Database) {
```
benbucksch, line 27: I thought we had some kind of "version" field in the database? How does the bettersqlite "migrate" function work?

jermy-c: Yes, we do have you access it with `PRAGMA user_version` and the `migrate()` function should be incrementing it with each successful migration SQL or function arg. I was including the migrations in a single function but instead we should be appending a separate function or argument for each migration. It's working correctly.

benbucksch: What I mean is: If we have versioning, we don't need the entire `isEmailIDUnique()` function, do we?

benbucksch: What I'm concerned about is that this function would run on every startup.

jermy-c: > What I mean is: If we have versioning, we don't need the entire `isEmailIDUnique()` function, do we?

Yes, you're right. We just have to make sure all migration `sql` and function calls are successful so `user_version` increments and then subsequent startups wouldn't run any of the `sql` or functions.

app/logic/Mail/SQL/Source/SQLSourceEmailMigrate.ts
```diff
@@ -0,0 +1,42 @@
+import { mailSourceDatabaseSchema } from "./createSourceDatabase";
+import sql, { type Database } from "../../../../../lib/rs-sqlite";
+
+export async function migrateToNewSchema(database: Database) {
```
benbucksch, line 4:
```suggestion
export async function addUNIQUEEMailID(database: Database) {
```

benbucksch (approved):
Must better, thanks!

Could you please:
* Rename the function, as commented
* Split this into 3 PRs: One for each migration, and one for the delete?

? Thanks!

benbucksch:
I understand that this is how the migration system is supposed to work.
However, I do not like the idea that a new user has to iterate through all migrations. I do not want that, but rather, I want a new user to immediately have the correct, most recent schema.
That means that we need to check whether the migration is necessary before applying it. This also adds an extra layer of protection, which is good.

As an added benefit, the initial schema also acts as documentation for us developers how the DB schema is looking now.

So, Jeremy did it correctly here.

jermy-c:
I've split the changes into 3 PRs:

1. https://github.com/mustang-im/mustang/pull/1079: Has the changes for UNIQUE emailID with the migration function renamed
2. https://github.com/mustang-im/mustang/pull/1080: Has the changes for the folderID, dateSent DESC index
3. https://github.com/mustang-im/mustang/pull/1078: Has the changes for the deleting email from the content DB

But https://github.com/mustang-im/mustang/pull/1077 seems to be needed also because I saw the database getting locked sometimes with parallel connections when I was testing https://github.com/mustang-im/mustang/pull/1080.

Issue #1059, opened 2026-03-04
JMAP: Calendar: Support recurrence etc.


benbucksch:
I wrote some basic implementation of `JSCalendarEvent`. However, the recurrence is just a stub, only frequence and interval are supported, but not "4th Thursday", and no exclusion and exceptions. Please add that.

If you see other things missing, feel free to add them as well.

Please *do* keep the code structure and style as-is. Readability is more important than brevity.

RFC 8984 is outdated. Please implement [JSCalendar 2.0](https://www.ietf.org/archive/id/draft-ietf-calext-jscalendarbis-15.html).

PR #1075, opened 2026-03-10
JMAP: Support recurring events #1059

NeilRashbrook:
- [X] Support recurrence rules
  - [X] Reading
  - [X] Writing
- [X] Support exclusions
  - [X] Reading
  - [X] Writing
- [X] Support exceptions
  - [X] Reading
  - [X] Writing

I also added time zone support.

### DIFF
--- a/app/logic/Calendar/JMAP/JMAPCalendar.ts
+++ b/app/logic/Calendar/JMAP/JMAPCalendar.ts
@@ -1,4 +1,5 @@
 import { Calendar } from "../Calendar";
+import type { Event } from "../Event";
 import { JMAPEvent } from "./JMAPEvent";
 import type { TJMAPCalendar } from "./TJMAPCalendar";
 import type { TJMAPCalendarEvent } from "./TJSCalendar";
@@ -68,6 +69,7 @@ export class JMAPCalendar extends Calendar {
       let { newEvents, updatedEvents } = await this.fetchEvents(i, batchSize + 1);
       this.events.addAll(newEvents);
       await this.saveEvents(newEvents);
+      await this.updateRecurrenceOverrides(newEvents);
       allNewEvents.addAll(newEvents);
       hasMore = newEvents.length + updatedEvents.length > batchSize;
     }
@@ -205,7 +207,9 @@ export class JMAPCalendar extends Calendar {
           await event.deleteLocally();
         }
         await this.saveEvents(changedResult.updatedEvents);
+        await this.updateRecurrenceOverrides(changedResult.updatedEvents);
         await this.saveEvents(addedResult.newEvents);
+        await this.updateRecurrenceOverrides(addedResult.newEvents);
         if (this === calendar) {
           newEventsOfThisCal = addedResult.newEvents;
         }
@@ -270,10 +274,18 @@ export class JMAPCalendar extends Calendar {
     return { newEvents, updatedEvents };
   }
 
-  protected async saveEvents(events: Collection<JMAPEvent>) {
+  protected async saveEvents(events: Collection<Event>) {
     for (let event of events) {
       await event.save();
     }
+  }
+
+  protected async updateRecurrenceOverrides(events: Collection<JMAPEvent>) {
+    let exceptions = new ArrayColl<Event>();
+    for (let event of events) {
+      event.updateRecurrenceOverrides(exceptions);
+    }
+    await this.saveEvents(exceptions);
   }
 
   getEventByJMAPID(jmapID: string): JMAPEvent | undefined {
--- a/app/logic/Calendar/JMAP/JSCalendarEvent.ts
+++ b/app/logic/Calendar/JMAP/JSCalendarEvent.ts
@@ -1,5 +1,5 @@
 import { Event, RecurrenceCase } from "../Event";
-import type { TJMAPCalendarEvent } from "./TJSCalendar";
+import type { TJMAPCalendarEvent, TJSCalendarRecurrenceRule, TJSCalendarRecurrenceFrequency } from "./TJSCalendar";
 import { Frequency, RecurrenceRule, type RecurrenceInit } from "../RecurrenceRule";
 import { InvitationResponse } from "../Invitation/InvitationStatus";
 import { Participant } from "../Participant";
@@ -9,7 +9,9 @@ import { NotImplemented, assert } from "../../util/util";
 export class JSCalendarEvent {
   static toEvent(jmap: TJMAPCalendarEvent, event: Event) {
     // ID and metadata
-    event.calUID = sanitize.nonemptystring(jmap.uid, null);
+    if (!event.parentEvent) {
+      event.calUID = sanitize.nonemptystring(jmap.uid, null);
+    }
     event.lastUpdateTime = sanitize.date(jmap.updated, null);
     event.original = jmap;
 
@@ -24,7 +26,7 @@ export class JSCalendarEvent {
     }
 
     // Time
-    event.startTime = sanitize.date(jmap.start);
+    event.startTime = this.toDate(jmap.start, jmap);
     event.endTime = getEndTimeFromDuration(event.startTime, sanitize.nonemptystring(jmap.duration));
     event.timezone = jmap.timeZone ?? null;
     event.allDay = sanitize.boolean(jmap.showWithoutTime, false);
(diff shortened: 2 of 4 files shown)


app/logic/Calendar/JMAP/JSCalendarEvent.ts
```diff
@@ -1,11 +1,22 @@
...
-import type { TJMAPCalendarEvent } from "./TJSCalendar";
+import type { TJMAPCalendarEvent, TJSCalendarRecurrenceRule } from "./TJSCalendar";
 import { Frequency, RecurrenceRule, type RecurrenceInit } from "../RecurrenceRule";
 import { InvitationResponse } from "../Invitation/InvitationStatus";
 import { Participant } from "../Participant";
 import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
 import { NotImplemented, assert } from "../../util/util";
 
+/** Maps between jsCal names and JavaScript day of week values */
+enum jsCalWeekday {
```
benbucksch: Please _always_ start the file with the main class. Everything else is at the bottom.

app/logic/Calendar/JMAP/JSCalendarEvent.ts
```diff
@@ -33,30 +44,33 @@ export class JSCalendarEvent {
...
-              r.frequency == "daily" ? Frequency.Daily :
-                Frequency.Yearly;
+      let rrinit: RecurrenceInit = {
+        masterDuration: event.duration,
+        seriesStartTime: event.startTime,
+        seriesEndTime: sanitize.date(r.until, null),
+        count: sanitize.integer(r.count, Infinity),
       // iCal and we don't support hourly, minutely or secondly. Esp. the latter is pretty silly
-      rrinit.seriesStartTime = event.startTime;
+        frequency: sanitize.translate(r.frequency, { yearly: Frequency.Yearly, monthly: Frequency.Monthly, weekly: Frequency.Weekly, daily: Frequency.Daily }, Frequency.Yearly),
```
benbucksch: Nice.

But the line is too long. Please format the map with one line per value (and end with `,`)

app/logic/Calendar/RecurrenceRule.ts
```diff
@@ -4,7 +4,6 @@ import { getDateTimeFormatPref, gt } from "../../l10n/l10n";
 import { assert } from "../util/util";
 
 export enum Frequency {
-  None = "NONE",
```
benbucksch, line 7: This is used by the UI and needs to stay.

app/logic/Calendar/JMAP/JSCalendarEvent.ts
```diff
@@ -130,27 +144,31 @@ export class JSCalendarEvent {
...
-      } else if (er.frequency == Frequency.Weekly) {
-        jr.frequency = "weekly";
-      } else if (er.frequency == Frequency.Daily) {
-        jr.frequency = "daily";
-      } // else Unknown `Frequency` -> keep as-is
+      let jr: TJSCalendarRecurrenceRule = jmap.recurrenceRule = {
+        frequency: er.frequency.toLowerCase() as Lowercase<Frequency>,
+        interval: er.interval,
+        firstDayOfWeek: jsCalWeekday[er.first],
+      };
```
benbucksch: Can you please preserve unknown properties, as I did?

```suggestion
      let jr: TJSCalendarRecurrenceRule = jmap.recurrenceRule ??= {};
      jr.frequency = er.frequency.toLowerCase() as Lowercase<Frequency>;
      jr.interval = er.interval;
      jr.firstDayOfWeek = jsCalWeekday[er.first];
```

app/logic/Calendar/JMAP/JSCalendarEvent.ts
NeilRashbrook: Now if only this had been an import it would have been at the top 😉 

(Having noticed it in the code, I had actually been wondering where `kLocationNameDescriptionSeparator` was defined. I guess I'm used to C/C++?)

app/logic/Calendar/JMAP/JSCalendarEvent.ts
NeilRashbrook: We use less than half of the _known_ sub-properties. We'll have to delete all of the unused known properties, because we don't want to preserve them, because they could conflict with any changes to used properties. (I don't think it makes sense to preserve unknown _sub_-properties.)

app/logic/Calendar/RecurrenceRule.ts
NeilRashbrook: Oops. I knew this was too easy.

app/logic/Calendar/JMAP/JSCalendarEvent.ts
benbucksch: If your editor supports TypeScript, you just need to press F12 (or the comparable function).

app/logic/Calendar/RecurrenceRule.ts
benbucksch: FWIW, an editor supporting TypeScript can do "Show all occurrences" (not a grep, but anything specifically using this variable/constant/function/class etc.).

app/logic/Calendar/JMAP/JSCalendarEvent.ts
benbucksch: Are you talking about properties that we do support, or those that we don't support?

Supported properties are easy: Write the current value from `Event`. If that current value is nothing, then that nothing (e.g. null or undefined or default value) will be written to JSCalendar.

Unsupported properties: Do you mean that they would influence the time, in a way that we don't understand, so the event would happen in other clients at a different time than in ours? That's a legitimate concern, and a special case, given that it's about the time of the meeting.

In all other cases, we have to leave unsupported properties in place. This is super important for compatibility with other clients, and even with future versions of Mustang.

app/logic/Calendar/JMAP/JMAPCalendar.ts
```diff
@@ -68,6 +69,7 @@ export class JMAPCalendar extends Calendar {
       let { newEvents, updatedEvents } = await this.fetchEvents(i, batchSize + 1);
       this.events.addAll(newEvents);
       await this.saveEvents(newEvents);
+      await this.updateRecurrenceOverrides(newEvents);
```
benbucksch, line 72: That's odd. This should not be necessary.

Events can come from all kinds of sources, and should not require some manual special "update" function.

* How does that work in the other protocols?
* How do we calculate recurrences in the calendar app?

NeilRashbrook: > Events can come from all kinds of sources

???

> How does that work in the other protocols?

We download the single and master events. For each event, the event is parsed and saved; this includes exclusions. At this point, ActiveSync has already given us full exception data as a property of the master, but in the case of EWS/OWA, the exceptions have to be downloaded in a separate request. Either way these exceptions are then parsed and saved.

Note that there's a slight logic error in that master events with exclusions get saved multiple times, because `makeExclusionLocally` saves the master event. This is a regression from 2d32757 in that the old code did not save the master event when making an exclusion while parsing the event.

> How do we calculate recurrences in the calendar app?

???

app/logic/Calendar/JMAP/JSCalendarEvent.ts
NeilRashbrook: No, I don't use Visual Studio.

app/logic/Calendar/RecurrenceRule.ts
NeilRashbrook: Unfortunately the occurrences are in Svelte, not TypeScript, so I forget about them. (Particularly since the 3.0.0 @tsconfig/svelte config file is incompatible with TypeScript 5.9.3 in the package.json... should that have been updated when Svelte was?)

app/logic/Calendar/JMAP/JSCalendarEvent.ts
NeilRashbrook: Yes, I'm talking about properties that are known to influence the time or dates at which recurrences happen. If the user edits the recurrence in Mustang, we definitely need to remove those properties, otherwise it's possible that other clients would be confused by them. There are also some properties that are documented but I wasn't sure whether they would confuse other clients if we left them in.

benbucksch (approved):

PR #1077, opened 2026-03-11
Mail: Email: Load and parse MIME once

jermy-c:
- `loadEvent` and `loadForDisplay` both call `loadMIME` or `parseMIME` in parallel
- Changing the `loadedBody` property causes an extra call in parallel also
- This wraps `loadMIME` and `parseMIME` in runOnce to prevent extra calls in parallel
- This also fixes: 
> For some reason the code even after the existing .migrate() runs more than once in parallel so I had to wrap it in a runOnce.

from: https://github.com/mustang-im/mustang/pull/1074#issuecomment-4035113601
With extra call and I also see the database is being opened multiple times in parallel in the backend.

### DIFF
--- a/app/logic/Mail/EMail.ts
+++ b/app/logic/Mail/EMail.ts
@@ -251,88 +251,92 @@ export class EMail extends Message {
     // indirectly calls @see `ICalEMailProcessor.process()`
   }
 
+  readonly parseMIMERunOnce = new RunOnce<void>();
   async parseMIME() {
     assert(this.mime?.length, "MIME source not yet downloaded");
     assert(this.mime instanceof Uint8Array, "MIME source should be a byte array");
     //console.log("MIME source", this.mime, new TextDecoder("utf-8").decode(this.mime));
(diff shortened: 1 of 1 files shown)


app/logic/Mail/EMail.ts
```diff
@@ -283,100 +251,92 @@ export class EMail extends Message {
...
-          inline: ContentDisposition.inline,
-        }, ContentDisposition.unknown);
-        attachment.related = sanitize.boolean(a.related, false);
-        attachment.content = new File([a.content], attachment.filename, { type: attachment.mimeType });
-        attachment.size = sanitize.integer(attachment.content.size, -1);
-        return attachment;
-      } catch (ex) {
-        this.folder.account.errorCallback(ex);
-        return null;
+    await this.parseMIMERunOnce.runOnce(async () => {
```
benbucksch, line 260: parseMIME() now has a return value. I've re-done the patch and will push it directly to master.

PR #1081, opened 2026-03-12
Date: Split getDateTimeFormat into 2 functions

jermy-c:
- We need to always have a language code for the `toLocaleString()` functions otherwise an error will be thrown
- But we also need to get the preference in local storage so the `LanguageDropDown` displays the correct selected language.
- `getDateTimeFormat()` is used for functions that must always have a valid language code
- `getDateTimeFormatPref()` is used for `LanguageDropDown` who needs an empty string if `System` is selected. Before it was returning `en-US` if `System` was selected which made selected blank.

### DIFF
--- a/app/l10n/l10n.ts
+++ b/app/l10n/l10n.ts
@@ -108,7 +108,7 @@ export function getDateTimeLocalePref(): string {
 }
 
 export function saveDateTimeLocale(language: string) {
-  cachedDateTimeLocale = language;
+  cachedDateTimeLocale = null;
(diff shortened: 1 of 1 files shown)


benbucksch (changes requested):
* For most callers across the app, please keep the function name and meaning the same. Just fix it to no longer break.
* Not sure why the implementation has to change so much. Just fix that one case that fails, please.
* The language dropdown is just a single caller, right? Just make a new function name for that, and document in the function signature that it's just for this.

Issue #1018, opened 2026-02-10
Mail: Rules: Move action etc. doesn't work


benbucksch:
Some mail filter rule actions don't work, esp. the ones that move messages into another folder, but a few others as well. Setting a message as read works, IIRC. The code is in `EMail.processMessage()`.

PR #1086, opened 2026-03-13
Lazily load move/copy to folder #1018

NeilRashbrook:
Filters get loaded very early in account creation via `MailAccount.fromConfigJSON`. Unfortunately this is before the account has loaded its folders.

I don't know whether it's worth caching the `toFolder` but I can implement that if you like.

### DIFF
--- a/app/logic/Mail/FilterRules/FilterRuleAction.ts
+++ b/app/logic/Mail/FilterRules/FilterRuleAction.ts
@@ -28,7 +28,7 @@ export class FilterRuleAction extends Observable {
   @notifyChangedProperty
   toFolderID: string | undefined;
   get toFolder(): Folder | null {
-    return this.toFolderID && (this._toFolder ??= this.account.findFolder(folder => folder.id == this.toFolderID)) || null;
+    return this.toFolderID && (this._toFolder ??= this.account.findFolder(folder => folder.id == this.toFolderID)) ?? null;
   }
   set toFolder(folder: Folder | null) {
     this._toFolder = folder;


benbucksch:
Yes, #1088 was better

NeilRashbrook:
#1088 was for a different variable... when I get back to work I'll rewrite this using the #1088 approach.

benbucksch:
Oh.

Thank you

app/logic/Mail/FilterRules/FilterRuleAction.ts
```diff
@@ -112,7 +120,8 @@ export class FilterRuleAction extends Observable {
     this.deleteImmediately = boolean(json.deleteImmediately);
     this.copy = boolean(json.copy);
     // let folderAccount = appGlobal.emailAccounts.find(acc => acc.id == json.toFolderAccountID);
-    this.toFolder = this.account.findFolder(folder => folder.id == json.toFolder) ?? undefined;
+    this.toFolderID = sanitize.nonemptystring(json.toFolder, null);
```
NeilRashbrook, line 123: (I know this value is "stored", but it's still only used for a comparison.)

benbucksch: Yes. The rule is: We validate *all* input from the network or from storage, immediately when we read it, in the same function where we read it. If the comparison is in the same function, then fine. If it's stored as variable, it has to be checked. It does not matter how the variable is used right now.

The scope of knowledge is the function. A function must never assume that another function or another class does something or doesn't do something, but we must assume that the other code is doing something else than what we want. Every function ensures state itself.

app/logic/Mail/FilterRules/FilterRuleAction.ts
```diff
@@ -24,8 +24,16 @@ export class FilterRuleAction extends Observable {
    * - Delete/spam actions with toFolder or markAsStarred
    */
   /** Move or copy (`copyToFolder`) the email to this folder. */
+  _toFolder: Folder | null = null;
   @notifyChangedProperty
-  toFolder: Folder | null = null;
+  toFolderID: string | undefined;
+  get toFolder(): Folder | null {
+    return this.toFolderID && (this._toFolder ??= this.account.findFolder(folder => folder.id == this.toFolderID)) || null;
```
benbucksch: Please use `??` unless you want to specifically handle `""` or `false` or `0`


```suggestion
    return this.toFolderID && (this._toFolder ??= this.account.findFolder(folder => folder.id == this.toFolderID)) ?? null;
```

app/logic/Mail/FilterRules/FilterRuleAction.ts
benbucksch: Given that you validate at least the string, that's fine for me.

app/logic/Mail/FilterRules/FilterRuleAction.ts
NeilRashbrook: Should have mentioned this back in #1088...

NeilRashbrook: And in fact this turns out to be a SyntaxError, because `??` can't be mixed with `&&`.

benbucksch: OK, then leave it here in this case.
(alternatively, just add `()` brackets)

benbucksch: Fixed in commit b51b50c1a2
Sorry for breaking the PR.

PR #1087, opened 2026-03-13
Hide account selector from filter criteria #1018

NeilRashbrook:
I forced the rule to specify the criteria account because the UI needs an account to be able to offer the folders; the alternative would be something like `<SearchCriteria search={rule.criteria} forceAccount={rule.account} />`.


NeilRashbrook:
Before the fix, you could specify that a filter on an account only ran on messages from another account, which would be useless; with the fix the UI only allows you to pick a folder from the filter's account.

PR #1088, opened 2026-03-13
Lazily load filter criteria folder #1018

NeilRashbrook:
Filters and their criteria get loaded very early in account creation via MailAccount.fromConfigJSON. Unfortunately this is before the account has loaded its folders.

I decided to cache the folder this time by comparison with #1086.

### DIFF
--- a/app/logic/Mail/Store/SearchEMail.ts
+++ b/app/logic/Mail/Store/SearchEMail.ts
@@ -90,7 +90,7 @@ export class SearchEMail extends Observable {
 
     this.includesPerson = findPerson(json.includesPersonEMail);
     this.account = appGlobal.emailAccounts.find(acc => acc.id == json.accountID);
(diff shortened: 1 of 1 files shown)


app/logic/Mail/Store/SearchEMail.ts
```diff
@@ -82,7 +90,8 @@ export class SearchEMail extends Observable {
 
     this.includesPerson = findPerson(json.includesPersonEMail);
     this.account = appGlobal.emailAccounts.find(acc => acc.id == json.accountID);
-    this.folder = this.account?.findFolder(folder => folder.id == json.folderID) ?? null;
+    this.folderID = json.folderID;
```
benbucksch:
```suggestion
    this.folderID = sanitize.nonemptystring(json.folderID, null);
```

PR #1098, opened 2026-03-16
EWS: Honour schema

NeilRashbrook:
I know Exchange 2010 (which we don't support) definitely requires us to order elements per the schema.

And I know that Exchange 2019 doesn't care about the schema order, since we've tested against it.

But there's no point relying on that when we can fix the schema on our end.

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -192,6 +192,7 @@ export class EWSAccount extends MailAccount {
         t$Value: value,
       })), null);
     }
+    // Older versions of Exchange require a specific order of parameters
     addRecipients(request, "ToRecipients", email.to.contents);
     addRecipients(request, "CcReipients", email.cc.contents);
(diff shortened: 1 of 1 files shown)


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -189,16 +192,13 @@ export class EWSAccount extends MailAccount {
...
       })), null);
     }
-    if (email.inReplyTo) {
-      request.addField("Message", "InReplyTo", email.inReplyTo, "item:InReplyTo");
-    }
-    if (email.replyTo) {
-      addRecipients(request, "ReplyTo", [email.replyTo]);
-    }
-    addRecipients(request, "From", [email.from]);
     addRecipients(request, "ToRecipients", email.to.contents);
```
benbucksch, line 196:
```suggestion
    // Older versions of Exchange require a specific order of parameters
    addRecipients(request, "ToRecipients", email.to.contents);
```

Issue #1084, opened 2026-03-13
Router sets browser location to URL that we cannot load


benbucksch:
_Originally posted by @jermy-c in [#1076](https://github.com/mustang-im/mustang/issues/1076#issuecomment-4057438840)_

Clicking Command-R for refresh renders the application all white.

### Reproduction

1. Remove lines

https://github.com/mustang-im/mustang/blob/b31e59237a20070766a01533051e7c258c9c49ea/desktop/src/main/index.ts#L79-L82

2. `cd desktop`
3. `yarn electron-vite preview`
4. Refresh the app after it is started

### Actual Result

- White screen

### Expected Result

- Goes back to the start screen

### Cause

- The initial URL is `../renderer/index.html`
- Then the URL is changed to `file://mail`
- When you refresh it tries to fetch `file://mail`
- It seems like it must be `../renderer/index.html`

jermy-c:
Fixed by #1102, #1176

PR #1102, opened 2026-03-17
Router: Hash history #1084

jermy-c:
- Use hash history so the base URL or `../renderer/index.html` is not changed
- We still need to have the URL changing because we need the URL history for the back button to work
- The created history needs to passed to all callers either as an argument or using the method of the history object
- Tested on Desktop and Android Emulator

### DIFF
--- a/app/frontend/Util/hashHistory.ts
+++ b/app/frontend/Util/hashHistory.ts
@@ -1,4 +1,7 @@
-// <copied from="https://github.com/mefechoel/svelte-navigator/blob/main/example/custom-hash-history/src/hashHistory.js">
+// From <https://github.com/mefechoel/svelte-navigator/blob/main/example/custom-hash-history/src/hashHistory.js>
+// Copyright (c) 2020 Michel Strelow
+// MIT License (MIT)
+
 import { createHashHistory as _createHashHistory } from "history";
 import { createHistory, type HistorySource } from "svelte-navigator";
 


app/frontend/Util/hashHistory.ts
```diff
@@ -0,0 +1,46 @@
+// <copied from="https://github.com/mefechoel/svelte-navigator/blob/main/example/custom-hash-history/src/hashHistory.js">
```
benbucksch: Is this a verbatim copy?
Why do we need to copy and cannot import?
If we copy, we need to put proper license and attribution

benbucksch (changes requested):
Code looks good to me, apart from the copy of the third party code file.

This needs very careful testing before merging.

app/frontend/Util/hashHistory.ts
jermy-c: > Is this a verbatim copy?

No, I did merge merge createHashSource and createHashHistory so it can easily be used.

> Why do we need to copy and cannot import?

Because there's no hash history feature in `svelte-navigator` only an example. The creator of `svelte-navigator` does have a [`svelte-navigator-history`](https://github.com/mefechoel/svelte-navigator-history) package which has the hash history feature for `svelte-navigator` but I couldn't get it to work even it was designed for `svelte-navigator`.

benbucksch: > The creator of svelte-navigator does have a [svelte-navigator-history](https://github.com/mefechoel/svelte-navigator-history) package which has the hash history feature for svelte-navigator but I couldn't get it to work even it was designed for svelte-navigator.


ditto. I tried the same and failed.

Thanks for getting this to work!

jermy-c:
I've added the license and attribution in https://github.com/mustang-im/mustang/pull/1102/commits/af862feb4c55ac32629e5a16f90274be4ead80d0.

PR #1104, opened 2026-03-17
EWS, OWA: Create only the main addressbook and calendar automatically

benbucksch:
Create only the main addressbook and calendar automatically.
We do *not* want "ExternalContacts" addressbook, "Birthdays" calendar, "Public holidays" calendar etc.pp. by default.
Other addressbooks and calendars can be added in settings on user request.

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -702,7 +702,7 @@ export class EWSAccount extends MailAccount {
     let haveAddressbook = appGlobal.addressbooks.some(addressbook => addressbook.mainAccount == this);
     if (!haveAddressbook) {
       let folder = ensureArray(result.RootFolder.Folders.ContactsFolder).find(folder =>
-        folder.DistinguishedFolderId == "contacts" && folder.ExtendedProperty.Value != "true");
+        folder.DistinguishedFolderId == "contacts" && folder.ExtendedProperty?.Value != "true");
       let addressbook = this.createAddressbookAccount(folder);
       await addressbook.save();
       appGlobal.addressbooks.add(addressbook);


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -698,25 +698,22 @@ export class EWSAccount extends MailAccount {
...
     if (!haveCalendar) {
-      let calendars = ensureArray(result.RootFolder.Folders.CalendarFolder)
-        .filter(folder => folder.FolderClass == "IPF.Appointment");
-      for (let folder of calendars) {
-        let calendar = this.createCalendarAccount(folder);
-        await calendar.save();
-        appGlobal.calendars.add(calendar);
-      }
+      let folder = ensureArray(result.RootFolder.Folders.CalendarFolder)
+        .find(folder => folder.DistinguishedFolderId == "calendar" && folder.FolderClass == "IPF.Appointment");
```
NeilRashbrook: I see you didn't check that the default address book was visible, yet you did leave in the check that the default calendar folder was a calendar...

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -178,15 +178,13 @@ export class OWAAccount extends MailAccount {
...
       let response = await this.callOWA(new OWAGetPeopleFiltersRequest());
-      let contacts = response.filter(ab => !ab.IsReadOnly && ab.FolderId?.Id);
-      let i = 0;
-      for (let folder of contacts) {
-        let addressbook = this.createAddressbookAccount(folder, i++ == 0);
-        await addressbook.save();
-      }
+      let folder = response.find(ab => !ab.IsReadOnly && ab.FolderId?.Id); // first one is main addressbook
+      let addressbook = this.createAddressbookAccount(folder, true);
+      await addressbook.save();
```
NeilRashbrook, line 188: This is my fault; in PR #1068 I accidentally removed the following line:
```
```

But I didn't notice because the addressbook was saved anyway so it automatically appeared on the next restart.

NeilRashbrook (changes requested):

NeilRashbrook:
> a) Was my change correct?

Yes; I should have tested on more servers.

> b) What is that `folder.ExtendedProperty.Value == true` anyways? What is this overly generic name? Why do we make it a condition for EWS addressbooks?

Office 365 sets that value to `true` for its internal addressbooks. It sets it to `false` for the main addressbook, but as that's the default value, older servers just leave it off and the whole property goes missing. The name is overly generic because Microsoft didn't implement a more specific name for it.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook:
```suggestion
      appGlobal.addressbooks.add(addressbook);
      await addressbook.save();
```

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: > you didn't check that the default address book was visible

Where did I remove that check for "visible"?

Do you mean the check for `folder.ExtendedProperty.Value != "true"` ? That's because I assumed it means that `ExtendedProperty.Value` means this is a "special-purpose" address book. Which should never be true for the main user addressbook.

That's why I specifically asked you whether I use `folder.ExtendedProperty.Value` correctly. I also referred to the address book place here.

I'm still confused. In your answer to me, you say that it means "internal". Which should always be false for the main address book, so no point in checking that for `folder.DistinguishedFolderId == "contacts"`. **But** here, and in the assert, you imply (but don't clearly state) that `folder.ExtendedProperty.Value != true` means "visible". These are contradictionaly definitions.

I really don't understand. Can you please clarify:
a) What does ExtendedProperty.Value mean? internal? not visible? what exactly?
b) can ExtendedProperty.Value be true for the default address book? if so, why?
c) is my code correctly? if not, please give me the correct code, and explain why it is correct. Please make the connection between the code, and its meaning explicit. E.g. in the way I do here in this comment.

benbucksch:
> I should have tested on more servers.

I tested with neil@be on Office365.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: You've misunderstood me. You are correct to assume that `ExtendedProperty.Value == "true"` never happens on the main user addressbook. For calendars, we again have a check to ensure that we don't try to add Birthdays as a calendar; this check obviously won't exclude the main user calendar, but you left it in anyway.

I test with `neil@b.onm` which always seemed to supply the `ExtendedProperty.Value` in my testing; to double-check that the `.?` was needed I had to try the `bexchange` server.

benbucksch: > You are correct to assume that ExtendedProperty.Value == "true" never happens on the main user addressbook.

OK. I just re-added it, because I wasn't sure anymore. If you want to make PR to straighten this out, whatever you say here, feel free. But please do it later on, we have important bugs right now.

benbucksch: FWIW, I'm still unclear about this:

> a) What does ExtendedProperty.Value mean? internal? not visible? what exactly?

NeilRashbrook: If ExtendedProperty.Value is true, this means that the addressbook is hidden, because it is internal.

Issue #1105, opened 2026-03-18
ActiveSync: Office365: ActiveSync version(s) null not supported


benbucksch:
### Reproduction
* Set up neil@be Office365 account with ActiveSync

### Actual result
* Error message "ActiveSync version(s) null not supported
*  `response.headers.get("MS-ASProtocolVersions")` is null
* see full headers below

### Expected result
* Setup works
* Get mail

### Debug info
Headers:
```
(2) ['content-length', '0']
(2) ['date', 'Wed, 18 Mar 2026 01:58:29 GMT']
(2) ['nel', '{"report_to":"NelOfficeUpload1","max_age":7200,"in…e,"failure_fraction":1.0,"success_fraction":0.01}']
(2) ['report-to', '{"group":"NelOfficeUpload1","max_age":7200,"endpoi…AS&MiraPartnerName="}],"include_subdomains":true}']
(2) ['request-id', '...']
(2) ['server', 'Microsoft-HTTPAPI/2.0']
(2) ['strict-transport-security', 'max-age=31536000; includeSubDomains']
(2) ['x-feserver', 'AM0P190CA0008']
(2) ['x-nanoproxy', '1']
(2) ['x-proxyerrorfromnanoproxy', '0']
(2) ['x-proxyerrorhresult', '0x80070036']
(2) ['x-proxyerrorlabel', 'Microsoft::M365::RoutingPlane::NanoProxy::HttpProxy::OnHttpRequest']
(2) ['x-proxyerrormessage', 'The network is busy.']
```

PR #1107, opened 2026-03-18
ActiveSync: Fix version is null #1105

benbucksch:
* HACK to pass setup
* Login still fails

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -176,17 +176,22 @@ export class ActiveSyncAccount extends MailAccount {
     }
     let response = await fetch(this.url, options);
     if (response.ok) {
-      let versions = (response.headers.get("MS-ASProtocolVersions") || "").split(",");
-      if (versions.includes("14.1")) {
-        this.protocolVersion = versions.includes("16.1") ? "16.1" : "14.1";
-        this.setStorageItem("protocolVersion", this.protocolVersion);
-        return;
+      let versionsHeader = response.headers.get("MS-ASProtocolVersions");
+      let versions = (versionsHeader || "").split(",");
+      if (!versionsHeader && response.headers.get("x-feserver")) {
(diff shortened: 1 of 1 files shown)


jermy-c:
> @jermy-c If you try to add your `@outlook.com` account without this change, does it just say that 16.1 isn't supported? If that's the case, we could drop support for 14.x which would massively simplify the code.

No, it says `null not supported`.

It responds with these headers:

```
x-ms-location: https://outlook.office365.com/Microsoft-Server-ActiveSync
x-nanoproxy: 1,1
x-proxy-backendserverstatus: 451
x-proxy-routingcorrectness: 1
x-rum-notupdatequerieddbcopy: 1
x-rum-notupdatequeriedpath: 1
x-rum-validated: 1
```

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -176,17 +176,22 @@ export class ActiveSyncAccount extends MailAccount {
     }
     let response = await fetch(this.url, options);
     if (response.ok) {
-      let versions = (response.headers.get("MS-ASProtocolVersions") || "").split(",");
-      if (versions.includes("14.1")) {
-        this.protocolVersion = versions.includes("16.1") ? "16.1" : "14.1";
```
benbucksch, line 179: @NeilRashbrook Here, in the old version, we were already setting `this.protocolVersion = "16.1"`. What does this PR change regarding 16.1 that you say it won't work? I merely tried to improve code readability, and fix the "null"

NeilRashbrook: In the current code base (i.e. without PR #1124), 16.1 is a fictitious version meaning "the server supports both 14.1 and 16.1 and we're going to use one 16.1 feature while using 14.1 everywhere else". (Whether ActiveSync technically allows us to do that is a different matter...)

Not that that's relevant because this is just wallpaper anyway (PR #1126 is the real fix).

benbucksch:
Fixed in PR #1124

Issue #1108, opened 2026-03-18
EWS: Send: permissions is null


benbucksch:
With 0.9.21 release:

```
TypeError: Cannot read properties of undefined (reading 'exchangePermissions')
  at EWSAccount.send (../../../../app/logic/Mail/EWS/EWSAccount.ts:163:24)
  at ComposeActions.send (../../../../app/logic/Mail/ComposeActions.ts:323:5)
  at onSend (../../../../app/frontend/Mail/Composer/MailComposer.svelte:344:11)
  at myOnClick (../../../../app/frontend/Shared/RoundButton.svelte:67:13)
```

benbucksch:
The user is still running into this bug, with 0.9.27.

<img width="1404" height="65" alt="Image" src="https://github.com/user-attachments/assets/3b3f3f33-14b8-491b-b721-edeae1abfacb" />

benbucksch:
Same reproduction steps as #1108

PR #1112, opened 2026-03-18
Exchange permissions: Check email addresses case-insensitively #1108

NeilRashbrook:
#1108

### DIFF
--- a/app/logic/Mail/EWS/EWSFolder.ts
+++ b/app/logic/Mail/EWS/EWSFolder.ts
@@ -639,12 +639,16 @@ type ExchangeUser = { UserId: { DisplayName: string; DistinguishedUser: string; 
 export class ExchangePermission {
   exchangePermissions: ExchangePermissions;
   distinguishedUser?: string;
-  emailAddress?: string;
+  protected emailAddress?: string;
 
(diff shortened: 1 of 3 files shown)


app/logic/Mail/EWS/EWSFolder.ts
```diff
@@ -639,14 +639,18 @@ type ExchangeUser = { UserId: { DisplayName: string; DistinguishedUser: string;
...
-  emailAddress?: string;
+  protected emailAddress?: string;
 
   constructor(permissions: Partial<ExchangePermissions> & ExchangeUser) {
     this.exchangePermissions = new ExchangePermissions(permissions);
     this.distinguishedUser = permissions.UserId.DistinguishedUser;
     this.emailAddress = permissions.UserId.PrimarySmtpAddress;
   }
 
+  matchesEmailAddress(emailAddress: string) {
```
benbucksch:
```suggestion
  matchesEMailAddress(emailAddress: string) {
```

E**M**ail

NeilRashbrook: Why is this word not quite camelCase? (account.emailAddress instead of account.eMailAddress)

Issue #1109, opened 2026-03-18
EWS: Too many concurrent connections opened., Cannot open mailbox


benbucksch:
```
Too many concurrent connections opened., Cannot open mailbox. Server = [GV0P278MB2089.CHEP278.PROD.OUTLOOK.COM](http://gv0p278mb2089.chep278.prod.outlook.com/), mdbGuid = ..., maiboxGuid = ...
```
Stack:
```
  at EWSAccount.checkResponse (../../../../app/logic/Mail/EWS/EWSAccount.ts:251:13)
  at EWSAccount.callEWS (../../../../app/logic/Mail/EWS/EWSAccount.ts:346:21)
  at EWSGAL.quickSearchAsync (../../../../app/logic/Contacts/EWS/EWSGAL.ts:32:22)
  at None (/opt/Parula/resources/app.asar/out/renderer/assets/index-C7dQE1uH.js:169435:9)
  at async Promise.all (index 3)
```

An important customer ran into this.

Error was gone after a restart of the app.

benbucksch:
Possible causes and ideas:
* Too many streaming connections. With 3 delegate accounts, we open 10 connections already. If there are any other connection, possibly from other clients, they might count into the limit? Or: What if I have 3 devices with Mustang?
* Connections not closed. What happens when Mustang crashes, freezes, or even if it's shut down cleanly? Do the connections stay open? Do they count for the server?
* Do normal EWS REST calls count?
* Are there places where we're not propertly throttling the EWS connections?
* Is out throttle too high?
* Do we recognize the error from the server here, and decrease the throttle?
* Do we increase the throttle again, if it got really low?

PR #1115, opened 2026-03-19
EWS: Only use one notification stream per mailbox #1109

NeilRashbrook:
- When two subaccounts (addressbook, calendar, email) are for the same mailbox, only one stream is used for the three subscriptions
- Unsubscribes notifications when a subaccount is removed
- Subscribes to notifications when a subaccount is added


app/logic/Calendar/EWS/EWSCalendar.ts
```diff
@@ -33,6 +33,24 @@ export class EWSCalendar extends Calendar {
...
+    await this.account.login(interactive);
+  }
+
+  async disconnect(): Promise<void> {
+    await this.account.unsubscribeNotifications(this);
+  }
+
+  async startup(): Promise<void> {
+    await super.startup();
+    if (this.username != this.account.emailAddress) {
```
benbucksch: ewhy comparing username with email address?

app/logic/Contacts/EWS/EWSAddressbook.ts
```diff
@@ -32,6 +32,24 @@ export class EWSAddressbook extends Addressbook {
...
+    await this.account.login(interactive);
+  }
+
+  async disconnect(): Promise<void> {
+    await this.account.unsubscribeNotifications(this);
+  }
+
+  async startup(): Promise<void> {
+    await super.startup();
+    if (this.username != this.account.emailAddress) {
```
benbucksch: ditto

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -30,10 +30,17 @@ import { assert, blobToBase64, ensureArray, NotReached, NotSupported, type Json
 import { gt } from "../../../l10n/l10n";
 import { ArrayColl } from "svelte-collections";
 
-// <https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxprops/01b52d3c-d194-4a8c-83ee-4ac7506339da>
+/** @see <https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxprops/01b52d3c-d194-4a8c-83ee-4ac7506339da> */
 const HiddenPidTag = "0x10F4";
 
-export class EWSAccount extends MailAccount {
+export interface EWSSubscribable extends Account {
```
benbucksch: move down

app/logic/Contacts/EWS/EWSAddressbook.ts
```diff
@@ -32,6 +32,24 @@ export class EWSAddressbook extends Addressbook {
...
 
+  async login(interactive: boolean) {
+    if (this.isLoggedIn) {
+      return;
+    }
+    await this.account.login(interactive);
+  }
+
+  async disconnect(): Promise<void> {
+    await this.account.unsubscribeNotifications(this);
```
benbucksch, line 36: This doesn't make sense to me.

Neither `EWSAddressbook` nor `EWSCalendar` have a `subscriptionID` property on the class. It's optional on the *interface* `EWSSubscribable`, but an interface is just a type, a contract, but doesn't create actual class variables, because it's a not a class.

```
EWSAccount.unsubscribeNotifications(account: EWSSubscribable) {
    if (!account.subscriptionID) {
      return;
```

So, this condition should always be true for `EWSAddressbook` and `EWSCalendar`, meaning this code line in `disconnect()` that you added here will never do anything.

So, that seems to not work as intended.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -33,24 +32,29 @@ import { ArrayColl } from "svelte-collections";
 // <https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxprops/01b52d3c-d194-4a8c-83ee-4ac7506339da>
 const HiddenPidTag = "0x10F4";
 
-export class EWSAccount extends MailAccount {
+export interface EWSSubscribable extends Account {
+  // FolderId to subscribe to
+  readonly folderID: string;
+  // SubscriptionId for unsubscribing on disconnect
+  subscriptionID?: string;
```
benbucksch: I am unclear whether `EWSAddressbook` and `EWSCalendar` should have a `subscriptionID` or not. They currently do *not* (see my other comment in EWSAddressbook).
If not, then you can remove it from `EWSSubscribable` here, and a lot of the other code that uses `subscriptionID` could cast to `EWSAccount` instead of `EWSSubscribable`.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -527,19 +538,34 @@ export class EWSAccount extends MailAccount {
...
     };
     let response = await this.callEWS(subscribe);
-    this.subscriptions.push(sanitize.nonemptystring(response.SubscriptionId));
+    assert(!(account ?? this).subscriptionID, "startup code is broken");
+    (account ?? this).subscriptionID = sanitize.nonemptystring(response.SubscriptionId);
+    await this.streamNotifications(account?.username);
+  }
+
+  async streamNotifications(username = "") {
+    this.notificationAbort[username]?.abort("Restarting stream due to changed subscription");
```
benbucksch, line 543: Why not put that on the subaccount, and make it a single value?

My understanding is that we have:
* EWSAccount (mail of main user) <-- has a stream
  * EWSAddressbook (primary of main user)
  * EWSAddressbook (other AB of main user)
  * EWSAccount (mail of delegated account) <-- has a stream
     * EWSAddressbook (primary of delegated account)
     * EWSAddressbook (other AB of delegated account)

benbucksch (commented):
@NeilRashbrook I'm not sure I did the merge with master correctly. I think there were are parts where I accidentally took the master version and lost your changes. Can you please check?

I made a few review changes. I could not test them. Please review them, and test after you fixed the other review comments.

Pariticularily, if you try to make it work both for dependent accounts and main accounts, it becomes hard to read. Esp. lots of conditionals (`if` or `?:`) make it hard to read. I've split out `subscribeToNotificationsForSubaccount()`, because the previously unified function was consisting mostly of a huge `if` block. But `EWSAccount.streamNotifications(username = "")` still suffers from this, and I made a review comment for one option that might help to clear this case up a little.

I found a few things during review that make no sense to me. Comments below. Could you please think about them and fix them?

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: Wow, you can implement an interface before declaring it? Whatever next?

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: It's an optional declaration, so TypeScript doesn't complain, so I forgot to add it manually. And the whole point is that no, you don't want to cast an `EWSCalendar` to `EWSAccount`. That would be bad.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: No, it's not hierarchical, all of the delegated accounts are subaccounts of the main mail account. (You can share just the calendar and not the inbox for example.)

app/logic/Calendar/EWS/EWSCalendar.ts
NeilRashbrook: Calendars (and addressbooks) don't have an email address, so I put the account's address in the username field instead.

app/logic/Contacts/EWS/EWSAddressbook.ts
NeilRashbrook: `subscriptionID` does of course get set (in `subscribeToNotifications`).

NeilRashbrook (commented):
(I haven't actually looked at the merge; I just wanted to reply to your comments.)

app/logic/Calendar/EWS/EWSCalendar.ts
```diff
@@ -33,6 +33,24 @@ export class EWSCalendar extends Calendar {
     return this.account.isLoggedIn;
   }
 
+  async login(interactive: boolean) {
+    if (this.isLoggedIn) {
+      return;
+    }
+    await this.account.login(interactive);
+  }
```
NeilRashbrook: This override is no longer needed.

app/logic/Contacts/EWS/EWSAddressbook.ts
```diff
@@ -32,6 +32,24 @@ export class EWSAddressbook extends Addressbook {
     return this.account.isLoggedIn;
   }
 
+  async login(interactive: boolean) {
+    if (this.isLoggedIn) {
+      return;
+    }
+    await this.account.login(interactive);
+  }
```
NeilRashbrook: Ditto.

NeilRashbrook (commented):
> @NeilRashbrook I'm not sure I did the merge with master correctly. I think there were are parts where I accidentally took the master version and lost your changes. Can you please check?

I checked the merge and what actually happened is that you failed to take the master version in a couple of places 😉 

> I made a few review changes. I could not test them. Please review them, and test after you fixed the other review comments.

Mustang won't currently start for me since #1146 😢 

> Particularly, if you try to make it work both for dependent accounts and main accounts, it becomes hard to read.

It used to only have one conditional, but I agree that I should have split the subscription part out into two functions instead of just one.

jermy-c:
> Mustang won't currently start for me since #1146 😢

Sorry about that, what errors are you seeing?

app/logic/Calendar/EWS/EWSCalendar.ts
benbucksch: can you at least compare username with username, please?

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: JavaScript parsers make 2 passes:
1. First pass only *declares*, i.e. makes the function signatures and classes known. Imagine it like a C++ header file.
2. Second pass *defines*, i.e. parses the function's code.

That's why you can use a function before it's declared. Just imagine that the JavaScript parser automatically creates a .h header file for you, and uses that, before really compiling the file.

app/logic/Calendar/EWS/EWSCalendar.ts
NeilRashbrook: I hadn't noticed that you changed the default addressbook and calendar to use the account's username way back in 4a5e74e, so I thought it was still using the email address...

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: Well, it's tricky for classes - they work like `let` and `const`, so JavaScript knows that you're *going* to declare the class, but disallows you from using it anyway. (On the other hand, `import` declarations are also hoisted, so you don't actually need to put them at the beginning...)

app/logic/Calendar/EWS/EWSCalendar.ts
NeilRashbrook: While fixing that I noticed that you hadn't changed `subscribeToNotifications` to `subscribeToNotificationsForSubaccount` here and in `EWSAddressbook.ts`.

NeilRashbrook (commented):
I've pushed a commit containing:
- Removed mismerged `login` methods
- Fixed use of `username` instead of `emailAddress` when comparing subaccounts
- Fixed use of `subscribeToNotificationsForSubaccount` for subaccounts

NeilRashbrook:
> Mustang won't currently start for me since #1146 😢

@jermy-c Turns out this is electron/electron#48859.

The workaround that works for me is to replace
```typescript
    mainWindow.on('ready-to-show', () => {
```
with
```typescript
    mainWindow.webContents.on('did-finish-load', () => {
```

benbucksch:
@NeilRashbrook thanks for finding the cause and workaround.
@jermy-c and Neil: Can you please test this very specific workaround and then make a PR?
https://github.com/FreeTubeApp/FreeTube/pull/8294/changes

jermy-c:
> @jermy-c Turns out this is [electron/electron#48859](https://github.com/electron/electron/issues/48859).

Thanks for finding this Neil!

> @jermy-c and Neil: Can you please test this very specific workaround and then make a PR?

Sorry, I cannot reproduce the problem so I will not know if the workaround works or not.

benbucksch:
@NeilRashbrook This bitrotted dramatically. Can you please merge with master and test it again?

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -529,19 +507,61 @@ export class EWSAccount extends MailAccount {
...
+                "CreatedEvent",
+                "DeletedEvent",
+                "ModifiedEvent",
+                "MovedEvent",
+              ],
+            },
+          },
+        },
+      };
+    let response = await this.callEWS(subscribe);
```
NeilRashbrook, line 536: (Indentation here is wrong.)

benbucksch:
Thanks for merging. Let me know when it's well-tested and working and ready to ship.

NeilRashbrook:
I did find one issue.

One of my profiles was broken because it had an addressbook from a deleted account.

Trying to delete the addressbook fails with this patch because it tries and fails to unsubscribe notifications.

(Presumably the code on master is more wrong because it fails to try to unsubscribe notifications?)

NeilRashbrook:
Also, in relation to https://github.com/mustang-im/mustang/issues/1036#issuecomment-3970963353 the current startup code tries to run `startup` after `login`, which I didn't notice at the time, but this code notices the problem. I'm not sure why you would want an account that's logged in but not started up (e.g. it wouldn't be polling/pinging for new messages) but the quick fix would be to add an `else` so that `startup` only runs if the account didn't need to login.

NeilRashbrook:
I did notice one related change; this code expects notification streams to be aborted and so doesn't report them to the error callback, while on master abort errors have a read-only `message` property, so when the background error callback tries to prepend the account name to the message it fails.

benbucksch:
### Environment
* I have set up ben@be as main account, with EWS
* neil@be as delegated account (also EWS obviously)

### Actions
* I clicked on neil@be
* -> it was not logged in
* I clicked [Login]
* I was asked to log in as ben@be. (Correct)
* I did that
* I clicked on account ben@be
* I clicked on account neil@be again (and maybe on Login?)

### Actual result
Error:

```
logHistory.ts:43 Error: stream notification started twice
    at assert (util.ts:4:11)
    at EWSAccount.subscribeToNotifications (EWSAccount.ts:515:5)
    at async EWSAccount.ts:124:7
    at async RunOnce.runOnce (RunOnce.ts:16:14)
    at async EWSAccount.startup (EWSAccount.ts:113:5)
    at async EWSAccount.login (EWSAccount.ts:109:5)
    at async EWSAccount.ts:101:9
    at async RunOnce.runOnce (RunOnce.ts:16:14)
    at async EWSAccount.login (EWSAccount.ts:99:5)
```

NeilRashbrook:
This could be a regression from commit d149eff. For dependent mail accounts, the old code just forwarded the call to the main account. However, commit d149eff allowed dependent mail accounts to separately run `startup` after the login completes, although the main account already runs startup for all dependent accounts (mail, contacts, calendar). If you're lucky, both calls try to execute at the same time and the `RunOnce` saves you, but if not, you get a double startup, triggering the assertion.

> (and maybe on Login?)

Logging in an account that's already started up will cause a double startup. If this is what you did, that would definitely trigger the assertion.

(While trying to reproduce this, I ran into a display issue where it looked as if the correct account's inbox was selected but in fact it was actually a different account's emails being listed, but I don't know how to reproduce it.)

(While trying to reproduce this, I ran into a startup issue where it looks as if the dependent account is logged in even though the main account wasn't. This happens because the account hasn't had its `mainAccount` set yet.)

Issue #1118, opened 2026-03-24
OWA: Contacts: Error when creating contact


jermy-c:
### Reproduction

1. Have an OWA account setup
2. Start the app
3. Click the login button
4. Go to contacts
5. Select the OWA Addressbook
6. Click [+] to create a new contact
7. Type a name
8. Click the check mark to save

### Actual Results

- Yellow banner with `Id is malformed.`
- Contact not created remotely
- Stack
```
POST https://outlook.office.com/owa/service.svc 400 (Bad Request) OWAAccount.ts:319 
 
OWAError: Id is malformed.
    at _OWAAccount.callOWA (OWAAccount.ts:350:13)
    at async OWAPerson.saveToServer (OWAPerson.ts:69:20)
    at async OWAPerson.save (Person.ts:58:5)
    at async EditableSimpleText.save (NameBox.svelte:92:18)
```
- Network payload
```json
{
    "request": {
        "__type": "UpdatePersonaJsonRequest:#Exchange",
        "Header": {
            "__type": "JsonRequestHeaders:#Exchange",
            "RequestServerVersion": "Exchange2013"
        },
        "Body": {
            "__type": "UpdatePersonaRequest:#Exchange",
            "PersonTypeString": "Person",
            "PropertyUpdates": [
                {
                    "__type": "PersonaPropertyUpdate:#Exchange",
                    "Path": {
                        "__type": "PropertyUri:#Exchange",
                        "FieldURI": "PersonaGivenNames"
                    },
                    "NewValue": "Steve"
                },
                {
                    "__type": "PersonaPropertyUpdate:#Exchange",
                    "Path": {
                        "__type": "PropertyUri:#Exchange",
                        "FieldURI": "PersonaSurnames"
                    },
                    "NewValue": "Jones"
                }
            ],
            "PersonaId": {
                "__type": "ItemId:#Exchange",
                "Id": "1774371462794-537950"
            }
        }
    }
}
```
- Network reponse
```json
{
    "Body": {
        "ErrorCode": 400,
        "ExceptionName": "ErrorInvalidIdMalformedException",
        "FaultMessage": "Id is malformed.",
        "IsTransient": false,
        "ResponseCode": "ErrorInvalidIdMalformedException"
    }
}
```

### Expected Results

- Contact created remotely
- No errors

### Cause

- Not a regression from 7164d14f006bc9eca022cde26b409867b3bd8a34

jermy-c:
[...]

Issue #1106, opened 2026-03-18
EWS, OWA, ActiveSync: Sanitize all network input


benbucksch:
Whenever we read values from a server response, we need to sanitize the values before storing them in variables. We do that at the moment when we parse and read the server response.

We use `sanitize` class for that. Use the most specific function that still passes all valid input values. If there is a function that expressed the meaning, e.g. `emailAddress()` or `label()` (for names and other strings displayed to the user), then use that, not the generic `string()`.

EWS, OWA and ActiveSync have not sanitized all input. For some parts, we have already added it, but there are other parts where it's missing, e.g. `ActiveSyncPerson` and likely others.

* Go through all EWS, OWA, ActiveSync classes, where they parse and read server-returned data
* Before the value is used, wrap the property in the appropriate `sanitize.foo()` function
* In most cases, we want a fallback value. It may be `null` or the current time, or whatever is appropriate for the data.
* However, if the data set makes no sense without this value, then do not provide a fallback, which will make the `sanitize` function throw. Check that the caller, at the point where it loops over the data, catches the exception and continues to parse the next data object.
* If we are merely compating the input value for equality with another value, and that happens within the same function, then we don't need to sanitize. If the comparison happens in another function, we need to sanitize.
* Test that you didn't break anything

Check `IMAPEMail.fromFlow()`, `JSContact.toPerson()` etc. for examples

benbucksch:
Ask @NeilRashbrook and me for review of the PR

benbucksch:
@jermy-c Is this done?

(@jermy-c @NeilRashbrook If an issue is assigned to you and you think that it was completely fixed/resolved and the fixes landed on master, please close it.)

jermy-c:
> [@jermy-c](https://github.com/jermy-c) Is this done?

No, there's 2 more items:

- EWS and OWA ids are base64-encoded binary values. ActiveSync ids are opaque strings. https://github.com/mustang-im/mustang/pull/1123#issuecomment-4183766675
- Fix: the UI does not prevent you from creating contacts with these values (values with errors). https://github.com/mustang-im/mustang/pull/1123#issuecomment-4183963376

PR #1119, opened 2026-03-24
OWA: Sanitize inputs #1106

jermy-c:
- Most of the inputs were sanitized already
- Tested but saving contacts doesn't work #1118

### DIFF
--- a/app/logic/Mail/OWA/OWAEMail.ts
+++ b/app/logic/Mail/OWA/OWAEMail.ts
@@ -66,7 +66,7 @@ export class OWAEMail extends EMail {
     setPersons(this.cc, json.CcRecipients);
     setPersons(this.bcc, json.BccRecipients);
     this.contact = this.outgoing ? this.to.first : this.from;
-    this.invitationMessage = sanitize.integer(ExchangeScheduling[json.ItemClass], InvitationMessage.None);
+    this.invitationMessage = ExchangeScheduling[sanitize.string(json.ItemClass)] || InvitationMessage.None;
   }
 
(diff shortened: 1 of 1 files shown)


app/logic/Mail/OWA/OWAEMail.ts
```diff
@@ -66,7 +66,7 @@ export class OWAEMail extends EMail {
     setPersons(this.cc, json.CcRecipients);
     setPersons(this.bcc, json.BccRecipients);
     this.contact = this.outgoing ? this.to.first : this.from;
-    this.invitationMessage = ExchangeScheduling[json.ItemClass] || InvitationMessage.None;
+    this.invitationMessage = sanitize.integer(ExchangeScheduling[json.ItemClass], InvitationMessage.None);
```
benbucksch: santize the json property, not the lookup result

app/logic/Mail/OWA/OWAEMail.ts
```diff
@@ -75,7 +75,7 @@ export class OWAEMail extends EMail {
     this.isStarred = json.Flag?.FlagStatus == "Flagged";
     // can't work out how to find junk status
     this.isDraft = sanitize.boolean(json.IsDraft);
-    this.tags.replaceAll((json.Categories || []).map(name => getTagByName(name)));
+    this.tags.replaceAll(ensureArray(json.Categories).map(name => getTagByName(sanitize.string(name))));
```
benbucksch, line 78: Could have used sanitize.array(), but alas....

PR #1121, opened 2026-03-25
EWS: Sanitize server inputs #1106

jermy-c:
- Sanitize all server inputs for EWS

### DIFF
--- a/app/logic/Mail/EWS/EWSEMail.ts
+++ b/app/logic/Mail/EWS/EWSEMail.ts
@@ -86,7 +86,7 @@ export class EWSEMail extends EMail {
     setPersons(this.cc, xmljs.CcRecipients?.Mailbox);
     setPersons(this.bcc, xmljs.BccRecipients?.Mailbox);
     this.contact = this.outgoing ? this.to.first : this.from;
-    this.invitationMessage = ExchangeScheduling[sanitize.string(xmljs.ItemClass, "")] || InvitationMessage.None;
+    this.invitationMessage = ExchangeScheduling[sanitize.string(xmljs.ItemClass, null)] || InvitationMessage.None;
   }
 
   /** Get body and attachments from Exchange.
@@ -112,7 +112,7 @@ export class EWSEMail extends EMail {
         if (a.ContentId) {
           attachment.contentID = sanitize.nonemptystring(a.ContentId);
         }
-        attachment.size = sanitize.integer(a.Size);
+        attachment.size = sanitize.integer(a.Size, null);
         return attachment;
       }));
     }


app/logic/Calendar/EWS/EWSCalendar.ts
```diff
@@ -68,8 +68,8 @@ export class EWSCalendar extends Calendar {
     return participants.map((participant, i) => ({
       participant,
       availability: ensureArray(results[i].FreeBusyView.CalendarEventArray?.CalendarEvent).map(event => ({
-        from: new Date(event.StartTime + "Z"),
-        to: new Date(event.EndTime + "Z"),
+        from: new Date(sanitize.date(event.StartTime, new Date()) + "Z"),
+        to: new Date(sanitize.date(event.EndTime, new Date()) + "Z"),
```
NeilRashbrook: This changes behaviour. Parsing the date without the ISO "Z" suffix is incorrect. (Why Exchange doesn't send the "Z" suffix for this one API call Gates only knows.)

benbucksch: Can you remove this change from the PR, please?

a) We don't want to mix this kind of behavior change with systematic changes, that makes it really hard to find later.

b) I am not sure that the change is correct.

(Reason a) alone would be sufficient reason to remove it.)

benbucksch: Thanks that you pointed this out in the PR. This is helpful.

jermy-c: I've removed the change in 26eaaf1ccf69ea32a23c7692f06b8c7cd1c501e2, thanks for pointing it out.

benbucksch: I think it should be:
```
from: sanitize.date(event.StartTime + "Z"),
to: sanitize.date(event.EndTime + "Z"),
```

i.e. the ` + "Z"` inside, not outside. And `sanitize.date()` already returns a `Date` object.

@NeilRashbrook the values must always be present, and it's an error when they are missing, right?

jermy-c: I've made the change, I think there will be an error anyways because passing `undefined + "Z"`, `null + "Z"`, `"" + Z` or any other case will give an invalid date.

benbucksch: You're right. How about:
`sanitize.date(sanitize.nonemptystring(event.StartTime) + "Z"),`

@NeilRashbrook The "Z" is needed, because the server returns UTC, and new Date() defaults to local time, is that correct?

[new Date() spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date): "When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time forms are interpreted as a local time. The interpretation as a UTC time is due to a historical spec error"

NeilRashbrook: A real Exchange server will always provide the values because the schema says that it will. The "Z" is needed to force UTC, yes.

jermy-c: > `sanitize.date(sanitize.nonemptystring(event.StartTime) + "Z"),`

Change made in 2a3f37ba364a0b115af8c630212d2eb702452a3a. I thought we wouldn't need to sanitize the `event.StartTime` string since we are already sanitizing the combined. After thinking about it'll be helpful if we ever get an empty string and `"" + "Z"` is passed and the sanitize function just throws `Invalid date` instead of `String empty`.

app/logic/Contacts/EWS/EWSAddressbook.ts
```diff
@@ -178,7 +178,7 @@ export class EWSAddressbook extends Addressbook {
       let results = ensureArray(await this.account.callEWS(request));
       for (let result of results) {
         try {
-          let person = this.getPersonByItemID(result.Items.Contact.ItemId.Id);
+          let person = this.getPersonByItemID(sanitize.nonemptystring(result.Items.Contact.ItemId.Id));
```
benbucksch, line 181: @NeilRashbrook Are EWS IDs alphanumdash() ?

NeilRashbrook: No; they're a base64 format, so they contain `0-9`, `A-Z`, `a-z` and `+/=`.

benbucksch: Might be worth making a dedicated sanitize function for them.

app/logic/Mail/EWS/EWSEMail.ts
```diff
@@ -86,7 +86,7 @@ export class EWSEMail extends EMail {
     setPersons(this.cc, xmljs.CcRecipients?.Mailbox);
     setPersons(this.bcc, xmljs.BccRecipients?.Mailbox);
     this.contact = this.outgoing ? this.to.first : this.from;
-    this.invitationMessage = ExchangeScheduling[xmljs.ItemClass] || InvitationMessage.None;
+    this.invitationMessage = ExchangeScheduling[sanitize.string(xmljs.ItemClass, "")] || InvitationMessage.None;
```
benbucksch: same effect, but clearer: nonemptystring() with null fallback, then ??

app/logic/Mail/EWS/EWSEMail.ts
```diff
@@ -106,13 +106,13 @@ export class EWSEMail extends EMail {
...
       this.attachments.replaceAll(attachments.map(a => {
         let attachment = new Attachment();
-        attachment.filename = a.Name;
-        attachment.mimeType = a.ContentType;
+        attachment.filename = sanitize.filename(a.Name);
+        attachment.mimeType = sanitize.nonemptystring(a.ContentType);
         attachment.disposition = a.IsInline == "true" ? ContentDisposition.inline : ContentDisposition.attachment;
         if (a.ContentId) {
-          attachment.contentID = a.ContentId;
+          attachment.contentID = sanitize.nonemptystring(a.ContentId);
```
benbucksch, line 113: content ID can be null, I think,
so nonemptystring with null fallback

app/logic/Mail/EWS/EWSEMail.ts
```diff
@@ -106,13 +106,13 @@ export class EWSEMail extends EMail {
...
         }
-        attachment.size = Number(a.Size);
+        attachment.size = sanitize.integer(a.Size);
```
benbucksch: Is this guaranteed to be there? Do we handle size == null? if so, we should default to null

app/logic/Mail/EWS/EWSEMail.ts
jermy-c: I had set it without the null fallback because there's an `if (a.ContentId)` condition before it. Unless there's possibility that a.ContentId is not `null`, `undefined`, `""` or a valid string.

app/logic/Mail/EWS/EWSEMail.ts
jermy-c: This is an alternative to `parseMIME()` so if the attachments aren't  downloaded until it should be null.

benbucksch: What I mean is: Should there be a fallback to null, or is `a.Size` always populated here at this code line? If it's always there, the code should stay as-is. If not, it must have a fallback.

app/logic/Mail/EWS/EWSEMail.ts
benbucksch: Oh, I didn't see that, you're right, sorry.

PR #1123, opened 2026-03-26
ActiveSync: Sanitize server inputs #1106

jermy-c:
- Sanitize ActiveSync server inputs

### DIFF
--- a/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
@@ -267,8 +267,8 @@ export class ActiveSyncEvent extends Event {
   }
 }
 
-function extractWeekdays(dayOfWeek: number): Weekday[] | null {
-  return dayOfWeek ? [0, 1, 2, 3, 4, 5, 6].filter(day => dayOfWeek & 1 << day) : null;
+function extractWeekdays(daysOfWeek: number): Weekday[] | null {
+  return daysOfWeek ? [0, 1, 2, 3, 4, 5, 6].filter(day => daysOfWeek & 1 << day) : null;
 }
 
 /// Returns the compact date time of a date


app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -95,11 +95,11 @@ export class ActiveSyncEvent extends Event {
...
     let masterDuration = this.duration;
     let seriesStartTime = this.startTime;
-    let seriesEndTime = wbxmljs.Until ? fromCompact(wbxmljs.Until) : null;
+    let seriesEndTime = wbxmljs.Until ? fromCompact(sanitize.nonemptystring(wbxmljs.Until)) : null;
     let count = sanitize.integer(wbxmljs.Occurrences, Infinity);
-    let frequency = kRecurrenceTypes[wbxmljs.Type];
+    let frequency = kRecurrenceTypes[sanitize.integer(wbxmljs.Type, 0)];
     let interval = sanitize.integer(wbxmljs.Interval, 1);
-    let weekdays = extractWeekdays(wbxmljs.DayOfWeek);
+    let weekdays = extractWeekdays(sanitize.nonemptystring(wbxmljs.DayOfWeek, ""));
```
NeilRashbrook: `extractWeekdays` already calls `sanitize.integer`. Or alternatively, if you think it's more readable, move that call here instead.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -43,26 +43,26 @@ export class ActiveSyncEvent extends Event {
       this.rawHTMLDangerous = null;
     }
     if (wbxmljs.DtStamp) {
-      this.lastUpdateTime = fromCompact(wbxmljs.DtStamp);
+      this.lastUpdateTime = fromCompact(sanitize.nonemptystring(wbxmljs.DtStamp));
```
NeilRashbrook, line 46: Would it be worth moving the check into the `fromCompact` function? Depends on which is more readable I guess; maybe leave this like it is.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
jermy-c: I moved the call here. Thanks for pointing that out.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
jermy-c: I'm going leave it as-is, I think it'll be easier to find all sanitization in the `fromWBXML` functions.

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -267,9 +267,8 @@ export class ActiveSyncEvent extends Event {
   }
 }
 
-function extractWeekdays(dayOfWeek: string): Weekday[] | null {
-  let daysOfWeek = sanitize.integer(dayOfWeek, 0);
-  return daysOfWeek ? [0, 1, 2, 3, 4, 5, 6].filter(day => daysOfWeek & 1 << day) : null;
+function extractWeekdays(dayOfWeek: number): Weekday[] | null {
+  return dayOfWeek ? [0, 1, 2, 3, 4, 5, 6].filter(day => dayOfWeek & 1 << day) : null;
```
NeilRashbrook:
```suggestion
function extractWeekdays(daysOfWeek: number): Weekday[] | null {
  return daysOfWeek ? [0, 1, 2, 3, 4, 5, 6].filter(day => daysOfWeek & 1 << day) : null;
```
(The field is named `DayOfWeek` because it's usually only one day, but for weekly recurring events it can be more than one day of the week.)

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
benbucksch: Agreed! For consistency, all input sanitization should happen in the same function where we read the value from the server or file.

Issue #1120, opened 2026-03-24
Calendar: Cannot invite again when organizer is removed


jermy-c:
### Reproduction

1. Go to calendar
2. Create an event
3. Click on [Invite]
4. Click on the organizer and remove
5. Click on [Invite] again

### Actual Results

- Clicking [Invite] again doesn't open the Invite UI
- UI does not freeze

### Expected Result

- Clicking [Invite] again opens the Invite UI again
- When the Invite UI is opened you are the organizer again

benbucksch:
The organizer should not appear at all in the list. It should always be there implicitly, but not in the list.

@NeilRashbrook This should ideally be fixed in the backend, so that the backend doesn't even list the organizer.

@jermy-c  Short-term we could fix it in the frontend, that we don't know the organizer in the UI list. The organizer should have a specific role.

benbucksch:
(BTW: @NeilRashbrook Same is true for folder permissions: Our user should always have access, and be there implicitly, but not appear in the list of users with access.)

PR #1125, opened 2026-03-26
Calendar: ParticipantBox: Don't show organizer in UI list #1120

jermy-c:
- Don't show self in self organized events

### DIFF
--- a/app/frontend/Calendar/EditEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/EditEvent/DialogHeader.svelte
@@ -176,7 +176,7 @@
   $: canSaveSingle = hasMinimalPropsChanged || // Single changed
     hasMinimalProps && event.recurrenceRule && !event.parentEvent || // Change single event into series
     newCalendar != event.calendar;
-  $: participants = event.participants;
+  $: participants = $event.displayParticipants;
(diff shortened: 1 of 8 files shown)


benbucksch (approved):
Nice

benbucksch (changes requested):
wait a second... that won't work for editing  (adding/removing) participants, right?

Please make sure to test *all* cases, not only the ones in your focus.

jermy-c:
> Or, actually, add might just work, thanks to `onAddPerson`, but remove might fail.

Yeah, remove only removed from the UI but not from the actual collection. I fixed that and other parts of the code that was missing the change.

app/frontend/Calendar/EditEvent/DialogHeader.svelte
```diff
@@ -227,6 +227,9 @@
       event = await event.moveToCalendar(newCalendar);
     } else {
       if (!event.calendar.events.contains(event)) {
+        if (event.isOutgoingMeeting && participants.isEmpty) {
+          event.participants.clear();
```
jermy-c: This is for when a new event is created and you click invite but don't invite anyone and it saves as an event with an invite.

benbucksch:
Back to #1120

PR #1126, opened 2026-03-27
Hotmail: Send DefaultAnchorMailbox in the backend #1105

NeilRashbrook:
`Cookie` is a banned CORS header, so we can't `fetch` it without some major shenanigans.

@jermy-c If you try to add your `@outlook.com` account with this change, does it now say that 16.1 isn't supported? If that's the case, we could drop support for 14.x which would massively simplify PR #1124.

### DIFF
--- a/desktop/src/main/index.ts
+++ b/desktop/src/main/index.ts
@@ -217,9 +217,10 @@ function allowCrossDomainRequestsFromFrontend() {
         case "referer":
           delete requestHeaders[name];
           break;
-        case "defaultanchormailbox":
+        case "cookie-bypass":
           // Fake out the Cookie on all ActiveSync requests, because Hotmail.
-          requestHeaders.Cookie = name + "=" + requestHeaders[name];
+          requestHeaders.Cookie = requestHeaders[name];
+          delete requestHeaders[name];
           break;
         case "user-agent":
           // Fake out the User-Agent on all ActiveSync requests, because Office.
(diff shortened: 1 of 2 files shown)


jermy-c (approved):
It is working after this change.

Response headers:

```
ms-asprotocolversions: 2.5,14.0,14.1,16.0,16.1
ms-server-activesync: 15.20
x-proxy-backendserverstatus: 200
```

desktop/src/main/index.ts
```diff
@@ -217,6 +217,10 @@ function allowCrossDomainRequestsFromFrontend() {
         case "referer":
           delete requestHeaders[name];
           break;
+        case "defaultanchormailbox":
```
benbucksch: That's not obvious in the app code.

Can you please forward `Cookie-Bypass` (or better name) as `Cookie` header *verbatim*?

benbucksch (changes requested):

desktop/src/main/index.ts
NeilRashbrook: You want **me** to think of a better name? :wink: how about `Cookie-DefaultAnchorMailbox`? Then I could write something like this:
```typescript
if (/^Cookie-/i.test(name)) {
  requestHeaders.Cookie = RegExp.rightContext + "=" + requestHeaders[name];
} else switch (name.toLowerCase()) {
```

benbucksch: > You want me to think of a better name? 😉 

😜 

It should not be specific to this case,  but simply the generic Cookie header, only with the name changed.

benbucksch (approved):

Issue #1127, opened 2026-03-27
Calendar: `Event.saveLocally` on an instance causes the master to start editing


NeilRashbrook:
If you call `Event.saveLocally` on an instance, it moves the event from its parent's `instances` array to its `exceptions` array, and also changes its `recurrenceCase` from `Instance` to `Exception`.

Unfortunately if the event is open in the UI this then triggers an update of the UI. If the `UI` called `Event.save`, this update happens between the `await this.saveLocally();` and the `await this.saveToServer();`.

One of the things that this update does is to call `startEditing` on the event's parent to give it a new recurrence rule.

This means that the event can no longer find itself in its parent's recurrences, and it's unable to save itself to the server.

PR #1128, opened 2026-03-27
Make it possible to edit and delete unedited instances again

NeilRashbrook:
While trying to test PR #1124 I found that editing and deleting unedited instances of recurring events was broken, because we weren't cancelling the master event's editing, so the instances were getting detached from their master.

Additionally, the code to edit and cancel editing for a master event wasn't working properly anyway, so instances were still getting detached from their master.

I also ran across issue #1127 but I don't know how to fix that properly so I just added a workaround for now so that editing actually works.

### DIFF
--- a/app/frontend/Calendar/DisplayEvent/DialogHeader.svelte
+++ b/app/frontend/Calendar/DisplayEvent/DialogHeader.svelte
@@ -113,6 +113,7 @@
     if (event.seriesStatus == "only") {
       await event.parentEvent.deleteIt();
     } else {
+      event.parentEvent?.cancelEditing();
       await event.deleteIt();
     }
     $selectedEvent = null;
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -364,9 +364,12 @@ export class Event extends Observable {
    * copyEditableFieldsFrom and recurring events need extra care. */
   copyFrom(original: Event) {
     this.copyEditableFieldsFrom(original);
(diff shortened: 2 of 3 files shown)


app/frontend/Calendar/DisplayEvent/DialogHeader.svelte
```diff
@@ -113,6 +113,7 @@
     if (event.seriesStatus == "only") {
       await event.parentEvent.deleteIt();
     } else {
+      event.parentEvent?.cancelEditing();
       await event.deleteIt();
```
benbucksch, line 117: Should this be inside deleteIt()?

NeilRashbrook: I think of `cancelEditing` as a front-end helper function, not something that `deleteIt` should need to be aware of, so that's why I put it here. Otherwise you then have to think about whether `save` should call `finishEditing` etc.

app/logic/Calendar/Event.ts
```diff
@@ -580,8 +583,10 @@ export class Event extends Observable {
    * Saves the event to the server and to the database.
    */
   async save() {
-    await this.saveLocally();
+    // Temporary to work around #1127 - save to server first,
```
benbucksch, line 586: What would be the proper fix? (not "temporary")

NeilRashbrook: Sadly I think #1138 already fixed this... I'll verify that and open a new PR.

PR #1137, opened 2026-04-02
Support recurring events in other time zones #1135

NeilRashbrook:
The major changes are:
- `timezone` field added to `RecurrenceRule` and `RecurrenceInit`
- `RecurrenceRule.seriesStartTime` is no longer a true date

Ideally we would use a `Temporal.ZonedDateTime` to store the series start time, but they're not available yet. Instead, I'm using a regular `Date`, but using the `UTC` fields to store the series start time for the given time zone.

Let's take an example. A recurring event's start time is 2026-03-31T22:00:00 in zone America/New_York. This is 2026-04-01T02:00:00Z or 2026-04-01T04:00:00 in your time zone Europe/Berlin. When you're viewing that event, you will see it listed in Wednesday April 1 (or should I say Mittwoch?) in your calendar (because for you that's where it is), but for the creator of the event, it's in Tuesday March 31. In particular, if it's a weekly recurring event, their recurrence pattern will say "Every Tuesday", but in your calendar you want it to appear on Wednesdays.

To handle this, the `RecurrenceRule` will store the start date as 2026-03-31T22:00:00Z. As you can see, if I format that date in zone UTC, I will get the same result as if I format the original date in zone America/New_York: "10PM on March 31st 2026". Additionally, I can use functions such as `getUTCDay()` which will tell me that the day of the week is Tuesday.

The `RecurrenceRule` object will then use this to generate occurrences every week on Tuesday, still in the notional America/New_York zone but using the UTC fields, before converting to true time to save the actual occurrence dates, which due to the time zone difference will end up on Wednesdays as required, despite the rule being an "Every Tuesday" rule.

I've had to tweak the UI to show the recurrence as given by the event's time zone. In the case of calls to `toLocaleDateString`, this is done by using the `seriesStartTime` and passing a `timeZone` of "UTC"; I could have used `event.startTime` and `event.timezone` but I felt this was clearer since `seriesStartTime` had to be used for the other calls (e.g. to find the day of the week or week of the month) anyway.

For most of the rest of the callers it's just necessary to a) pass in the time zone when creating the `RecurrenceRule` b) using the `UTC` series of functions on the `recurrenceRule.seriesStartDate` c) where the UTC date is needed, switching back to `event.startDate`. To ensure `recurrenceRule.seriesStartDate` is used correctly I've given it its own interface, so that you get a TypeScript error if you try to use regular `Date` functions on it.

The one tricky caller was `newRecurrenceRule` where it's tricky to validate the `init` object before we know the day of week and/or month of the date in the event's time zone.

### DIFF
--- a/app/frontend/Calendar/DisplayEvent/DetailsGrid.svelte
+++ b/app/frontend/Calendar/DisplayEvent/DetailsGrid.svelte
@@ -6,7 +6,7 @@
 
   {#if $event.recurrenceCase == RecurrenceCase.Master}
     <hbox />
-    <hbox class="recurrence">{ruleAsLabel($event.recurrenceRule, $event.startTime)}</hbox>
+    <hbox class="recurrence">{ruleAsLabel($event.recurrenceRule}</hbox>
   {/if}
 
   {#if $participants?.hasItems}
--- a/app/frontend/Calendar/EditEvent/RepeatBox.svelte
+++ b/app/frontend/Calendar/EditEvent/RepeatBox.svelte
@@ -108,21 +108,24 @@
   let yearWeekOptions: RadioOption[];
 
   $: startTime = $event.startTime.getTime();
+  $: timezone = $event.timezone;
   $: duration = $event.duration;
-  $: startTime, duration, updateDateUI();
+  $: startTime, timezone, duration, updateDateUI();
   function updateDateUI() {
     master.startEditing();
     master.startTime = event.startTime;
+    master.timezone = event.timezone;
     master.duration = event.duration;
     master.newRecurrenceRule(frequency, interval, week, weekdays);
+    let localStartTime = master.recurrenceRule.seriesStartTime;
     week = master.recurrenceRule.week;
-    weekdays = master.recurrenceRule.weekdays?.slice() ?? [$event.startTime.getDay()];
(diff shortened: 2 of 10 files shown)


benbucksch (approved):

app/frontend/Calendar/DisplayEvent/DetailsGrid.svelte
```diff
@@ -6,7 +6,7 @@
 
   {#if $event.recurrenceCase == RecurrenceCase.Master}
     <hbox />
-    <hbox class="recurrence">{ruleAsLabel($event.recurrenceRule, $event.startTime)}</hbox>
+    <hbox class="recurrence">{ruleAsLabel($event.recurrenceRule}</hbox>
```
benbucksch, line 9: The missing `)` is a parse error and breaks the app at startup.

@NeilRashbrook Can we make it so that there's a PR up, that it's tested and working, please?

I do **not** have the capacity to carefully test all PRs, in all scenarios. This is the responsibility of the developer creating the PR. The more often this happens, the more I'm going to be scared to merge PRs, and the more effort I have to put in reviews and tests, and the more I will lag behind with merging PRs, because it's so much effort to test them. If you want the PRs to merge quickly, please build a reputation that they always work, without exception. Please do not rely on me to catch bugs.

benbucksch:
I found #1178 while testing this.

Issue #1154, opened 2026-04-09
Keyboard shortcuts


benbucksch:

benbucksch:
I've added many more keyboard shortcuts for messages, both in message list and in message viewer.
| Key | Function | Alternate key|
|----------|----------|----------|
| S | Toggle starred | Insert |
| M | Toggle read/unread |
| Ctrl-U | Mark as unread |
| Ctrl-Q | Mark as read -- Linux: Quit the app |
| J | Mark as Junk |
| Shift+J | Mark as Not Junk |
| Delete | Delete (Move to trash) | Backspace |
| Shift-Delete | Delete immediately (Don't even move to trash) | Shift-Backspace |
| Arrow down | Go to next message | F |
| Arrow up | Go to previous message | B |
| F5 | Get new messages in folder | F9 or Ctrl-M |
| Ctrl+R | Reply to author |
| Ctrl+Shift+R | Reply to all |
| Ctrl+L | Forward inline |
| Ctrl+Shift+L | Forward as attachment |
| Ctrl-E | Edit as new - Create a new email with the same content, attachments, and recipients as the selected one |
| Ctrl+Shift+N | New email, with the same recipients, but not a reply |
| Ctrl+N | New email (independent of selected email) |

benbucksch:
@jermy-c Ctrl-R works for me in Mustang in my dev build (Linux), but not in the release build (Parula Linux).
(Mac: We should react to Ctrl and Cmd the same.)
And Ctrl-Shift-I for Dev Tools doesn't work anymore, either.
Can you please take a look what's going on?

jermy-c:
I made a fix which removes the function that was disabling `Ctrl-R` and `Ctrl-Shift-I` in #1156.

But there's another possible issue but from testing I don't see the issue. The Electron InputEvent and Browser Keyboard map shift, control, ctrl, alt, meta different.

Electron: `shift, control, ctrl, alt, meta`
Browser: `shiftKey / ctrlKey / altKey / metaKey`

benbucksch:
What I see from WebView is actually much worse: I first get an event with key = "ctrl", then another event with key = "b", but the latter doesn't have any hint that ctrl is still pressed.

jermy-c:
### Note

Docs:

Electron:
https://www.electronjs.org/docs/latest/api/structures/input-event
https://www.electronjs.org/docs/latest/api/structures/keyboard-event
[...]

PR #1156, opened 2026-04-09
Electron: Allow Ctrl-R and Ctrl-Shift-I in production #1154

jermy-c:
- `Ctrl-Shift-I` and `Ctrl-R` disabled by `optimizer.watchWindowShortcuts(window)`
- But it also enables opening DevTools by `F12`
- I removed `optimizer.watchWindowShortcuts(window)` and wrote our own `watchWindowShortcuts(window)` for DevTool by `F12`

### DIFF
--- a/desktop/src/main/index.ts
+++ b/desktop/src/main/index.ts
@@ -3,7 +3,7 @@ import { app, shell, BrowserWindow, session, Menu, MenuItemConstructorOptions, t
 import { ipcMain } from 'electron/main';
 import { join } from 'path'
 import electronUpdater from 'electron-updater';
-import { electronApp, optimizer, is } from '@electron-toolkit/utils'
+import { electronApp, is } from '@electron-toolkit/utils'
 import icon from '../../build/icon.png?asset'
 const { autoUpdater } = electronUpdater;
 
@@ -134,11 +134,8 @@ async function whenReady() {
 
   allowCrossDomainRequestsFromFrontend();
 
-  // Default open or close DevTools by F12 in development
-  // and ignore CommandOrControl + R in production.
-  // see https://github.com/alex8088/electron-toolkit/tree/master/packages/utils
   app.on('browser-window-created', (_, window) => {
-    optimizer.watchWindowShortcuts(window)
+    watchWindowShortcuts(window);
   })
 
   createMenu();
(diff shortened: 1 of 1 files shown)


desktop/src/main/index.ts
```diff
@@ -252,5 +249,23 @@ function allowCrossDomainRequestsFromFrontend() {
   );
 }
 
+/**
+ * Enable DevTools by F12 shortcut in development only.
+ */
+function watchWindowShortcuts(window: BrowserWindow) {
+  if (is.dev) {
```
benbucksch, line 264: in production?

jermy-c: It doesn't work in production, or should it be enabled in production also?

benbucksch: I need ctrl-shift-i in production

benbucksch:
expected:
ctrl-shift-i opens the dev tools, in dev and production builds
ctrl-r does not reload the app, but is handled in app code. we need it for "reply"

actual in 0.9.23:
neither do anything in prod, but they work in dev.

desktop/src/main/index.ts
jermy-c: Yes, that works in production. It is only the `F12` shortcut that's disabled in production.

jermy-c:
> ctrl-r does not reload the app, but is handled in app code. we need it for "reply"

 ctrl-r for reload was disabled in #1089

desktop/src/main/index.ts
benbucksch: ctrl-shift-i doesn't work for me in prod

benbucksch:
> > ctrl-r does not reload the app, but is handled in app code. we need it for "reply"
> 
> ctrl-r for reload was disabled in #1089

right. but it also doesn't work in app code for "reply", in production. it works in dev.

benbucksch:
In the patch, I see only handling for F12. How does this fix
* Ctrl-Shift-I -> devtools and
* Ctrl-R -> handled by app code -> Reply to message?

jermy-c:
> * Ctrl-Shift-I -> devtools and

The problem is because watchWindowShortcuts is applied:

https://github.com/mustang-im/mustang/blob/f082ac601e45d6f312b33270ee73c7925587695c/desktop/src/main/index.ts#L141

So removing it would stop the key press from being ignored in Production.

https://github.com/alex8088/electron-toolkit/blob/13874c6e6ed70248846763ec89dabe97dcaa98a9/packages/utils/src/optimizer.ts#L55



> * Ctrl-R -> handled by app code -> Reply to message?

It has the same problem as `Ctrl-Shift-I` but there might another problem where the Event.alt and Event.altKey mapping is different when it goes from Electron to the Browser.

https://github.com/alex8088/electron-toolkit/blob/13874c6e6ed70248846763ec89dabe97dcaa98a9/packages/utils/src/optimizer.ts#L54


The patch for F12 is not necessary if we don't use F12 to open devTools, I just added it there to keep it same as before. But I don't use that shortcut.

jermy-c:
Fixed by: 8598f9653d993944c1a918911aafaec305cb0a42

PR #1157, opened 2026-04-10
Properly handle participants with unknown availability #1150

NeilRashbrook:
`arePersonsFree` now returns:
- An empty array if the function is not supported
- An array of objects, one for each participant
- If the participant availability is unknown, then no availability array is returned
- If the participant availability is known, then an array with at least the time ranges that the participant is busy (an empty array means that the participant has no appointments at all in that time range)

I've also updated `AvailabilityGrid.svelte` to properly query the availability array.


benbucksch (approved):
Now it works for me. It didn't before. Thank you.

PR #1159, opened 2026-04-13
WebView: Use correct MouseInputEvent conditions

jermy-c:
- Use the correct MouseInputEvent conditions
- https://www.electronjs.org/docs/latest/api/structures/mouse-input-event
- Fixes URL not opening in browser when clicked

### DIFF
--- a/app/frontend/Shared/WebView.svelte
+++ b/app/frontend/Shared/WebView.svelte
@@ -170,9 +170,7 @@
       if (!url) {
         return;
       }
-      let modifiers = event.modifiers.map(m => m.toLowerCase());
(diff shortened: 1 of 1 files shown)


benbucksch (approved):

benbucksch:
Did the API change?

jermy-c:
> Did the API change?

From reading the git history for the docs, it seems it was like that since v17 and we were on v32 before. So the bug might have have been there before or electron just allowed it.
https://github.com/electron/electron/blob/663773a677b85dccf02b481752319d857cb136c1/docs/api/structures/mouse-input-event.md

Issue #794, opened 2025-07-19
Mail: Show mail in separate standalone window


benbucksch:
### Reproduction
1. Double-click on message

### Actual result
* Same as single click: Message opens on the right pane

### Expected result
* New desktop window opens, which contains only this message. It's smaller than the main window.

jermy-c:
### Investigations

- We can share the same instance between 2 windows if you set `globalThis.appGlobal` in the parent and then access it with `window.opener.appGlobal`, I see the same instance but there are some errors while using it.
- We might need to use #1084 to let the new Window know to open the app in a specific location or hash-based route
- We need to modify `setWindowOpenHandler` to allow itself to create a new window

PR #1160, opened 2026-04-13
Electron: Don't open new windows for other windows of the app

jermy-c:
- When you Ctrl/Cmd+Click it opens the link in a new Electron Window along with in the System Browser
- We don't set the windowHandler for all child or later created windows
- This sets the windowHandler to prevent open all windows that's not the main window from opening new Electron windows
- You have to check if the windows is destroyed before checking the web contents otherwise it will throw an error

### DIFF
--- a/desktop/src/main/index.ts
+++ b/desktop/src/main/index.ts
@@ -165,7 +165,6 @@ async function whenReady() {
   }
 }
 
-// Reject opening new windows initially. It can be overriden to allow later.
 app.on('web-contents-created', (event, webContents) => setWindowOpenHandler(webContents));
 
 // Quit when all windows are closed, except on macOS. There, it's common
--- a/desktop/src/main/index.ts
+++ b/desktop/src/main/index.ts
@@ -1,5 +1,5 @@
 import { setMainWindow, startupBackend, shutdownBackend, startupArgs } from '../../backend/backend';
-import { app, shell, BrowserWindow, session, Menu, MenuItemConstructorOptions } from 'electron'
+import { app, shell, BrowserWindow, session, Menu, MenuItemConstructorOptions, type WebContents } from 'electron'
 import { ipcMain } from 'electron/main';
 import { join } from 'path'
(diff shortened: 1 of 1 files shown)


desktop/src/main/index.ts
```diff
@@ -60,6 +60,15 @@ function createWindow(): void {
       return { action: 'deny' }
     })
 
+    app.on('web-contents-created', (event, webContents) => {
+      if (!mainWindow.isDestroyed() && webContents === mainWindow.webContents) {
```
benbucksch:
```suggestion
      if (webContents === mainWindow.webContents && !mainWindow.isDestroyed()) {
```

Does it also work the other way around, or does this throw?

desktop/src/main/index.ts
```diff
@@ -60,6 +60,15 @@ function createWindow(): void {
...
+        return;
+      }
+      webContents.setWindowOpenHandler((details) => {
+        return { action: 'deny' };
```
benbucksch: How would we do #794?

benbucksch (approved):

desktop/src/main/index.ts
jermy-c: Sorry, it is not needed. The app.on() listeners should be outside. I made a push applying to all new windows created maybe it will better for security also. createWindow() sets it again which overrides it.

desktop/src/main/index.ts
```diff
@@ -165,6 +165,9 @@ async function whenReady() {
   }
 }
 
+// Reject opening new windows initially. It can be overriden to allow later.
```
benbucksch:
```suggestion
```

Comment seems like in the wrong place.

benbucksch (approved):

desktop/src/main/index.ts
jermy-c: I think we might be able to do it in `mainWindow`, windowHandler and with `window.open()`.

PR #1161, opened 2026-04-13
WebView: Map Electron InputObject to Browser KeyboardEvent correctly #1154

jermy-c:
- Electron and the Browser has different property names for the Keyboard event
- Electron uses the ones listed in [`before-input-event`](https://www.electronjs.org/docs/latest/api/web-contents#event-before-input-event) not the [`KeyboardInputEvent`](https://www.electronjs.org/docs/latest/api/structures/keyboard-input-event) or [`KeyboardEvent`](https://www.electronjs.org/docs/latest/api/structures/keyboard-event)
- The browser uses [`KeyboardEvent`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent)

### DIFF
--- a/app/frontend/Shared/WebView.svelte
+++ b/app/frontend/Shared/WebView.svelte
@@ -152,7 +152,21 @@
       if (event.type == "mouseDown") {
         webviewE.click();
       } else if (event.type == "rawKeyDown") {
-        onKeyOnMessage(new KeyboardEvent("keydown", event));
+        // Electron InputEvent uses shift / control / alt / meta, but
+        // KeyboardEvent init needs shiftKey / ctrlKey / altKey / metaKey.
+        // see https://www.electronjs.org/docs/latest/api/web-contents#event-before-input-event
+        // see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
+        onKeyOnMessage(new KeyboardEvent("keydown", {
+          key: event.key,
+          code: event.code,
+          repeat: event.isAutoRepeat,
+          isComposing: event.isComposing,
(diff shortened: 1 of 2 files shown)


app/frontend/Shared/WebView.svelte
```diff
@@ -152,7 +152,21 @@
       if (event.type == "mouseDown") {
         webviewE.click();
       } else if (event.type == "rawKeyDown") {
-        onKeyOnMessage(new KeyboardEvent("keydown", event));
+        // Electron InputEvent uses shift / control / alt / meta, but
+        // KeyboardEvent init needs shiftKey / ctrlKey / altKey / metaKey.
+        // see https://www.electronjs.org/docs/latest/api/web-contents#event-before-input-event
+        // see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
+        onKeyOnMessage(new KeyboardEvent("keydown", {
```
benbucksch: Can you please make a function for newElectronKeyboardEvent(), which does the `new Keyboard()`, but also does the mapping? Then it's re-usable in other places.

benbucksch (changes requested):

app/frontend/Shared/WebView.svelte
jermy-c: I've made the change in 87cebf2cc3b9ea932344bf759c6e1ef9ae87718e

benbucksch (approved):

benbucksch:
Thank you. Nicely done. Thanks also for the JSDoc comment.

Utility functions should be at the bottom, not the top. I moved it.

PR #1162, opened 2026-04-14
Electron: Fix starting dev app with single yarn dev command

jermy-c:
- Fix Electron to start dev application with a single `yarn dev` in `/desktop`
- `yarn start` also works now for testing compiled code without a complete production build
- Replace `is.dev` with native Electron `app.isPackaged`


benbucksch:
Where does this load from? Local files from the code directory?

Note: With vite, we currently get auto-reload, which allows me to modify the source code, and the changes are immediately visible in the running app. I don't have to restart the app. Only the one component reloads, but the data stays, which aviods restarts and helps a lot during development. It appears we'd lose this with this PR?

jermy-c:
> Where does this load from? Local files from the code directory?

It loads the `index.html` from `/renderer` instead of `/app`. Then the TS and other assets are loaded from `/app`.

> It appears we'd lose this with this PR?

No, it still auto-reloads when you make a change in the source code. The only difference is that the frontend vite build log, backend vite build and electron logs are in a single terminal together.

jermy-c:
> With vite

Under the hood, `electron-vite` uses vite for the dev server also.

benbucksch:
> Under the hood, electron-vite uses vite for the dev server also.

Ah, right. Thanks. That's the part that I was missing. I thought it was simply pointing to a local dir, serving static files.

benbucksch:
@jermy-c I tried the branch, and yes, it works fine. Auto-reload of changed code also works.

However, startup time is drastically regressed, because app/ files now have to be recompiled on *every* start.

Before PR:
* (Warm) start of desktop/, app/ already running: 4 seconds
* (Cold) start of desktop/, app/ also starting: 10 seconds

After PR:
* (Warm) start of desktop/: 10 seconds
* (Cold) start of desktop/: 10 seconds

I do the warm start all the time, and the cold start only once every few days, so the start now takes 2.5 times as long as before.

jermy-c:
> However, startup time is drastically regressed, because app/ files now have to be recompiled on _every_ start.

Ok, I'll try to find if there's a config to speed it up a bit. But yes, it has to recompile everything on start now because it only uses `electron-vite` to compile it. Before we still had a renderer dev server by `electron-vite` at `localhost:5173` but we ignored it and used the one in `/app`. Now, it is just one dev server by `electron-vite`.

benbucksch:
I see 2 options:
* Let the dev server run (and we need some manual way to restart it, e.g. when we add new packages)
* On backend start, check whether the port for the dev server is open. If so, use it. Otherwise, use the internal dev server.

jermy-c:
### Changes

- Reverted to having frontend served from `/app`, so you have to do `yarn dev` there also
- Fixed the condition so `yarn start` works with `import.meta.env.DEV` which is only true with `electron-vite dev`
- Made the renderer config `undefined` when only when running `electron-vite dev` so there's no dev server for the frontend created in `electron-vite`
- Silenced the missing config for `dev` only

### Improvements

- No redundant dev server in `electron-vite`
- Remove use of `is.dev` condition from the toolkit
- Fix `yarn start` for testing that requires a build of the main and preload scripts but not a full production build

PR #1163, opened 2026-04-15
Router: Memory history

jermy-c:
- Use memory history so the browser URL is not changed


benbucksch:
same as #1090

app/frontend/AppsBar/selectedApp.ts
```diff
@@ -3,7 +3,8 @@ import { appGlobal } from "../../logic/app";
 import { backgroundError } from "../Util/error";
 import { ArrayColl, MapColl } from "svelte-collections";
 import { writable, type Writable } from "svelte/store";
-import { createHashHistory } from "../Util/hashHistory";
+import { navigate, createHistory, createMemorySource } from "svelte-navigator";
+
```
benbucksch:
```suggestion
import { createHistory, createMemorySource } from "svelte-navigator";
```

PR #1172, opened 2026-04-16
Composer: Tab key to indent #1165

### LINKED ISSUE #1165: Composer: Tab does not indent the text
### Reproduction
* Write email
* On your keyboard, press the [Tab] key
* Press the [Shift-Tab] key

### Actual result
* Focus moves away from the editor, to another control of the composer UI

### Expected result
* If the cursor is in a list item (at any point in the item, not necessarily at the start):
  * [Tab] is pressed, then make a new sublist, or add to the existing that follows - `sinkListItem()`.
  * If [Shift]+[Tab] is pressed: `liftListItem()`
* When inside blockquote:
  * If [Tab] is pressed, wrap the blockquote inside another blockquote.
  * If [Shift]+[Tab] is pressed, remove the inner-most blockquote.
* When not in any list, blockquote, or similar:
  * [Tab] should insert 4 spaces at the cursor position
  * [Shift]+[Tab]: if there are more than 2 spaces right before the cursor, otherwise do nothing.

### Implementation
If there is an existing TipTap extension that is similar to the above, use that one. In this case, don't bother getting the details exactly as stated above.

jermy-c:
Performs indent actions based on the parent nodeType.

 Tab:
 - BulletList or OrderedList: sinks the list item, if the item is already nested then it will do nothing
 - Blockquote: wraps content in blockquote
 - Other nodeType: inserts 4 spaces

Shift-Tab :
 - BulletList or OrderedList: lifts the list item
 - Blockquote: unwraps content in blockquote
 - Other nodeType: removes 4 spaces or less if cursor is at start of line, if it's not at the start or there's less than 2 spaces it will do nothing

### DIFF
--- a/app/frontend/Shared/Editor/HTMLEditor.svelte
+++ b/app/frontend/Shared/Editor/HTMLEditor.svelte
@@ -12,6 +12,7 @@
   import { Footer } from './Footer';
   import { BoldStar, ItalicSlash } from './StdConventions';
   import { ParagraphNewLine } from './ParagraphNewLine';
+  import { TabIndent } from './TabIndent';
   // import CodeBlockLowlightFeature from '@tiptap/extension-code-block-lowlight';
   // import { common as lowlightCommon, createLowlight } from 'lowlight'
   import { onMount, onDestroy } from 'svelte';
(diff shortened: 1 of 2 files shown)


benbucksch (approved):
Beautiful!

Nice code, and works flawlessly. Great work!

PR #1173, opened 2026-04-17
S/MIME: Using node-forge

NeilRashbrook:
Using the `node-forge` library. Functions:

- Import password-protected private key
- Import certificate(s) for private key
- Import public key
- Sign messages
- Encrypt messages
- Decrypt messages
- Verify signatures

Note: When Thunderbird encrypts a message, it only includes the Content-Type header in the encrypted message. Currently this code generates an encrypted message with all headers and expects that when decrypting.


NeilRashbrook:
Obsoleted by #1196.

PR #1181, opened 2026-04-21
Workaround ready-to-show not firing consistently on Wayland

NeilRashbrook:
The app definitely starts up on my system with this change...

(Did something change to stop the F12 key from working?)

### DIFF
--- a/desktop/src/main/index.ts
+++ b/desktop/src/main/index.ts
@@ -28,9 +28,18 @@ function createWindow(): void {
     })
     setMainWindow(mainWindow);
 
-    mainWindow.on('ready-to-show', () => {
-      mainWindow.show()
-    })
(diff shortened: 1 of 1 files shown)


benbucksch:
> Did something change to stop the F12 key from working?

Yes. Try Ctrl-Shift-i .

benbucksch (approved):

benbucksch:
And there we go the way all the widely used apps go.... Lots of "if system this, else do that".

NeilRashbrook:
(Also the app used to show up as `mustang` in my activity bar, but it now says `electron`. Or nothing at all.)

(Also the developer tools are broken for me; I can't click on the sources » "tab" selector chevron any more.)

Issue #1187, opened 2026-04-27
EWS: Delegated send: matchesEMailAddress() runs with email address = null


benbucksch:
### Environment
* EWS
* Delegated account (functional account, brought in via account sharing)

### Reproduction
* Send email with delegated account

### Actual result
* "Cannot read properties of undefined (reading 'toLowerCase')" in `ExchangePermission.matchesEMailAddress()` in `EWSFolder.ts`
* Stack
```
../../../../app/logic/Mail/EWS/EWSFolder.ts:651:30inExchangePermission.matchesEMailAddress
In App
  matchesEMailAddress(emailAddress: string) {
    return this.emailAddress.toLowerCase() == emailAddress.toLowerCase();
  }
../../../../app/logic/Mail/EWS/EWSAccount.ts:173:89inpermissions
In App
<anonymous>inArray.find
In App
../../../../app/logic/Mail/EWS/EWSAccount.ts:173:57inEWSAccount.send
In App
../../../../app/logic/Mail/ComposeActions.ts:330:5inComposeActions.send
In App
../../../../app/frontend/Mail/Composer/MailComposer.svelte:334:11inonSend
```

### Expected result
* Mail is send with delegated account as From
* Sent mail is saved in the Sent folder of the delegated account, not the user account

benbucksch:
A null pointer check would be trivial:
```ts
return this.emailAddress?.toLowerCase() == emailAddress?.toLowerCase();
```
but I don't know which of the email addresses is null, why it's null, and what would happen (from end user perspective) when the comparison returns false.

It seems the root problem is that one of the email addresses is null.

NeilRashbrook:
This is a regression from PR #1112. It fails because the permissions for the special users "Default" and "Anonymous" don't have an email address. (This wasn't the case in my initial implementation of PR #885 where each `ExchangePermission` was derived from `PersonUID` and did have an email address value.)

benbucksch:
@NeilRashbrook Good to know that you found the cause. Can you fix it, please?

benbucksch:
Thanks for the fix. Were you able to reproduce the error before the change, and confirmed that the PR fixes it?

NeilRashbrook:
Yes, I have a profile for `neil@bonm` with `ben@` as a delegated account and I was able to reproduce the issue while with the patch I was able to send an email with the subject `#1187`.

benbucksch:
It appears that this bug was fixed, but now he runs into #1108 again, with the same reproduction.

PR #1188, opened 2026-04-27
Exchange permissions: Check email addresses case-insensitively where applicable #1187

NeilRashbrook:


app/logic/Mail/EWS/EWSFolder.ts
```diff
@@ -648,7 +648,7 @@ export class ExchangePermission {
   }
 
   matchesEMailAddress(emailAddress: string) {
-    return this.emailAddress.toLowerCase() == emailAddress.toLowerCase();
+    return this.emailAddress?.toLowerCase() == emailAddress.toLowerCase();
```
benbucksch, line 652:
```suggestion
    // (Null check: Permissions for the special users "Default" and "Anonymous" don't have an email address)
    return this.emailAddress?.toLowerCase() == emailAddress.toLowerCase();
```

benbucksch (approved):

benbucksch:
Please always make code comments for non-obvious cases like this.

PR #1191, opened 2026-04-30
OWA: Support large attachments on Office 365

NeilRashbrook:
Apart from failing to notice that the Action has to be set to CreateAttachmentFromLocalFile, the code changes were easier here because we were adding attachments through the API anyway.

(I still don't like that ? : format.)

### DIFF
--- a/app/logic/Mail/OWA/OWAAccount.ts
+++ b/app/logic/Mail/OWA/OWAAccount.ts
@@ -317,7 +317,7 @@ export class OWAAccount extends MailAccount {
         },
       });
       if (this.authorizationHeader) {
-        await this.callOWA(request, { attachment });
+        await this.callOWAWithOffice365Attachment(request, attachment);
       } else {
         request.Body.Attachments[0].Content = await blobToBase64(attachment.content);
         await this.callOWA(request);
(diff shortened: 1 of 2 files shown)


benbucksch:
> I still don't like that ? : format.

You mean the bug onenin callOWA()? Yes, I agree.

Generally we should avoid `if` and `?:` where easily possible.
When most of the function is in ifs, that suggests that it should be 2 separate functions. This is the case here. Can you please split it up?

This is all the more true as callOWA() is a core
function and should stay lean, not loaden with special cases.

In general, instead of multiple cases into a single function, rather make 2 functions with dedicated purpose, and where necessary share code, again using functions that have a clearly defined purpose and are reusable beyond that current case.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -299,19 +294,79 @@ export class OWAAccount extends MailAccount {
...
       let mainAccount = this.mainAccount as OWAAccount;
-      return await mainAccount.callOWA(aRequest, { mailbox: this.username });
+      return await mainAccount.callOWA(aRequest, { mailbox: this.username, attachment });
     }
     if (!this.hasLoggedIn) {
       throw new LoginError(null, gt`Please login`);
     }
-    let url = this.url + 'service.svc';
-    let options = {
+    let url = this.url + (attachment ? "service.svc/CreateAttachmentFromLocalFile" : "service.svc");
```
benbucksch: This whole section can move out of the function.

The rest is trivial (a few lines) and can be copied over.

benbucksch (changes requested):

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: I'm not sure what you mean by that... `callOWA` is a 75-line function, and I don't want to maintain two copies of it if I don't need to.

NeilRashbrook (commented):
When I said that ? : format, I meant the way the lines are formatted.

app/logic/Mail/OWA/OWAAccount.ts
benbucksch: Ah, the rest was hidden in the PR UI.

What I meant was that this whole block here is at the beginning of the functionc, and is for one special case only. It should be factored out.

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -299,19 +294,76 @@ export class OWAAccount extends MailAccount {
     }
     addRecipients(request, "BccRecipients", email.bcc.contents);
     request.addField("Message", "From", { Mailbox: { Name: email.from.name, EmailAddress: email.from.emailAddress } }, "message:From");
-    await this.callOWA(request);
+    let response = await this.callOWA(request);
+    if (request.Body.MessageDisposition == "SendAndSaveCopy") {
+      return;
+    }
+    let itemID = response.Items[0].ItemId.Id;
```
benbucksch: This from here on could easily be moved into a function.

Please try not to make functions that have 40 or more lines.

benbucksch (changes requested):

PR #1194, opened 2026-05-01
Throw if attempting to get capabilities before login

NeilRashbrook:
Drafting because this depends on PR #1193.

The race condition happens because startup code (e.g. `loadFolder`) can try to access `IMAPAccount.hasCapability` before `IMAPAccount.startup` has refreshed it.

This also allows `this.connections.get(ConnectionPurpose.Main)?.authenticated` to work (see comment in `isLoggedIn`).

### DIFF
--- a/app/logic/Mail/IMAP/IMAPAccount.ts
+++ b/app/logic/Mail/IMAP/IMAPAccount.ts
@@ -66,7 +66,6 @@ export class IMAPAccount extends MailAccount {
   }
 
   async startup() {
-    this.namespaces = await this.getNamespaces();
     await super.startup();
     this.notifyObservers();
     (this.inbox as IMAPFolder).startPolling();
@@ -157,6 +156,8 @@ export class IMAPAccount extends MailAccount {
       this.connections.set(purpose, connection);
       this.connectionLock.set(connection, new Lock());
       if (purpose == ConnectionPurpose.Main) {
+        // Get namespaces and refresh connection early to avoid race condition.
+        this.namespaces = await this.getNamespaces();
         this.notifyObservers();
       }
       this.log(null, connection, "connected");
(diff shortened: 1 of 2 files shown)


app/logic/Mail/IMAP/IMAPAccount.ts
```diff
@@ -157,6 +156,8 @@ export class IMAPAccount extends MailAccount {
       this.connections.set(purpose, connection);
       this.connectionLock.set(connection, new Lock());
       if (purpose == ConnectionPurpose.Main) {
+        // Get namespaces and refresh connection early to avoid race condition.
+        this.namespaces = await this.getNamespaces();
```
benbucksch: This does not belong in this function. This function should only connect, and otherwise not talk to the server.

Functions are not opportunistic, but have a single purpose. They do exactly what they say in the function name, all of it, and **only** that.

benbucksch (changes requested):

benbucksch:
What code is checking `isLoggedIn` after `connect()` and before `startup()`?

NeilRashbrook:
> What code is checking `isLoggedIn` after `connect()` and before `startup()`?

I don't know, I just mentioned `isLoggedIn` because I happened to notice the comment. (It might even work already as of #1193, but I didn't check.)

NeilRashbrook:
New approach checks the `authenticated` property to see whether login has refreshed the connection before trying to get its capabilities. Except the UI doesn't show the logged-in state, despite the `notifyObservers` in `startup`...

NeilRashbrook:
Revised new approach: just throw if we're not logged in; it's not safe to try to log in at this point.

benbucksch (approved):

benbucksch (approved):
Makes sense

PR #1196, opened 2026-05-08
S/MIME: Implementation

NeilRashbrook:
Based on the "asn1.js" library (not to be confused with any other asn1 js library). I took the library's core code files, turned them into ES6 modules, updated them to use `bigint` and `Uint8Array`, and made them somewhat type-safe (the original code used all sorts of tricks that TypeScript really doesn't like). I made at least one other change (to allow encoding of objects that share keys with their parent, which already decodes correctly) but the code doesn't currently use that feature. Password protection and hashing are done using WebCrypto but the actual RSAES encryption and decryption aren't supported by WebCrypto so I had to implement those manually. I tried to use WebCrypto RSAES for verification but I couldn't get it to work but I was able to reuse the encryption and decryption code. As such I didn't try using WebCrypto RSAES for signing. Functions:

- Import password-protected private key
- Import certificate(s) for private key
- Import public key
- Sign messages
- Encrypt messages
- Decrypt messages
- Verify signatures

Currently only supports PKCS#8 format private keys (which can optionally be password-protected), not raw RSA private keys (PKCS#1).


benbucksch:
> Based on the "asn1.js" library

Could you please add an appropriate credits and license file in lib/asn1/?

app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
```diff
@@ -29,10 +32,71 @@ export class SMIMEPrivateKey extends SMIMEPublicKey implements PrivateKey {
...
+      let cryptoKey = await this.decryptCryptoKey("SHA-256");
+      privateKeyInfo = PrivateKeyInfo.decode(new Uint8Array(await crypto.subtle.exportKey("pkcs8", cryptoKey)));
+    }
+    if (privateKeyInfo.privateKeyAlgorithm.algorithm != "rsaEncryption") {
+      throw new Error("Unsupported private key algorithm");
+    }
+    return RSAPrivateKey.decode(privateKeyInfo.privateKey);
+  }
+
+  async decryptCryptoKey(hash: "SHA-256" | "SHA-384" | "SHA-512"): Promise<CryptoKey> {
```
benbucksch: Make this type an enum, please

app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
```diff
@@ -29,10 +32,71 @@ export class SMIMEPrivateKey extends SMIMEPublicKey implements PrivateKey {
...
+    let pbes2 = PBES2Params.decode(encryptedKey.encryptionAlgorithm.parameters);
+    if (pbes2.keyDerivationFunc.algorithm != "pkcs5PBKDF2") {
+      throw new Error("Unsupported private key derivation function");
+    }
+    let length = sanitize.translate(pbes2.encryptionScheme.algorithm, { aes128cbc: 128, aes192cbc: 192, aes256cbc: 256 });
+    let pbkdf2 = PBKDF2Params.decode(pbes2.keyDerivationFunc.parameters);
+    if (pbkdf2.salt.type != "specified") {
+      throw new Error("Unsupported private key derivation salt");
+    }
+    let prfhash = sanitize.translate(pbkdf2.prf.algorithm, { hmacWithSHA1: "SHA-1", hmacWithSHA256: "SHA-256", hmacWithSHA384: "SHA-384", hmacWithSHA512: "SHA-512" });
```
benbucksch: use the enum values here

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
```diff
@@ -0,0 +1,127 @@
...
+  runOn = ProcessingStartOn.Parse;
+  async process(email: EMail, postal: PostalEmail) {
+    // There is a direct accessor to the content type in the PostalMIME object
+    // itself, but we don't have that here. And to get the headers in the
+    // EMail object requires an extra parse, which seems wasteful. So
+    // let's fish the header out of the PostalEmail object...
+    let contentTypeHeader = postal.headers.find(header => header.key == "content-type")?.value ?? "";
+    let contentType = contentTypeHeader.split(";")[0].trim().toLowerCase();
+    if (contentType == "application/pkcs7-mime") {
+      // Encrypted messages only have a body part but fsr this is an attachment.
```
benbucksch, line 23: fsr?

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
```diff
@@ -0,0 +1,127 @@
...
+          console.log("signature did not contain a message digest");
+          return;
+        }
+        if (indexedDB.cmp(OctetString.decode(digestAttribute.attrValue[0]), messageDigest)) {
+          console.log("signed digest did not match message");
+          return;
+        }
+        let signedAttrs = Attributes.encode(signerInfo.signedAttrs);
+        let attributesDigest = new Uint8Array(await crypto.subtle.digest(digestAlgorithm, signedAttrs));
+        /* This doesn't work fsr
```
benbucksch: fsr?

app/logic/Mail/Encryption/SMIME/SMIMESend.ts
```diff
@@ -14,8 +21,198 @@ export class SMIMESend {
...
+            contentType: "data",
+          },
+          certificates: [myCertificate],
+          signerInfos: [signerInfo],
+        },
+      };
+      for (let cert of privateKey.chain) {
+        signedData.content.certificates.push(Certificate.decodePEM(cert.certificate, { label: "CERTIFICATE" }));
+      }
+      /*
```
benbucksch: do we need this commented-out code?

app/logic/Mail/Encryption/PublicKey.ts
```diff
@@ -100,7 +100,7 @@ export class PublicKey extends Observable {
     json.name = this.name;
     json.id = this.id;
     json.fingerprint = this.fingerprint;
-    json.created = this.created.toISOString();
+    json.created = this.created?.toISOString();
```
benbucksch: This exists since God existed?

app/logic/Mail/Encryption/PublicKey.ts
```diff
@@ -117,8 +117,8 @@ export class PublicKey extends Observable {
     this.system = sanitize.enum<EncryptionSystem>(json.system, Object.values(EncryptionSystem));
     this.id = sanitize.alphanumdash(json.id);
     this.name = sanitize.label(json.name, this.id.substring(0, 4));
-    this.fingerprint = sanitize.alphanumdash(json.fingerprint);
-    this.created = sanitize.date(json.created);
+    this.fingerprint = sanitize.alphanumdash(json.fingerprint, "");
+    this.created = sanitize.date(json.created, null);
```
benbucksch: ditto

app/logic/Mail/Encryption/PublicKey.ts
```diff
@@ -117,8 +117,8 @@ export class PublicKey extends Observable {
     this.system = sanitize.enum<EncryptionSystem>(json.system, Object.values(EncryptionSystem));
     this.id = sanitize.alphanumdash(json.id);
     this.name = sanitize.label(json.name, this.id.substring(0, 4));
-    this.fingerprint = sanitize.alphanumdash(json.fingerprint);
-    this.created = sanitize.date(json.created);
+    this.fingerprint = sanitize.alphanumdash(json.fingerprint, "");
```
benbucksch: When would it not have a fingerprint?

app/frontend/Contacts/PersonPage/EncryptionKey.svelte
```diff
@@ -120,7 +120,7 @@
         </hbox>
         <hbox>
           <hbox class="label">{$t`ID`}</hbox>
-          <hbox class="value">{key.id}</hbox>
+          <hbox class="value">{key.id.slice(-16)}</hbox>
```
benbucksch: How long is it? And who creates such long IDs, for what purpose?

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
```diff
@@ -98,6 +98,29 @@
...
+        <vbox>
+          <hbox class="label">{$t`Certificate chain *=> Any additional certificates that might be required to authenticate this key`}</hbox>
+          {#each $chain.each as certificate}
+            <hbox class="verification-code">
+              <pre wrap class="value">{certificate.fingerprintDisplay}</pre>
+              <vbox>
+                <RoundButton
+                  label={$t`Delete`}
+                  icon={DeleteIcon}
+                  onClick={() => onRemoveFromChain(certificate)}
```
benbucksch: Why would I do that?

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
```diff
@@ -98,6 +98,29 @@
           {/each}
         </vbox>
       </hbox>
+      {#if chain}
+        <vbox>
+          <hbox class="label">{$t`Certificate chain *=> Any additional certificates that might be required to authenticate this key`}</hbox>
+          {#each $chain.each as certificate}
+            <hbox class="verification-code">
+              <pre wrap class="value">{certificate.fingerprintDisplay}</pre>
```
benbucksch: Would users compare these?

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
```diff
@@ -200,6 +229,21 @@
     identity.encryptionPrivateKeys.remove(key);
     await identity.account.save();
   }
+  let fileSelector: FileSelector;
+  const acceptFileTypes = [ "application/x-x509-email-cert", "application/x-x509-ca-cert", "application/pkix-cert", ".crt" ];
+  async function onFileCertificate() {
```
benbucksch:
```suggestion
  async function onAddCertificateSMIME() {
```

benbucksch (changes requested):

app/frontend/Contacts/PersonPage/EncryptionKey.svelte
NeilRashbrook: RSA keys themselves don't have an ID; the nearest equivalent is that once you have obtained a certificate then you can use its fingerprint. All you have to play with is the modulus and the exponent, and the latter is always `0x10001`, which is unhelpful, and the modulus for a typical 4096-bit key is of course 1024 hex digits, which is a bit long to display.

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
NeilRashbrook: Certificates don't exactly have a `displayName` property. The fingerprint is something that users do get asked to check on occasion. CAs should have a subject with a common name though, so I could display that if it's there.

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
NeilRashbrook: Added the wrong certificate? Intermediate certificate has expired?

app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
NeilRashbrook: Actually I ended up not using this parameter, and forgot to remove it.

app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
NeilRashbrook: Not exactly sure what you're looking for here.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
NeilRashbrook: Abbreviation of "for some reason".

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (contentType == "application/pkcs7-mime") {
+      // Encrypted messages only have a body part but fsr this is an attachment.
+      let encrypted = email.attachments.first;
+      let envelopedData = EnvelopedData.decode(new Uint8Array(await encrypted.content.arrayBuffer()));
+      if (!["aes128cbc", "aes192cbc", "aes256cbc"].includes(envelopedData.content.encryptedContentInfo.contentEncryptionAlgorithm.algorithm)) {
+        return;
+      }
+      let vector = OctetString.decode(envelopedData.content.encryptedContentInfo.contentEncryptionAlgorithm.parameters);
+      let encryptedContent = envelopedData.content.encryptedContentInfo.encryptedContent;
+      // XXX what if you were BCC'd?
```
NeilRashbrook, line 31: While that's true, that's not the issue here, where we're receiving the message, and need to know which identity's private key to use to decrypt it.

app/logic/Mail/Encryption/SMIME/SMIMESend.ts
NeilRashbrook: My bad; I meant to delete it once I had the asn1.js version working, but I left it in while I was developing so that I could see what needed to be done.

app/logic/Mail/Encryption/PublicKey.ts
NeilRashbrook: S/MIME private keys and certificates are typically stored in separate files (there is a binary format that combines them but I don't know whether you want me to look into supporting that). There is therefore a window of opportunity where a private key object doesn't have a certificate and therefore a fingerprint.

app/logic/Mail/Encryption/PublicKey.ts
NeilRashbrook: I don't actually know what this variable is for, so I don't know what I should be setting it to (always supposing that there is something suitable for me to set it to).

NeilRashbrook (commented):
> Could you please add an appropriate credits and license file in lib/asn1/?

Please can you point to an existing example so that I know what you're looking for?

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
benbucksch: Oh, sorry.
In an incoming BCC, we should normally be listed in the BCC.

If it's not there, then check all private keyse of all identities of this account where the message came in.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
benbucksch: Never heard that abbr.

The comment still makes no sense to me even with that expanded, esp. the part after "but". Please write more words to make clearer what you mean.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
benbucksch: Don't write "doesn't work", but what you expected and what you actually happens.

app/logic/Mail/Encryption/PublicKey.ts
benbucksch: While we import private keys, shouldn't we make sure that we have both parts?

app/frontend/Contacts/PersonPage/EncryptionKey.svelte
benbucksch: Same as above. Private keys should have a certificate that comes with it.

benbucksch: You're not using (the modulus of) the actual private key as ID, do you? Please make sure to properly populate the ID field on import.

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
benbucksch: If the user added the wrong certificate (which doesn't chain up), then we should bail already during addition.

benbucksch: If parts of the certificate chain are missing during import, can you please make sure to ask during import?

Additionally, you can keep the "add cert" button here for now, but I'm not sure it will stay here.

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
benbucksch: > The fingerprint is something that users do get asked to check on occasion.

The fingerprint of the CA?

> CAs should have a subject with a common name though, so I could display that if it's there.

Yes, please. Is the CA cert valid without common name?

Also, if the cert is only valid for a domain, show that. (I think it's subjectName?)

benbucksch:
> Please can you point to an existing example so that I know what you're looking for?

You've never seen a license file and author credits in a JS library?

Well, this is third party code that you copied from somewhere (and then modified), right? What's the credits and license of that origin? Keep those. Put them in files in dir lib/asn1/ .

For example, if you got the code from npm asn1, then make a file lib/asn1/LICENSE and put the contents of https://github.com/TritonDataCenter/node-asn1?tab=License-1-ov-file == https://github.com/TritonDataCenter/node-asn1/blob/master/LICENSE in there.

Furthermore, you find more authors in the `contributors` field of the package.json file in https://github.com/TritonDataCenter/node-asn1/blob/master/package.json .

Of course, if you based it on another library, then please use the license and credit of that one.

app/logic/Mail/Encryption/PublicKey.ts
benbucksch: It's a propery `created`, type Date, on a Public or Private Key. You have 3 guesses what it is for... It's the time when the public or private key material was originally generated.

I'm fairly certain that every S/MIME certificate has one, and it's important, because the certificate is valid only between the time of creation and the expiry, not before and not after. So, yes, it has to be populated.

It cannot be null. Please remove this change here and make sure that you during import that we have a complete certificate (also for a private key) with all required data, not only half.

app/frontend/Contacts/PersonPage/EncryptionKey.svelte
NeilRashbrook: > Private keys should have a certificate that comes with it.

Wrong way around; a single private key can have multiple certificates, as they they expire and can be replaced without having to change the private key. If you want a shorter ID I could always create a SHA-256 hash of the RSAPublicKey structure. (I think the only other part of the certificate that is guaranteed be the same is the subject email address, but even if it is I'm not sure that qualifies as an ID.)

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
NeilRashbrook: > Also, if the cert is only valid for a domain, show that. (I think it's subjectName?)

That's server SSL certificates, which doesn't apply here; S/MIME uses email certificates; they can have a subjectAltName, but it should be the same as the subject email address.

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
NeilRashbrook: > If the user added the wrong certificate (which doesn't chain up), then we should bail already during addition.

OK, I think I can implement that.

> If parts of the certificate chain are missing during import, can you please make sure to ask during import?

To know how long the chain is we would need an API for root certificates. (When you find a certificate whose issuer is a root certificate you recognise then you know you've got to the end of the chain). I don't know how to do that in Electron.

> Additionally, you can keep the "add cert" button here for now, but I'm not sure it will stay here.

(For now, the Add Certificate button also allows the user to replace an expired certificate.)

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
NeilRashbrook: > In an incoming BCC, we should normally be listed in the BCC.

Excuse me? That's never how it's worked. (Note that some clients, e.g. Gmail web, show you as in the BCC list if you don't appear to be in the To or CC list.)

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
NeilRashbrook: Steps to reproduce problem: View a correctly signed message.

Expected result (works): `crypto.subtle.verify` returns `Promise<true>`.

Actual result (doesn't work): `crypto.subtle.verify` returns `Promise<false>`.

I'm not sure what else you were expecting it to do...

app/logic/Mail/Encryption/PublicKey.ts
NeilRashbrook: How are you proposing to do that, an immediate second filepicker prompt for the certificate?

app/logic/Mail/Encryption/PublicKey.ts
NeilRashbrook: Sorry, I do actually set created (and expires) from the certificate but then forgot about them. That just leaves the issue of how to ensure there's a certificate when the private key is imported.

NeilRashbrook (commented):
Corrections to my initial description:
- I did use the change to allow nested ASN.1 objects to share the same key; an RDNSequence is an array of RDNs, which are themselves in theory an array of SingleAttribute values. So far, I've never seen anyone put more than one SingleAttribute in an RDN, so I don't know whether it's possible, and if so, I would have to tweak the code, but in the mean time I just omit the intervening array.
- I also use WebCrypto for the AES-CBC encryption and decryption.

> You've never seen a license file and author credits in a JS library?

I've rarely interacted with JS libraries at all; other than for some of your projects I don't think I've deliberately used them.

> For example, if you got the code from npm asn1
>
>> the "asn1.js" library (not to be confused with any other asn1 js library)

Anyway, I found the LICENCE and I'll add it in. There weren't any contributors listed though.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
benbucksch: I meant to put that into the code comment:
```suggestion
        /* Doesn't work: `await crypto.subtle.verify()` returns `false` instead of `true`.
```

benbucksch:
> That just leaves the issue of how to ensure there's a certificate when the private key is imported.

You ensure that in the UI for importing S/MIME keys.

> How are you proposing to do that, an immediate second filepicker prompt for the certificate?

Yes, exactly. If you get a private key *without* cert, you show another step in the import process where you ask the user to select the corresponding certificate file. Then, you check whether they match. Only then do you add the private key.

That's all the more important, given that S/MIME depends on a CA signing the private key. Without the CA signature, it's just a self-signed key, which is useless.

app/frontend/Contacts/PersonPage/EncryptionKey.svelte
benbucksch: Please google a bit what other apps show as identifier for a private key.

*Not* the email address, please

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
benbucksch: > When you find a certificate whose issuer is a root certificate you recognise then you know you've got to the end of the chain

Correct. We should ensure that the certificate chains to a known CA. Only then is the cert actually usable.

> I don't know how to do that in Electron

We're in the chrome browser, not Electron. So, WebCrypto. If there is no API for root CAs, then so be it. But there should be. Rhere should be a way to validate whether a cert is valid.

benbucksch: When you're *reading* incoming S/MIME signatures, how do you validate that the certificates is chained to a CA that *we* accept? Otherwise, the signature would not be valid.

If you're not checking that, that is absolutely necessary.

Once you can check certificates for incoming messages, you can also check imported certificates for private keys.

app/frontend/Contacts/PersonPage/EncryptionKey.svelte
NeilRashbrook: I wasn't able to find an email application that showed you the private key. Instead, you get to choose between certificates that had been imported with their private key. (In some cases, you see a list of all certificates, and the ones with a private key are marked with a key icon.)

I had trouble finding suitable screenshots to see how other apps display your selected certificate, so I think this is accurate but it's hard to be sure:

Outlook for Windows: Common name
Outlook for Mac, Apple mail for Mac: Email address
Apple mail for iOS: Email address
Thunderbird: Common name + Serial number
CipherMail: Email address + Common name
Android: User names certificate on import, default name is a hexadecimal blob

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
NeilRashbrook: (By 'Electron' I mean anything available to the application process, whether front or back end. If I meant only the back end, I would say 'Node', and if I meant only the front end, I would say 'browser'.)

There is no API for root CAs in WebCrypto. Or indeed any API for certificates at all. (Fortunately there is an API for decrypting password-protected private keys.)

And you're right, I'm not validating the certificate on incoming messages.

benbucksch: > validating the certificate on incoming messages.

That is crucially important. Let's do this before we merge, as discussed.

We're lucky, and node.js 22.15 just added an API which does *precisely* what we need:
https://nodejs.org/api/tls.html#tlsgetcacertificatestype

Add this to the backend as a 1-liner, simply exposing this function to the frontend. It returns PEMs as array of strings, and you can already read PEMs, so that's easy. Keep the CAs as global variable. Read them lazily.

Please distinguish between Mozilla CA = "bundled" (node.js) = `TrustLevel.ThirdParty` (our enum) and OS CA store = "system" (node.js) = `TrustLevel.OS` (new enum value).

Check all places where we check for `TrustLevel.ThirdParty` and treat `TrustLevel.OS` the same as ThirdParty for now. I will later change the UI to make it slightly different, e.g. a different color and label.

NeilRashbrook:
(Re-implementation?)

- Added support for verifying certificate chains against built-in, OS and user-provided CA certificates
- Added support for creating private keys and certificate signing requests
- Various other fixes

(I happen to remember another change I made to asn1.js - I renamed `iso646str` to `visstr` since the documentation I was reading referred to it as VisibleString.)

app/frontend/Contacts/PersonPage/EncryptionKey.svelte
benbucksch: Still, please fix the ID (e.g. limit to 16 chars) at the place where you set it, not here in the display.

app/logic/Mail/Encryption/PublicKey.ts
```diff
@@ -172,13 +175,15 @@ export function trustOrder(trustLevel: TrustLevel): number {
 export const trustColor = {
   [TrustLevel.Personal]: "green",
   [TrustLevel.ThirdParty]: "blue",
+  [TrustLevel.OS]: "orange",
```
benbucksch:
```suggestion
  [TrustLevel.OS]: "#005eff", // lighter blue
```

app/logic/Mail/Encryption/PublicKey.ts
```diff
@@ -172,13 +175,15 @@ export function trustOrder(trustLevel: TrustLevel): number {
...
   [TrustLevel.Sender]: "yellow",
   [TrustLevel.Distrusted]: "red",
 }
 
 export const trustColorFG = {
   [TrustLevel.Personal]: "white",
   [TrustLevel.ThirdParty]: "white",
+  [TrustLevel.OS]: "black",
```
benbucksch:
```suggestion
  [TrustLevel.OS]: "white",
```

desktop/backend/backend.ts
```diff
@@ -217,6 +219,15 @@ function newHTTPServer() {
   return new HTTPServer();
 }
 
+function getCACertificates(type: string) {
+  try {
+    return tls.getCACertificates(type);
+  } catch (ex) {
+    console.error(ex);
+    return type == "bundled" ? tls.rootCertificates : [];
```
benbucksch: Be more specific to the problem you try to solve, please:

```suggestion
    // Fallback for old node.js
    if (!tls.getCACertificates && tls.rootCertificates && type == "bundled") {
      return tls.rootCertificates;
    }
    return [];
```

app/logic/Mail/Encryption/SMIME/SMIMERSAES.ts
```diff
@@ -0,0 +1,157 @@
+// RSAES-PKCS1.5-v1_5 operations. See RFC8017 section 7.2
```
benbucksch, line 1: Did you copy this from somewhere? If so, please add copyright and license.

In general, please always do that when you copy code.

benbucksch (changes requested):

NeilRashbrook:
- Added support for importing certificate chains from one file
- Various other fixes

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
benbucksch: ping

benbucksch:
@NeilRashbrook I've made an automatic merge with master. Can you please check that the merge is correct?

This is still marked as Draft. Can we finish this, please? I'd like to merge this.

benbucksch:
Here's a review by Claude Fable:

What it is: A substantial clean-room S/MIME implementation (+3352 lines, draft). It vendors a
TS/bigint/Uint8Array port of asn1.js (lib/asn1/), builds CMS structures (SignedData/EnvelopedData/CSR)
via an ASN.1 DSL, implements RSA primitives by hand (WebCrypto won't expose raw RSA), and uses WebCrypto
for hashing, AES-CBC, and PBES2/PBKDF2 key import. Full send/receive sign+verify+encrypt+decrypt, key
generation, and CSR export. It fits the project's clean-room conventions (no protocol libs). Overall the
cryptographic construction is correct — sign-then-encrypt ordering, correct RSAES/RSASSA PKCS#1 v1.5
padding, CRT decryption with proper Garner recombination, encrypt-to-self, and signed-attributes hashed
as SET(0x31). Solid work; the issues below are mostly hardening, one real fallback bug, and polish.

Security

1. [Medium] Lenient PKCS#1 v1.5 signature unpadding (BERserk surface). unpadPKCS (SMIMERSAES.ts:60) for
Signed blocks only checks 00 01, ≥8 non-zero bytes, then a zero — it does not require all padding octets
be 0xFF, and callers then DigestInfo.decode(...) which ignores trailing bytes. Robust verifiers
reconstruct 00 01 FF..FF 00 ‖ DigestInfo and compare. Exploitable only against small-exponent (e=3)
keys, i.e. CA chain verification (SMIMEPublicKey.verifySignature), not message signatures. Low
real-world risk (CAs use 65537) but worth hardening.
2. [Medium] Non-constant-time RSA + PKCS#1v1.5 padding oracle. UbigintModPow (SMIMERSAES.ts:124)
branches on secret exponent bits; no blinding; the envelope-decrypt path is a classic Bleichenbacher
shape. Errors are uniform ("Decryption error") which helps, and a local mail client has no adaptive
network oracle, so practical risk is low — but note it as an inherent limitation of hand-rolled RSA.
3. [Medium] keyStatus() trust upgrade never fires for normal certs. SMIMEPublicKey.ts:119 upgrades
trustLevel only if (this.trustLevel == TrustLevel.Distrusted), but imported/received certs default to
TrustLevel.Sender. So a cert that validly chains to a bundled/system CA stays "Sender" (yellow) in the
message UI even though keyStatus() returns Valid. Looks like a logic bug — likely intended
trustOrder(current) < trustOrder(new). 
4. [Low/Info] No basicConstraints/keyUsage checks; chain not auto-validated on read. The signature model
itself is sound — email.signed records the embedded cert's key-id and the UI only shows "signed" when
the sender's stored key (getPublicKeyByKeyID, scoped to email.from) has a matching id, so spoofing needs
a 64-bit modulus collision (infeasible). But verifySignature/keyStatus never check CA:TRUE or 
emailProtection, so if chain-building is later trusted for identity, any leaf could sign for another.
5. [Info] Private-key passphrase stored in plaintext JSON. SMIMEPrivateKey.toJSON writes both the
PBES2-encrypted key and its passphrase in the clear (DB is the security boundary); createNewPrivateKey
uses 2048 PBKDF2 iterations (fine only because the passphrase is a random UUID). Probably intentional —
flagging for awareness.

Correctness

6. [Medium — real bug] toBase64 fallback overflows the stack on large messages. SMIMESend.ts:172/188 and
the PEM encoder fall back to btoa(String.fromCharCode(...buf)). For an enveloped email with attachments
(der = hundreds of KB–MB), spreading every byte as an argument throws RangeError: Maximum call stack 
size exceeded. Native Uint8Array.prototype.toBase64 avoids it, but the fallback needs a chunked loop.
7. [Low] _decodeInt throws uncaught RangeError on a zero-length integer body (der.ts:458, getInt8(0)),
escaping the reporter path — only on malformed DER.
8. [Low, latent] new DataView(raw.buffer) (der.ts:338/458) ignores byteOffset/length; correct only
because raw() copies via .slice(). Any future switch to .subarray() silently corrupts int/bmpstr
decoding — pass explicit offset/length.

Style / conventions / minor

9. [Easy, do before merge] Stray debug console.log committed at Encryption.svelte:72 ($: 
console.log("override", …)), fires every reactive tick.
10. [Medium] Untranslated, dev-style error strings in both EncryptionKey.svelte {:catch} blocks
(Something went wrong: {error.message}, Unexpected private key status) — should use $t\`/showError`.
11. [Low] onCreateNewSMIME lacks the !identity.isCatchAll guard that onCreateNewPGP has (Settings
EncryptionImport.svelte).
12. [Low] Unused NotReached import (SMIMESend.ts:9); implicit-any params
(onRemoveFromChain(certificate), getCACertificatesLazy(type)).
13. [Low] Interop nits: micalg=sha256 (RFC 5751 wants sha-256); header splitting via /\r\n\b/ is fragile
for folded headers; PKCS#8-only private keys (per PR description).
14. [Info] No tests. A from-scratch ASN.1/RSA/CMS stack ships with zero unit tests — round-trip +
OpenSSL known-answer vectors would materially de-risk it before un-drafting.

Verified non-issues: ASN.1 integer sign/two's-complement/leading-zero, length, and tag encoding are
faithful ports; time decode/encode units are consistently ms; the reactive keyStatusAsync +
manual-reassign pattern is intentional and safe; backend.ts getCACertificates is a legitimate thin
Node-only wrapper (not an app-logic violation); getRandomValues().filter(Boolean) padding loop is
correct.

The applied changes are in the worktree at .claude/worktrees/pr-1196-smime (branch
worktree-pr-1196-smime) if you want to inspect or build on the resolution. Nothing is pushed.

result: Applied PR #1196 (S/MIME) onto master — resolved 4 enums-relocation merge conflicts, typecheck
clean — and reviewed it: crypto construction is sound, but flagged a real large-message toBase64
stack-overflow, a keyStatus() trust-upgrade guard that never fires for default-Sender certs, lenient
PKCS#1v1.5 signature unpadding, a stray debug console.log, untranslated catch-block errors, and absence
of tests.

benbucksch:
@NeilRashbrook I've merged it, because it seems solid, but please address my review comments, and Claude Fable's.

benbucksch:
If you have fully addressed a review comment of mine, please close it as resolved. If you have not, please explain clearly why you did not and leave it unresolved.

benbucksch:
@NeilRashbrook I've had Fable fix its own review comment. git commit 87be0cb79 . Can you please:
* Review that commit 87be0cb79 . Feel free to adjust the code and to make the code comments much shorter (50% or more)
* Go through Fable's review above, and address any comments that are not yet addressed and that are significant. I leave judgement to you to freely ignore some smaller review comments.
* Address my review comments, if any are still outstanding.
* Make a new PR with these changes

app/logic/Mail/Encryption/SMIME/SMIMERSAES.ts
NeilRashbrook: Copyright and licence added in 5671f20.

app/frontend/Settings/Mail/Account/EncryptionKey.svelte
NeilRashbrook: Use of `getCACertificates` added in d3e1aa1.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
NeilRashbrook: I don't like the way Postal MIME messes with the MIME structure, moving body parts into its attachment array.

NeilRashbrook:
Review of 87be0cb:

> Security
> 
> 1. Fixed.
> 2. Commented. No further action needed.
> 3. Fixed.
> 4. Commented. No action needed right now.
> 5. I don't remember where I got the suggestion to use 2048 PBKDF2 iterations. No action needed right now.
> 
> Correctness
> 
> 6. At the time, the Electron version did not have native `Uint8Array.prototype.toBase64` but I wrote the code to take advantage of any upgrade you might make. This has happened on the Electron side, but I don't know what version of Node you run for the tests (needs Node 25 to be able to remove the fallback code).
> 7. Fixed. The original asn.1 library incorrectly treats a malformed DER as a value of zero.
> 8. Fixed.
> 
> Style / conventions / minor
> 
> 9. Fixed. Wait, this bug wasn't even in this PR, it was in 64f4367...
> 10. Fixed. Did it actually use ErrorMessageInline by itself? It hadn't mentioned it before.
> 11. Fixed.
> 12. Fixed.
> 13. Fixed. (No attempt to address PKCS#8-only private keys.)
> 14. Fixed, except the test for 6. may no longer be necessary.

benbucksch:
> Did it actually use ErrorMessageInline by itself? It hadn't mentioned it before.

Yes. It tries to adapt to our codebase.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
benbucksch: I agree. If you have a better MIME parser in TypeScript (or JavaScript), please let me know.

We still need the high-level logic of "this is the HTML/text body" and "these are attachments", but there are places like this one where we need access to the real MIME structure, and not having that is a real problem. Either we find a better lib, or we add the feature to this lib.

NeilRashbrook:
Any news on the minimum Node version the tests need to support?

PR #1204, opened 2026-05-14
EWS, OWA: Subaccounts: Make double startups less likely

NeilRashbrook:
As commented in PR #1115, there is a login bug where subaccounts can try to startup twice. Additionally, there are some problems with the UI for subaccounts:
- the UI does not always show the correct login status on startup
- the UI does not update the login status after login/logout

While I was here I fixed a regression from PR #1115 which prevented logging out from the UI for an EWS subaccount.

### DIFF
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -277,6 +277,7 @@ export function setMainAccounts(): void {
     let mainID = account._mainAccountID;
     if (mainID && !account.mainAccount) {
       account.mainAccount = accounts.find(acc => acc.id == mainID);
+      account.notifyObservers();
     }
   }
 }
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -36,6 +36,7 @@ export class Account extends Observable {
   @notifyChangedProperty
   realname: string = appGlobal.me?.name;
   /** @see `mainAccount` */
+  @notifyChangedProperty
   _mainAccount: Account | null = null;
   /** Internal. Used only during load. */
   _mainAccountID: string | null = null;
@@ -277,7 +278,6 @@ export function setMainAccounts(): void {
     let mainID = account._mainAccountID;
     if (mainID && !account.mainAccount) {
       account.mainAccount = accounts.find(acc => acc.id == mainID);
-      account.notifyObservers();
     }
   }
 }
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -139,9 +139,7 @@ export class ActiveSyncAccount extends MailAccount {
 
   notifyObserversOfSubaccounts() {
     for (let account of this.dependentAccounts()) {
-      if (account.mainAccount == this) {
-        account.notifyObservers();
-      }
+      account.notifyObservers();
     }
   }
 
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -102,9 +102,9 @@ export class ActiveSyncAccount extends MailAccount {
           throw new ActiveSyncError("Settings", response.DeviceInformation.Status, this);
         }
       }
-    });
 
-    await this.startup();
+      await this.startup();
+    });
   }
 
   async startup() {
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -85,7 +85,10 @@ export class ActiveSyncAccount extends MailAccount {
       if (this.authMethod == AuthMethod.OAuth2) {
         this.oAuth2 ??= getOAuth2BuiltIn(this);
         assert(this.oAuth2, gt`Could not find OAuth2 config for ${this.hostname}`);
-        this.oAuth2.subscribe(() => this.notifyObservers());
+        this.oAuth2.subscribe(() => {
+          this.notifyObservers();
+          this.notifyObserversOfSubaccounts();
+        });
         await this.oAuth2.login(interactive);
       }
       this.protocolVersion = this.getStorageItem("protocolVersion");
@@ -131,6 +134,14 @@ export class ActiveSyncAccount extends MailAccount {
     let galAB = appGlobal.searchOnlyAddressbooks.find(ab => ab.mainAccount == this);
     if (galAB) {
       appGlobal.searchOnlyAddressbooks.remove(galAB);
+    }
+  }
+
+  notifyObserversOfSubaccounts() {
+    for (let account of this.dependentAccounts()) {
+      if (account.mainAccount == this) {
+        account.notifyObservers();
+      }
     }
   }
 
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -129,7 +129,7 @@ export class EWSAccount extends MailAccount implements EWSSubscribable {
   }
 
   async logout(): Promise<void> {
-    if (this.mainAccount) {
+    if (this.isDependentAccount) {
       await this.mainAccount.logout();
       return;
     }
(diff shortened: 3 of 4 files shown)


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -138,6 +146,15 @@ export class EWSAccount extends MailAccount implements EWSSubscribable {
     }
   }
 
+  notifyObserversIncludingSubaccounts() {
+    this.notifyObservers();
+    for (let account of appGlobal.emailAccounts) {
```
benbucksch: I think we have a function to list subaccounts

app/logic/Abstract/Account.ts
```diff
@@ -277,6 +277,7 @@ export function setMainAccounts(): void {
     let mainID = account._mainAccountID;
     if (mainID && !account.mainAccount) {
       account.mainAccount = accounts.find(acc => acc.id == mainID);
+      account.notifyObservers();
```
benbucksch: is mainAccount not an observed property?

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -138,6 +146,15 @@ export class EWSAccount extends MailAccount implements EWSSubscribable {
     }
   }
 
+  notifyObserversIncludingSubaccounts() {
```
benbucksch: Make it "ofSubaccounts()" to match OWA, and then call notifyObservers() directly from the caller.

benbucksch (changes requested):
minor nits

app/logic/Abstract/Account.ts
NeilRashbrook: Not at the moment. I'm assuming you wanted me to change that.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: I wasn't sure whether you wanted me to put this in a common file at this point...

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: I was only considering listing email accounts; I've change this to all subaccounts but then I've done this for ActiveSync too for consistency.

app/logic/Abstract/Account.ts
benbucksch: Yes, it would make sense to trigger the notification from the object itself. That's the pattern used in general.

app/logic/Mail/EWS/EWSAccount.ts
benbucksch: Not yet. If this happens for more situations, then maybe.

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
```diff
@@ -134,6 +137,14 @@ export class ActiveSyncAccount extends MailAccount {
     }
   }
 
+  notifyObserversOfSubaccounts() {
+    for (let account of this.dependentAccounts()) {
+      if (account.mainAccount == this) {
```
benbucksch: Why this condition? Thhat's what dependentAccounts() is for, by definition, no?

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -125,6 +128,14 @@ export class EWSAccount extends MailAccount implements EWSSubscribable {
     });
   }
 
+  async logout(): Promise<void> {
+    if (this.mainAccount) {
```
```suggestion
    if (this.isDependentAccount) {
```

Please never use `if (this.mainAccount)` (it reads in English: "If this is the main account", but it means the exact opposite, so it's confusing), but always `if (this.isDependentAccount)`. That's why we have the property.

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -138,6 +149,14 @@ export class EWSAccount extends MailAccount implements EWSSubscribable {
     }
   }
 
+  notifyObserversOfSubaccounts() {
+    for (let account of this.dependentAccounts()) {
+      if (account.mainAccount == this) {
```
benbucksch: ditto

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -234,6 +236,14 @@ export class OWAAccount extends MailAccount {
     }
   }
 
+  notifyObserversOfSubaccounts() {
+    for (let account of this.dependentAccounts()) {
+      if (account.mainAccount == this) {
```
benbucksch: ditto

benbucksch (changes requested):

benbucksch:
Can you please be more specific in your PR description?

> UI is not always correct

What exactly is not correct?

> UI does not update

What exactly does not update?

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: It's a good thing there isn't an `isMainAccount` property 😉

app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
NeilRashbrook: D'oh!

NeilRashbrook (commented):
> Can you please be more specific in your PR description?

Sorry, yes, I should have specifically referred to the login status. I've edited the description now.

benbucksch:
E.g. "If the main account was logged out at start up and the user logs into the dependent account first, we automatically log into the main account, but the UI still shows the main account as logged out"

Can you please adapt this sentence to make it correct?

Then make a similar description for the second problem

app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -125,6 +128,14 @@ export class EWSAccount extends MailAccount implements EWSSubscribable {
     });
   }
 
+  async logout(): Promise<void> {
+    if (this.isDependentAccount) {
+      await this.mainAccount.logout();
+      return;
```
benbucksch, line 134: Is the "return" correct? Do we have no cleanup to do for this account specifically?

app/logic/Mail/OWA/OWAAccount.ts
```diff
@@ -216,11 +217,12 @@ export class OWAAccount extends MailAccount {
   }
 
   async logout(): Promise<void> {
-    if (this.mainAccount) { // TODO Why?
+    if (this.isDependentAccount) {
       await this.mainAccount.logout();
       return;
     }
     this.hasLoggedIn = false;
```
benbucksch, line 224: Same here. Are you sure that we don't need to set these properties?

benbucksch (approved):
Looks good to me, apart from the question here.

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: Any cleanup has to be done by the main account, in case it's the one whose logout was requested.

app/logic/Mail/OWA/OWAAccount.ts
NeilRashbrook: (We never set `hasLoggedIn` on a dependent account.)

NeilRashbrook (commented):
Sorry, I find it very hard to elaborate on an incorrect Boolean value.

benbucksch (approved):

benbucksch:
> Sorry, I find it very hard to elaborate on an incorrect Boolean value.

Hey, I went to all the length of giving you a concrete example.

What's missing is:
* The exact situation when the bug happens.
  * You: "not always" ... "on startup"
  * Better: "If the main account was logged out at start up and the user logs into the dependent account first"
* What is the wrong result and why, and what is the expected result
  * You: "not ... the correct login status"
  * Better: "we automatically log into the main account, but the UI still shows the main account as logged out"

I hope you would agree that "If the main account was logged out at start up and the user logs into the dependent account first" is important info to capture and more helpful than "not always".

NeilRashbrook:
>  * Better: "If the main account was logged out at start up and the user logs into the dependent account first"
>  * Better: "we automatically log into the main account, but the UI still shows the main account as logged out"

Well that does at least prove that what I wrote wasn't helpful because you completely misunderstood it...

The first part happens when dependent accounts are added to the UI; they don't have their dependencies set up yet so the login status is garbage at this point because the dependent account isn't looking at the correct data yet. Sometimes this garbage value happens to be or become correct, but that's dependent on a number of factors that would be tedious to enumerate. At the time I didn't know the cause of the problem, so I couldn't be more specific.

I still don't know of a clearer way to explain that you should expect the UI to update after performing whichever of a number of actions that updates the backend value that the UI is supposed to be representing, so I'm sticking with "the [subaccount] UI does not update the login status after login/logout".

In all cases there is only one correct status (UI matches backend) and one incorrect status (UI does not match backend), given that the login status is a Boolean value.

Issue #1200, opened 2026-05-11
Calendar: Cannot edit event anymore


benbucksch:
### Reproduction
* Create calendar event in your personal calendar
* Close it

a)
1. Double-click on it in the month view

b) 
1. Double-click on it in the day view

c)
1. Click on it, to show it in the sidebar
2. In the sidebar, click the [Pencil] button to edit it
3. add a description using [Description (+)]
4. add some text in the description
5. try to save

### Actual result
* a) and b): The Edit event screen opens with a *new* event
* c) The Edit event screen opens with the correct event. But in step c) 5., there is no [Save] button, only the (x) close button

benbucksch:
c) is not always reproducible. It worked for me after I switched apps to mail and came back to calendar.

jermy-c:
> 1. Double-click on it in the month view

Same problem with event propagation we had for #1197. The outer function in MonthView addEvent() triggers because the double click is propagated there.

NeilRashbrook:
The `DayCell` is seeing the `dblclick` event, causing it to start editing a new event; before the Svelte 5 upgrade, only the `EventContent` would see the `dblclick` event.

benbucksch:
@NeilRashbrook Thank you. Jeremy is fixing it right now.

benbucksch:
@jermy-c I merged the PR, but I can still reproduced the bugs a) and c) above. Both open a new event instead of editing the existing event.

jermy-c:
> [@jermy-c](https://github.com/jermy-c) I merged the PR, but I can still reproduced the bugs a) and c) above. Both open a new event instead of editing the existing event.

Because the components in Events don’t use the Buttons we need to add the `event.stopPropagation()` to the Calendar components.

PR #1205, opened 2026-05-15
UI: Clickable: Wrap all elements that don't need on:click or on:dblclick to be propagated #1200

jermy-c:
- Wrap all elements that don't need `on:click` or `on:dblclick` to be propagated
- Remove the `catchErrors()` wrapping for handler functions
- Remove `event.stopPropagation()` and `event.preventDefault()` that were in handler functions
- Fixed: `<Clickable>` by using `display: contents` which removes the box model or sizing from the element but uses the child box model or size from its children
- `ListViewLine` needs the `on:click` to be propagated for the `<FastList>` to highlight the selected item

### DIFF
--- a/app/frontend/Calendar/DayView/EventContent.svelte
+++ b/app/frontend/Calendar/DayView/EventContent.svelte
@@ -1,23 +1,23 @@
-<hbox class="event font-small" flex
-  on:click
-  on:click={onSelect}
-  on:dblclick={onOpen}
-  title={eventAsText}>
-  {#if showTime}
-    <!--{event.startTime.toLocaleTimeString(getDateTimeLocale(), { hour: "numeric", minute: "numeric" })}-->
-    <hbox class="time">{startTime}</hbox>
-  {:else if forceShowText}
-    <hbox class="time">…</hbox>
-  {/if}
-  {#if showTitle}
-    <hbox class="title">{$event.title}</hbox>
-  {/if}
-</hbox>
+<Clickable onClick={onSelect} onDoubleClick={onSelect}>
+  <hbox class="event font-small" flex
+    title={eventAsText}>
+    {#if showTime}
+      <!--{event.startTime.toLocaleTimeString(getDateTimeLocale(), { hour: "numeric", minute: "numeric" })}-->
+      <hbox class="time">{startTime}</hbox>
+    {:else if forceShowText}
+      <hbox class="time">…</hbox>
+    {/if}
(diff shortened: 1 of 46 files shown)


app/frontend/Calendar/DayView/EventContent.svelte
```diff
@@ -1,23 +1,23 @@
-<hbox class="event font-small" flex
-  on:click
```
benbucksch, line 2: Did you check why we wanted this to bubble?

app/frontend/Contacts/PersonPage/ContactEntryUI.svelte
```diff
@@ -45,17 +45,23 @@
   </hbox>
 {:else}
   {#if protocolLabels}
-    <hbox class="protocol display" on:click={startEditing}>
-      {displayProtocol(entry.protocol)}
-    </hbox>
+    <Clickable onClick={startEditing}>
```
benbucksch: You can move the `<Clickable>` outside `{#if protocolLabels}`, to avoid duplication.

benbucksch (changes requested):

benbucksch:
Did you test *all* instances changed here, both for the visual layout, and for the click handler functionality, whether it looks and works as before? We want to avoid regressions.

app/frontend/Contacts/PersonPage/ContactEntryUI.svelte
jermy-c: Change made in f573cebc3685f0f8be239503bd4956c4d78ac7f6

app/frontend/Calendar/DayView/EventContent.svelte
jermy-c: Yes, I did but I couldn't find a reason, it just worked as before. And this component is not used anywhere else other than DayView. My only guess was that before it was for a FastList slot and now it is in a <Scroll>

benbucksch: This component is also used in SML.

Reproduction
1. New email
2. Click on SML button (second to left on the formamtting toolbar, left of attachments button)
3. Click add [Meeting time poll] or [Book me] (try both)
4. In the calendar view, click on times to propose them (this is where the click handler might matter)
5. Send to yourself
6. In the received email, in the calendar view, click on meeting times to accept/reject proposals (this is where the click handler might matter)

PR #1209, opened 2026-05-20
Mail: SearchPane: Don't listen to changes from $search because it causes an infinitive loop - #1203

### LINKED ISSUE #1203: Mail: Search pane flashing
### Reproduction

1. Go to Mail
2. Go to the Search Tab
3. Don't input anything in the search field

### Actual Result

- Switches between 0 results and some results
- Flashes

### Expected Result

- No flashing

### Possible Cause

**Theory 1**

> Consequently, there's no such thing as a "before update"/"after update" hook.

https://svelte.dev/docs/svelte/lifecycle-hooks

- In Svelte 4, before the component is mounted the function runs the first time and then the subsequent run within the given period is skipped
- In Svelte 5, the function only runs after the function is mounted, and since the there's an update it will cause it to run again

**Theory 2**

> Everything in-between — when certain state is updated — is not related to the component as a whole; only the parts that need to react to the state change are notified. This is because under the hood the smallest unit of change is actually not a component, it's the (render) effects that the component sets up upon component initialization.

https://svelte.dev/docs/svelte/lifecycle-hooks

- In Svelte 4 all reactive changes are based on the component
- In Svelte 5 all reactive changes are based signals which are independent which might be why the debounce doesn't when to skip a call

### Problem Code
https://github.com/mustang-im/mustang/blob/dfcebea9223191308f34085f23d220b2e766ec30/app/frontend/Mail/Search/SearchPane.svelte#L77-L78

jermy-c:
- SearchPane: Don't listen to changes from $search because it causes an infinitive loop

When `startSearch()` it sets the `bodyText` property which causes an update and triggers another update in the UI

https://github.com/mustang-im/mustang/blob/544eca4e8fef48c962fd2e463c6cf586053bbe8c/app/frontend/Mail/Search/SearchPane.svelte#L89

### DIFF
--- a/app/frontend/Mail/Search/SearchPane.svelte
+++ b/app/frontend/Mail/Search/SearchPane.svelte
@@ -74,7 +74,7 @@
   let tags = search.tags;
   let attachmentTypes = search.hasAttachmentMIMETypes;
 
-  $: isOpen && $globalSearchTerm, $search, $tags, $attachmentTypes, startSearchDebounced();
+  $: isOpen && $globalSearchTerm, $tags, $attachmentTypes, startSearchDebounced();
   const startSearchDebounced = debounce(() => startSearch(), 300);
   async function startSearch() {
     try {


benbucksch:
Yes, but this will break the search, because we won't notice when the user changes the search fields (anything other than the text).

jermy-c:
> Yes, but this will break the search, because we won't notice when the user changes the search fields (anything other than the text).

You're right, I'll see if there's another way.

PR #1210, opened 2026-05-20
Svelte 5: Don't subscribe to objects that are re-assigned which also use .filterObservable()

jermy-c:
- Svelte 5: Don't subscribe to objects that are re-assigned which also use .filterObservable()
- MailChat was the only one that froze silently but the other's didn't I did notice some lag though. But I'm not sure if there's some other bug causing it.

### DIFF
--- a/app/frontend/Mail/MailChat/MailChat.svelte
+++ b/app/frontend/Mail/MailChat/MailChat.svelte
@@ -13,7 +13,7 @@
     <ViewSwitcher />
   </vbox>
   <vbox class="right-pane" slot="right">
-    {#if $filteredMessages.hasItems && selectedPerson }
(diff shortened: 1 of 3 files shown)


benbucksch (approved):
Did you test whether the observers break (miss changes)?

benbucksch:
"Don't to"? (in description here)

jermy-c:
> Did you test whether the observers break (miss changes)?

I tested MailChat and InMeeting and checked the child components if they had `$` but it's just the WebAppShopButton I couldn't test.

benbucksch:
> Just the WebAppShopButton I couldn't test.

Go to [Apps] app, click [+] at the bottom, select an app, and click (+) there.

PR #1215, opened 2026-05-21
Build: Upgrade GitHub actions to new versions with node:24 - #1213

### LINKED ISSUE #1213: Build: Upgrade github actions to new versions with node:24
Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/setup-java@v4, actions/setup-node@v3, actions/upload-artifact@v4, gradle/actions/setup-gradle@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Note: This is not about the node:24 that *we* set, but the node:20 that the predefined actions set.

@jermy-c Can you please update all github actions (whether mentioned above or not) to the latest version?

jermy-c:
- Build: Upgrade GitHub actions to new versions with node:24

### DIFF
--- a/.github/workflows/auto-translate.yml
+++ b/.github/workflows/auto-translate.yml
@@ -10,7 +10,7 @@ jobs:
       contents: write
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - name: Auto Translate
       uses: actions/setup-node@v3
       with:
--- a/.github/workflows/auto-translate.yml
+++ b/.github/workflows/auto-translate.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - name: Auto Translate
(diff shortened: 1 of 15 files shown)


benbucksch (approved):

benbucksch:
Thanks

PR #1216, opened 2026-05-21
Mobile: Checkbox: Fix styling - #1199

### LINKED ISSUE #1199: Mobile: Svelte 5 broke some styles
### Environment
* Mobile (Android)
* master 0.9.28-dev

### Reproduction
Compare with 0.9.27:
1. Look at app bar at bottom
2. Go to Mail Search (via app menu)

### Actual result
1. App bar at bottom has black text color.
2. Toggles are too huge. When enabled, they look like eyes 👀 

### Expected result
1. App bar at bottom has white text color.
2. Toggles look like toggles

### Implementation
* There are likely other problems across the app. I checked only superficially.
* Some of the style rules must have broken and no longer apply. Find out why and fix it. Make sure that all CSS applies as before.
* This is a regression from Svelte 5 #990

jermy-c:
- The Toggle checkbox seems to be broken even in Svelte 4 so I could only fix the component itself
- There was also a floating icon because of the `||` condition
- However since the toggle is bigger on mobile it makes it appear as if there's no space around the checkbox
image

### DIFF
--- a/app/frontend/Shared/Checkbox.svelte
+++ b/app/frontend/Shared/Checkbox.svelte
@@ -34,7 +34,7 @@
     {/if}
   {/if}
 
-  {#if !toggle && allowIndetermined}
+  {#if !toggle || allowIndetermined}
     <hbox class="state-icon-wrapper">
       {#if checked === true}
         <CheckIcon strokeWidth={5} size={12} />
--- a/app/frontend/Shared/Checkbox.svelte
+++ b/app/frontend/Shared/Checkbox.svelte
@@ -183,15 +183,16 @@
       inset 0 1px 2px rgba(255, 255, 255, 0.2);
   }
   :global(.mobile) .checkbox {
-    min-height: 44px;
+    min-height: 32px;
   }
   :global(.mobile) .knob {
-    top: 14px;
+    height: 32px;
+    width: 32px;
   }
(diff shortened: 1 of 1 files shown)


app/frontend/Shared/Checkbox.svelte
```diff
@@ -34,7 +34,7 @@
     {/if}
   {/if}
 
-  {#if !toggle || allowIndetermined}
+  {#if !toggle && allowIndetermined}
```
benbucksch:
```suggestion
  {#if !toggle || allowIndetermined}
```

Why this change? This looks wrong. See that there's a check for `allowIndermined` within this `if`, which would make no sense, if this is always true.

benbucksch:
Why do you make them smaller? On mobile, widgets need to be at least 44px to be tappable with the finger.

app/frontend/Shared/Checkbox.svelte
jermy-c: image

Because without this change it would always show a floating icon as seen in the screenshot.

jermy-c:
> Why do you make them smaller? On mobile, widgets need to be at least 44px to be tappable with the finger.

image

It looks too big compared to the labels on the side at 44px.

PR #1217, opened 2026-05-23
Upgrade `@tsconfig/svelte` and `svelte-check`

NeilRashbrook:
Currently `yarn install` gives the following warnings because of a mismatch between version 2.10.3 of `svelte-check` and `svelte`. Two of these warnings also apply to version 3.5.0 of `svelte-check`:
- warning svelte-check > svelte-preprocess > magic-string > sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
- warning svelte-check > svelte-preprocess > sorcery > sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
- warning svelte-check > svelte-preprocess > @types/sass@1.45.0: This is a stub types definition. sass provides its own type definitions, so you do not need this installed.
- warning svelte-check > svelte-preprocess > sorcery > sander > rimraf@2.7.1: Rimraf versions prior to v4 are no longer supported [both 2.10.3 and 3.5.0]
- warning svelte-check > svelte-preprocess > sorcery > sander > rimraf > glob@7.2.3: Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me [both 2.10.3 and 3.5.0]
- warning svelte-check > svelte-preprocess > sorcery > sander > rimraf > glob > inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
- warning " > svelte-check@2.10.3" has incorrect peer dependency "svelte@^3.24.0".
- warning "svelte-check > svelte-preprocess@4.10.7" has incorrect peer dependency "svelte@^3.23.0".

`@tsconfig/svelte` version 3.0.8 was presumably designed for TypeScript 3.0.8; it does work with TypeScript 4 and early versions of TypeScript with a compatibility flag but not with TypeScript 5.9.3 which is the version in `package.json`.

Note that `@tsconfig/svelte` changed the setting of `strict` from `false` to `true` so we need to change it back again or get even more error messages than the almost two hundred we currently get.

For me, `tsc` also complains about the duplication of `logic/build.ts` in both `tsconfig.json` and `tsconfig.node.json`. Removing it from `tsconfig.node.json` seems to make it happiest.


NeilRashbrook:
(When I said error messages, I meant error message lines. Some errors take up multiple lines.)

As I mentioned in PR #1218 I checked the new error list.

The new thing TypeScript now likes complaining about is the difference between an regular `Uint8Array` and a shared `Uint8Array` which some functions won't accept. This is typically fixed by changing `Uint8Array` to `Uint8Array<ArrayBuffer>` in all relevant type declarations.

I don't know whether this is new with this change but I also noticed that TypeScript doesn't seem to like the construct `import type OpenPGP from "openpgp";`; instead it wants `import type * as OpenPGP from "openpgp";
`. (Similarly in `netUtil.ts`.)

benbucksch (approved):

PR #1230, opened 2026-06-05
Calendar: DurationSelector, DurationUnit: Use function bindings #1229

### LINKED ISSUE #1229: Calendar: Edit event: Cannot change duration
### Reproduction
1. Create a new calendar event
2. Enter a time 8:00
3. Use the duration dropdown to change the duration to 2 hours or 30 minutes

### Actual result
* The duration resets to 1 hour

### Expected result
* The duration is saved as you selected it
* The end time adapts, e.g. 10:00 or 8:30 in the examples above

### Importance
* Obviously,  this makes the calendar unusable

### Other cases
* Regresssion. This worked a few weeks ago. Caused by Svelte 5?
* When I select 2+ days,  it switches to multi day, so the input of the dropdown is processed

jermy-c:
- Use [Function-bindings](https://svelte.dev/docs/svelte/bind#Function-bindings) to emit event before changing the value
- The on:change was removed because it causes one extra update with same value because it runs independently (1 for on:change, bind:value), in Svelte 4 updates with same values and in the same component are batched and that's other problem

What happens:

1. `adaptAllDays()` is called and the value is still the same `1hour`
2. The parent receives the event and updates the `Event` object
3. The `Event` object is updated and `Event.duration` with the old value is written back to the component, <- this is an extra update
4. `bind:value` writes back to the `Event.duration`

### DIFF
--- a/app/frontend/Calendar/EditEvent/DurationSelector.svelte
+++ b/app/frontend/Calendar/EditEvent/DurationSelector.svelte
@@ -1,6 +1,5 @@
 <select
-  on:change={event => adaptAllDays(event.target.value)}
-  bind:value={durationInSeconds}><!-- `on:change` must be before `bind:value` in the code, to get the value before the `Event.duration` setter adapts it -->
+  bind:value={getDurationInSeconds, adaptAllDays}><!-- `on:change` must be before `bind:value` in the code, to get the value before the `Event.duration` setter adapts it -->
   {#if allDay}
     <option value={1 * k1HourS}>{$plural(1, { one: 'hour', other: 'hours' })}</option>
     {#each kDayOptions as day}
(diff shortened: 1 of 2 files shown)


jermy-c:
There are 2 other components in Calendar, that seem to have a similar problem. I'll include those fixes in this PR also.

jermy-c:
DurationUnit was the only other component that seems to have the same problem. You can change to minutes and hours but not days. It would reset back to hours.

benbucksch:
> adaptAllDays() is called and the value is still the same 1hour

Actually, on:change calls already with the new value, as you can see with the `console.log("adapt", newValue, durationInSeconds);` in `adaptAllDays()`.

The problem was the implementation of the function: It unconditionally changed to `setAllDay true`. Which then put the chain in motion that you described above, starting with step 1. You analylsis was correct. You were just wrong about step 1, which changes the fix. Fixed in git commit 9f0727dba771002736f0fe0cffd21ab0243339fe  .

benbucksch:
Thanks for your analysis here. This helped to find the right fix.

Issue #1228, opened 2026-06-04
Contacts: Error when deleting contact immediately after creating it


jermy-c:
### Reproduction

1. Create a contact
2. Save the contact
3. Don't select or click on anything else
4. Click delete

### Actual Result

- The contact is not deleted
- There are 2 contacts selected when you click [+]
- Even after saving it still shows 2 contacts selected

### Expected Result

- The selected contact is deleted
- The selections is showing the correct contact selected
- Only one is selected

### Stack

```
TypeError: Cannot read properties of null (reading 'deleteIt')
    at deleteIt (PersonMenu.svelte:33:3)
    at HTMLButtonElement.myOnClick (MenuItem.svelte:80:7)
```

### Related

I saw another bug where the contact is deleted but the wrong contact is deleted. I don't have a reliable reproduction for this one yet, however.

jermy-c:
Findings so far:

- In `PersonMenu.svelte` it receives the correct person
- It seems like it's because we don't update the collection I see `.save()` in `Contact.ts` and `Person.ts` but none of them add or remove from the address book. I did a test and it stops the error but it doesn't close `PersonDetails` afterwards and the person is still displayed there after deleting.

jermy-c:
### Update

#### [All] Addressbooks
- It's the personList not updating correctly, might be the `mergeColls`

#### Specific Addressbooks

- The +1 index or the last item which might be the current one being deleted
https://github.com/mustang-im/mustang/blob/ca25d8c64ea885932a76a13b63aff7d890058015/app/frontend/Contacts/PersonPage/PersonMenu.svelte#L25

- After that is selected, the `person` is updated because `$selectedPerson` is updated
- Now we delete the latest `$selectedPerson` which is the wrong one because it's index +1
https://github.com/mustang-im/mustang/blob/ca25d8c64ea885932a76a13b63aff7d890058015/app/frontend/Contacts/PersonPage/PersonMenu.svelte#L31

jermy-c:
There are 3 problems here:

1. The wrong contact being selected and deleted: Fix #1232 
2. The PersonList for [All] not being updated when something is deleted but added it is and on mobile it happens for all Addressbooks: Fix #1233 
3. There are multiple selected Persons when you click the [+]

PR #1232, opened 2026-06-08
Contacts: PersonMenu: Select the correct person during and after delete #1228

jermy-c:
- The person variable changed while the delete was not complete
- The fix is to save the reference to the person we're going to delete and use it throughout
- The other problem was that next might be the person we're going to delete so we should not select that as next
- On mobile we need to navigate back to main Contacts screen if there's no other contacts

### DIFF
--- a/app/frontend/Contacts/PersonPage/PersonMenu.svelte
+++ b/app/frontend/Contacts/PersonPage/PersonMenu.svelte
@@ -21,13 +21,15 @@
   export let person: Person;
 
   async function deleteIt() {
+    let toDelete = person;
     if (person == $selectedPerson) {
-      $selectedPerson = getNext(person.addressbook?.persons, person);
+      let next = getNext(person.addressbook?.persons, person);
+      $selectedPerson = next === toDelete ? null : next;
       if (appGlobal.isMobile) {
         goTo(URLPart`/contacts/person/${person.id}/edit`, { person });
       }
     }
 
-    await person.deleteIt();
+    await toDelete.deleteIt();
   }
 </script>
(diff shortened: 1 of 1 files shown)


app/frontend/Contacts/PersonPage/PersonMenu.svelte
```diff
@@ -21,13 +21,19 @@
   export let person: Person;
 
   async function deleteIt() {
+    let toDelete = person;
     if (person == $selectedPerson) {
-      $selectedPerson = getNext(person.addressbook?.persons, person);
+      let next = getNext(person.addressbook?.persons, person);
+      $selectedPerson = next === toDelete ? null : next;
```
benbucksch:
```suggestion
      $selectedPerson = getNext(person.addressbook?.persons, toDelete);
```

Did you have a case where `getNext()` returned the same person? It should not. If so, that's a bug in `getNext()`.

benbucksch (changes requested):
OK with that one minor change

app/frontend/Contacts/PersonPage/PersonMenu.svelte
jermy-c: Yes, when the person is the last item on the list and this happens very often since it's the newest addition to the collection.

benbucksch: In this case, return the first item in the list.

jermy-c: Fix made in #1235

benbucksch (approved):

benbucksch:
Thanks. Good fix.

Can you please check the source code in other places where we might have the same problem? Start with delete actions. But maybe also for others.

PR #1233, opened 2026-06-08
Contacts: Person: Add to Addressbook if not present #1228

jermy-c:
- The person was only being added to the persons prop which updated the PersonsList on add
- However, onDelete we remove directly from the address book and since it's not there it doesn't update the UI


benbucksch (approved):

benbucksch:
@jermy-c This caused a regression that we are adding Persons twice.

https://github.com/mustang-im/mustang/commit/f3eee72364afd131d14943590779155271b12857#commitcomment-191041808

jermy-c:
Sorry about that.

PR #1235, opened 2026-06-09
Collections: getNext() try all options first and make sure it's not the same object #1232

jermy-c:
-  `getNext()`: Try all options first and make sure it's not the same object before null is returned
- We need to find the first option that's that same object
- list.last ?? list.first and then the check if the it is same object would return null even though there's a first object. Because list.last is not null and the next check sets it to null because it's the same object.

### DIFF
--- a/app/logic/util/collections.ts
+++ b/app/logic/util/collections.ts
@@ -34,10 +34,9 @@ export function getNext<T>(list: ArrayColl<T>, curObject: T): T | null {
     return null;
   }
   let position = list.indexOf(curObject);
-  let next = list.getIndex(position + 1)
-    ?? list.last
-    ?? list.first;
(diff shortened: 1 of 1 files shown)


NeilRashbrook:
If the item at `position + 1` doesn't exist doesn't that mean that this already is the last item so there's probably no point going on to check the last item? (But do check the first item in case the list only has one item. The other alternative would have been to try `position + 1` and `position - 1` as neither should be the same item.)

benbucksch:
I'm happy for Neil to do the review here.

jermy-c:
Thanks for the review Neil! That's works even better because now the next selected item doesn't jump to the bottom in the UI.

NeilRashbrook (approved):

Issue #1236, opened 2026-06-09
EWS: Contacts: Zimbra: Missing DistinguishedFolderId for contacts causes error "Cannot read properties of undefined (reading 'ExtendedProperty')"


benbucksch:
### Reproduction
* Set up email account on Zimbra using EWS

### Actual result
* Immediately on first login, error:
```
TypeError: Cannot read properties of undefined (reading 'ExtendedProperty')
    at EWSAccount.createAddressbookAccount (EWSAccount.ts:909:19)
    at EWSAccount.listFolders (EWSAccount.ts:771:30)
    at async EWSAccount.startup (MailAccount.ts:45:5)
    at async EWSAccount.ts:117:7
    at async RunOnce.runOnce (RunOnce.ts:16:14)
    at async EWSAccount.startup (EWSAccount.ts:116:5)
    at async EWSAccount.ts:111:7
    at async RunOnce.runOnce (RunOnce.ts:16:14)
    at async EWSAccount.login (EWSAccount.ts:102:5)
    at async getFirstMessages (saveConfig.ts:116:3)
```

### Expected result
* No error
* Address books are read
* Optional properties are skipped

benbucksch:
This errors:
```assert(folder.ExtendedProperty?.Value != "true", "Need visible addressbook");```
which means `folder` is null, which means the caller `listFolders()` is at fault:
```
let folder = ensureArray(result.RootFolder.Folders.ContactsFolder).find(folder =>
        folder.DistinguishedFolderId == "contacts" && folder.ExtendedProperty?.Value != "true");
```

I think I can see what the problem is: Zimbra doesn't support `DistinguishedFolderId`. Neither does Grommunio, BTW. So, we need to deal with that.

benbucksch:
BTW: I also have doubts that the `ExtendedProperty?.Value == true` meaning "visible" will hold true for other EWS implementations.

NeilRashbrook:
Fortunately `folder.PidTagAttributeHidden == "true"` doesn't mean "visible"; we're asserting that it *isn't* `"true"`...

benbucksch:
I meant not `PidTagAttributeHidden`, but this assert here:
```assert(folder.ExtendedProperty?.Value != "true", "Need visible addressbook");```
That `ExtendedProperty?.Value != true` freaks me out, due to the generic name.

But, as I said, that's not problem here. The problem is that the server doesn't support `DistinguishedFolderId`.

[...]

PR #1240, opened 2026-06-09
Fill in DistinguishedFolderId for those servers that don't support it #1236

NeilRashbrook:
(Tested locally by commenting out the code requesting DistinguishedFolderId as I don't have any current credentials for a non-standard server.)

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -715,8 +715,6 @@ export class EWSAccount extends MailAccount implements EWSSubscribable {
                 FieldURI: "folder:FolderClass",
               }, {
                 FieldURI: "folder:ParentFolderId",
-              }, {
-                FieldURI: "folder:DistinguishedFolderId",
               }],
               t$ExtendedFieldURI: {
                 PropertyTag: HiddenPidTag,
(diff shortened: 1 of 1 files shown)


benbucksch:
Sorry, I don't want us to make one extra server call for every folder, on 99% of the servers, for this fringe server. It's not worth even the amount of code.

The broken server likely won't respond to that other call, either, because it doesn't return that DistinguishedFolderID.

I think an `if (!folder!) { return; }` kind of fix is sufficient here, given that it's rare.

benbucksch:
One factor is that the Zimbra EWS is likely going EOL in October. This was more about reliability and not failing.

benbucksch:
Thanks for the effort, though! Sorry that it was not needed in this case. But I appreciate the attempt. In other cases, it will be useful.

NeilRashbrook:
> Sorry, I don't want us to make one extra server call for every folder, on 99% of the servers, for this fringe server. It's not worth even the amount of code.

It's not for every folder for every server. I don't know why you would think that. The very first step is to check whether the server does actually return a `DistinguishedFolderId`. Even if it doesn't, there's only one further server call to find the relevant folders.

> The broken server likely won't respond to that other call, either, because it doesn't return that DistinguishedFolderID.

It had better respond to that other call; even the first versions of EWS included that functionality. (Owl uses it to support Exchange 2000 for instance, as it won't let you request `DistinguishedFolderId` in search results.)

> I think an `if (!folder!) { return; }` kind of fix is sufficient here, given that it's rare.

Except it doesn't fix the other problems, such as not knowing which folder is Sent Items etc.

Issue #1241, opened 2026-06-10
Settings: "Part of account" feature makes the settings tree stop responding


SKSx:
### Reproduction

- Set up an Exchange (O365) account
- Go to Settings
- Under either "Calendar" or "Contacts", select the Exchange account
- Click on the transparent label that displays the account name beside "Part of account"
 <img width="990" height="424" alt="Image" src="https://github.com/user-attachments/assets/06904d5f-a81f-4f7d-9334-47d29b303304" />

### Actual result

- Clicking on any node in the settings tree does nothing

### Expected result

- Clicking on a node should display the settings for that node

### Related
Clicking on one of the primary tabs on the left sidebar (People, Mail, Calendar, Files, Apps) fixes the problem.

jermy-c:
While I was testing this yesterday, I encountered another bug but it seems to be only triggered by this. I don't see it anywhere else using the FastList.

```
TypeError: Failed to execute 'unobserve' on 'ResizeObserver': parameter 1 is not of type 'Element'.
    at FastList.svelte:313:54
```

And the workaround/fix was to use `.disconnect()` instead of `unobserve(listE)`.

jermy-c:
Should be fixed by #1242

PR #1242, opened 2026-06-10
Settings: Use goTo() instead of setting  directly #1241

benbucksch:
SettingsRoutes.svelte
```
    {params?.category ? $selectedCategory = params.category : null, "" }
    <SettingsApp />
```
If something changed to a settings category using `goTo()` with the `param` set to a category, and then something else changed to another settings category merely by setting `$selectedCategory = `, then the above like would overwrite `$selectedCategory` with the old value from `param`.

The fix is to always use `goTo()` with param.

### DIFF
--- a/app/frontend/Settings/Window/AccountItem.svelte
+++ b/app/frontend/Settings/Window/AccountItem.svelte
@@ -14,6 +14,7 @@
   import { MailAccount } from "../../../logic/Mail/MailAccount";
   import { accountSettings, type SettingsCategory } from "../SettingsCategory";
   import { selectedCategory, selectedAccount } from "./selected";
+  import { getSettingsCategoryForAccount, openSettingsCategoryForAccount } from "./CategoriesUtils";
   import SubCategoriesList from "./SubCategoriesList.svelte";
   import Clickable from "../../Shared/Clickable.svelte";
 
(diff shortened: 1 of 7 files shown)


app/frontend/Settings/Window/AccountItem.svelte
```diff
@@ -23,7 +24,7 @@
 
   $: accountSelected = account == $selectedAccount;
   $: itemSelected = account == $selectedAccount && $selectedCategory == mainAccountCategory;
-  $: mainAccountCategory = accountSettings.find(cat => account instanceof cat.type && cat.isMain);
+  $: mainAccountCategory = getSettingsCategoryForAccount(account);
```
jermy-c, line 27: Where is still used? Line 38 doesn't seem to use it anymore.

benbucksch: Yes, in the line above.

jermy-c: Sorry, I see it used above now.

jermy-c (approved):

benbucksch:
@jermy-c Would you mind to be so kind to test all the places touched here, tomorrow?

PR #1245, opened 2026-06-11
Calendar: open.ts: Rename to openEvent and remove extra $selectedEvent

jermy-c:
- Rename `openEventFromOtherApp` to `openEvent`
- Remove extra `$selectedEvent = event` before it


benbucksch (approved):

benbucksch:
So, this was already all fixed, correct? This is just renaming the function.

jermy-c:
> So, this was already all fixed, correct? This is just renaming the function.

Yes, it mostly. It was just Dayview/MonthView EventContent.svelte that was still setting $selectedEvent before `openEvent()`

Issue #1263, opened 2026-06-23
Contacts: Clicking [+] creates a new contact every time even if there's one already and not saved


jermy-c:
### Reproduction

1. Go to contacts
2. Click [+] to create a new contact
3. Click [+] again

### Actual Result

- Creates a new contact every time
- When a contact is created and isn't saved yet it doesn't selected that instead of creating a new one

### Expected Result

- Creates a new contact only if there's no new contact that's unsaved
- Selects the new contact if one exists instead of creating a new one

jermy-c:
This also makes 2 contacts selected.

jermy-c:
You cannot close and remove this new unsaved contact from the persons list.

benbucksch:
Fixed in PR #1267

benbucksch:
> You cannot close and remove this new unsaved contact from the persons list.

Can you fix this, too, please?

benbucksch:
Actually, I filed #1269 about that (I assume that's what you meant? If not, do whatever you found here in that bug.)

PR #1267, opened 2026-06-23
Contacts: Selected unsaved new contact if it exists #1263

jermy-c:
- Finds and selects an unsaved new contact before creating another new one

### DIFF
--- a/app/frontend/Contacts/PersonsToolbar.svelte
+++ b/app/frontend/Contacts/PersonsToolbar.svelte
@@ -49,7 +49,7 @@
     let addressbook = selectedAddressbook ?? appGlobal.addressbooks.first;
     assert(addressbook, $t`Please add an addressbook first`);
     //assert(persons instanceof ArrayColl, "Please exit the search before adding a person");
-    let person = persons.find(p => !p.dbID);
+    let person = persons.find(p => !p.dbID && !p.name);
     if (!person) {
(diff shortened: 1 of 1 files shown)


app/frontend/Contacts/PersonsToolbar.svelte
```diff
@@ -49,9 +49,12 @@
     let addressbook = selectedAddressbook ?? appGlobal.addressbooks.first;
     assert(addressbook, $t`Please add an addressbook first`);
     //assert(persons instanceof ArrayColl, "Please exit the search before adding a person");
-    let person = addressbook.newPerson();
-    person.name = "";
-    persons.add(person);
+    let person = persons.find(p => !p.dbID);
```
benbucksch:
```suggestion
    let person = persons.find(p => !p.dbID && !p.name);
```

benbucksch: How about doing that only when you didn't add a name yet?

benbucksch (approved):

app/frontend/Contacts/PersonsToolbar.svelte
jermy-c: I had that before but it might get confusing if you have typed a name but clicked somewhere else and maybe that makes you think you're just editing the contact.

Issue #1265, opened 2026-06-23
Contacts: Contact not saved to the correct address book when it is changed


jermy-c:
### Reproduction

1. Go to Contacts
2. Click [+] to create a contact
3. Change the address book of the contact
4. Fill in the name
5. Click save

### Actual Result

- Contact is not saved to the new selected address book
- When you select the saved contact it displays it is saved to the initial selected address book

### Expected Result

- Contact is saved to the new selected address book
- When you select the saved contact it displays it is saved to the new selected address book

benbucksch:
Is this fixed by #1268 ?

jermy-c:
> Is this fixed by [#1268](https://github.com/mustang-im/mustang/pull/1268) ?

Yes, it is.

PR #1268, opened 2026-06-23
Contacts: Only change addressBook for unsaved contact #1265

jermy-c:
- Kind of a regression from #1255 because we weren't getting the latest address book to change to, `$: selectedAddress` is already assigned before onChangeAddressbook and is updated after onChangeAddressbook which never happens in this case
- `person.moveToAddressbook(newAddressbook)` causes the UI to create another contact because it cannot remove an unsaved contact

### DIFF
--- a/app/frontend/Contacts/AddressbookChanger.svelte
+++ b/app/frontend/Contacts/AddressbookChanger.svelte
@@ -4,7 +4,7 @@
   filterByWorkspace={false}
   icon={selectedAddressbook?.icon ?? AddressbookIcon}
   {withLabel}
-  on:select={() => catchErrors(() => onChangeAddressbook(selectedAddressbook))}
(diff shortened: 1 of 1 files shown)


benbucksch (approved):

app/frontend/Contacts/AddressbookChanger.svelte
```diff
@@ -24,6 +24,10 @@
 
   async function onChangeAddressbook(newAddressbook: Addressbook) {
     console.log("Selected addressbook", newAddressbook?.name, "old", person?.addressbook?.name);
-    person.moveToAddressbook(newAddressbook);
+    if (!person.dbID) {
+      person.addressbook = newAddressbook;
```
NeilRashbrook, line 28: The problem here is that you'll only be able to save the new person if its addressbook is of the same type.

jermy-c: Ok, I see what you mean, I'll fix that.

Issue #1271, opened 2026-06-24
Calendar: Outgoing Invitation: When sending an invitation with 1 day length, Office365 shows it as going until 2:00 the next day


benbucksch:
Invitation sent from Stalwart, event duration 1 day. all day event (!).

<img width="444" height="439" alt="Image" src="https://github.com/user-attachments/assets/b5f4390d-8506-44fd-846d-d94428c4ed1a" />

But in the invitation sent, we show it with time:

<img width="400" height="291" alt="Image" src="https://github.com/user-attachments/assets/a6b23016-95e7-4d28-ad36-a0a15671dc75" />

Here's the source of the ICS invitation:
```
BEGIN:VCALENDAR
METHOD:REQUEST
VERSION:2.0
PRODID:-//Beonex//Parula//EN
BEGIN:VEVENT
DTSTAMP:20260624T074112Z
UID:3dffa60f-4d7e-46a7-8254-fa5dce0eb2de
SUMMARY:Going to the airport FRA
LOCATION:FRA airport
DTSTART;VALUE=DATE:20260706
DTEND;VALUE=DATE:20260707
ORGANIZER:MAILTO:elena@u
ATTENDEE;ROLE=CHAIR;PARTSTAT=ACCEPTED;CN=Elena:MAILTO:elena@u
ATTENDEE;RSVP=TRUE;CN=elena@b:MAILTO:elena@b
ATTENDEE;RSVP=TRUE;CN=mustang@k:MAILTO:mustang@k
END:VEVENT
END:VCALENDAR
```

Received via IMAP, put in local calendar: Shows as 0:00 for 1 day:

<img width="313" height="237" alt="Image" src="https://github.com/user-attachments/assets/b32509c1-6770-4a4c-a6ba-0d1ba968f093" />

Received via Office365, put in local calendar: Shows as 2:00 for 1 day:

<img width="313" height="237" alt="Image" src="https://github.com/user-attachments/assets/427bc79c-4a90-4576-9f97-008ed942cbc4" />

and spills over to next day:

<img width="221" height="163" alt="Image" src="https://github.com/user-attachments/assets/6d39f571-d1a9-4464-8147-f87e4861e904" />

benbucksch:
There are 3 bugs here:
- [x] Our event *display* shows all-day events with time. It should show only the date. The duration "for 1 day" is correct, though.
[...]

PR #1275, opened 2026-06-24
Calendar: All day event: Make Exchange also understand that there's no time - #1271

benbucksch:


NeilRashbrook:
I don't think this is relevant, the event already shows up as all-day in the outlook.cloud.microsoft UI.

benbucksch:
OK. We definitely show it wrong, but only on Exchange recipients, see #1271.

benbucksch:
Maybe it's in our "reading event from Exchange server" code?

Issue #1276, opened 2026-06-24
Calendar: Settings: Can't choose a different account to send from my personal calendar


NeilRashbrook:
Steps to reproduce problem:
- Create an Exchange account
- Create an IMAP account
- Open settings for the personal calendar

Expected results: Able to link the personal calendar to the IMAP account

Actual results: Personal calendar will only link to the Exchange account

benbucksch:
@NeilRashbrook Can you figure out why that is and fix it, please? This is similar to #1273.

benbucksch:
@NeilRashbrook PR #1276 wouldn't fix this bug, would it, right? Because the PR only narrows the available identities further, whereas the problem here was that eligible accounts were not selectable.

NeilRashbrook:
If you mean PR #1283, then no, that widens the available identities to be from all eligible accounts rather than always the first account (which might not even be an eligible account), so from my point of view this issue is now fixed.

PR #1283, opened 2026-06-25
Calendar: Settings: Allow identities from all available accounts #1276

NeilRashbrook:


app/logic/Calendar/Calendar.ts
```diff
@@ -110,8 +110,10 @@ export class Calendar extends Account {
       .catch(this.errorCallback);
   }
   get identitiesAvailable(): Collection<MailIdentity> {
-    return (this.mainAccount as MailAccount)?.identities
-      ?? appGlobal.emailAccounts.first?.identities;
+    if (this.mainAccount instanceof MailAccount) {
+      return this.mainAccount.identities;
+    }
```
benbucksch, line 115: While your new code is fine, just a question: Did the old `(this.mainAccount as MailAccount)?.identities ?? ` not work? If it's not a MailAccount, it would not have an `identities` property, which would then evaluate to null, and it would take the expression after `??`, no? Of course your new code is more correct.

benbucksch (approved):

app/logic/Calendar/Calendar.ts
NeilRashbrook: Yes, the old code worked, but with the longer code below I thought it would be more readable to split it up into a separate condition statement, at which point I just picked the optimal condition.

Issue #1272, opened 2026-06-24
Calendar: Outgoing invitation: When changing event time, people still shows as having accepted, but they didn't accept the new time yet


benbucksch:
### Reproduction
* Send an invitation from IMAP via iTIP to somebody. Event starts at 14:00.
* The other person confirms
* Change the time to 15:00

### Actual result
* The person is still shown as having Confirmed

### Expected result
* An invitation change goes out
* The person shows as not having responded at all
* The person confirms the new time
* The person shows as not having Confirmed

### Compat
* How does Exchange and Outlook handle this on the organizer side? (If you did the same thing with Office365 as organizer account.)
*  does Exchange and Outlook handle this on the invitee side? Do they still show as having Confirmed, even after the time change? Do they even get the chance to Confirm again after the time change?

NeilRashbrook:
> - How does Exchange and Outlook handle this on the organizer side? (If you did the same thing with Office365 as organizer account.)
> - does Exchange and Outlook handle this on the invitee side? Do they still show as having Confirmed, even after the time change? Do they even get the chance to Confirm again after the time change?

Both sides show as being not responded after the organiser changes the event's time.

PR #1285, opened 2026-06-25
Calendar: Outgoing Invitation: Reset recipients to not responded #1272

NeilRashbrook:
Changing e.g. the event alarm doesn't send any messages.
Changing the event participants only sends messages to those participants.
Major changes send messages to all participants.
This change makes it so that the participation status is reset to no response received only for those participants.
This is done in `event.saveLocally` because it needs to be saved.
`sendInvitationsTo` no longer needs to mark new participants as no response received.
I also refactored out some code that was shared with `sendInvitationsTo`.

This PR does not handle the case where you receive an updated invitation.

### DIFF
--- a/app/logic/Calendar/Event.ts
+++ b/app/logic/Calendar/Event.ts
@@ -598,6 +598,9 @@ export class Event extends Observable {
     assert(this.calendar, "To save an event, it needs to be in a calendar first");
     assert(this.calendar.storage, "To save an event, the calendar needs to be saved first");
     this.calUID ??= crypto.randomUUID();
+    if (!this.isIncomingMeeting && this.participants.hasItems && this.hasChanged()) {
+      await this.outgoingInvitation.markInvitationsNeeded();
+    }
     await this.calendar.storage.saveEvent(this);
 
     if (this.recurrenceCase == RecurrenceCase.Master) {
--- a/app/logic/Calendar/Invitation/OutgoingInvitation.ts
+++ b/app/logic/Calendar/Invitation/OutgoingInvitation.ts
@@ -109,8 +109,8 @@ export class OutgoingInvitation {
   participantsToNotify(): Collection<Participant> {
     let participants = this.event.participants;
     return this.changesNeedToNotify()
-      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer)
-      : participants.subtract(this.event.unedited.participants);
+      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer) // everybody
+      : participants.subtract(this.event.unedited.participants); // newly added participants
   }
 
   /**
--- a/app/logic/Calendar/Invitation/OutgoingInvitation.ts
+++ b/app/logic/Calendar/Invitation/OutgoingInvitation.ts
@@ -28,6 +28,7 @@ export class OutgoingInvitation {
   }
 
   async sendInvitationsTo(participants: Collection<Participant>) {
+    this.markInvitationsNeeded();
     let sendTo = participants.filterOnce(participant => participant.isInvitee);
     if (sendTo.isEmpty) {
       return;
@@ -109,7 +110,7 @@ export class OutgoingInvitation {
   participantsToNotify(): Collection<Participant> {
     let participants = this.event.participants;
     return this.changesNeedToNotify()
-      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer) // everybody
+      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer) // all invitees
       : participants.subtract(this.event.unedited.participants); // newly added participants
   }
 
--- a/app/logic/Calendar/Invitation/OutgoingInvitation.ts
+++ b/app/logic/Calendar/Invitation/OutgoingInvitation.ts
@@ -28,9 +28,6 @@ export class OutgoingInvitation {
   }
 
   async sendInvitationsTo(participants: Collection<Participant>) {
-    for (let participant of participants) {
-      participant.response ||= InvitationResponse.NoResponseReceived;
-    }
     let sendTo = participants.filterOnce(participant => participant.isInvitee);
     if (sendTo.isEmpty) {
       return;
@@ -56,10 +53,7 @@ export class OutgoingInvitation {
     let removed = unedited.participants.subtract(event.participants);
     // Use the original event when sending cancellations
     await unedited.outgoingInvitation.sendCancellationsTo(removed);
-    let notify = this.changesNeedToNotify()
-      ? event.participants
-      : event.participants.subtract(unedited.participants);
-    await this.sendInvitationsTo(notify);
+    await this.sendInvitationsTo(this.participantsToNotify());
   }
 
   async sendCancellations() {
@@ -96,6 +90,27 @@ export class OutgoingInvitation {
       email.html = this.event.descriptionHTML;
     }
     await email.folder.account.send(email);
+  }
+
+  /**
+   * Mark anyone we're notifying as no response recieved.
+   * This always applies for new recipients.
+   */
+  markInvitationsNeeded() {
+    for (let participant of this.participantsToNotify()) {
+      participant.response = InvitationResponse.NoResponseReceived;
+    }
+  }
+
+  /**
+   * Get the participants we need to notify about this change.
+   * Always includes newly added participnts.
+   */
+  participantsToNotify(): Collection<Participant> {
+    let participants = this.event.participants;
+    return this.changesNeedToNotify()
+      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer)
+      : participants.subtract(this.event.unedited.participants);
   }
 
   /**
--- a/app/frontend/Calendar/EditEvent/ParticipantConfirmIcon.svelte
+++ b/app/frontend/Calendar/EditEvent/ParticipantConfirmIcon.svelte
@@ -3,7 +3,7 @@
   class:declined={$participant.response == InvitationResponse.Decline}
   class:tentative={$participant.response == InvitationResponse.Tentative}
   class:organizer={$participant.response == InvitationResponse.Organizer}
-  class:unknown={$participant.response == InvitationResponse.Unknown}
+  class:noresponse={$participant.response == InvitationResponse.NoResponseReceived}
   title={$participant.responseLabel}
   >
   {#if $participant.response == InvitationResponse.Organizer}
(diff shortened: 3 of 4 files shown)


app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -98,6 +92,27 @@ export class OutgoingInvitation {
...
+
+  /**
+   * Get the participants we need to notify about this change.
+   * Always includes newly added participnts.
+   */
+  participantsToNotify(): Collection<Participant> {
+    let participants = this.event.participants;
+    return this.changesNeedToNotify()
+      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer)
+      : participants.subtract(this.event.unedited.participants);
```
```suggestion
      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer) // everybody
      : participants.subtract(this.event.unedited.participants); // newly added participants
```

app/logic/Calendar/Event.ts
```diff
@@ -598,6 +598,9 @@ export class Event extends Observable {
     assert(this.calendar, "To save an event, it needs to be in a calendar first");
     assert(this.calendar.storage, "To save an event, the calendar needs to be saved first");
     this.calUID ??= crypto.randomUUID();
+    if (!this.isIncomingMeeting && this.participants.hasItems && this.hasChanged()) {
+      await this.outgoingInvitation.markInvitationsNeeded();
+    }
```
benbucksch, line 603: Can we keep invitations logic out of the `Event` object, please, and in `OutgoingInvitation`?

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -28,9 +28,6 @@ export class OutgoingInvitation {
   }
 
   async sendInvitationsTo(participants: Collection<Participant>) {
-    for (let participant of participants) {
-      participant.response ||= InvitationResponse.NoResponseReceived;
```
benbucksch, line 32: Why is this removed? What replaces it?

NeilRashbrook: It's replaced by `markInvitationsNeeded`.

app/logic/Calendar/Event.ts
NeilRashbrook: Where does the logic for "is this event an outgoing invitation or not" belong? It seems odd to create an `outgoingInvitation` object for an event that isn't an outgoing invitation...

app/logic/Calendar/Invitation/OutgoingInvitation.ts
```diff
@@ -98,6 +92,27 @@ export class OutgoingInvitation {
...
+  }
+
+  /**
+   * Get the participants we need to notify about this change.
+   * Always includes newly added participnts.
+   */
+  participantsToNotify(): Collection<Participant> {
+    let participants = this.event.participants;
+    return this.changesNeedToNotify()
+      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer) // everybody
```
NeilRashbrook:
```suggestion
      ? participants.filterOnce(participant => participant.response != InvitationResponse.Organizer) // all invitees
```

app/logic/Calendar/Invitation/OutgoingInvitation.ts
benbucksch: But you don't call it here. Which call to mark() replaces this code here?

app/logic/Calendar/Event.ts
benbucksch: How is it not an outgoing invitation? It has participants and is not incoming invitation, so must be an outgoing invitation, no?

There are only 3 types of events that I know of:
* Personal events not involving anybody else
* incoming invitations
* outgoing invitations

Or are you referring to the difference between the event in the calendar, and the email that is sent out for that event?

This all needs much much clearer design.

app/logic/Calendar/Invitation/OutgoingInvitation.ts
NeilRashbrook: There is only the one caller, in `Event.saveLocally`.

app/logic/Calendar/Event.ts
NeilRashbrook: Then I don't understand which logic you want to keep out of the `Event` object.

app/logic/Calendar/Invitation/OutgoingInvitation.ts
benbucksch: You see, I cannot map those two callers in my mind. Here's what I am actually asking: I would like you to prove that the new code runs in *all* cases where the old code ran. That requires to show all callers of sendInvitationsTo() and why they are guaranteed to run through Event.saveLocally() before sendInvitationsTo() . I'd like you to consider *all* possible cases and situations.

Either you have already considered all situations, then it's just a matter of communicating your reasoning to me in a way I can understand.
Or you have missed some cases and situations, which you will discover now, then this exersize stopped bugs.

NeilRashbrook: `sendInvitationsTo` is called by `sendInvitations` is (if the event is an edited invitation) called by `sendInvitationsDirectly` is (for some protocols) called by `saveToServer` is called by `save`.

`markInvitationsNeeded` is (if the event is an edited invitation) called by `saveLocally` is called by `save`.

benbucksch: OK.

So, why not leave it in sendInvitations() ? Because you'd have to save the event again after that change?

The problem I have is that this change moves code from where it belongs to a place where it doesn't belong. It currently happens to be that saveToServer() calls sendInvitations(), but that's always been a HACK, not a correct design. This is very likely to change. And then your assumption above that saveLocally() is called before sendInvitations() is no longer true and this will break. It happens to be true now, but nothing ensures that. We need to build code that does *not* fall apart unexpectedly just because we change one call around.

We cannot have such assumptions in class A that class B will do something in a certain moment and situation. That makes a fragile codebase.

NeilRashbrook: > It currently happens to be that saveToServer() calls sendInvitations(), but that's always been a HACK

It has? When would you prefer to send invitations?

There are a number of options:
- Always add new recipients as NoResponseReceived. (This wasn't originally the case due to a limitation at the time of the autocomplete widget.)
- Assert here that none of the receipients have an unknown status.
- Duplicate the call to `markInvitationsNeeded`. (I don't like this option because it will hide future errors.)

benbucksch: > It has?

Yes, of course. "Send to other people" is semantically very different from "save it on my own calendar" (on my own server).

What Exchange does is not correct.

As a user, I might want to create an event, keep the fact that it involves certain other people, by adding "participants" (which then allows me to find the event later), but *not* send them an invitation to it.

> When would you prefer to send invitations?

Right after save, but conditional on the user choosing to invite. I would invoke it directly from the UI.

But that's a refactor for another time.

> Always add new recipients as NoResponseReceived.

Yes, please. That's a no-brainer. What has been the value for new participants so far?

> Duplicate the call to markInvitationsNeeded. (I don't like this option because it will hide future errors.)

Makes sense to me, because it merely brings the object in a known-good state.

Which errors would it hide?

-----

Generally, it's *very* important that

a) the code for a certain functionality is exactly where you'd expect it. Code for incoming invitations should be in `IncomingInvitation`.

b) code does *not* assume that some other code (other functions and *esp* other classes) does something. Assume that other classes are hostile and actively out to destroy your logic. Build your logic so that it works no matter what the other class does. Other classes only use your API, and your own function ensures that everything goes according to other. You do *not* trust other classes. And other functions in the same class only to a very limited degree, only when it's documented or obvious from the code structure.

The objective is to isolate the code that cares about a specific aspect to the absolute minimum.

NeilRashbrook: > > When would you prefer to send invitations?
> 
> Right after save

So we are already sending the invitations at the right time?

> What Exchange does is not correct.

- EWS, OWA: At point of saving to the server, you can choose not to notify any attendees
- ActiveSync 14.x: Does not generate invitations, client must notify attendees separately
- ActiveSync 16.x, Graph: Always generates invitations (responses are optional)

> > Always add new recipients as NoResponseReceived.
> 
> Yes, please. That's a no-brainer. What has been the value for new participants so far?

`Unknown`. (As opposed to "unknown".)

> > Duplicate the call to markInvitationsNeeded. (I don't like this option because it will hide future errors.)
> 
> Makes sense to me, because it merely brings the object in a known-good state.
> 
> Which errors would it hide?

Someone might remove the other call, not realising that it's that call that's actually doing the work, and without it, the changes won't get saved.

benbucksch: > So we are already sending the invitations at the right time?

It's not about timing at all. See my last comment. I cannot explain it any better.

"works for me" is a necessary condition, but not a suffient one.

NeilRashbrook: Ah, so what you want to be able to do is write:
```
async save() {
  await this.saveLocally();
  await this.saveToServer();
  await this.sendInivtationsDirectly();
}
```
Your problem is that ActiveSync 16.x and Graph always send invitations when you save to server.

benbucksch: More like:
```
async save() {
  await this.save();
  if (shouldSendInvitations) { // Some kind of UI for user to decide
    await this.sendInivtations(); // protocol-specific implementation
  }
}
```

> Your problem is that ActiveSync 16.x and Graph always send invitations when you save to server.

Yes, that's why I said that Exchange design is completely broken when it comes to invitations. Exchange also automatically adds incoming invitations to my calendar, without me asking for it, which I find outright offensive and rude. Exchange is *not* the right design guide here.

benbucksch: You will now say that Exchange already sends the outgoing invitations in the moment when I save. EWS has a flag for it, but that flag has to be set at save time. Calling the function after save will be too late. Right?

So, Exchange forces our hand here, and we'd have to make it a flag on the event, whether or not to send invitations out. But I don't think that's the right design.

NeilRashbrook: Right: once you've saved an event, the only way to send more invitations is to alter the event. (On the other hand, responses are much more flexible, unless you're declining, as on-premises exchange will delete a declined incoming invitation, so you have decide whether you want to notify at that point.)

PR #1286, opened 2026-06-25
Calendar: Incoming Invitation: Reset participation to not responded #1272

NeilRashbrook:
This makes it so that all incoming iMip messages have automatic processing:
- Requests reset your participation had you already replied to the previous invitation
- Cancellations and replies as before

### DIFF
--- a/app/logic/Calendar/ICal/ICalIncomingInvitation.ts
+++ b/app/logic/Calendar/ICal/ICalIncomingInvitation.ts
@@ -32,6 +32,23 @@ export class ICalIncomingInvitation extends IncomingInvitation {
     }
   }
 
+  async updateInvitation() {
+    assert(this.invitationMessage == InvitationMessage.Invitation, "Not an invitation");
+    let event = this.calEvent();
+    if (!event) {
+      // Event is not yet in the user's calendar, nothing to do.
+      return;
+    }
+    let { myParticipant } = ICalIncomingInvitation.participantMe(event, this.message.folder.account);
+    let timestamp = this.event.lastUpdateTime;
+    if (myParticipant &&
+      (!myParticipant.lastUpdateTime || myParticipant.lastUpdateTime < timestamp)) {
+      this.myParticipation = myParticipant.response = InvitationResponse.NoResponseReceived;
+      myParticipant.lastUpdateTime = timestamp;
+      await event.save();
+    }
+  }
+
   static async respondToInvitationFromCalEvent(calEvent: Event, response: InvitationResponseInMessage, mailAccount?: MailAccount) {
     let { identity, myParticipant } = this.participantMe(calEvent, mailAccount);
     let hasChanged = myParticipant.response != response;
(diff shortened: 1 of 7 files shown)


app/logic/Calendar/ICal/ICalIncomingInvitation.ts
```diff
@@ -32,6 +32,23 @@ export class ICalIncomingInvitation extends IncomingInvitation {
...
+    let event = this.calEvent();
+    if (!event) {
+      // Event is not yet in the user's calendar, nothing to do.
+      return;
+    }
+    let { myParticipant } = ICalIncomingInvitation.participantMe(event, this.message.folder.account);
+    let timestamp = this.event.lastUpdateTime;
+    if (myParticipant &&
+      (!myParticipant.lastUpdateTime || myParticipant.lastUpdateTime < timestamp)) {
+      this.myParticipation = myParticipant.response = InvitationResponse.NoResponseReceived;
```
benbucksch, line 46: Why are you assuming that all "updates" to invitations need to clear my response? Shouldn't we add a conditional here?

benbucksch (changes requested):

app/logic/Calendar/ICal/ICalIncomingInvitation.ts
NeilRashbrook: We wouldn't have been sent an update if the organiser didn't want a new response?

(And it's all very well saying conditional but what would it be conditional on?)

benbucksch: > We wouldn't have been sent an update if the organiser didn't want a new response?

That's an assumption, and it's in fact wrong.

The organizer might have updated the description, added attachments etc.pp. We don't control why we receive a new invitation. All kinds of reasons why an updated invitation might be sent out, many of which do not invalidate my response.

NeilRashbrook: Ah right, the ICS file won't have `RSVP=TRUE` if a response isn't required, so we need to check for that.

benbucksch: Can you instead make a condition where we null the response only in the specific situation that we mean here? And ignore all others, please? That's the safe thing to do.

FWIW, we want participants to re-approve once either the time or the physical location changes. We don't need re-approvals for title, description, participant (other than us), online meeting URL etc. changes.

benbucksch: What would make sense to me is to clear the response when we're changing the calendar event. Would you agree?
I'd do that from the code that calls the event, not within `Event`, but from `IncomingInvitation`. (Compare https://github.com/mustang-im/mustang/pull/1285#discussion_r3483924392 )

NeilRashbrook: I don't agree; iCal says that RSVP=TRUE means that the organiser wants a new response.

benbucksch: Fair enough.

Can you log when the 2 sides don't agree? I.e. `RSVP=TRUE`, but we don't see a reason based on the above rules?

NeilRashbrook: When we get an invitation update (newer than the calendar event that we previously saved), should we:
1. Always update the outdated calendar event
2. Only update the outdated calendar event if no response was requested
3. As 2. but also delete the outdated calendar event if a response was requested

Issue #1269, opened 2026-06-23
Contacts: Cannot delete (get rid of) a new contact


benbucksch:
### Reproduction
* Create a new contact
* Delete it

### Actual result
* It stays

### Expected result
* It's gone

jermy-c:
This requires the `persons` prop passed down all the way to this component. Because `addPersons` adds it to the persons prop from ContactAppD. And you cannot remove it from the real collection because it doesn't exist there.

PR #1287, opened 2026-06-25
Contacts: Edit persons collection directly #1269

jermy-c:
- We were adding to the `persons` collection in the PersonsToolBar which made it impossible to remove a person if it was unsaved since it didn't belong to any address book persons collection
- The other problem was that after moving to another address book we didn't select the new instance
- `moveToAddressbook()` would unknowingly save to the DB or server by just calling it even if the contact was not saved yet
- This PR fixes the above problems, however, after changing the Addressbook it would move to the bottom if it's not saved yet and it makes it confusing because on unsaved contacts should always be at the top because it's still being edited.

### DIFF
--- a/app/frontend/Contacts/AddressbookChanger.svelte
+++ b/app/frontend/Contacts/AddressbookChanger.svelte
@@ -10,6 +10,7 @@
 <script lang="ts">
   import type { Addressbook } from "../../logic/Contacts/Addressbook";
   import type { Person } from "../../logic/Abstract/Person";
+  import { selectedPerson } from "./Person/Selected";
   import { appGlobal } from "../../logic/app";
   import { selectedWorkspace } from "../MainWindow/Selected";
   import AccountDropDown from "../Shared/AccountDropDown.svelte";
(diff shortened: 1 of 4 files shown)


benbucksch:
Wouldn't it be cleaner, if new persons were not added to the persons list or address book at all yet, until the user saves?
It means that the person won't show up in the persons list on the left until saving, but I think that's OK.

jermy-c:
> Wouldn't it be cleaner, if new persons were not added to the persons list or address book at all yet, until the user saves?

Yes, but what if the user accidentally selects another person on the left while the new person is unsaved. The user would loose all the information he has inputted because there's no where to select it again, unless there's a popup warning about it.

benbucksch:
> what if the user accidentally selects another person on the left while the new person is unsaved.

Great point, thanks.

I discussed this with Elena, and here is what we would like to do:

a) add `newPerson` to frontend/Contacts/selected.ts , set to the new `Person`. If that is set:
b) At the bottom of PersonHeader, add a line for the new person. With a new icon, so that it stands out.
c) If I selected a different person or a different app, a sidebar appears that shows the new person, including the header with Save and Delete buttons. The mechanism is identical to Meet, which shows an ongoing meeting. The reason is that the pohone number, email address, street etc. is most likely copied from somewhere else (another contact, an email, a chat message), so the sidebar would make this copying easier. The sidebar stays until the user either saved or deleted the new contact.

We could do the same with new events.

As part of the first PR, please do a) and b). Part c) can be left to another PR.

PR #1288, opened 2026-06-25
Contacts: Person: Update Name with FirstName and LastName #1281

### LINKED ISSUE #1281: Contacts: New contact: Surname field loses focus after first typed character
Macbook - Parula 0.9.31-dev

Steps to Reproduce

1. Open the Contact creation form.
2. Click into the Surname field.
3. Type the first character.

Actual results
- Cursor/focus moves to the first row
- This behavior affects all fields in the Name header section
- What you type on the second row doesn't update the first raw (Full name)

Expected results
Focus remains in the active field while the user is typing

https://github.com/user-attachments/assets/321f664a-5d89-40f1-8886-fd3b664ccce6

Image

jermy-c:
- Update Name with FirstName and LastName
- Update it so it's not null and it doesn't trigger a focus on the Name input
- There's one problem with this which I haven't managed to fix where if you only input the lastName it should leave a space infront. If you do `trim()` for the firstName or lastName or both in the function it would make it hard add a space in the Name  input.
- This doesn't seem to cause a reactive loop.
- Fixes #1281

### DIFF
--- a/app/logic/Abstract/Person.ts
+++ b/app/logic/Abstract/Person.ts
@@ -110,6 +110,11 @@ export class Person extends ContactBase {
   notifyObservers(propertyName?: string, oldValue?: any): void {
     if (propertyName == "name" && this.name && typeof (this.name) == "string") {
       this.fixName();
+    } else if ((propertyName == "firstName" &&
+      this.firstName && typeof (this.firstName) == "string") ||
+      (propertyName == "lastName" &&
+        this.lastName && typeof (this.lastName) == "string")) {
+      this.joinName();
     }
     super.notifyObservers(propertyName, oldValue);
   }
@@ -136,6 +141,10 @@ export class Person extends ContactBase {
         }
       }
     }
+  }
+
+  protected joinName() {
+    this.name = [this.firstName ?? "", this.lastName ?? ""].join(" ");
   }
 
   async merge(other: Person) {


benbucksch:
Which bug does this fix?

Note that it's valid to have a name different from first name and last name. E.g. your wife could have first name = "Maria" lastname = "Sanchez", but the display name is "Birdy", because that's your nick name for your wife and you want to see that instead of her full name.

jermy-c:
> Which bug does this fix?

It fixes #1281

jermy-c:
> Note that it's valid to have a name different from first name and last name.

That's a valid point, maybe just auto-add the name if it's empty.

benbucksch:
> > Which bug does this fix?
> 
> It fixes #1281

Ah, I see it in the title now. It's good that you put it into the title, so that it becomes part of the title of the git commit. Thanks. But could you please *also* put it into the initial description? a) it's more visible there and b) it will be linked there, so I can navigate c) github will then automatically also post a link to the PR into the bug report.

jermy-c:
> That's a valid point, maybe just auto-add the name if it's empty.

This doesn't work because it stops updating the display name after the first letter is typed. e.g. LastName: "Smith", DisplayName: "S".

jermy-c:
This fix won't work when you input other fields first e.g. company, position because it won't update DisplayName.

PR #1292, opened 2026-06-26
Not let `login` call `startup`, but let the caller decide #1036 #1291

### LINKED ISSUE #1291: EWS: stream notification started twice
* https://mustang-jq.sentry.io/issues/121715387

And I encounter it myself

### Environment
* neil@be as EWS
* End user ran into it as well

### Reproduction
* Start app

### Actual result
* On error console:
Error: EWS: stream notification started twice
    at assert (￼util.ts:4:11)
    at EWSAccount.subscribeToNotifications (￼EWSAccount.ts:528:5)
    at async ￼EWSAccount.ts:123:7
    at async RunOnce.runOnce (￼RunOnce.ts:16:14)
    at async EWSAccount.startup (￼EWSAccount.ts:112:5)
    at async ￼startup.ts:71:9

NeilRashbrook:
- Assumes all UI callers want the account to start up
- Removes the automatic startup from those mail accounts that have it
- Fixes #1291

### DIFF
--- a/app/frontend/Settings/Mail/Account/Folders.svelte
+++ b/app/frontend/Settings/Mail/Account/Folders.svelte
@@ -31,7 +31,7 @@
   {:else if !$mailAccount?.isLoggedIn}
     <vbox class="login">
       <hbox class="label">{$t`Please log in to account ${mailAccount?.name} first`}</hbox>
-      <Button label={$t`Login`} onClick={async () => await mailAccount.loginAndStartup()} />
+      <Button label={$t`Login`} onClick={async () => await mailAccount.loginAndSync()} />
     </vbox>
   {/if}
 </vbox>
--- a/app/frontend/Settings/Mail/Account/Folders.svelte
+++ b/app/frontend/Settings/Mail/Account/Folders.svelte
@@ -31,7 +31,7 @@
   {:else if !$mailAccount?.isLoggedIn}
     <vbox class="login">
       <hbox class="label">{$t`Please log in to account ${mailAccount?.name} first`}</hbox>
-      <Button label={$t`Login`} onClick={async () => await account.loginAndStartup()} />
+      <Button label={$t`Login`} onClick={async () => await mailAccount.loginAndStartup()} />
     </vbox>
   {/if}
 </vbox>
--- a/app/frontend/Settings/Mail/Account/Folders.svelte
+++ b/app/frontend/Settings/Mail/Account/Folders.svelte
@@ -31,7 +31,7 @@
   {:else if !$mailAccount?.isLoggedIn}
     <vbox class="login">
       <hbox class="label">{$t`Please log in to account ${mailAccount?.name} first`}</hbox>
-      <Button label={$t`Login`} onClick={login} />
+      <Button label={$t`Login`} onClick={async () => await account.loginAndStartup()} />
     </vbox>
   {/if}
 </vbox>
@@ -58,10 +58,6 @@
   $: mailAccount = account as MailAccount;
   let selectedFolders: ArrayColl<Folder>;
   let isCreating: "toplevel" | "subfolder" | false = false;
-
-  async function login() {
-    await account.loginAndStartup();
-  }
 </script>
 
 <style>
--- a/app/frontend/Settings/Mail/Account/Folders.svelte
+++ b/app/frontend/Settings/Mail/Account/Folders.svelte
@@ -60,8 +60,7 @@
   let isCreating: "toplevel" | "subfolder" | false = false;
 
   async function login() {
-    await account.login(true);
-    await account.startup();
+    await account.loginAndStartup();
   }
 </script>
 
--- a/app/frontend/Settings/Mail/Account/Folders.svelte
+++ b/app/frontend/Settings/Mail/Account/Folders.svelte
@@ -31,7 +31,7 @@
   {:else if !$mailAccount?.isLoggedIn}
     <vbox class="login">
       <hbox class="label">{$t`Please log in to account ${mailAccount?.name} first`}</hbox>
-      <Button label={$t`Login`} onClick={async () => await mailAccount.login(true)} />
+      <Button label={$t`Login`} onClick={login} />
     </vbox>
   {/if}
 </vbox>
@@ -58,6 +58,11 @@
   $: mailAccount = account as MailAccount;
   let selectedFolders: ArrayColl<Folder>;
   let isCreating: "toplevel" | "subfolder" | false = false;
+
+  async function login() {
+    await account.login(true);
+    await account.startup();
+  }
 </script>
 
 <style>
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -106,8 +106,12 @@ export class Account extends Observable {
   }
 
   async loginAndStartup(interactive = true): Promise<void> {
-    await this.login(interactive);
-    await this.startup();
+    if (this.isDependentAccount) {
+      await this.mainAccount.loginAndStartup(interactive);
+    } else {
+      await this.login(interactive);
+      await this.startup();
+    }
   }
 
   /** For setup only. Test that the login works. */
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -60,22 +60,18 @@ export class Account extends Observable {
   }
 
   /**
-   * For local accounts, this is the startup entry point. It is overridden
-   * e.g. for contacts and calendar to load data from the database.
-   * Startup for accounts that need to log in will attempt to use a
-   * noninteractive `login` call instead. However if this is successful
-   * then the account's login code should finish by running its startup code.
+   * This is the entry point to perform post-login sync on this account.
    */
-  async startup() {
+  async initialSync() {
   }
 
   /**
-   * Convenience method for accounts to use to start up dependent accounts.
-   * This should be called after the account has finished its own startup.
+   * Convenience method for accounts to use to sync all dependent accounts.
+   * This should be called after the account has finished its own sync.
    */
-  protected async startupDependentAccounts() {
+  protected async syncDependentAccounts() {
     for (let dependent of this.dependentAccounts()) {
-      dependent.startup()
+      dependent.initialSync()
         .catch(dependent.errorCallback);
     }
   }
@@ -105,13 +101,13 @@ export class Account extends Observable {
     }
   }
 
-  async loginAndStartup(interactive = true): Promise<void> {
+  async loginAndSync(interactive = true): Promise<void> {
     if (this.isDependentAccount) {
-      await this.mainAccount.loginAndStartup(interactive);
+      await this.mainAccount.loginAndSync(interactive);
       return;
     }
     await this.login(interactive);
-    await this.startup();
+    await this.initialSync();
   }
 
   /** For setup only. Test that the login works. */
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -76,6 +76,28 @@ export class Account extends Observable {
     }
   }
 
+  /**
+   * Entry point to notify the account that login has completed.
+   * Could potentially be used in conjunction with network offline events
+   * (`disconnect()` when going offline and `connected()` when back online)?
+   */
+  async connected() {
+  }
+
+  /**
+   * Convenience method for accounts to use to notify all dependent accounts
+   * that the login has completed.
+   */
+  protected async dependentAccountsConnected() {
+    for (let dependent of this.dependentAccounts()) {
+      try {
+        await dependent.connected();
+      } catch (ex) {
+        dependent.errorCallback(ex);
+      }
+    }
+  }
+
   get isLoggedIn(): boolean {
     return false;
   }
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -105,6 +105,11 @@ export class Account extends Observable {
     }
   }
 
+  async loginAndStartup(interactive = true): Promise<void> {
+    await this.login(interactive);
+    await this.startup();
+  }
+
   /** For setup only. Test that the login works. */
   async verifyLogin(): Promise<void> {
     await this.login(true);
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -108,10 +108,10 @@ export class Account extends Observable {
   async loginAndStartup(interactive = true): Promise<void> {
     if (this.isDependentAccount) {
       await this.mainAccount.loginAndStartup(interactive);
-    } else {
-      await this.login(interactive);
-      await this.startup();
+      return;
     }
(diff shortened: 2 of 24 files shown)


benbucksch:
Nice change. I like it. I wanted to do that anyway. Thanks for doing this.

1. Please add in class `Account`
```
  async loginAndStartup(interactive: boolean = true) {
    await this.login(interactive);
    await this.startup();
  }
```

and then you can change all callers to `account.loginAndStartup();`

2. Please check *all* classes that *none* of them assume that actions in `startup()` will always happen as part of login. I can imagine that `startup()` might have some init or reset code that the rest of the code depends on running after every login. This requires to read through all FooAccount classes, whether there are such assumptiong.

NeilRashbrook:
> 2. Please check _all_ classes that _none_ of them assume that actions in `startup()` will always happen as part of login. I can imagine that `startup()` might have some init or reset code that the rest of the code depends on running after every login. This requires to read through all FooAccount classes, whether there are such assumptiong.

I don't know whether these count as init code:
- All mail accounts use `startup` to list folders and update the Inbox
  - ActiveSync ensures that its contacts and addressbook exist at this point
  - EWS ensures that its contacts and addressbook exist at this point
  - OWA ensures that its addressbook exists at this point
- All contacts accounts use `startup` to list contacts
  - EWS accounts subscribe to notifications at this point
- All calendar accounts use `startup` to list contacts
  - EWS accounts subscribe to notifications at this point
- ActiveSync uses `startup` to add its GAL
- EWS uses `startup` to add its GAL and subscribe to notifications
- JMAP uses `startup` to start polling the Inbox and ensure that its contacts and addressbook exist
- OWA uses `startup` to add its GAL, subscribe to notifications and ensure that its contacts exists
- Graph uses `startup` to ensure that its contacts and addressbook exist
- IMAP uses `startup` to query the server's namespaces and start polling the Inbox
- WhatsApp uses `startup` to list rooms
- GraphChat uses `startup` to list rooms (it's a dependent account so it doesn't have its own `login`)

app/frontend/Settings/Mail/Account/Folders.svelte
```diff
@@ -31,7 +31,7 @@
   {:else if !$mailAccount?.isLoggedIn}
     <vbox class="login">
       <hbox class="label">{$t`Please log in to account ${mailAccount?.name} first`}</hbox>
-      <Button label={$t`Login`} onClick={async () => await mailAccount.login(true)} />
+      <Button label={$t`Login`} onClick={login} />
```
benbucksch:
```suggestion
      <Button label={$t`Login`} onClick={async () => await account.loginAndStartup()} />
```

app/frontend/Settings/Mail/Account/Folders.svelte
```diff
@@ -58,6 +58,10 @@
   $: mailAccount = account as MailAccount;
   let selectedFolders: ArrayColl<Folder>;
   let isCreating: "toplevel" | "subfolder" | false = false;
+
+  async function login() {
+    await account.loginAndStartup();
+  }
```
benbucksch:
```suggestion
```

benbucksch (approved):

app/frontend/Settings/Mail/Account/Folders.svelte
```diff
@@ -31,7 +31,7 @@
   {:else if !$mailAccount?.isLoggedIn}
     <vbox class="login">
       <hbox class="label">{$t`Please log in to account ${mailAccount?.name} first`}</hbox>
-      <Button label={$t`Login`} onClick={async () => await mailAccount.login(true)} />
+      <Button label={$t`Login`} onClick={async () => await account.loginAndStartup()} />
```
benbucksch:
```suggestion
      <Button label={$t`Login`} onClick={async () => await mailAccount.loginAndStartup()} />
```

benbucksch (approved):

benbucksch:
The key part was this subphrase:

> that the rest of the code depends on running after every login

Can you please check that?

NeilRashbrook:
What about `getFirstMessages`? It assumes that `login` will list folders, but not fetch messages; currently `login` does neither, while `startup` does both (because dependent accounts need it to).

benbucksch:
We may need a third sync()  function.

We have login, startup like list folders, and fetch/sync like get new mail, sync contacts and calendar events, upload of changes.

Frankly, startup is not clearly defined to me. For me, stream listener would belong to login, but I think you put them in startup, right? Why would I log in and not start listeners? What does "login" even mean, then? Unclear to me.

Please first make a very sharp definition of login, startup, and first sync. the definition should be generic, but able to answer without doubt where steam listeners, folder list, inbox get messsages, fetch and push of contacts, and sending outgoing waiting mail and upload of locally changed calendar entries and sending of invitations belongs. Note that we will soon have to implement full offline capability.

NeilRashbrook:
(Not intended as a definition)

Login currently means authenticating to the server, so that we know that we can perform requests against it. Dependent accounts don't login because their main account does that.

Startup currently means any one-time actions that an account does after it or its main account (as appropriate) has logged in. This might include sync or registering for notifications.

The reverse operations are called disconnect and logout respectively.

app/logic/Abstract/Account.ts
```diff
@@ -106,6 +105,15 @@ export class Account extends Observable {
     }
   }
 
+  async loginAndStartup(interactive = true): Promise<void> {
+    if (this.isDependentAccount) {
+      await this.mainAccount.loginAndStartup(interactive);
+    } else {
+      await this.login(interactive);
+      await this.startup();
+    }
```
benbucksch, line 130:
```suggestion
    if (this.isDependentAccount) {
      await this.mainAccount.loginAndStartup(interactive);
      return;
    }
    await this.login(interactive);
    await this.startup();
```

Special code inside an `if`, normal code straight and outside `if`

benbucksch:
I'd like to merge this, but I am still unclear on how carefully you checked (via code reading) that this won't violate any assumptions in the code and won't regress anything.

We cannot affort regressions right now.

NeilRashbrook:
(Still not intended as a definition)

Originally, Mustang only supported main accounts and SMTP accounts, which were a special sort of dependent account for IMAP accounts (but POP accounts will eventually use them too).

Then Exchange reared its ugly head. Here we have a main Exchange email account, which is the one that actually logs into the server, but not only can this account have child contacts and calendar accounts, it went on to support child email accounts which read from a shared mailbox (and further contacts and calendar accounts for those shared mailboxes but these are still child accounts of the main account). Furthermore these accounts can be added and deleted at will, without logging out of the main account.

This however caused a conflict between the UI and the startup code.

The startup code wanted to be able to login to all of the accounts to start them up. The Exchange child accounts simply did nothing when you tried this, since the main Exchange account would start up the child accounts once it had finished logging in.

The UI code wanted to be able to log in to any account that wasn't logged in, so a change was made so that child accounts would forward the login call to the main account. This created a startup headache as the startup code now would try to start up all of the Exchange accounts.

I then tried to streamline startup in PR #1036 so that each child account type would know how to start itself up after the main account had logged in, plus this also provided a startup point for local accounts (which previously had to be special-cased by account type in the startup code). At the time I made it so that the main account would start up itself and all of its child accounts when it logged in, but this conflicted with the startup code as that wants to log in and startup accounts separately, thus this PR.

So in summary:
- Local accounts initialise in `startup`.
- Standalone server-based accounts historically initialise in `login` and don't do anything in `startup`
  - for instance, MatrixAccount could rename `loginOnly` to `login` and move the initialisation code to `startup`
- Main Exchange email accounts *only* authenticate in `login`. Once that is successful, initialisation of it and all of its child accounts happens in `startup`.
- IMAP happens to authenticate in `login` and initialise in `startup`.

Changes made by this PR:
- Email accounts no longer automatically run `startup` after `login`.
- UI now runs `startup` after `login` for those accounts that need it (i.e. email accounts).

Effects of this  PR:
- On app startup all accounts are still logged in and started up, but now email accounts only start up once, because their login isn't triggering a second startup.
- On UI login accounts are started up but via `loginAndStartup` rather than relying on `login` running `startup`.
- Accounts that don't have their own `startup` are unaffected.

Ideal goal:
- Login is just whatever is needed to turn `isLoggedIn` from `false` to `true`.
- Startup is everything else.

While checking I did find a bug in the ActiveSync 14.0 support, but I'll file a separate PR for that.

benbucksch:
Thanks for this summary, Neil. This is the kind of discussion and explanation that I needed.

> the main account would start up itself and all of its child accounts when it logged in,

Yes, that makes sense to me.

> but his conflicted with the startup code as that wants to log in and startup accounts separately, thus this PR.

I don't understand this part, specifically the words "startup code" and "startup accounts separately".

Likewise, below that, you use terms that I don't know exactly what you mean with that, e.g. "initialize".

> Main Exchange email accounts only authenticate in login.

Given that EWS is a REST kind of protocol, what does "authenticate" mean, given that you authenticate with every single server request.

> Once that is successful, initialisation of it and all of its child accounts happens in startup.

(ditto, what is "initialisation"?)

I think the issue is still centering around an unclear definition of what "init", "login" and "startup" mean. I'll make an attempt in my next comment.

benbucksch:
Here's the definition I run with: (REVISED below)
* login() is the preparation work which allows me to issue any normal call to the server.
  * Examples:
    * For IMAP, that means creating the TCP connection, logging in, doing CAPABILITIES etc.
    * For JMAP, that means getting the Session object, account information etc.
    * For EWS, that means starting the stream listeners. If the dependent account needs its own stream listener, it should do that in login().
    * For ActiveSync, that includes checking the server version #1312
    * For all mail accounts, that includes listing mail folders.
  * Rules:
    * There are no network calls to the server before login().
    * login() does all startup work until I can ask the server any question I want. That includes all server back and forth, until normal server calls can be issued.
* startup() are the calls that the UI could make at any time, but we choose to do automatically after a typical login, on startup or a login triggered by the user.
  * If it's clearer, we could rename `startup()` to `initialSync()`.
  * Examples
     * Fetching the inbox and getting the newest emails.
     * Receiving and sending the changes to contacts and calendar events.
   * Rationale: Earlier, we did explict calendar.listEvents() after login, now we renamed it and generalized it for all account types to startup(), but the purpose and timing is still the same.
   * Rule:
     * The startup() code can be skipped, and server commands can still be issued. That allows the UI to only issue a specific call, so it would do login(), then that call, and not do startup().
* The UI then makes sync or list or send or receive calls as it pleases, in any order.

benbucksch:
So, if you need to do any of the work falling under the scope of `login()` for every dependent account, then do that in the `login()` function of the dependent account, after ensuring mainAccount.login() .

NeilRashbrook:
> > but this conflicted with the startup code as that wants to log in and startup accounts separately, thus this PR.
> 
> I don't understand this part, specifically the words "startup code" and "startup accounts separately".

"startup code" basically means `loginOnStartup` in `startup.ts`.

> Main Exchange email accounts only authenticate in login.
> 
> Given that EWS is a REST kind of protocol, what does "authenticate" mean, given that you authenticate with every single server request.

For those accounts not using OAuth2, it's already authenticated; `isLoggedIn` returns `true`. It always has.

> For EWS, that means starting the stream listeners. If the dependent account needs its own stream listener, it should do that in login().

I'm not seeing how that would work, given that you can only log in one dependent account, after which all of the accounts now think they're logged in.

Is there a UI point which should run login and initial sync for it and all dependent subaccounts, for an account which was unable to login on startup?

There was one startup task which you didn't mention, so I would also like to know at what point you want to start polling the Inbox (or whatever the equivalent for that account is)?

benbucksch:
Am I reading this right that with this PR, we won't fetch new mail on startup anymore? see startup.ts (also the webmail version of it).

NeilRashbrook:
Before the PR, we could try to fetch new mail twice on startup. After the PR, we will only fetch it once.

(I'm still hoping for an answer to my two questions, so I can properly separate login, tasks that should happen once per account after login, and initial sync.)

benbucksch:
> Is there a UI point which should run login and initial sync for it and all dependent subaccounts, for an account which was unable to login on startup?

The [Login] button above the msg list when you're logged out, or the [Get Mail] button, which also should do login first.

> at what point you want to start polling the Inbox (or whatever the equivalent for that account is)?

Ideally: After login and initial mail sync.

benbucksch:
> Here's the definition I run with:

I still need a clear answer from you on [those definitions](https://github.com/mustang-im/mustang/pull/1292#issuecomment-4861013957). We *first* need high-level definitions of what each function does, *independent* of protocol and dependent account and any technical mechanics. Only once we have that clear definition that these and future (!) questions can be answered on solidly, then we match that with our technical problems, and we get a straight answer. If the answer is not straight and clear, then the definition was not good.

benbucksch:
Here's the new definition discussed today with Neil:
* `login()` is the preparation work which allows me to issue any normal call to the server.
  * Examples:
    * For IMAP, that means creating the TCP connection, logging in, doing CAPABILITIES etc.
    * For JMAP, that means getting the Session object, account information etc.
    * For EWS, that does nothing.
    * For ActiveSync, that includes checking the server version #1312
  * Rules:
    * There are no network calls to the server before login().
    * `login()`  does all startup work until I can ask the server any question I want. That includes all server back and forth, until normal server calls can be issued.
  * When called:
    * Before `startupSync()`
    * A second parallel server connection to the same account would do `login()`, but not `startupSync()`
* `startupSync()` are the calls that the UI could make at any time, but we choose to do automatically after a typical login, on startup or a login triggered by the user. It also includes setting up listeners so that we are notified of new mails, calendar events etc..
  * Examples
    * For all mail accounts, that includes listing mail folders.
    * Fetching the inbox and getting the newest emails.
    * Receiving and sending the changes to contacts and calendar events.
    * For EWS, that means starting the stream listeners. If the dependent account needs its own stream listener, it should do that in login().
  * Rationale: Earlier, we did explict calendar.listEvents() after login, now we renamed it and generalized it for all account types to startup(), but the purpose and timing is still the same.
  * Rule:
    * The startup() code can be skipped, and server commands can still be issued. That allows the UI to only issue a specific call, so it would do login(), then that call, and not do startup().
  * When called:
    * After application startup
    * After manual login
    * After computer sleep for a longer time
* The UI then makes sync or list or send or receive calls as it pleases, in any order.

benbucksch:
Should now be fixed in:
* 1a5b4ff92 Mail: ActiveSync: Ask the server for its protocol version during the login
* 2c8b55ec5 Mail: OWA: Shared mailboxes listed their folders during the login
* 59ff01f31 EWS: Shared mailbox showed as not logged in
* 24982d36a Accounts: Split login() and startup() - fixes #1291
particularly the last one.
It is the same as this PR in sprit.

@NeilRashbrook Are these changes correct, esp 24982d36a ?

benbucksch:
* c8520bd3a Accounts: Rename startup() to syncOnStartup()

and still open:
* PR #1425

NeilRashbrook:
These changes seem correct, although see my https://github.com/mustang-im/mustang/pull/1425#issuecomment-5641118794 in case I've misunderstood something.

PR #1293, opened 2026-06-26
Contacts: newPerson store instead of adding to the persons collection #1287

jermy-c:
- store newPerson store instead of adding to the persons collection
- The UI is right below the action bar but right above the search which seems odd
- I didn't add an icon beside it because it looked odd
- As mentioned in https://github.com/mustang-im/mustang/pull/1287#issuecomment-4811984394

### DIFF
--- a/app/frontend/Contacts/Person/PersonLine.svelte
+++ b/app/frontend/Contacts/Person/PersonLine.svelte
@@ -1,11 +1,8 @@
 <hbox flex class="person" class:selected={isSelected} on:click
   style="--account-color: {addressbook?.color ?? "transparent"};"
   >
-  {#if $$slots.icon}
     <slot name="icon" />
-  {:else}
     <PersonPicture {person} size={pictureSize} />
-  {/if}
   <vbox flex class="main">
     <hbox class="first-row">
       <hbox flex class="name {fontClass}">{$person.name}</hbox>
--- a/app/frontend/Contacts/Person/PersonLine.svelte
+++ b/app/frontend/Contacts/Person/PersonLine.svelte
@@ -1,7 +1,11 @@
 <hbox flex class="person" class:selected={isSelected} on:click
   style="--account-color: {addressbook?.color ?? "transparent"};"
   >
-  <PersonPicture {person} size={pictureSize} />
+  {#if $$slots.icon}
+    <slot name="icon" />
+  {:else}
+    <PersonPicture {person} size={pictureSize} />
+  {/if}
   <vbox flex class="main">
     <hbox class="first-row">
       <hbox flex class="name {fontClass}">{$person.name}</hbox>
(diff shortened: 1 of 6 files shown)


app/frontend/Contacts/Person/Selected.ts
```diff
@@ -3,6 +3,7 @@ import { ArrayColl } from "svelte-collections";
 import { writable } from "svelte/store";
 
 export const selectedPerson = writable<Person>(null);
+export const newPerson = writable<Person>(null);
```
benbucksch: move above `lastPersons`, not close to `selectedFoo`

app/frontend/Contacts/PersonsToolbar.svelte
```diff
@@ -25,17 +25,24 @@
       />
   </hbox>
 </hbox>
+{#if $newPerson}
+  <hbox class="new-person">
+    <PersonLine person={$newPerson} isSelected {pictureSize} />
```
benbucksch: Add `SparklesIcon` for this person.

benbucksch (changes requested):
Looks nice and minimal

benbucksch:
> The UI is right below the action bar but right above the search which seems odd

That's fine

benbucksch:
> I didn't add an icon beside it because it looked odd

Can you show screenshots, please?

jermy-c:
> Can you show screenshots, please?

image

benbucksch:
Yeah, that's in the wrong spot. It should show in the same place as the avatar.

That is: in the gap before the name.

benbucksch:
In PersonLine.svelte, add:
```
  <slot name="icon" />
  <PersonPicture {person} size={pictureSize} />
```

jermy-c:
> Yeah, that's in the wrong spot. It should show in the same place as the avatar.
> 
> That is: in the gap before the name.

image

app/frontend/Contacts/Person/PersonLine.svelte
```diff
@@ -1,7 +1,11 @@
 <hbox flex class="person" class:selected={isSelected} on:click
   style="--account-color: {addressbook?.color ?? "transparent"};"
   >
-  <PersonPicture {person} size={pictureSize} />
+  {#if $$slots.icon}
+    <slot name="icon" />
+  {:else}
+    <PersonPicture {person} size={pictureSize} />
+  {/if}
```
benbucksch:
```suggestion
    <slot name="icon" />
    <PersonPicture {person} size={pictureSize} />
```

benbucksch (changes requested):

app/frontend/Contacts/Person/PersonLine.svelte
jermy-c: I had to add condition because it should render an extra empty space.

image

Issue #1295, opened 2026-06-29
EWS: Handling HTTP errors


benbucksch:
```
Could not read error response from server,
message: Cannot read properties of null (reading 'querySelector'),
name: TypeError,
stack:
TypeError: Cannot read properties of null (reading 'querySelector')
    at new EWSError 
    at EWSAccount.callEWS
    at async repeat
    at async EWSAccount.verifyLogin
    at async checkConfig
```

Looking at `EWSAccount` and `EWSError`, it seems to be way off. I can see at least 3 big problems:

1. `callEWS()`
`response.responseXML = this.parseXML(await response.text());`
*before* checking the HTTP status code. With HTTP 401 etc, that call is likely to fail, right?

2. `EWSError` ctor starts with `let responseXML = aResponse.responseXML; let messageNode = responseXML.querySelector("Message");`, but if there was no XML (as is the case in a 401), this is going to fail as well. Yes, there's a try/catch around, but would make sense to add an `if` guard as well instead of provoking a very likely exception and then logging it.

3. `EWSError` ctor has duplicated code - large parts of it.

jermy-c:
### Reproduction

1. Setup on-premise Exchange account with EWS
2. Go to Mail
3. Select a folder

### Actual Result

- Folder's are loaded with messages loaded also
- A lot of 401s before the following error
- Error
```
Could not read error response from server TypeError: Cannot read properties of null (reading 'querySelector')
    at new EWSError (EWSError.ts:18:37)
    at EWSAccount.callEWS (EWSAccount.ts:375:18)
    at async repeat (EWSAccount.ts:372:16)
    at async EWSFolder.updateChangedMessages (EWSFolder.ts:102:20)
    at async EWSFolder.listMessages (EWSFolder.ts:60:12)
    at async EWSFolder.getNewMessages (EWSFolder.ts:375:19)
    at async EWSAccount.startup (MailAccount.ts:51:5)
    at async EWSAccount.ts:113:7
    at async RunOnce.runOnce (RunOnce.ts:16:14)
    at async EWSAccount.startup (EWSAccount.ts:112:5)
```
- Login failed error afterwards

PR #1298, opened 2026-06-29
Refactor EWSError #1295

NeilRashbrook:
1. Don't bother trying to parse XML on HTTP 401 errors. (Note that Exchange does return XML for HTTP 500; I don't know about other errors offhand.)
2. Check for non-null `responseXML` to avoid likely exception.
3. Unduplicate code:

```typescript
        let errorNode = responseXML.querySelector("[ResponseClass=\"Error\"]");
        if (errorNode) {
          let errorResponse = XML2JSON(errorNode) as any;
```

### DIFF
--- a/app/logic/Mail/EWS/EWSAccount.ts
+++ b/app/logic/Mail/EWS/EWSAccount.ts
@@ -352,7 +352,11 @@ export class EWSAccount extends ExchangeMailAccount implements EWSSubscribable {
     } finally {
       lock.release();
     }
-    response.responseText = await response.text();
+    try {
+      response.responseText = await response.text();
+    } catch (ex) {
+      response.responseText = "";
+    }
     this.fatalError = null;
     if (response.status == 200) {
       try {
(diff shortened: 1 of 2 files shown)


app/logic/Mail/EWS/EWSAccount.ts
```diff
@@ -352,10 +352,11 @@ export class EWSAccount extends ExchangeMailAccount implements EWSSubscribable {
     } finally {
       lock.release();
     }
-    response.responseXML = this.parseXML(await response.text());
+    response.responseText = await response.text();
```
benbucksch: This can throw as well

benbucksch (changes requested):

app/logic/Mail/EWS/EWSAccount.ts
NeilRashbrook: I thought it was just the XML parsing "error" that you wanted to avoid. Also, what's the problem if this throws anyway? It would be a network error, not a protocol error.

benbucksch: If it's a network error, we'd want a proper "HTTP error 403" error message, not "Could not read .text()" or something like that, right?

NeilRashbrook: HTTP is a protocol.

benbucksch:
Merged to master

PR #1299, opened 2026-06-29
Contacts: Reference newPerson instead of the oldPerson #1269

### LINKED ISSUE #1269: Contacts: Cannot delete (get rid of) a new contact

jermy-c:
- Contacts: Reference newPerson instead of the oldPerson in the old addressbook
- Fixes #1269 for remote addressbooks, but OWA is throwing a error `Id is malformed` for some reason

### DIFF
--- a/app/logic/Abstract/Person.ts
+++ b/app/logic/Abstract/Person.ts
@@ -154,7 +154,11 @@ export class Person extends ContactBase {
     await other.deleteIt();
   }
 
-  /** Person class needs to match account class, so need to clone. */
+  /** Person class needs to match account class, so need to clone.
+   * @returns reference to the new person created or the existing
+   * person if it was already in the new addressbook or the
+   * same addressbook type
+   */
   async moveToAddressbook(newAddressbook: Addressbook): Promise<Person> {
     if (this.addressbook == newAddressbook || !newAddressbook) {
       return this;
--- a/app/logic/Abstract/Person.ts
+++ b/app/logic/Abstract/Person.ts
@@ -155,9 +155,9 @@ export class Person extends ContactBase {
   }
 
   /** Person class needs to match account class, so need to clone. */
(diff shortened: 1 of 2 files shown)


app/logic/Abstract/Person.ts
```diff
@@ -155,24 +155,25 @@ export class Person extends ContactBase {
   }
 
   /** Person class needs to match account class, so need to clone. */
-  async moveToAddressbook(newAddressbook: Addressbook): Promise<void> {
```
benbucksch, line 158: Define the return value.  it's not obvious.

benbucksch (approved):

benbucksch:
> OWA is throwing a error Id is malformed for some reason

Can you please give some context (reproduction, actual result, expectes result, and what do other protocols do), so that @NeilRashbrook can answer?

app/logic/Abstract/Person.ts
jermy-c: committed c0425f02762cc75aaac02eb9dfdbe5f1e5f62965

benbucksch (approved):

benbucksch:
@jermy-c What about OWA now? Is it broken now? We still need the reproduction. Can you please file a bug (if not yet done) and reference it here? And see that it is fixed, please?

benbucksch:
Ah, it's #1300 . Please make a habit of cross-referencing the issues, as a linked comment,  otherwise it's nearly impossible to track.

UPDATE: I see now that there's an automatic link here,  but it's buried under all the other notification spam.

jermy-c:
> Ah, it's #1300 . Please make a habit of cross-referencing the issues, as a linked comment, otherwise it's nearly impossible to track. Note that titles are not linked.

Ok, sorry about that. Will do.

PR #1303, opened 2026-06-30
Calendar: Exchange: Read all-day events as using local midnight #1271

NeilRashbrook:
TODO:
- Should this go in shared code somewhere? If so, where?
- Need to check to see how we should send all-day events to the server
- Not tested yet.

### DIFF
--- a/app/logic/Calendar/EWS/ExchangeEvent.ts
+++ b/app/logic/Calendar/EWS/ExchangeEvent.ts
@@ -1,11 +1,10 @@
 import { Event } from "../Event";
 
 export class ExchangeEvent extends Event {
-  // ActiveSync 16.x gives all-day events as a UTC-style string with no time
-  // part and no time zone. This is special-cased in the ActiveSync code.
-  // Other Exchange protocols provide a time zone and the time in UTC that
+  // Except for ActiveSync 16.x, whose all-day events are special-cased in
+  // fromCompact(), Exchange provides a time zone and the time in UTC that
   // would be midnight in that time zone, but we need them to use local time.
-  protected fixupExchangeAllDayEvent() {
+  protected setAllDayTimeToLocalMidnight() {
     if (this.allDay && this.timezone) {
       this.startTime = new Date(this.startTime.toLocaleDateString('lt', { timeZone: this.timezone }) + "T00:00:00");
       this.endTime = new Date(this.endTime.toLocaleDateString('lt', { timeZone: this.timezone }) + "T00:00:00");
--- a/app/logic/Calendar/EWS/ExchangeEvent.ts
+++ b/app/logic/Calendar/EWS/ExchangeEvent.ts
@@ -1,4 +1,15 @@
 import { Event } from "../Event";
 
 export class ExchangeEvent extends Event {
+  // ActiveSync 16.x gives all-day events as a UTC-style string with no time
+  // part and no time zone. This is special-cased in the ActiveSync code.
+  // Other Exchange protocols provide a time zone and the time in UTC that
+  // would be midnight in that time zone, but we need them to use local time.
+  protected fixupExchangeAllDayEvent() {
+    if (this.allDay && this.timezone) {
+      this.startTime = new Date(this.startTime.toLocaleDateString('lt', { timeZone: this.timezone }) + "T00:00:00");
+      this.endTime = new Date(this.endTime.toLocaleDateString('lt', { timeZone: this.timezone }) + "T00:00:00");
+      this.timezone = null;
+    }
+  }
 }
--- a/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
@@ -73,7 +73,7 @@ export class ActiveSyncEvent extends ExchangeEvent {
     if (wbxmljs.Timezone) { // Omitted in 16.1 for all day events
       this.timezone = fromActiveSyncZone(sanitize.nonemptystring(wbxmljs.Timezone, null));
     }
-    this.fixupExchangeAllDayEvent();
+    this.setAllDayTimeToLocalMidnight();
     if (wbxmljs.Recurrence) {
       this.recurrenceRule = this.newRecurrenceRuleFromWBXML(wbxmljs.Recurrence);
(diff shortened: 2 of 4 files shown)


benbucksch:
This is when you read it from the server. What does "all day" with timezone mean? By definition of Exchange and Outlook?

Is this round--tripping correctly, or do we destroy data for Outlook and Exchange?

> Should this go in shared code somewhere? If so, where?

Would JMAPEvent also need that? If so, you can put it in `Event` class.

NeilRashbrook:
> This is when you read it from the server. What does "all day" with timezone mean? By definition of Exchange and Outlook?

Sorry, I should have copied out of #1271: when reading an all-day event from the server, Exchange can show it as being in a different time zone, in which case the start and end dates are midnight in that zone. (By comparison ICal doesn't allow all-day events to have a time zone.)

> Is this round--tripping correctly, or do we destroy data for Outlook and Exchange?

For invitations this isn't data we can change. For meetings this is in the TODO list.

> > Should this go in shared code somewhere? If so, where?
> 
> Would JMAPEvent also need that? If so, you can put it in `Event` class.

No, JMAP always uses `localDateTime`, which is preferred for all-day events (and the time component should be `T00:00:00` for these).

benbucksch:
> Should this go in shared code somewhere? If so, where?

Don't we have an Exchange base class now? Put it there, please.

benbucksch:
I see we have ExchangeCalendar, but not ExchangeEvent yet. Make a separate PR to add those (compare the commit that creates ExchangeCalendar et al), including ExchangePerson, Folder, EMail etc pp., and move shared code there.

Make that PR simple and straight forward, so that it can merge quickly without discussion or regression. After, you can base this PR on master with that change.

NeilRashbrook:
I double-checked and we were already sending all-day events using just local dates for EWS and OWA, which seems to work fine. I know EAS 16.1 wants a weird all-day event format which was implemented as part of EAS 16.1 support, and I think EAS 14.1 wants the UTC date of the local midnight, which we were also already supplying (I think EWS and OWA also support this format).

NeilRashbrook:
Now rebased after #1329, plus missed some changes to `OWAEvent.ts` that were overlooked in that PR.

app/logic/Calendar/EWS/ExchangeEvent.ts
```diff
@@ -1,4 +1,15 @@
 import { Event } from "../Event";
 
 export class ExchangeEvent extends Event {
+  // ActiveSync 16.x gives all-day events as a UTC-style string with no time
+  // part and no time zone. This is special-cased in the ActiveSync code.
+  // Other Exchange protocols provide a time zone and the time in UTC that
```
benbucksch: 1. // Exchange protocols (other than ActiveSync) ...
2. Remove the 2 comment lines above, and put them in ActiveSyncEvent
3. If the comment is true, why is ActiveSyncEvent calling fix() anyways? Don't explain as reply in github, but make it clear in the code comments in ActiveSyncEvent (not in fix(), because the latter contains the generic case)

app/logic/Calendar/EWS/ExchangeEvent.ts
```diff
@@ -1,4 +1,15 @@
...
+  // would be midnight in that time zone, but we need them to use local time.
+  protected fixupExchangeAllDayEvent() {
```
benbucksch: setAllDayTimeToLocalMidnight()

app/logic/Calendar/EWS/ExchangeEvent.ts
```diff
@@ -1,4 +1,15 @@
...
+    if (this.allDay && this.timezone) {
+      this.startTime = new Date(this.startTime.toLocaleDateString('lt', { timeZone: this.timezone }) + "T00:00:00");
```
benbucksch, line 9: I wish there was an API way, not going via time string parsing.

But Date has no timezone conversions other than toLocaleString(), does it?

If there's no straight forward way, you can leave it as-is.

It's not a performance issue, because there are not millions of all day events.

benbucksch (changes requested):
Looks tons better than before

Only minor comments

app/logic/Calendar/EWS/ExchangeEvent.ts
benbucksch: Explanation:

1. "fixup" is a black box, and worse, undefined what it does. The "how" (implementation) is up to the function,  but the "what" needs to defined.
2. Exchange is redundant, because this is ExchangeEvent.

app/logic/Calendar/EWS/ExchangeEvent.ts
NeilRashbrook: Correct, `toLocaleXXXXString` is the only API where you can choose the time zone. (But fortunately this is the easy case where you want to know what the time was in that zone.)

app/logic/Calendar/EWS/ExchangeEvent.ts
NeilRashbrook: Yes I realise I suck at naming things.

app/logic/Calendar/EWS/ExchangeEvent.ts
NeilRashbrook: I'll reply here anyway for convenience.

ActiveSync 14.x behaves the same way as EWS and OWA in that it gives us a time zone and a time that's midnight in that zone. ActiveSync 16.x is a special snowflake. It does not give a time zone, so nothing happens here, instead `fromCompact` deals with the time string, which is a time that is midnight UTC but should actually be interpreted as local time.

benbucksch (approved):

benbucksch:
@NeilRashbrook Sorry that this PR rotted. The conflict might be inherent, because there might be another fix for the same problem, not sure. If so, when merging, please pick the best solution.

NeilRashbrook:
It looks like 5ae0034 was landed on the 18th as ced6d1f and then other commits since caused the bitrot, but b863b72 would still apply cleanly, and I could rebase it if you want to use GitHub to merge this PR.

benbucksch:
> https://github.com/mustang-im/mustang/commit/b863b7216640f957809c4c9c6866808a8b089666 would still apply cleanly,

Merged on master

PR #1304, opened 2026-06-30
Settings: MailConnectedAccounts: Don't set calendars on every account update #1294

### LINKED ISSUE #1294: Settings: Mail: All accounts: Calendar section freezes Parula and breaks UI
### Reproduction

- Click on the "Calendar" section on the left sidebar
- Click on the "Mail" section on the left sidebar
- Click on the "Account settings" gear button beside "All accounts"
Image

### Actual result

1. Parula freezes for about 3-10 seconds
2. When Parula becomes responsive, the error message "https://svelte.dev/e/effect_update_depth_exceeded" appears in the Settings section

### Expected result

- Displays "All accounts" preferences (Account name, Workspace, Your name)

### Error console
Image

### Related

1. Must go to the "Calendar" section before the "Mail" section to trigger the bug
2. Restarting Parula clears the error message that appears in the "Settings" section
3. After restarting Parula, the "Account settings" gear button beside "All accounts" work as expected, until we visit the "Calendar" section

jermy-c:
- The calendars collection was being updated on every account update, especially whenever `account.calendar` was updated because of the binding
- https://github.com/mustang-im/mustang/blob/0b9ecfc23038ab90e83af8d6312ce8b696efb2cd/app/frontend/Settings/Mail/ConnectedAccounts.svelte#L8
- It happened more because the collection was empty which triggers setting `account.calendar` to `undefined` and since it's `undefined` there will be an infinite loop, however placing a condition to not set it if the collection is empty didn't help.
https://github.com/mustang-im/mustang/blob/0b9ecfc23038ab90e83af8d6312ce8b696efb2cd/app/frontend/Shared/AccountDropDown.svelte#L62-L65
- Fixes #1294

### DIFF
--- a/app/frontend/Settings/Mail/ConnectedAccounts.svelte
+++ b/app/frontend/Settings/Mail/ConnectedAccounts.svelte
@@ -24,7 +24,12 @@
 
   export let account: MailAccount;
 
-  $: calendars = account.calendarsAvailable;
+  let lastAccount = account;
+  let calendars = account.calendarsAvailable;
+  $: if (account !== lastAccount) {
+    lastAccount = account;
+    calendars = account.calendarsAvailable;
+  }
 </script>
 
 <style>
--- a/app/frontend/Settings/Mail/ConnectedAccounts.svelte
+++ b/app/frontend/Settings/Mail/ConnectedAccounts.svelte
@@ -27,8 +27,8 @@
   let lastAccount = account;
   let calendars = account.calendarsAvailable;
   $: if (account !== lastAccount) {
-    lastAccount = account;
     calendars = account.calendarsAvailable;
+    lastAccount = account;
   }
 </script>
 


jermy-c:
Maybe we should also have a condition here to prevent loops? Because it causes an update regardless if it's the same value. But it doesn't help for this case.
https://github.com/mustang-im/mustang/blob/0b9ecfc23038ab90e83af8d6312ce8b696efb2cd/app/frontend/Shared/AccountDropDown.svelte#L62-L65

benbucksch:
> Maybe we should also have a condition here to prevent loops? Because it causes an update regardless if it's the same value

If you are certain that this is the cause for loops, then add the protection, yes.
Otherwise, no.

jermy-c:
> If you are certain that this is the cause for loops, then add the protection, yes. Otherwise, no.

Turns out the root cause is not having the protection of checking whether or not there are items in `showAccounts` which keeps setting `selectedAccount` to `undefined` which tiggers `defaultSelection`.

Better fix #1308

jermy-c:
Better fix #1308 is merged and it should fix #1294

Issue #1300, opened 2026-06-29
OWA: Contacts: Id is malformed in PR #1299


jermy-c:
### Reproduction

1. Go to Contacts
2. Click [+] to create a new contact
3. Select an OWA addressbook
4. Input a name
5. Click save

### Actual Result

- `Id is malformed.`, yellow banner
```
POST https://outlook.office.com/owa/service.svc 400 (Bad Request) OWAAccount.ts:396
OWAError: Id is malformed.
    at _OWAAccount.callOWAShared (OWAAccount.ts:427:13)
    at async _OWAAccount.callOWA (OWAAccount.ts:373:12)
    at async OWAPerson.saveToServer (OWAPerson.ts:69:20)
    at async OWAPerson.save (Person.ts:58:5)
    at async Module.track_reactivity_loss (chunk-N7R74ERE.js?v=6c14ecbc:1765:15)
    at async save (NameBox.svelte:123:7)
    at async Module.track_reactivity_loss (chunk-N7R74ERE.js?v=6c14ecbc:1765:15)
    at async HTMLButtonElement.myOnClick (RoundButton.svelte:83:7)
```

### Expected Result

- Saves to the server
- No error

### Related

- NextCloud/CardDav works, when you click save there's the spinning wheel then it saves
- We had an error like this before #1118

NeilRashbrook:
I'm still waiting for Ben to tell me how to distinguish a new contact from an existing contact.

benbucksch:
Why is that a problem? You can check whether you have a server ID. Why does that not work?
Worst case, we do what Event does with Event.isNew. But maybe we don't even need that here. I don't really understand the problem.

NeilRashbrook:
> Why is that a problem? You can check whether you have a server ID. Why does that not work?

Contacts don't have a dedicated serverID field. (I tried giving them one in PR #968, but you said that was the wrong approach.)

> Worst case, we do what Event does with Event.isNew. But maybe we don't even need that here. I don't really understand the problem.

That doesn't work either; `save` calls `saveLocally` first, so `isNew` is always false in `saveToServer`.

benbucksch:
OK, then do what you suggested in #968, but at least document it in the variable definition that `null = new contact`.

benbucksch:
My main problem in #968 was that you removed `id`. If you just want to save the server-assigned ID in `pID`, that's fine with me. As said, please document the special value `null = new contact`.

I've approved #968

[...]

PR #1311, opened 2026-07-01
Exchange: Give contacts their own server/item/personaID

NeilRashbrook:
Supersedes #968. #1118 and #1300 should be fixed by this (i.e. works in local testing).

### DIFF
--- a/app/logic/Contacts/ActiveSync/ActiveSyncPerson.ts
+++ b/app/logic/Contacts/ActiveSync/ActiveSyncPerson.ts
@@ -8,12 +8,9 @@ import { parseOneAddress, type ParsedMailbox } from "email-addresses";
 export class ActiveSyncPerson extends Person {
   declare addressbook: ActiveSyncAddressbook | null;
 
-  get serverID() {
-    return this.id;
-  }
-  set serverID(val) {
-    this.id = val;
-  }
+  /** The ActiveSync ServerId,
+   * or the empty string if the item has not been saved to the server. */
+  serverID = "";
 
   fromWBXML(wbxmljs: any) {
     this.firstName = sanitize.nonemptystring(wbxmljs.FirstName, "");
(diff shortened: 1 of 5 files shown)


benbucksch:
There was one thing very good about #968: the DB column is called `pID` and defined as "protocol-specific ID". So, it makes sense to have that in a `Person.pID` variable, to match the DB column. And then use `pID` to store the protocol-specific server-side ID, depending on protocol, like we do in other classes as well. I don't see a reason why `Person` should be different.

benbucksch:
Still missing the `isNew()` getter, see https://github.com/mustang-im/mustang/pull/968#issuecomment-4854945247

jermy-c:
This PR is needed for creating contacts in OWA and EWS. I can't create any contact in OWA or EWS at the moment.

benbucksch:
@jermy-c This PR was waiting for Neil. See my last comment.

benbucksch:
(Sorry, waiting for Neil)

benbucksch:
Merged now, due to need

jermy-c:
> @jermy-c This PR was waiting for Neil. See my last comment.

Sorry, I just wanted to mention that creating contacts isn't working at all and this could fix it when it's ready.

NeilRashbrook:
Sorry, I don't think `isNew` makes sense in this PR (I had added it to #968 though).

benbucksch:
Good, thanks.

benbucksch:
OWA contacts lose their server ID across restarts: toExtraJSON writes personaID but fromExtraJSON reads json.serverID (OWAPerson.ts:119/124, same in OWAGroup.ts). Every restart treats all stored contacts as deleted and re-creates them; edits between restart and re-list fail.

benbucksch:
Fixed in git commit dfd9cbc7a

PR #1312, opened 2026-07-01
ActiveSync: Always read protocol version on startup

NeilRashbrook:
Reading the protocol version is startup init code, not authentication code. In the case of (say) a bexchange account, the account could be added and would work, but after a restart it would fail because the protocol version was unset.

### DIFF
--- a/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
+++ b/app/logic/Mail/ActiveSync/ActiveSyncAccount.ts
@@ -77,6 +77,14 @@ export class ActiveSyncAccount extends ExchangeMailAccount {
         });
         await this.oAuth2.login(interactive);
       }
+
+      await this.startup();
+    });
+  }
+
+  async startup() {
+    await this.startupRunOnce.runOnce(async () => {
+      // This needs to run first before trying to sync.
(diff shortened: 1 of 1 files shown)


benbucksch:
I don't understand the fix. You move the code from login to startup. So, you will have cases where it now will no longer run. But there are no cases where it now runs and didn't run before.

`login()` always runs before `startup()`, by definition, right? That's how it is defined right now. If we want to change that, we need to have a clear, explicit discussion about that precise point.

That's one of the reasons why I asked you to suggest a *very precise* definition of login() and startup() in #1292. Startup order stuff like this cannot be left undocumented ("this is how the code works right now"), otherwise it will lead to utter chaos and hairball code and startup races that are impossible to debug.

benbucksch:
This breaks my understanding of the meaning of login():
https://github.com/mustang-im/mustang/pull/1292#issuecomment-4861013957

NeilRashbrook:
> you will have cases where it now will no longer run. But there are no cases where it now runs and didn't run before.

Right now these are both incorrect, but given your commit in #1292 I'll address that in separate PRs.

benbucksch:
Fixed in 1a5b4ff92c33b5

Issue #1305, opened 2026-07-01
Mail: Mail compose fails on “Edit as new” / “Forward” when removing video attachment TypeError: Failed to execute 'readAsDataURL' on 'FileReader':..


elenaux:
App: Parula Mail
Version: 0.9.31‑dev
Platform: macOS 

**Steps to reproduce**

1. Go to Mail → Sent.
2. Open an email you sent a few minutes ago that includes:
3. At least one regular file attachment (e.g. PDF invoice), and
4. A video attachment (e.g. screen recording).
5. Click the “…” message context menu in the toolbar.
6. Choose “Edit as new” (same also happens with “Forward”).
7. The message opens in the composer with subject, body, recipients, and attachments pre-filled.
8. In the compose view, delete the video attachment while leaving the other file attachment in place.
9. Press Send.

**Actual behavior**

A yellow error banner appears at the top of the compose window:
Failed to execute 'readAsDataURL' on 'FileReader': parameter 1 is not of type 'Blob'.

The message is not sent.
The send button remains available, but re-trying produces the same error.

**Expected behavior**

- The message should send successfully after removing the video attachment.
- If an attachment cannot be processed, the UI should:
  - Highlight the problematic attachment,
  - Provide a clear human-readable explanation (e.g., “We could not include this video attachment; please reattach it from disk”), and
  - Offer a safe path to send the email.

<img width="1600" height="963" alt="Image" src="https://github.com/user-attachments/assets/8ac65df5-c5c7-4d08-b754-ce9b1c050ceb" />

<img width="1600" height="980" alt="Image" src="https://github.com/user-attachments/assets/0a25c3c0-890b-421b-bbc3-a1494c3ccdc0" />

<img width="1076" height="448" alt="Image" src="https://github.com/user-attachments/assets/301d0cdd-8842-4e6a-b39e-6784e7de851e" />

jermy-c:
The attachments from the previous were not loaded because `loadAttachments()` only read the metadata from the DB. #1313 should fix it at least from my tests it does.

benbucksch:
@jermy-c Do you think #1313 was the root cause and fixed this bug here?


Either way, the error handling needs to be fixed as well. There might be other situations why the file is null. Elena is right that we need to give a better error message. An `assert(attachment.content, gt...);` in the right place might be sufficient.

jermy-c:
[...]

PR #1313, opened 2026-07-01
Mail: Attachments: Load attachment.content before saving #1307, #1305

### LINKED ISSUE #1307: Mail: Attachment saved via “Save as…” opens empty on macOS
**Steps to reproduce**

1. Open an email that contains at least two attachments.
2. Open the attachment menu (for example, via a three‑dot context menu) for one attachment.
3. Choose “Save as…”.
4. Enter a filename and confirm.
5. In the file manager (Finder) open the saved file with any application that can handle that file type.

**Expected result**
The saved file has the same size and content as the original attachment and opens normally.

**Actual result (macOS)**
On macOS, the saved file appears with a generic or unknown‑type icon in Finder, shows a size of 0 bytes in its file properties, and opens as an empty file in multiple applications. Opening the same attachment directly in an external application from Parula or in finder works and displays the correct content.

**Comparison**
On Windows (Parula Mustang 0.9.32‑dev), all three actions work correctly:

Opening the attachment in an external app
Opening the attachment’s location in a folder/file manager
Saving the attachment to disk via “Save as…”

Saved files on Windows have the correct size and content and open normally, so the issue appears only on macOS.

Image

Image

jermy-c:
- We were using `SQLSourceEMail` to load the attachments which didn't load the content
- The content was not loaded so an empty file was saved
- Fixes #1307, #1305

### DIFF
--- a/app/frontend/Mail/Message/AttachmentMenu.svelte
+++ b/app/frontend/Mail/Message/AttachmentMenu.svelte
@@ -27,7 +27,6 @@
 
 <script lang="ts">
   import type { Attachment } from "../../../logic/Abstract/Attachment";
-  import type { EMail } from "../../../logic/Mail/EMail";
   import ButtonMenu from "../../Shared/Menu/ButtonMenu.svelte";
   import MenuItem from "../../Shared/Menu/MenuItem.svelte";
   import OpenIcon from "lucide-svelte/icons/external-link";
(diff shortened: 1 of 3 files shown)


app/frontend/Mail/Message/AttachmentMenu.svelte
```diff
@@ -44,6 +45,10 @@
     await attachment.openOSFolder();
   }
   async function saveFile() {
+    if (!attachment.content) {
+      let email = attachment.message as EMail;
+      await email.loadAttachments();
+    }
```
benbucksch: Put this in logic in `Attachment.saveFile()`, please?

benbucksch (changes requested):

benbucksch (approved):
@jermy-c Can you please test the code again after my change?

PR #1319, opened 2026-07-02
EMail: Check for unLoaded attachments #1305

### LINKED ISSUE #1305: Mail: Mail compose fails on “Edit as new” / “Forward” when removing video attachment TypeError: Failed to execute 'readAsDataURL' on 'FileReader':..

jermy-c:
- Check for unloaded attachments after failing to load `.content` from `RawAttachmentsStorage` and `parseMIME`
- Mentioned in https://github.com/mustang-im/mustang/issues/1305#issuecomment-4860779778

### DIFF
--- a/app/logic/Mail/EMail.ts
+++ b/app/logic/Mail/EMail.ts
@@ -478,6 +478,8 @@ export class EMail extends Message {
       // fallback
       await this.loadMIME();
     }
+    let unLoaded = this.attachments.filterOnce(a => !a.content).map(a => a.filename);
+    assert(!unLoaded.hasItems, `Error loading attachments: ${unLoaded.join(", ")}`);
   }
 
   async loadBody() {


benbucksch:
maybe this is better in the UI. not sure that other callers would be happy about this. also, if only some attachments failed to load, shouldn't we give access to those that did load?

jermy-c:
> maybe this is better in the UI. not sure that ask callers would be happy about this. also, if only some attachments failed to load, shouldn't we give access to those that did load?

How about checking it on clicking the [send] button? And asking the use whether would like to continue without the attachment sent?

benbucksch:
No, on send is too late. Very annoying for user.

PR #1330, opened 2026-07-03
Contacts: Don't display contacts from other workspaces for [All] when workspace is selected

jermy-c:
- Contacts: Don't display contacts from other workspaces for [All] when workspace is selected

### DIFF
--- a/app/frontend/Contacts/ContactsAppD.svelte
+++ b/app/frontend/Contacts/ContactsAppD.svelte
@@ -37,8 +37,9 @@
     // This must be above the `$: persons` statement, so that `persons` will be adapted and then the search happens
   }
 
-  $: persons = (selectedAddressbook?.persons ??
-    appGlobal.persons.filterObservable(p => p.addressbook.workspace == $selectedWorkspace || !$selectedWorkspace)) as Collection<Person>;
+  $: persons = (selectedAddressbook?.persons ?? ($selectedWorkspace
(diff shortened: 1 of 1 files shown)


jermy-c:
There's the same problem for calendar events. I can see calendar events from other workspaces.

app/frontend/Contacts/ContactsAppD.svelte
```diff
@@ -36,7 +37,8 @@
     // This must be above the `$: persons` statement, so that `persons` will be adapted and then the search happens
   }
 
-  $: persons = (selectedAddressbook?.persons ?? appGlobal.persons) as Collection<Person>;
+  $: persons = (selectedAddressbook?.persons ??
+    appGlobal.persons.filterObservable(p => p.addressbook.workspace == $selectedWorkspace || !$selectedWorkspace)) as Collection<Person>;
```
benbucksch: Check other places where we do it.

First check selectedWorkspace is set, then either filter or return appGlobal.persons verbatim

benbucksch (changes requested):

benbucksch:
> There's the same problem for calendar events. I can see calendar events from other workspaces.

This is intentional. If you don't see your appointments, you miss them. Calendar events (display) are the only exception: workspaces do not hide them.

app/frontend/Contacts/ContactsAppD.svelte
jermy-c: Fixed in commit 5b85287120b1fcd4a60d022d4f672e3f1110123c

PR #1331, opened 2026-07-03
TestData: Use .soon() and .recent() for because minimum is 1 year for .future() and .past() #1318

jermy-c:
- TestData: Use .soon() and .recent() for because minimum is 1 year for .future() and .past()
- Found during testing Chat in Demo mode
- Followup for missed instances from #1318


jermy-c:
There's also, which I didn't include in this PR because I don't think it'll return a later date for `.endTime`:

https://github.com/mustang-im/mustang/blob/40d420a8487f0302f92dc33776fe787e72ea75b6/app/frontend/Meet/Start/start.ts#L56-L57

benbucksch (approved):

PR #1338, opened 2026-07-07
Files: Null check for file.account #1337

### LINKED ISSUE #1337: File: Cannot read properties of undefined (reading 'name')
### Reproduction

1. Setup n\*@b\*.com as OWA
2. Go to Files
3. Go to Persons tab
4. Select "Ben Test"
5. Select "dgh"
### Actual Result

- Right Pane shows an error with [Try again] button
- No file preview is shown
- Clicking [Try again] doesn't help
```
TypeError: Cannot read properties of undefined (reading 'name')
    at FileRightPane.svelte:48:66
    at Module.untrack (chunk-PBBPFN4W.js?v=0e5fc608:4545:12)
    at FileRightPane.svelte:98:15
    at update_reaction (chunk-PBBPFN4W.js?v=0e5fc608:4228:18)
    at execute_derived (chunk-PBBPFN4W.js?v=0e5fc608:2102:15)
    at update_derived (chunk-PBBPFN4W.js?v=0e5fc608:2120:15)
    at Module.get (chunk-PBBPFN4W.js?v=0e5fc608:4490:7)
    at get disabled (FileRightPane.svelte:46:28)
```

### Expected Result

- No error
- File preview is displayed

jermy-c:
- Files: file.account might be null especially for OWA if the account isn't loaded yet
- https://github.com/mustang-im/mustang/blob/da519c5d89da9823b9a43caa280816df5e92144e/app/logic/Files/File.ts#L40-L42

### DIFF
--- a/app/frontend/Files/RightSidePane/FileRightPane.svelte
+++ b/app/frontend/Files/RightSidePane/FileRightPane.svelte
@@ -45,7 +45,7 @@
         onClick={() => openFileInCloudApp(file)}
         icon={OpenCloudIcon}
         label={$t`Open with a cloud app on the web`}
-        disabled={!editors || editors.isEmpty ? $t`${file.account.name} does not offer a cloud app for this file type` : null}
+        disabled={!editors || editors.isEmpty ? $t`${file.account?.name} does not offer a cloud app for this file type` : null}
         iconSize="20px"
         padding="6px"
         classes="secondary"


benbucksch:
For null pointers, please don't just sprinkle `?` across the code base, but ask:

In which conditions is the account valid (!) to be null? Or should this normally be populated, but it's null due to a bug somewhere else? If the latter, fix the root cause, please.

More generally, always find and fix the root cause, not the symptom.

Even if this was the right fix:
* Which other places run into the same problem?
* This is a user visible string, so you need a translated fallback, so not for free.

benbucksch:
>     * https://github.com/mustang-im/mustang/blob/da519c5d89da9823b9a43caa280816df5e92144e/app/logic/Files/File.ts#L40-L42

Looking at the git blame, this came from:

https://github.com/mustang-im/mustang/commit/1bf0ef12387aec2ee64e0b2d3a26b258ce862272

You see that there are accesses `this.account.storage?.save()` . Note that there's no `?` after `account`,  even though there's one after `storage`, so either this is also a bug and hasn't been found yet, or the account is populated at this time of call. If the latter,  then why isn't it populated at the time of the UI view, given that account load should have completed by then?

Makes clear that this needs more investigation for the root cause.

jermy-c:
Ok, I'll investigate further.

jermy-c:
This is for #1337

jermy-c:
This is not the root cause indeed, it is: https://github.com/mustang-im/mustang/issues/1337#issuecomment-4919640440

benbucksch:
@jermy-c Do we still need this?

At the very least, you need a `?? ""` , otherwise you get a message "undefined does not offfer…"

jermy-c:
> @jermy-c Do we still need this?
> 
> At the very least, you need a `?? ""` , otherwise you get a message "undefined does not offfer…"

No, seems like it fixed by e4e14f76ca016dac39d71a90529773bcd153767a

PR #1346, opened 2026-07-17
PGP: Send: Protected headers in encrypted mails as m/mixed, like Thunderbird

benbucksch:
Wrap the encrypted part in multipart/mixed; protected-headers="v1" and move the message headers (esp. Subject) into it, so they are encrypted together with the body.


NeilRashbrook (approved):
I found the comments somewhat confusing; the message you're encrypting has always had the headers, and the final message has only ever had minimal headers. As such, `protectHeaders` does no such thing, instead what it does is wrap the message in an container which makes it possible for Thunderbird to find the protected headers.

However apart from the names and comments there does not seem to be a problem with the code itself.

PR #1351, opened 2026-07-19
ActiveSync: 4 fixes

benbucksch:
See the individual commits.

Please adopt the PR. Feel free to modify or drop changes as you see fit.

### DIFF
--- a/app/logic/Contacts/ActiveSync/ActiveSyncPerson.ts
+++ b/app/logic/Contacts/ActiveSync/ActiveSyncPerson.ts
@@ -51,7 +51,7 @@ export class ActiveSyncPerson extends ExchangePerson {
   }
 
   async saveToServer() {
-    let fields: Record<string, string | { Type: string, Data: string | {} }> = {
+    let fields: Record<string, any> = {
       FirstName: this.firstName,
       LastName: this.lastName,
       Body: {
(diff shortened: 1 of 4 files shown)


app/logic/Contacts/ActiveSync/ActiveSyncPerson.ts
```diff
@@ -51,7 +51,7 @@ export class ActiveSyncPerson extends ExchangePerson {
   }
 
   async saveToServer() {
-    let fields: Record<string, string | { Type: string, Data: string | {} }> = {
+    let fields: Record<string, any> = {
```
NeilRashbrook, line 54: Technically this should be
```suggestion
    let fields: Record<string, string | Record<string, never> | { Type: string, Data: string | Record<string, never> }> = {
```
but that's a bit of a mouthful...

NeilRashbrook (approved):
- 2f12157: Nice oversight; I must have only checked Calendar and not Contacts for this.
- 5e1447e: Nice fix; it only throws the first error but I'm not sure we can improve on that.
- 9c74ec3: Nice oversight; I knew ActiveSync 16.x saved exceptions differently, and I had at least fixed deleting exceptions, but I had left out this crucial step when saving them.
- 4426041: I had actually meant to check this as part of #1303 but forgot about it, so thanks for finding it.

I don't know whether you were thinking of squashing any of these commits but 5e1447e is obviously unrelated to the others which are all 16.x support bugs.

PR #1352, opened 2026-07-19
CalDAV: 5 fixes

benbucksch:
See the individual commits.

I realize that you didn't write the CalDAV code. But some of the iCal code might look familiar to you.

Please adopt the PR. Feel free to modify or drop changes as you see fit.

### DIFF
--- a/app/logic/Calendar/CalDAV/CalDAVEvent.ts
+++ b/app/logic/Calendar/CalDAV/CalDAVEvent.ts
@@ -1,10 +1,11 @@
 import { Event } from "../Event";
 import type { CalDAVCalendar } from "./CalDAVCalendar";
-import { convertICalToEvent } from "../ICal/ICalToEvent";
+import { convertICalToEvent, convertICalContainerToEvent, parseDate } from "../ICal/ICalToEvent";
+import { ICalParser } from "../ICal/ICalParser";
 import { getICal } from "../ICal/ICalGenerator";
 import { sanitize } from "../../../../lib/util/sanitizeDatatypes";
 import { assertHTTPResponseOK } from "../../util/netUtil";
-import type { URLString } from "../../util/util";
+import { ensureArray, type URLString } from "../../util/util";
 import { gt } from "../../../l10n/l10n";
 import type { DAVObject } from "tsdav";
 import type { ArrayColl } from "svelte-collections";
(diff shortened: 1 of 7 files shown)


app/logic/Calendar/CalDAV/CalDAVEvent.ts
```diff
@@ -49,22 +93,32 @@ export class CalDAVEvent extends Event {
...
+    if (this.parentEvent) {
+      // Exceptions are stored in the master's ics file.
+      // `saveLocally()` already attached this event to the master.
+      await this.parentEvent.saveToServer();
+      return;
+    }
+    await this.calendar.login(false);
     this.calUID ??= crypto.randomUUID();
     let iCal = getICal(this);
+    await this.calendar.login(false);
```
NeilRashbrook: Duplicate added code.

app/logic/Calendar/ICal/ICalGenerator.ts
```diff
@@ -16,6 +16,15 @@ export function getICal(event: Event, method?: iCalMethod): string | null {
   }
   lines.push(["VERSION", "2.0"]);
   lines.push(["PRODID", `-//Beonex//${appName}//EN`]);
+  addVEvent(lines, event);
+  for (let exception of event.exceptions) {
+    addVEvent(lines, exception);
+  }
```
NeilRashbrook, line 22: Are we supposed to send exceptions and exclusions when inviting an attendee to a master event? In which case, that means that we might expect to receive invitations to recurring meetings that contain exceptions and exclusions? If that's true, `convertICalContainerToEvent` needs to be able to handle this by using the code that was put in `updateExceptions`, and we have to work out when to save these exceptions as part of accepting the meeting...

app/logic/Calendar/CalDAV/CalDAVEvent.ts
```diff
@@ -39,6 +42,47 @@ export class CalDAVEvent extends Event {
...
+    }
+    let master = vevents.find(v => !v.entries.recurrenceid);
+    for (let entry of ensureArray(master?.entries.exdate)) {
+      for (let value of entry.value.split(",")) { // one EXDATE line can hold multiple dates
+        let [exclusionTime] = parseDate({ value, properties: entry.properties });
+        if (exclusionTime) {
+          this.makeExclusionLocally(exclusionTime);
+        }
+      }
+    }
```
NeilRashbrook, line 83: I thought this was wrong because it gets called after `saveLocally`, so the new exclusions never get saved, but I had forgotten that PR #663 had changed that. It does mean that the event gets saved an extra time for each exclusion, but we don't have a `makeExclusionsLocally`, and that would then need to be called before the event was saved anyway.

NeilRashbrook (approved):
Seems fine if you remove the duplicate line of code. Should I file issues for those other comments?

app/logic/Calendar/CalDAV/CalDAVEvent.ts
benbucksch:
```suggestion
```

app/logic/Calendar/ICal/ICalGenerator.ts
benbucksch: I'm not sure. I guess the exceptions and exclusions should be part of the invititation to the recurring event, no? iCal can contain multiple objects, right?

benbucksch:
> Should I file issues for those other comments?

Yes, please, and fix them, please.

app/logic/Calendar/ICal/ICalGenerator.ts
NeilRashbrook: I have verified that exclusions are expected to be included. I haven't checked exceptions yet; they would work differently so I can start working on exclusions anyway.

PR #1353, opened 2026-07-19
OWA: Contacts: Lists: Notes

benbucksch:
Minor, trivial change


NeilRashbrook:
The distribution list notes fix is still technically correct, even though we don't have any UI for distribution lists...

PR #1358, opened 2026-07-23
Docs: Document translation system

jermy-c:
- Docs: Document translation system

Is there anything I'm missing or that is unclear?

### DIFF
--- a/docs/systems/translation/l10n.md
+++ b/docs/systems/translation/l10n.md
@@ -1,0 +1,123 @@
+# l10n
+
+`app/l10n` is module that extracts strings for translation from `.svelte`, `.js` and `.ts` files.
+
+## Why did we choose this?
+
+We initially used the [svelte-i18n-lingui](https://github.com/HenryLie/svelte-i18n-lingui) package for extracting and displaying the translations because:
+- it had a clean syntax `$t~Hi ${name}~` which better than other systems using `t("Hi ${name}", { name: "Jack" })`
+- it extracted strings from `.svelte` files most other libraries only support `.js` and `.ts`
+- it offer interpolation for Svelte components and HTML e.g. `<p><T msg={~Click #~}><a>${$t~here~}</a></T></p>`. The `#` is replaced with `<a>` link.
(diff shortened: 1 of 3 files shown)


benbucksch (approved):

docs/systems/translation/l10n.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Why did we choose this?
+
+We initially used the [svelte-i18n-lingui](https://github.com/HenryLie/svelte-i18n-lingui) package for extracting and displaying the translations because:
+- it had a clean syntax `$t~Hi ${name}~` which better than other systems using `t("Hi ${name}", { name: "Jack" })`
+- it extracted strings from `.svelte` files most other libraries only support `.js` and `.ts`
+- it offer interpolation for Svelte components and HTML e.g. `<p><T msg={~Click #~}><a>${$t~here~}</a></T></p>`. The `#` is replaced with `<a>` link.
+- it generated unique IDs which made the keys much shorter and the need to think a new ID for each string.
+
+~ is used instead of tidle in the docs because tidle cannot be escaped.
```
NeilRashbrook, line 13: You can quote the `` ` `` character in inline code by using ``` `` ``` delimiters, like this:
```suggestion
We initially used the [svelte-i18n-lingui](https://github.com/HenryLie/svelte-i18n-lingui) package for extracting and displaying the translations because:
- it had a clean syntax `` $t`Hi ${name}` `` which better than other systems using `t("Hi ${name}", { name: "Jack" })`
- it extracted strings from `.svelte` files most other libraries only support `.js` and `.ts`
- it offer interpolation for Svelte components and HTML e.g. ``<p><T msg={`Click #`}><a>${$t`here`}</a></T></p>``. The `#` is replaced with `<a>` link.
- it generated unique IDs which made the keys much shorter and the need to think a new ID for each string.
```

jermy-c: Ok, thanks. I wanted to do that but didn't how to.

Issue #1356, opened 2026-07-21
OWA, EWS: Contacts: Address instructions not saved


jermy-c:
### Reproduction

1. Set an OWA or EWS addressbook
2. Create a contact
3. Fill in a name
4. Click [Street Address] to add a street address
5. Fill in all fields including the optional ones and also instructions
6. Click save
7. Click [Sync] or restart
8. Select the contact again

### Actual Result

- The instructions are not saved and disappear after sync or restart
- After click [Save] it remains there until sync or restart
- The instructions stay there for other protocols

### Expected Result

- The instructions are saved
- You see the instructions after sync or restart

NeilRashbrook:
Exchange doesn't have a way of saving the instructions. I guess we could look for an existing address in our contact, and if so just synchronise the fields supported by Exchange to it?

benbucksch:
> look for an existing address in our contact, and if so just synchronise the fields supported by Exchange to it

I don't understand

benbucksch:
> Exchange doesn't have a way of saving the instructions

In MAPI, doesn't Exchange store everything as properties that the client says? In EWS, Imisn't there a way to store additional properties in an object?

NeilRashbrook:
Yes, you can use an extended field in both EWS and OWA to store additional data that only Mustang knows about, but that wouldn't help ActiveSync.

I looked up and the way to do this is to define the property as an `<ExtendedFieldURI DistinguishedPropertySetId="PublicStrings" PropertyName="customNameHere" PropertyType="String"/>`. Note that "PublicStrings" is a special distinguished property set id which works with shared mailboxes.

NeilRashbrook:
Except that doesn't work with OWA, because there we're using Personas and not Contacts...

benbucksch:
The other alternative would be to hide the corresponding UI. We'd need meta-data on the EWSPerson class which properties are not supported, and then the UI hides them.

NeilRashbrook:
_Originally posted by @benbucksch in https://github.com/mustang-im/mustang/issues/1363#issuecomment-5212872813_

> > > EWS supports only 3 street addresses, exactly those home, business, other? If I add 2 business addresses, what happens?
> 
> > Ever since the Exchange code was written, in all cases where Exchange partially supports a Mustang field, the part of the field that Exchange supports is saved to the server, and the rest is unceremoniously lost. This PR is the first to attempt to save additional Mustang fields that Exchange does not natively support.
> 
[...]

PR #1363, opened 2026-07-29
EWS: Contacts: Save street address instructions #1356

NeilRashbrook:


app/logic/Mail/EWS/Request/EWSCreateItemRequest.ts
```diff
@@ -19,7 +19,7 @@ export class EWSCreateItemRequest {
           array.push(value.t$Entry);
         }
       } else {
-        item["t$" + key] = value;
+        (item["t$" + key] ??= []).push(value);
```
benbucksch: Can you please untangle this? Try to avoid such expressions, please. Variables are for free and increase readability.

app/logic/Contacts/EWS/EWSPerson.ts
```diff
@@ -66,6 +66,17 @@ export class EWSPerson extends ExchangePerson {
     this.chatAccounts.replaceAll(ensureArray(xmljs.ImAddresses?.Entry).filter(entry => entry.Value).map(entry => new ContactEntry(sanitize.nonemptystring(entry.Value), "other")));
     this.streetAddresses.replaceAll(ensureArray(xmljs.PhysicalAddresses?.Entry).map(entry =>
       new ContactEntry(EWSPerson.ewsToStreetAddress(entry).toString(), PhysicalAddressPurposes[entry.Key])));
+    for (let property of ensureArray(xmljs.ExtendedProperty).filter(property => ["BusinessInstructions", "HomeInstructions", "OtherInstructions"].includes(property.ExtendedFieldURI.PropertyName) && property.Value)) {
+      let key = property.ExtendedFieldURI.PropertyName.replace("Instructions", "");
```
benbucksch:
```suggestion
      let key: "Business" | "Home" | "Other"  = property.ExtendedFieldURI.PropertyName.replace("Instructions", "");
```

Not for the compiler, but for the human reader. It's totally non-obvious.

*Please* make an effort to write readable code.

benbucksch (commented):
So, EWS supports only 3 street addresses, exactly those home, business, other?
If I add 2 business addresses, what happens?

NeilRashbrook:
> So, EWS supports only 3 street addresses, exactly those home, business, other? If I add 2 business addresses, what happens?

Ever since the Exchange code was written, in all cases where Exchange partially supports a Mustang field, the part of the field that Exchange supports is saved to the server, and the rest is unceremoniously lost. This PR is the first to attempt to save additional Mustang fields that Exchange does not natively support.

benbucksch:
The other alternative would be to hide the corresponding UI. We'd need meta-data on the EWSPerson class which properties are not supported, and then the UI hides the corresponding fields.

benbucksch:
> > EWS supports only 3 street addresses, exactly those home, business, other? If I add 2 business addresses, what happens?

> Ever since the Exchange code was written, in all cases where Exchange partially supports a Mustang field, the part of the field that Exchange supports is saved to the server, and the rest is unceremoniously lost. This PR is the first to attempt to save additional Mustang fields that Exchange does not natively support.

I was not aware before that were silently dropping information. This here convinced me that we should take a different approach: Disable the UI for fields that the protocol does not support.

Could you please make a list of data that our model expresses that EWS, OWA and ActiveSync cannot express?

The bar for "cannot express": We should make an effort to map and save the info. But not to the extent that we invent be properties, and then Outlook cannot display it.

Exception: IDs that we need for internal mapping, or other classifications (office, Matrix, etc). If the protocol is missing these and we really really need them, then that's fair, if it's worth the extra effort.

E.g.
1a. if EWS has no field for street address instructions, then we should hide the field in the UI, because Outlook won't show the info anyways, and we'd have data loss this way.
1b. However, if there is a field in the protocol that fits semantically, and Outlook shows it, then we use that.
2a. If Outlook can show only 1 business address, then we need to limit our UI to that.
2b. If Outlook can show multiple business addresses, and the protocol just lacks the differentiating ID field, then we add that ID field.

PR #1368, opened 2026-08-03
CI: Linux: Run apt-get update before installing libarchive-tools

jermy-c:
- Run apt-get update before installing libarchive-tools
- The workflow was installing from an outdated package list
- Fixes: https://github.com/mustang-im/mustang/actions/runs/30784999896/job/91600093714
- Successful run: https://github.com/mustang-im/mustang/actions/runs/30839644276/job/91773294583


benbucksch (approved):
Duh. thanks for finding this

PR #1370, opened 2026-08-04
S/MIME: More bug fixes

benbucksch:
Here are some more bugs that Fable found. I looked through the first few and they look reasonable to me, and like real bugs, e.g. only self-signed certs with right now, not actual CA certs. Can you please review the fixes? Look at the individual commits.

### DIFF
--- a/app/logic/Mail/Encryption/KeyUtils.ts
+++ b/app/logic/Mail/Encryption/KeyUtils.ts
@@ -43,7 +43,10 @@ export async function getPublicKeyByKeyID(id: string | null, email?: EMail): Pro
     }
   }
   if (email) {
-    return await readAutoCryptKeys(email);
+    let key = await readAutoCryptKeys(email);
+    if (key?.id == id) {
+      return key;
+    }
   }
   return null;
 }
--- a/app/logic/Mail/Encryption/SMIME/SMIMEASN1.ts
+++ b/app/logic/Mail/Encryption/SMIME/SMIMEASN1.ts
@@ -353,7 +353,7 @@ const RecipientInfo = define("RecipientInfo", function() {
           this.key("issuer").seqof(AttributeValue),
           this.key("serialNumber").int(),
         ),
-        subjectKeyIdentifier: this.explicit(0).octstr(), // TODO
+        subjectKeyIdentifier: this.implicit(0).octstr(), // TODO
       }),
       this.key("keyEncryptionAlgorithm").use(AlgorithmIdentifier),
       this.key("encryptedKey").octstr(),
@@ -415,7 +415,7 @@ export const SignerInfo = define("SignerInfo", function() {
         this.key("issuer").seqof(AttributeValue),
         this.key("serialNumber").int(),
       ),
-      subjectKeyIdentifier: this.explicit(0).octstr(), // TODO
+      subjectKeyIdentifier: this.implicit(0).octstr(), // TODO
     }),
     this.key("digestAlgorithm").use(AlgorithmIdentifier),
     this.key("signedAttrs").implicit(0).optional().use(Attributes),
--- a/app/logic/Mail/Encryption/SMIME/SMIMEASN1.ts
+++ b/app/logic/Mail/Encryption/SMIME/SMIMEASN1.ts
@@ -126,6 +126,10 @@ export const GeneralTime = define<number>("GeneralTime", function() {
   this.gentime();
 });
 
+export const UTCTime = define<number>("UTCTime", function() {
+  this.utctime();
+});
+
 /** An algorithm */
 export interface AlgorithmIdentifier {
   algorithm: string | number[];
--- a/app/logic/Mail/Encryption/SMIME/SMIMEASN1.ts
+++ b/app/logic/Mail/Encryption/SMIME/SMIMEASN1.ts
@@ -393,8 +393,7 @@ export const EnvelopedData = define("EnvelopedData", function() {
       this.key("encryptedContentInfo").seq().obj(
         this.key("contentType").objid(oids),
         this.key("contentEncryptionAlgorithm").use(AlgorithmIdentifier),
-        // RFC5652 says this is implicit?
-        this.key("encryptedContent").explicit(0).optional().octstr(),
+        this.key("encryptedContent").implicit(0).optional().octstr(),
       ),
       this.key("unprotectedAttrs").implicit(1).optional().use(Attributes),
     ),
--- a/app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
+++ b/app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
@@ -53,10 +53,16 @@ export class SMIMEReadProcessor extends EMailProcessor {
                 let key = await crypto.subtle.importKey("raw", symmetricKey, "AES-CBC", false, ["decrypt"]);
                 let decryptedContent = await crypto.subtle.decrypt({ name: "AES-CBC", iv: vector }, key, encryptedContent);
                 let mimeAsText = new TextDecoder().decode(email.mime);
-                let otherHeaders = mimeAsText.slice(0, mimeAsText.indexOf("\r\n\r\n")).split("\r\n").filter(header => !/Content-/i.test(header)).join("\r\n");
+                // Split on CRLF, but keep folded continuation lines (those starting with
+                // whitespace) attached to their header.
+                let otherHeaders = mimeAsText.slice(0, mimeAsText.indexOf("\r\n\r\n")).split(/\r\n(?![ \t])/).filter(header => !/^Content-/i.test(header)).join("\r\n");
                 email.wasEncrypted = true;
                 email.downloadComplete = false;
-                email.mime = new TextEncoder().encode(otherHeaders + "\r\n" + new TextDecoder().decode(decryptedContent));
+                let headerBytes = new TextEncoder().encode(otherHeaders + "\r\n");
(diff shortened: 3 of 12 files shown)


benbucksch:
Interop-breaking (verified against openssl cms):
- EnvelopedData.encryptedContent was declared EXPLICIT [0] instead of IMPLICIT [0] (the code even had a doubting comment). Encrypted mail was incompatible with every other CMS implementation, in both directions.
- The read side matched a recipient info's rid.issuer against the certificate's subject,
so decryption only ever worked with self-signed certificates mail encrypted to a
CA-issued cert was never decrypted.
- The sign side put the certificate subject instead of its issuer into the SignerInfo sid. OpenSSL rejected our signatures with "signer certificate not found"; our own reader
never noticed because it ignored the sid.
- Verification assumed certificates[0] is the signer; with chain certs first (legitimate
ordering), it used the wrong key and the padding exception broke parsing of the whole
message. Now looked up via the sid, with a fallback for our previously-signed mail.
- The signed part was digested after a UTF-8 decode/encode round-trip, which corrupts
8-bit content. valid signatures on ISO-8859-1 messages failed. New
parseMIMEDirectSubpartsBytes() digests the raw bytes.
- subjectKeyIdentifier rid/sid variants used the wrong tag mode, so SKI-keyed messages
(e.g. openssl cms -keyid, some Outlook setups) crashed the whole decode instead of being skipped.
 
Spec compliance and robustness:
* signed attributes now in canonical DER SET OF order (the ASN.1 encoder doesn't sort)
* signingTime as UTCTime per RFC 5652
* shaNwithRSAEncryption accepted as signature algorithm on verify
* closing MIME boundary accepted without trailing CRLF (our own PGP sender produces that)
* a failing mail processor no longer breaks message parsing
* decrypted messages are rebuilt without orphaning folded header lines, dropping innocent headers, or re-corrupting 8-bit bodies.
 
Periphery:
* signed-but-unencrypted mail silently lost its BCC recipients (cloneEMail didn't copy bcc)
* expired keys were used forever because obsolete was computed only at import. now derived live from expires
* the Settings encryption import panel couldn't be closed once opened (wrong variable in the button handler)
* the Autocrypt fallback returned a key without checking its ID
* importing the same key twice created persistent duplicates
* a stray debug console.log

app/logic/Mail/Encryption/SMIME/SMIMEASN1.ts
```diff
@@ -393,8 +393,7 @@ export const EnvelopedData = define("EnvelopedData", function() {
       this.key("encryptedContentInfo").seq().obj(
         this.key("contentType").objid(oids),
         this.key("contentEncryptionAlgorithm").use(AlgorithmIdentifier),
-        // RFC5652 says this is implicit?
-        this.key("encryptedContent").explicit(0).optional().octstr(),
+        this.key("encryptedContent").implicit(0).optional().octstr(),
```
NeilRashbrook, line 396: Huh, I commented here because I thought I had been having problems with `implicit` and had to use `explicit` to get things to work. I'll re-test this.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
```diff
@@ -76,13 +76,21 @@ export class SMIMEReadProcessor extends EMailProcessor {
           console.log("signed data has no certificate and/or signature");
           return;
         }
+        let signerInfo = signedData.content.signerInfos[0];
```
NeilRashbrook, line 79: I would have moved the signature algorithm check up along with this line.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
```diff
@@ -97,7 +97,12 @@ export class SMIMEReadProcessor extends EMailProcessor {
           console.log("certificate does not contain an RSA public key");
           return;
         }
-        if (signerInfo.signatureAlgorithm.algorithm != "rsaEncryption") {
+        let digestAlgorithm = sanitize.translate(signerInfo.digestAlgorithm.algorithm, DigestAlgorithm);
+        // RFC 5652 section 10.1.2: signers write rsaEncryption, but verifiers
+        // should also accept e.g. sha256WithRSAEncryption, which some clients
+        // write instead. Its hash must match the digest algorithm.
+        if (signerInfo.signatureAlgorithm.algorithm != "rsaEncryption" &&
+            SignatureAlgorithm[signerInfo.signatureAlgorithm.algorithm] != digestAlgorithm) {
```
NeilRashbrook, line 105: I take it this is allowable since you're only comparing it?

app/logic/Mail/Encryption/SMIME/SMIMESend.ts
```diff
@@ -46,7 +46,8 @@ export class SMIMESend {
         attrValue: [OctetString.encode(messageDigest)],
       }, {
         attrType: "signingTime",
-        attrValue: [GeneralTime.encode(Date.now())],
+        // RFC 5652 section 11.3: MUST be UTCTime for dates through 2049
+        attrValue: [UTCTime.encode(Date.now())],
```
NeilRashbrook, line 50: Fair enough, I hadn't seen that clause. Let's hope nobody's still using Mustang 0.9.35 in 2050 😉

app/logic/Mail/Encryption/KeyUtils.ts
```diff
@@ -43,7 +43,10 @@ export async function getPublicKeyByKeyID(id: string | null, email?: EMail): Pro
     }
   }
   if (email) {
-    return await readAutoCryptKeys(email);
+    let key = await readAutoCryptKeys(email);
+    if (key?.id == id) {
+      return key;
+    }
```
NeilRashbrook, line 49: I can't comment on this change, because I have no idea what auto crypt keys are.

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
```diff
@@ -95,8 +95,7 @@ export class SMIMEReadProcessor extends EMailProcessor {
           console.log("signature was not signed with RSA");
           return;
         }
-        let digestAlgorithm = sanitize.translate(signerInfo.digestAlgorithm.algorithm, DigestAlgorithm);
-        let messageDigest = new Uint8Array(await crypto.subtle.digest(digestAlgorithm, new TextEncoder().encode(clearText)));
+        let messageDigest = new Uint8Array(await crypto.subtle.digest(digestAlgorithm, clearText as BufferSource));
```
NeilRashbrook, line 98: Ugh, `SharedArrayBuffer` making work for us... maybe there's a way to declare that we never use it and all `Uint8Array` objects have a proper `ArrayBuffer` and are therefore a valid `BufferSource`? (Apart from changing *all* of our uses from `Uint8Array` to `Uint8Array<ArrayBuffer>`, just to fix 20 TypeScript errors...)

NeilRashbrook (approved):
> Interop-breaking (verified against openssl cms):
> * EnvelopedData.encryptedContent was declared EXPLICIT [0] instead of IMPLICIT [0] (the code even had a doubting comment). Encrypted mail was incompatible with every other CMS implementation, in both directions.

Actually Thunderbird could read it correctly. On the other hand, I can't read a Thunderbird encrypted message, either with or without the patch. (Without the patch I get a ⚠️ with no message...)

> * The read side matched a recipient info's rid.issuer against the certificate's subject
> * The sign side put the certificate subject instead of its issuer into the SignerInfo sid. OpenSSL rejected our signatures with "signer certificate not found"; our own reader never noticed because it ignored the sid.

Oops.

> * signed attributes now in canonical DER SET OF order (the ASN.1 encoder doesn't sort)

Heh, I got the original order from nodemailer-smime, a plugin for nodemailer that uses node-forge to encrypt a message.

> * the Autocrypt fallback returned a key without checking its ID

I can't comment on this because it's PGP-specific.

benbucksch:
@NeilRashbrook Feel free to adopt this PR and change it to how you think it's correct. Just make sure that it's correct, please.

app/logic/Mail/Encryption/KeyUtils.ts
benbucksch: AutoCrypt is an email header containing the public key of the sender of the message. Opportunistic signing and encryption.

benbucksch: Details https://docs.autocrypt.org/level1.html

benbucksch:
@NeilRashbrook I need to make a release and will merge this as-is. Please address your comments and concerns in a new PR. Consider that one of your TODOs for today, please.

NeilRashbrook:
I looked at them again and I don't feel it's worthwhile making a PR for any of them at this point.

benbucksch:
Can you please test them, please? Esp. the implicit etc

NeilRashbrook:
I can't actually tell the difference; both `openssl smime -decrypt -inform der -in smime.p7m -inkey smime.key > smime.eml` and `openssl smime -verify -in smime.eml -CAfile ca.crt` are successful on messages created with or without these changes.

NeilRashbrook:
Google AI tells me that NSS is actually using BER encoding of large binary blobs, not DER encoding; this is apparently allowed for certain fields. However, it's likely that the asn1.js code doesn't handle BER at all; this might then explain why it can't read the Thunderbird data correctly. I'll have to research how the BER encoding differs from the DER encoding as it might be possible to tweak the asn1.js code to allow this field to decode as BER.

benbucksch:
> NSS is actually using BER encoding of large binary blobs, not DER encoding

See #1378, which has an implementation of BER -> DER converter

PR #1375, opened 2026-08-06
Build: Mac: Explicitly set the ZIP target for macOS #1374

### LINKED ISSUE #1374: Mac: Auto-update: ZIP file not found
### Reproduction

1. Install  `v0.9.35` on Mac
2. Go to Setting -> About
3. Click `Install Update`

### Actual Result

- Error
```
Error: ZIP file not provided: [
  {
    "url": "https://github.com/mustang-im/mustang/releases/download/v0.9.36/parula-0.9.36.dmg",
    "info": {
      "url": "parula-0.9.36.dmg",
      "sha512": "W5q3Y8eplQCu+T9uS3OHkh6yvv+OTArXojVDw+ppuT84sqBF/6e1ljajsMVPv7BugqU5E0CnVda0twKTZbQVTA==",
      "size": 233776072
    }
  }
]
```

### Expect Result

- Shows spinning wheel then restarts the app

### Related

- Might be because of https://github.com/mustang-im/mustang/pull/1369
- We might need to explicitly state the ZIP target but it is strange that the update `.yml` was created with the variable without a ZIP being generated.

jermy-c:
- Explicitly set the ZIP target for macOS
- We need to explicitly set the ZIP target now because the PKG target was added
- The ZIP target is the distribution that is downloaded for updates
- Fixes #1374 
- For regression #1369

### DIFF
--- a/desktop/electron-builder.yml
+++ b/desktop/electron-builder.yml
@@ -35,6 +35,9 @@ mac:
     - target: dmg
       arch:
         - universal
+    - target: zip
+      arch:
+        - universal
     # Requires CSC_INSTALLER_LINK and CSC_INSTALLER_KEY_PASSWORD secrets on GitHub CI
     # pkg is intentionally arm only
     - target: pkg


benbucksch (approved):
Was this built by default before, or was it always broken, or is it only necessary for pkg builds?

jermy-c:
> Was this built by default before, or was it always broken, or is it only necessary for pkg builds?

This was built by default before but the recent addition of the PKG target removed it based on my testing.

PR #1378, opened 2026-08-10
S/MIME: Read Outlook signed messages

benbucksch:
by Claude Fable. I have not reviewed the changes.

Please review carefully.

FYI, it includes a BER to DER converter that you were meaning to write.

### DIFF
--- a/app/logic/Abstract/Attachment.ts
+++ b/app/logic/Abstract/Attachment.ts
@@ -198,6 +198,7 @@ const kHiddenMIMETypes = [
   "text/vcard", // vCard
   "text/calendar", // vCard
   "application/pkcs7-signature", // S/MIME signature
+  "application/x-pkcs7-signature", // S/MIME signature, legacy type name
   "application/pgp-signature", // PGP signature
   "application/pgp-keys", // Sender announcing his PGP keys
   // "application/pkcs7-mime", // S/MIME encrypted
--- a/app/logic/Mail/Encryption/SMIME/SMIME.test.ts
+++ b/app/logic/Mail/Encryption/SMIME/SMIME.test.ts
@@ -1,6 +1,23 @@
 import { expect, test, describe } from "vitest";
-import { PrivateKeyInfo, RSAPrivateKey, RSAPublicKey, DigestInfo, Null, OctetString } from "./SMIMEASN1";
+import { PrivateKeyInfo, RSAPrivateKey, RSAPublicKey, DigestInfo, Null, OctetString, ContentInfo, SignedData, EnvelopedData, Certificate } from "./SMIMEASN1";
 import { BlockType, padFF, padRandom, encrypt, decrypt, unpadPKCS } from "./SMIMERSAES";
+import { verifySignedData } from "./SMIMEVerify";
+import { berToDER } from "../../../../../lib/asn1/ber";
+import { base64ToBytes } from "../../../../../lib/asn1/decoders/pem";
+
+// `verifySignedData()` compares digests using `indexedDB.cmp()`,
+// which the browser has, but Node does not.
+globalThis.indexedDB ??= {
+  cmp(a: Uint8Array, b: Uint8Array): number {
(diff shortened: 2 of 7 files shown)


benbucksch:
@NeilRashbrook Feel free to change the code at will without asking me. But add it as new commits on top of it, don't overwrite the commits, so that I have a trace of what you changed.

benbucksch:
1. e4794d777 S/MIME: Read: Convert streaming BER to DER before decoding. Outlook and OpenSSL emit CMS blobs as streaming BER (indefinite lengths, strings split into chunks), which our DER decoder couldn't parse at all — I confirmed this empirically with openssl cms output. New lib/asn1/ber.ts canonicalizes BER to DER: constructed OCTET/BIT STRINGs are flattened always, IMPLICITly-tagged chunked strings only when indefinite-length, so EXPLICIT wrappers in normal DER stay intact. This also fixes decrypting Outlook encrypted mail (its chunked encryptedContent was silently undecodable before).

2. b406dd4e7 S/MIME: Read: Support opaque-signed messages — the read processor now sniffs the CMS ContentInfo OID (instead of assuming application/pkcs7-mime = encrypted), unwraps the SignedData eContent so the message displays, and verifies the signature over it. Verification is extracted into SMIMEVerify.ts, shared with multipart/signed, and now also accepts signatures without signed attributes (RFC 5652 §5.4). The message is shown even when the signature fails — it just isn't marked as signed.

3. 5fbb5f61c S/MIME: Read: Accept the legacy MIME type `application/x-pkcs7-mime/-signature`, which Outlook still uses.

Verification: 14 vitest tests pass, including new openssl-generated fixtures:
* opaque-signed in DER
* streaming BER
* no-signed-attrs form (all verify against the embedded cert)
* a tampered-content rejection
* a streaming-BER encrypted message decrypted end-to-end through the chunk-flattening path
* truncated-input robustness (malformed mail must throw cleanly, not loop).

Pre-existing failures:
* svelte-check shows 217 errors vs. 219 on master baseline with no new type errors
* the S/MIME BufferSource complaints are the pre-existing lib.dom typing noise also present in untouched files.

Caveats / open issues:
* verification still re-encodes signedAttrs in our canonical SET order. A sender whose wire order differs could theoretically fail to verify (OpenSSL interop is fine).
* the kari (ECDH) recipient type

NeilRashbrook:
Somehow I had overlooked that CMS is mostly only required to be BER encoded, with only certain fields such as signedAttrs actually required to be DER encoded. So a BER to DER converter is in fact the way to go.

While trying to review this, I looked at what the imported asn1.js code was doing, and it does in fact try to handle indefinite lengths. However, the subarrays it creates include the end-of-indefinite-container marker, which causes problems because it looks like an implicit [0] type. It also sort of expects to see this marker when decoding lists, but this only works in partial decode mode. I think I should PR those fixes anyway, even if the BER to DER converter will mask those errors. In particular, it might then only be necessary for the converter to handle chunked strings, although as yet I don't know whether that's a significant change.

(With my asn1.js fixes, I was even able to kludge the uses and definition of `encryptedContent` to be `implicit(0).optional().seqof(OctetString)`, allowing it to read the Thunderbird message I was having problems with, but that only works specifically for that one chunked string.)

NeilRashbrook:
The code mostly looks fine to me by inspection, but I don't like the way `SMIMEReadProcessor` has to call `berToDER`; I think `DERDecoder.decode` would be a better call site, using a flag in `options` to decide whether to convert from BER. I'll write and test that change tomorrow.

benbucksch:
> I'll write and test that change tomorrow.

Thanks.

NeilRashbrook (approved):
Nice, that removes those ugly imports from `SMIMEReadProcessor.ts`. I also tested that the branch can read a Thunderbird self-encrypted message and the provided tests still pass after my changes.

benbucksch:
> can read a Thunderbird self-encrypted message

Yay

PR #1380, opened 2026-08-11
SMIME: ASN.1: Correctly return buffer for indefinite length container

NeilRashbrook:
Fixes the following issues:
- Checks that the end-of-indefinite-length marker is two zeros, not just one
- Does not include the end-of-indefinite-length marker when decoding the tag
  - Allows seqof and setof to end properly at the end of an indefinite length tag
  - Prevents the end-of-indefinite-length marker from accidentally decoding as an empty implicit [0] tag.


benbucksch (approved):

benbucksch:
Thank you

PR #1381, opened 2026-08-11
Router: selectedApp: Compare the correct location URL #1359

### LINKED ISSUE #1359: Mail composer: Does not close on first try after switching back to composer window
### Reproduction

- Open the message compose window by clicking the "+" icon in Mail
Image

- Click on the pencil icon on the left navigation bar
Image

- Click on the "X" icon to close the message compose window
Image

### Actual result

- The message compose window does not close
- A second click on the "X" icon closes the message compose window, as expected

### Expected result

- The message compose window closes on the first click

### Error console

- No error messages

### Related

- Opening multiple message compose windows make the "X" icon work as expected
- The problem occurs when there is only one message compose window open and the pencil icon is clicked before the "X"
- If "X" is clicked without clicking on the pencil icon first, the message compose window closes correctly, as expected

jermy-c:
- `window.location.hash` included the `#` at the start while `pageURL` did not which meant replace was never `true`
- The location history was never replaced only pushed.
- Fixes #1359

### DIFF
--- a/app/frontend/AppsBar/selectedApp.ts
+++ b/app/frontend/AppsBar/selectedApp.ts
@@ -25,8 +25,9 @@ export function openApp(app: MustangApp, params: PageParams) {
 }
 
 export function goTo(pageURL: string, params: PageParams) {
-  let replace = pageURL == window.location.hash;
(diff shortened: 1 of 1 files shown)


benbucksch (approved):

benbucksch:
Nice. Thanks.

PR #1388, opened 2026-08-13
Calendar: Events: Support attachments

### LINKED ISSUE #1384: Calendar: Attachments to events
### Use case

In a big org, when creating meetings, the organizer often attaches a bunch of files, which the participants will need for the meeting. E.g. an agenda, a document that will be discussed, or any other supporting materials or background info. So, this is important. The attachments may be large.

### Implementation

We need to support attachments for calendar events. Currently, there is no support for   
  it at all yet. We need, at least:                                                        
  * Add table in DB schema, and a migration for existing databases. (the create schema     
  should have the new table in there, and the migration checks whether it was already      
  done, with a cheap check.)                                                               
  * Add directory to save the files on disk                                                
  * Add class to save attachments, or reuse existing `RawFilesAttachment` class (with a    
  different file path)                                                                     
  * in DisplayEvent: UI to show the attachments in events                                  
  * in EditEvent: UI to add and remove attachments                                         
  * Support in iCal to read and write attachments                                          
  * Support in EWS and OWA (and possibly ActiveSymc) to read and write attachments         
                                                                                           
  Please implement that. Follow the code pattern from email attachments - including DB     
  schema, how to save and load them etc. Use the same `Attachment` class that we use in    
  Mail and Chat (which proves that it's generic). If possible, reuse `RawFilesAttachment`  
  storage class. Definitely reuse the UI components to display attachments (plural), and   
  the mail composer sidebar to add attachments, and the file drag & drop component in mail 
  composer (just that we have no inline attachments in calendar events, so all             
  attachments are real attachhments and that UI to decide between inline or not is         
  unnecessary).                                                                            
                                                                                           
  In DisplayEvent, show the attachments above the Description (or in case that there's no  
  Description, then in the place where the Description would be).                          
  In EditEvent, show the sidebar on the right of description (full screen event edit) or   
  (when in sidebar or small screens) after the description.                                
  In EditEvent, add an ExpanderButton for "(+) Attachment". When clicked, a "file open"    
  dialog appears immediately. If the user successfully selected a file, it will be added   
  to the event. Consequently, the sidebar with list of attachments (same as in mail        
  composer) will appear, because it appears whenever there are attachments.                
  Likewise, if the user drags a file from the desktop into the EditEvent dialog            
  (anywhere), it creates an attachment (without further queries or decisions), and the     
  attachments sidebar appears.                                                             
                                                                                           
  Lastly, the attachments should appear in Files app | Persons view, exactly like email    
  attachments. They are sorted by the person who is the organizer of the meeting.          
                                                                                           
  Make minimal changes, adapt to the code style, and reuse existing code patterns that you 
  find in the Mail component (both logic and frontend).

benbucksch:
Fixes #1384

Events can now have attachments, using the same `Attachment` class as Mail and Chat, and the same UI components.

- New `eventAttachment` table in the calendar DB schema, plus a migration that creates it in pre-existing databases, guarded by a cheap `sqlite_schema` lookup.
- The files are saved on the local disk under `files/calendar/`, sorted by the organizer of the meeting, by `RawFilesEventAttachment`, a subclass of the email `RawFilesAttachment`.
- `DisplayEvent` shows the attachments above the description.
- `EditEvent` has an "Attachment" expander button that opens the file dialog right away, shows the mail composer attachments sidebar as soon as there is an attachment, and accepts files dropped anywhere in the dialog.
- iCal reads and writes them as inline `ATTACH;VALUE=BINARY` (RFC 5545), which also covers CalDAV. `getICal()` is async now, because reading the file contents is.
- EWS and OWA read the meta-data together with the event, and fetch, create and delete the contents with `GetAttachment`, `CreateAttachment` and `DeleteAttachment`.
- ActiveSync reads them (16.0 and later, contents via `ItemOperations`). The protocol has no command to write attachments of a calendar item.
- They appear in Files | Persons, like the email attachments do.


app/logic/Abstract/Attachment.ts
```diff
@@ -39,7 +44,10 @@ export class Attachment extends Observable {
   protected _blobURL: URLString;
   /** Exists while editing or displaying. */
   dataURL: URLString;
+  /** The email or chat message that this attachment is part of. @see event */
   message: Message;
+  /** The calendar event that this attachment is part of. @see message */
+  event: Event;
```
NeilRashbrook: Would be nice if this could turn into `owner: AttachmentOwner;`.

app/logic/Abstract/Attachment.ts
```diff
@@ -198,13 +211,28 @@ const kHiddenMIMETypes = [
...
   "application/pgp-keys", // Sender announcing his PGP keys
   // "application/pkcs7-mime", // S/MIME encrypted
   // "application/pgp-encrypted", // PGP encrypted
 ];
 
+/** An email, chat message or calendar event that can have attachments */
+export interface AttachmentOwner {
+  dbID: number | string;
+  readonly attachments: ArrayColl<Attachment>;
+  newAttachment(): Attachment;
```
NeilRashbrook, line 217: Could add `loadAttachments?(): Promise<void>;`?

app/logic/Calendar/ActiveSync/ActiveSyncEvent.ts
```diff
@@ -115,6 +116,43 @@ export class ActiveSyncEvent extends ExchangeEvent {
...
     }
+    if (wbxmljs.Attachments != undefined) {
+      this.attachmentsFromWBXML(wbxmljs.Attachments);
+    }
+  }
+
+  /** Only the meta-data. The contents are fetched later,
+   * @see downloadAttachmentsFromServer()
+   * ActiveSync 16.0 and later. Read-only: ActiveSync has no command to add or
+   * remove attachments of a calendar item, so `toFields()` cannot write them. */
```
NeilRashbrook: 2.2.2.2 [Add](https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-asairs/d60d864c-0730-4b20-bd74-cc0245d4cdc6)
2.2.2.21 [Delete](https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-asairs/b9f6e6cd-fe71-4014-a148-cb21694f9a80)

(I'm not asking you to implement them.)

app/logic/Calendar/EWS/EWSCalendar.ts
```diff
@@ -214,6 +214,8 @@ export class EWSCalendar extends ExchangeCalendar implements EWSSubscribable {
               FieldURI: "item:LastModifiedTime",
             }, {
               FieldURI: "item:TextBody",
+            }, {
+              FieldURI: "item:Attachments",
```
NeilRashbrook: Although not required here, please move this to after "item:Body" to match the field order in the XML calendar item. (Also applies elsewhere, e.g. in `EWSEvent.ts`. Since attachments aren't written though XML, there isn't really an schema issue here, but I've been trying to keep this in schema order for consistency. For OWA, there isn't even a schema, but I've still been trying keep the property list in EWS order for consistency with EWS, so I would prefer it moved there too please.)

app/logic/Calendar/EWS/EWSEvent.ts
```diff
@@ -122,6 +123,81 @@ export class EWSEvent extends ExchangeEvent {
...
+  }
+
+  protected async downloadAttachmentsFromServer(): Promise<void> {
+    for (let attachment of this.attachments) {
+      if (attachment.content || !attachment.pID) {
+        continue;
+      }
+      let response = await this.calendar.account.callEWS({
+        m$GetAttachment: {
+          m$AttachmentShape: {},
```
NeilRashbrook: `AttachmentShape` is listed as being optional in the schema.

app/logic/Calendar/EWS/EWSEvent.ts
```diff
@@ -122,6 +123,81 @@ export class EWSEvent extends ExchangeEvent {
...
+      await this.calendar.account.callEWS({
+        m$DeleteAttachment: {
+          m$AttachmentIds: {
+            t$AttachmentId: {
+              Id: removed.pID,
+            },
+          },
+        },
+      });
+    }
```
NeilRashbrook: All the attachments can be removed in a single call:

```suggestion
    if (this.removedAttachments.hasItems) {
      await this.calendar.account.callEWS({
        m$DeleteAttachment: {
          m$AttachmentIds: this.removedAttachments.contents.map(removed => ({
            t$AttachmentId: {
              Id: removed.pID,
            },
          },
        })),
      });
    }
```

(In theory you can also add multiple attachments in a single call, but I don't know how easy that would be to organise, and you might not want to do that to keep each call's size down anyway.)

app/logic/Calendar/ICal/ICalToEvent.ts
```diff
@@ -144,6 +146,28 @@ export function convertICalContainerToEvent(vevent: ICalContainer, event: Event)
...
+  for (let entry of vevent.entries.attach ?? []) {
+    if (entry.properties.value?.toUpperCase() != "BINARY") {
+      continue;
+    }
+    let attachment = event.newAttachment();
+    attachment.mimeType = sanitize.nonemptystring(entry.properties.fmttype, "application/octet-stream");
+    attachment.filename = sanitize.filename(
+      entry.properties.filename ?? entry.properties["x-filename"] ?? entry.properties["x-apple-filename"],
+      "attachment-" + ++fallbackID + "." + fileExtensionForMIMEType(attachment.mimeType));
+    attachment.content = new File([Uint8Array.from(atob(entry.value), c => c.charCodeAt(0))], attachment.filename, { type: attachment.mimeType });
```
NeilRashbrook: I assume you don't want to use `base64ToArrayBuffer` like you do elsewhere because it's async. Would `Uint8Array.fromBase64` work instead, given that we now have a new enough Electron to support it? (Should we use `Uint8Array.fromBase64` in other places too?)

app/logic/Calendar/OWA/OWAEvent.ts
```diff
@@ -120,6 +121,57 @@ export class OWAEvent extends ExchangeEvent {
...
+      attachment.size = attachment.content.size;
+    }
+  }
+
+  /** Exchange cannot save attachments as part of the event,
+   * but needs separate calls, after the event exists on the server. */
+  protected async saveAttachmentsToServer(): Promise<void> {
+    for (let removed of this.removedAttachments) {
+      await this.calendar.callOWA(owaDeleteAttachmentRequest(removed.pID));
+    }
```
NeilRashbrook: Again, should be possible to do this in a single call.

app/logic/Calendar/Event.ts
```diff
@@ -499,6 +507,52 @@ export class Event extends Observable {
...
+    for (let attachment of this.attachments) {
+      await attachment.save();
+    }
+  }
+
+  /** Fetches the attachment contents from the server.
+   * Only needed for protocols that send the attachment metadata,
+   * but not the contents, together with the event. */
+  protected async downloadAttachmentsFromServer(): Promise<void> {
+  }
```
NeilRashbrook, line 546: throw NotReached here perhaps?

app/logic/Mail/Store/RawFilesAttachment.ts
```diff
@@ -109,19 +115,20 @@ export class RawFilesAttachment implements MailContentStorage {
...
     let fileparts = attachment.filename.split(".");
     let ext = fileparts.length > 1 ? fileparts.pop() : fileExtensionForMIMEType(attachment.mimeType);
     let filename = fileparts.join(".").substring(0, 40) + "-" + id + "." + ext;
     return `${dir}/${sanitize.filename(filename, "unknownAttachment")}`;
   }
 
-  async getDirPath(message: Message): Promise<string> {
+  async getDirPath(owner: AttachmentOwner): Promise<string> {
     filesDir ??= await getFilesDir();
+    let message = owner as Message;
```
NeilRashbrook: Since you don't actually use `message` as a `Message`:

```suggestion
  async getDirPath(message: AttachmentOwner): Promise<string> {
    filesDir ??= await getFilesDir();
    
```

NeilRashbrook (changes requested):
There is unfortunately an Exchange limitation when mixing attachment changes with attendee changes for the same event. As you know, attachments have to be changed separately to attendees. It is possible to create an event with attachments and attendees in a single API call, but I can't imagine it's recommended. In such a case, it might be better to create the event, add attachments, then add attendees. This avoids sending invitations twice.

app/logic/Abstract/Attachment.ts
benbucksch: Nice idea. Done

app/logic/Calendar/Event.ts
benbucksch: `AbstractFunction()`, rather, for functions that are supposed to be overridden or not called.

benbucksch (commented):
Thanks for your review comments. Fixed.

NeilRashbrook (approved):
(I quite liked the old `AttachmentOwner` name but what do I know about naming things?)

PR #1389, opened 2026-08-13
EWS: NTLM: Authenticate the TCP connection, not the HTTP request

### LINKED ISSUE #1387: EWS: NTLM randomly fails/succeeds with on-premise Exchange
# Objective

Fix NTLM authentication in EWS to consider the TCP connection that a HTTP request runs on.

# Background
Our code is async through-out, so highly parallelized (without actual threads), so we make many calls at the same time, esp. at login, when we check the mailbox, address books, calendars, subscribe to push mail, etc., and all that possibly for multiple accounts, if the user has delegated accounts. And after setup, we make many server calls to fetch the inbox content (can be tends of thousands of mails, addressbook, calendar events etc., which adds even more calls in short succession.

# Server environment
We tested EWS with NTLM auth with on-premise Exchange servers. Tested with 2 different Exchange 2019 servers, one on-prem, the other hosted by a third-party hoster. We have not noticed problems with Exchange on the Microsoft Office 365 cloud, but that Exchange server is a fork (different server flavor) and might react differently.

# Bug
Our NTLM implementation (for EWS) does not work reliably. It's randomly succeeding and failing. On startup, we have reproducibly multiple "401 Unauthorized" messages in the renderer devtools console, apparently printed on the console by Chromium, not us. We have an auto-repeat feature in EWS to react to 401, re-login, and re-try the server call, but even that doesn't work. Mayb 40% of the server calls are randomly failing. E.g. sometimes the list folders call fails and there are no folders (other than inbox and sent), sometimes it works. Sometimes the calendar list events call fails and there are no calendar events, sometimes it works. Sometimes the GAL call fails, and we see one 401 on the console per keypress.

# NTLM
The main problem appears to be that NTLM is a) a 3-step authentication process, so it can't be done on every server call like HTTP Basic and b) it has to be done one per *TCP* connection, so it depends on which TCP channel the HTTP request goes out on.

# Our execution environment
We are running in Electron, and we don't have direct control over the TCP connection. We just call fetch(). We currently do that in the Chromium renderer process. We previously had an implementation in the node.js main process (you should be able to find it in the git log), but that had the same bug. We moved it to renderer, because we want to have as little lines of code in the backend (backend.js and friends) as possible. However, it's OK to create a node.js or node.js library object in the backend and then drive it via JPC from the frontend code, we just prefer to keep all the *code* in app/logic/ in the frontend, not necessarily the execution.

# What we tried so far
We already tried the node.js net.fetch(), session.defaultSession.fetch() and similar APIs, but we're not getting the "login" event (in main) called, so all our attempts failed. Maybe we did something wrong. Maybe you have better luck, give it a try. session API is particularly useful, because it allows us create one session per mail account and to isolate different mail accounts from each other, also in terms of cookies (if a server were to ever do that) and whatnot. There's also http.Agent API, which we did not try yet.

# Requirements
Whatever we do, it has to be reliable. So, for NTLM, that means that we at least *know* which TCP connection a HTTP request runs on, so that we know whether to do auth or not. As mentioned, re-auth on demand after 401 seems to not be working reliably, and it's an extra round-trip (for the failure) as well, so we'd have 1 round-trip for the failure, another round-trip for the server challenge, and a third round-trip for the actual login, and (I'm not sure) another round-trip for the actual server call. Given that we do tons of calls, the code needs to handle that and be efficient and make the minimum amount of server requests (in the normal case).

Do *not* depend on a call pool. We will build a pool on top, but the underlying primitive has to be reliable. If we merely limit to 4 calls in parallel, it might work with our test server, and we think we fixed it, but then it won't at another customer deployment with slightly different settings or setup, e.g. with an authenticating load balancer in front of the Exchange server. The authentication needs to work reliably in all cases.

benbucksch:
Fixes #1387 

NTLM is a connection-based authentication: The 3-step handshake logs in the TCP connection, and only requests on that same connection are authenticated. Our calls went through the renderer's fetch(), which runs requests over an arbitrary connection of Chromium's socket pool. With many parallel calls, the handshake steps and the actual calls ended up on different connections, so calls randomly failed with 401, and even the 401 re-login retry was equally blind, causing missing folders, missing calendar events, and failing GAL searches.

New backend HTTPConnection: a node.js HTTP(S) client where one object guarantees at most one TCP connection (Agent with maxSockets: 1), and every response reports which connection served it (socketID) and whether the connection was reused. It also streams responses via an awaited chunk callback (backpressure), decompresses gzip, and trusts the OS certificate store in addition to node's bundled CAs.

New NTLMConnection (app/logic/Auth/NTLM/) drives the handshake on exactly that connection: Type 1 with an empty body, Type 3 riding on the first real request, so a fresh connection costs exactly one extra round trip, and later requests cost none and carry no Authorization header. Because every response reports its socketID, we always know whether our login applies. When the server closes the connection, drops the auth state (load balancer), or kills the connection while a request is on the wire (keep-alive race, detected via reusedSocket), we re-authenticate and retry, bounded, and a rejected login throws LoginError. NTLMConnectionPool runs parallel requests over several such connections; correctness never depends on the pool size. All connections of an account share a cookie jar, so load-balancer affinity cookies (e.g. X-BackEndCookie, ISA/TMG) keep working.

EWSAccount routes NTLM calls through the pool, including the streaming notifications, which get their own dedicated connection instead of hogging Chromium's 6-connections-per-host budget. Other auth methods are unchanged. This also stops Chromium from logging 401s on the console, and forces HTTP/1.1, which NTLM requires anyway.

Tested against a strict mock Exchange server that binds the challenge to the TCP connection and verifies the NTLMv2 proof with real crypto: parallel storm of 40 calls, server-closed connections, killed connections, dropped auth state, wrong password, streaming, cookies.

### DIFF
--- a/app/frontend/Settings/Mail/Account/Server.svelte
+++ b/app/frontend/Settings/Mail/Account/Server.svelte
@@ -8,7 +8,10 @@
     <ServerIMAPAdvanced {account} />
   {/if}
   {#if account instanceof EWSAccount && $account.authMethod == AuthMethod.NTLM}
+    <!-- Mobile has no Chromium network stack, so there is nothing to choose -->
+    // #if [!MOBILE]
     <ServerEWSAdvanced {account} />
+    // #endif
   {/if}
 </vbox>
 
@@ -21,7 +24,9 @@
   import { catchErrors } from "../../../Util/error";
   import ManualConfig from "../Manual/ManualConfig.svelte";
   import ServerIMAPAdvanced from "./ServerIMAPAdvanced.svelte";
+  // #if [!MOBILE]
   import ServerEWSAdvanced from "./ServerEWSAdvanced.svelte";
+  // #endif
   import { t } from "../../../../l10n/l10n";
 
   export let account: Account;
--- a/app/frontend/Settings/Mail/Account/Server.svelte
+++ b/app/frontend/Settings/Mail/Account/Server.svelte
@@ -7,15 +7,21 @@
   {#if account instanceof IMAPAccount}
     <ServerIMAPAdvanced {account} />
   {/if}
+  {#if account instanceof EWSAccount && account.authMethod == AuthMethod.NTLM}
+    <ServerEWSAdvanced {account} />
+  {/if}
 </vbox>
 
 <script lang="ts">
   import type { Account } from "../../../../logic/Abstract/Account";
+  import { AuthMethod } from "../../../../logic/Abstract/Account";
   import type { MailAccount } from "../../../../logic/Mail/MailAccount";
   import { IMAPAccount } from "../../../../logic/Mail/IMAP/IMAPAccount";
+  import { EWSAccount } from "../../../../logic/Mail/EWS/EWSAccount";
   import { catchErrors } from "../../../Util/error";
   import ManualConfig from "../Manual/ManualConfig.svelte";
   import ServerIMAPAdvanced from "./ServerIMAPAdvanced.svelte";
+  import ServerEWSAdvanced from "./ServerEWSAdvanced.svelte";
   import { t } from "../../../../l10n/l10n";
 
   export let account: Account;
--- a/app/frontend/Settings/Mail/Account/Server.svelte
+++ b/app/frontend/Settings/Mail/Account/Server.svelte
@@ -7,7 +7,7 @@
   {#if account instanceof IMAPAccount}
     <ServerIMAPAdvanced {account} />
   {/if}
-  {#if account instanceof EWSAccount && account.authMethod == AuthMethod.NTLM}
+  {#if account instanceof EWSAccount && $account.authMethod == AuthMethod.NTLM}
     <ServerEWSAdvanced {account} />
   {/if}
 </vbox>
--- a/app/logic/Auth/NTLM/NTLMResponse.ts
+++ b/app/logic/Auth/NTLM/NTLMResponse.ts
@@ -1,14 +1,3 @@
-import type { NTLMChromiumSession } from "./NTLMChromiumSession";
-import type { NTLMConnectionPool } from "./NTLMConnectionPool";
-
-/**
- * Our two alternative NTLM implementations:
- * - `NTLMChromiumSession`: Chromium's network stack does the NTLM login
- * - `NTLMConnectionPool`: our own NTLM implementation, over `HTTPConnection`
- * `EWSAccount` picks one, based on platform and user setting.
- */
-export type NTLMTransport = NTLMChromiumSession | NTLMConnectionPool;
-
 export interface NTLMRequestOptions {
   headers?: Record<string, string>;
   signal?: AbortSignal;
--- a/app/logic/Auth/NTLM/NTLMResponse.ts
+++ b/app/logic/Auth/NTLM/NTLMResponse.ts
@@ -4,29 +4,25 @@ export interface NTLMRequestOptions {
   onChunk?: (chunk: string) => Promise<void>;
 }
 
-/** Duck-typed like a `fetch()` `Response`, as far as `EWSAccount` needs it */
-export class NTLMResponse {
-  readonly status: number;
-  readonly statusText: string;
-  readonly ok: boolean;
+/** A `fetch()` `Response`, plus the TCP connection that it came in on */
+export class NTLMResponse extends Response {
+  /** Which TCP connection served this response. Only our own NTLM
+   * implementation knows it, and only it needs it. @see `HTTPConnection` */
   readonly socketID: number;
-  readonly headers: { get: (name: string) => string | null };
-  protected readonly bodyText: string;
 
+  /** @param response as the backend returned it over JPC */
   constructor(response: any) {
-    this.status = response.status;
-    this.statusText = response.statusText;
-    this.ok = response.ok;
+    let headers = new Headers();
+    for (let name in response.headers) {
+      headers.set(name, joinHeader(response.headers[name]));
+    }
+    // An empty body must be `null`: `Response` forbids a body for 204 etc.
+    super(response.body || null, {
+      status: response.status,
+      statusText: response.statusText,
+      headers,
+    });
     this.socketID = response.socketID;
-    this.bodyText = response.body;
-    let rawHeaders = response.headers;
-    this.headers = {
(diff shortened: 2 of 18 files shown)


benbucksch:
Didn't test yet, want to do a review first.

benbucksch:
@NeilRashbrook Can you tell me what you think of this? The pool looks alright to me. I didn't look at the connection itself yet, nor the calling code.

benbucksch:
I do not like the custom `HTTPConnection` implementation, nor the custom `CookieJar`.

benbucksch:
Issues found by Opus 5, not yet fixed:

1. Retry re-sends a POST the server already processed. 
  NTLMConnection.ts:113-119 retries on ECONNRESET/EPIPE whenever reusedSocket is true,
  without checking whether a response had already begun arriving. I proved it with a mock
  server that RSTs after sending part of a 200: the server processed the same <SendItem> 
  POST three times. Real-world that's duplicate sent mail / duplicate CreateItem. The
  reusedSocket heuristic is only safe when the server never saw the request. Fix: have
  HTTPConnection also report whether any response byte arrived (flag set in
  req.on("response"), attached to the error like reusedSocket), and retry only when it
  didn't.

2. Same bug corrupts the notification stream. With onChunk, an RST after the first chunk
  delivered <Envelope>1</Envelope> three times over 3 handshakes on 3 sockets. In
  callStream, the retry re-POSTs GetStreamingEvents while data still holds the partial
  envelope from the prior attempt, so the first envelope after a retry fails to parse. A
  clean FIN mid-stream behaves correctly (no retry) — it's specifically the RST path.

benbucksch:
This now incorporates Neil's https://github.com/mustang-im/mustang/issues/1387#issuecomment-5286712545 , as 2 classes `NTLMChromium*`

desktop/backend/HTTPConnection.ts
```diff
@@ -0,0 +1,227 @@
...
+  protected socketIDs = new WeakMap<object, number>();
+  protected lastSocketID = 0;
+  /** In-flight requests, so that `close()` can abort them */
+  protected requests = new Set<http.ClientRequest>();
+  protected _closed = false;
+
+  constructor(url: string, options?: HTTPConnectionOptions) {
+    this.url = new URL(url);
+    let secure = this.url.protocol == "https:";
+    this.protocolModule = secure ? https : http;
```
NeilRashbrook, line 40: Huh, do we even allow EWS over http?

app/frontend/Settings/Mail/Account/Server.svelte
```diff
@@ -7,15 +7,21 @@
   {#if account instanceof IMAPAccount}
     <ServerIMAPAdvanced {account} />
   {/if}
+  {#if account instanceof EWSAccount && $account.authMethod == AuthMethod.NTLM}
+    <ServerEWSAdvanced {account} />
+  {/if}
```
NeilRashbrook, line 15: Does this need a `[MOBILE]` check?

desktop/backend/HTTPConnection.ts
NeilRashbrook: Oh, I see, it's used for the tests.

desktop/backend/HTTPConnection.ts
```diff
@@ -0,0 +1,227 @@
...
+ */
+export class HTTPConnection {
+  protected url: URL;
+  protected agent: http.Agent;
+  protected protocolModule: typeof http | typeof https;
+  /** Identifies the TCP connection of each socket ever used, starting at 1 */
+  protected socketIDs = new WeakMap<object, number>();
+  protected lastSocketID = 0;
+  /** In-flight requests, so that `close()` can abort them */
+  protected requests = new Set<http.ClientRequest>();
```
NeilRashbrook, line 34: I am slightly confused as to how there can be more than one outstanding request at once, since the `NTLMConnection` is locked to prevent multiple requests at once, the `NTLMConnectionPool` doesn't reuse a connection until its request finishes, and the `Agent` only has one socket anyway.

app/logic/Auth/NTLM/NTLMResponse.ts
```diff
@@ -0,0 +1,36 @@
+export interface NTLMRequestOptions {
+  headers?: Record<string, string>;
+  signal?: AbortSignal;
+  onChunk?: (chunk: string) => Promise<void>;
+}
+
+/** Duck-typed like a `fetch()` `Response`, as far as `EWSAccount` needs it */
+export class NTLMResponse {
```
NeilRashbrook: Can't this just extend Response? Something like this:

```typescript
export class NTLMResponse extends Response {
  readonly socketID: number;
  constructor(response: any) {
    let headers = new Headers();
    for (let header in response.headers) {
      headers.set(header, joinHeader(response.headers[header]));
    }
    super(response.body, { status: response.status, statusText: response.statusText, headers });
    this.socketID = response.socketID;
  }
}
```

(I'm not even sure we look at the headers of the response, so that bit could get skipped too.)

NeilRashbrook (approved):
This seems reasonable to me although there are some nits that could be addressed before you land.

app/frontend/Settings/Mail/Account/Server.svelte
benbucksch: `[!MOBILE]`, yes.

desktop/backend/HTTPConnection.ts
benbucksch: This is `HTTPConnection`, not `NTLMConnection`.

Classes stand on their own and **MUST NOT** assume what other classes do, remember? This is a core principle of reliable code.

So, this is correct.

app/logic/Auth/NTLM/NTLMResponse.ts
benbucksch: Good idea, let's do that.

benbucksch: done

app/frontend/Settings/Mail/Account/Server.svelte
benbucksch: done

benbucksch:
It seems that I forgot to push the changes from my local branch, so they were not included in the PR and the merge here. DUH. Fixing in #1391.

PR #1390, opened 2026-08-14
EWS: NTLM: Move requests back to the backend

NeilRashbrook:
Electron's `net.request` can actually perform partitioned NTLM login. In particular, I can set up an account on a server, and try to set up a second account on the same server with a bogus username and password, and this will fail, although the first account is still performing EWS requests successfully.

Streaming notifications are not currently included in these changes. I would have to work out how to do streaming with `net.request`, and I can't test multiple streams (by default there is only one stream per account and so our existing NTLM will always succeed, although the developer console will log two HTTP 401 errors as it authenticates).

Annoyingly `net` is already imported from `node:net` as the default import (which here is equivalent to `*`) when possibly it should have been `import { Socket } from "node:net";` so I had to rename electron's `net`.

### DIFF
--- a/desktop/backend/backend.ts
+++ b/desktop/backend/backend.ts
@@ -8,7 +8,8 @@ import { ImapFlow } from 'imapflow';
 import { Database } from "@radically-straightforward/sqlite"; // formerly @leafac/sqlite
 import Zip from "adm-zip";
 import ky from 'ky';
-import { shell, nativeTheme, Notification, Tray, nativeImage, app, BrowserWindow, webContents, Menu, MenuItemConstructorOptions, clipboard, NativeImage, session, desktopCapturer, type DesktopCapturerSource, autoUpdater, systemPreferences, powerMonitor } from "electron";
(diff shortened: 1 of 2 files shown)


benbucksch:
This is nice and short.

Unfortunately, it won't work on mobile. Can you please review #1389?

desktop/backend/backend.ts
```diff
@@ -250,6 +252,45 @@ export class HTTPFetchError extends Error {
   }
 }
 
+function netRequest(url, options, partition, username, password) {
+  let triedAuth = false;
```
NeilRashbrook, line 256: The code I posted in issue #1387 did not have this but I hadn't tested the failure path at that point, and I found that incorrect auth would just trigger another login event, so I added the flag so that authentication would only be tried once, and if that failed then the 401 would bubble up to the caller.

benbucksch: Can we please work based on #1389?

NeilRashbrook:
This approach has since been incorporated into #1389.

PR #1391, opened 2026-08-15
EWS: NTLM: Simplify and clean up code #1387 #1389

benbucksch:
Improves code from #1389. By Opus 5.

* turn `netRequest()` function into a `NetSession` class. That avoids the free free closeFoo() function.
* Simplify the two transports
* Share the HTTP response body reader

Changes that I stupidly forgot to push to #1387 before merging:
* Settings: The NTLM implementation choice is desktop-only
* NTLMResponse is a Response
* Don't repeat a request that the server already answered

I think it's better, if this all lands as one, together with #1387, so I removed #1387 from master and put it all here. So, this PR now contains the entire change including #1387. If you want to see the changes since #1387, you can look at the commits. I left #1387 as the first commit, and then new changes on top as separate commits.

@NeilRashbrook Can you please review?


benbucksch:
I'm going to merge it to master as-is. But Neil, please review the code here *carefully*. If you have any change suggestions, please make the changes in a new branch and open a new PR for it. This way, we can a) have this important fix on master right now b) separate your changes from those here.

desktop/backend/HTTPConnection.ts
```diff
@@ -41,15 +42,14 @@ export class HTTPConnection {
...
-      maxFreeSockets: 1,
     };
     if (secure) {
       agentOptions.ca = getCACertificates();
       if (options?.acceptBrokenTLSCerts) {
         agentOptions.rejectUnauthorized = false;
       }
     }
-    this.agent = new (secure ? https : http).Agent(agentOptions);
+    this.agent = new this.protocolModule.Agent(agentOptions);
```
NeilRashbrook, line 52: Bah, I had meant to mention this, but I got distracted by wondering why http existed in the first place.

desktop/backend/httpBody.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export async function readBodyText(stream: AsyncIterable<any>,
+    onChunk?: (chunk: string) => Promise<void>): Promise<string> {
+  let decoder = new TextDecoder(); // EWS and friends are always UTF-8
+  if (!onChunk) {
+    let chunks: Buffer[] = [];
+    for await (let chunk of stream) {
+      chunks.push(chunk);
+    }
+    return decoder.decode(Buffer.concat(chunks));
```
NeilRashbrook, line 18: I meant to mention last time that I thought this was a bit awkward. I don't know whether it's safe to call `setEncoding` on the streams to make them automatically decode to strings. An alternative would be to call `Readable.toWeb(stream).pipeThough(new TextDecoderStream())`, which would also decode all of the chunks into strings. Another alternative would be `return await new Response(Readable.toWeb(stream)).text();` but this only works for the non-chunked case.

app/logic/Auth/NTLM/NTLMChromiumSession.ts
```diff
@@ -17,41 +16,45 @@ import { appGlobal } from "../../app";
...
 
   /** Cookies, HTTP auth state, and the pool of 6 TCP connections are
-   * per partition, which isolates the accounts from each other. */
+   * per partition, which isolates the accounts and streams from each other. */
   protected get partition(): string {
-    return this.account.webSessionID ?? "ntlm-setup";
+    let accountPartition = this.account.webSessionID ?? "ntlm-setup";
+    return this.streamID == null
+      ? accountPartition
+      : `${accountPartition}:stream:${this.streamID}`;
```
NeilRashbrook, line 33: Since this never changes, has it got so long that it's worth making it a readonly property computed in the constructor instead?

NeilRashbrook (commented):
Rather annoyingly, `git` seems to think that `netRequest.ts` and `NetSession.ts` are only 20% similar, so I had to manually download the commit and set the similarity cutoff to be able to get a proper diff.

desktop/backend/HTTPConnection.ts
benbucksch: So, you're saying it's good now.

app/logic/Auth/NTLM/NTLMChromiumSession.ts
benbucksch: Yes

PR #1394, opened 2026-08-18
NTLM: NetSession: Fix race condition creating conn

NeilRashbrook:
Two instances of `callEWS` can invoke `NTLMChromiumSession.request` simultaneously on a new `NTLMChromiumSession` object. Since `request` tries to create `conn` lazily, both invocations try to initialise `conn`. Instead, make `conn` a `Promise` to the `NetSession` object and definitely initialise the `Promise` once in the constructor, `await`ing it in `request` as necessary.

Because the creation of the `NetSession` object is asynchronous, the second invocation will see `conn` is still `null` because the first invocation hasn't assigned it yet, and it ends up creating a second `NetSession` object. Currently the only possible practical upshot seems to be limited to requests that are still pending on the first `NetSession` object as they can't be properly closed by the second `NetSession` object.

While I was here I folded the `partition` getter into the constructor as `partition` is only needed to create the `NetSession` object.

### DIFF
--- a/app/logic/Auth/NTLM/NTLMChromiumSession.ts
+++ b/app/logic/Auth/NTLM/NTLMChromiumSession.ts
@@ -16,29 +16,24 @@ import { appGlobal } from "../../app";
  */
 export class NTLMChromiumSession {
   protected readonly account: EWSAccount;
-  /** Only for a dedicated connection @see `newDedicatedConnection()` */
-  protected readonly streamID: string | null;
-  /** Backend `NetSession` (via JPC) */
-  protected conn: any = null;
+  /** Backend `NetSession` (via JPC).
+   * Keep the Promise here to avoid race conditions initialising it. */
+  protected conn: Promise<any>;
 
   constructor(account: EWSAccount, streamID: string | null = null) {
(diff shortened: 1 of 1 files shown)


benbucksch:
> the second invocation will see conn is still null because the first invocation hasn't assigned it yet, and it ends up creating a second NetSession object

So, this is still a race?

How about `RunOnce` in `request()`, and leave the `conn` as it was?
That means `this.streamID` changes to `this.partition`.

NeilRashbrook:
Sorry, I was describing the problem with the old code. The new code only calls `newNetSession` once per `NTLMChromiumSession` object.

benbucksch:
I don't like that `conn` is an Promise now, and that we have to await it every time.
At the very least, please rename `conn`.

NeilRashbrook:
OK, if you'd prefer to hide the `Promise` by wrapping it in a `RunOnce` I can do that.

NeilRashbrook:
Now using `RunOnce` instead of a manual `Promise`.

benbucksch (approved):

PR #1395, opened 2026-08-18
EWS: NTLM: NetSession: Fix main stream id

NeilRashbrook:
The main account wasn't passing a username to `callStream`, but that meant that its streaming notifications weren't getting their own session.

If you prefer I could reverse the condition on line 614 and swap over the two halves of the ternary.


benbucksch (approved):

benbucksch:
> If you prefer I could reverse the condition on line 614 and swap over the two halves of the ternary.

Yes, please. Common case (non-delegated) first.

NeilRashbrook:
Done.

benbucksch (approved):

PR #1396, opened 2026-08-18
Backend: Use Node stream consumers to consume Node streams

NeilRashbrook:
Instead of hand-rolled loops or messing around with `toWeb`.

(Stream consumers were added in Node 16.7.0)

Note that with chunked streams we want to push chunks eagerly to the front end, so there might not be a better way than our existing hand-rolled loop.

### DIFF
--- a/desktop/backend/httpBody.ts
+++ b/desktop/backend/httpBody.ts
@@ -1,3 +1,5 @@
+import { text } from 'node:stream/consumers';
+
 /**
  * Reads a HTTP response body as UTF-8 text.
  * Shared by our HTTP clients `HTTPConnection` and `NetSession`.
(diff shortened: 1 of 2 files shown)


desktop/backend/httpBody.ts
```diff
@@ -1,3 +1,5 @@
+import { text } from 'node:stream/consumers';
```
benbucksch:
```suggestion
import { text as textFromStream } from 'node:stream/consumers';
```

desktop/backend/owa.ts
```diff
@@ -1,5 +1,5 @@
-import { session as Session, BrowserWindow, net as Net } from "electron";
-import { Readable } from 'stream';
+import { session as Session, net as Net } from "electron";
+import { text } from 'node:stream/consumers';
```
benbucksch: ditto

benbucksch (approved):

desktop/backend/httpBody.ts
NeilRashbrook: Done. (In all four places.)

benbucksch (approved):

PR #1397, opened 2026-08-20
S/MIME: .p12 import

### LINKED ISSUE #1383: S/MIME: Import private key from .p12 file

benbucksch:
Fixes #1383 

2 commits:
1. S/MIME: Move the private key encryption to its own file — pure refactor: the
  duplicated PBES2 code in decryptCryptoKey() and createNewPrivateKey() moved to
  SMIME/pbes2.ts; both decryptKey() and decryptCryptoKey() now read the key through one
  decryptPKCS8() that checks the key algorithm on both paths.
2. S/MIME: Import the private key from a .p12 file - #1383 —
    - SMIME/PKCS12.ts: RFC 7292 reader — MAC verification (also the wrong-passphrase
  check), the PKCS#12-specific key derivation, all bag types incl. nested safe-contents,
  and toPEM() which re-encrypts the key with the same passphrase and appends every
  certificate, so the CA chain comes along and the key is never stored in the clear.
    - SMIME/legacyCiphers.ts: Triple DES (2- and 3-key), RC2 (40/128 bit) and RC4.
  WebCrypto and @noble implement none of them, and real .p12 files need them — Windows,
  macOS and OpenSSL 1 write RC2-40 + 3DES; only OpenSSL 3 and recent apps write AES.
    - KeyUtils.readKeyFile() + the Settings import UI: the picked file is read as bytes, a
  leading 0x30 (DER) routes to the PKCS#12 reader, anything else is decoded as text exactly
  as before; .p12/.pfx added to the file picker.

  Checks run: 28 tests pass in logic/Mail/Encryption (17 pre-existing S/MIME + 11 new). The
  new tests cover 7 openssl-generated fixtures — AES/PBES2, PBES2+3DES, RC2-40+3DES, 3DES
  with CA chain, RC2-128+2-key-3DES, RC4, unencrypted bags without MAC, non-ASCII
  passphrase — plus wrong-passphrase rejection and the end-to-end path into
  SMIMEPrivateKey. Triple DES was cross-checked against Node's OpenSSL, RC2 against all 8
  RFC 2268 vectors, RC4 against the standard vector, and each fixture's PEM output was read
  back by openssl pkey/openssl x509 with the public key matching the original.
  svelte-check reports 217 errors — identical to the baseline I measured before the change.
  The rest of the suite's failures (VCard, SQL, WebDAV, XMPP …) fail identically in the
  untouched checkout.
 
  Not verified: the change has not been run in the real app UI, only unit-tested — worth
  a manual import before landing. Also worth noting: the app now has a working 3DES
  implementation, which is what commit 96ddede20 said we lacked for decrypting
  3DES-encrypted S/MIME mail; wiring it into SMIMEReadProcessor would be a separate, easy
  follow-up.

### DIFF
--- a/app/frontend/Settings/Mail/Account/EncryptionImport.svelte
+++ b/app/frontend/Settings/Mail/Account/EncryptionImport.svelte
@@ -85,7 +85,7 @@
   import { SMIMEPrivateKey } from "../../../../logic/Mail/Encryption/SMIME/SMIMEPrivateKey";
   import { TrustLevel } from "../../../../logic/Mail/Encryption/enums";
   import { MailIdentity } from "../../../../logic/Mail/MailIdentity";
-  import { importPrivateKey } from "../../../../logic/Mail/Encryption/KeyUtils";
+  import { importPrivateKey, readKeyFile } from "../../../../logic/Mail/Encryption/KeyUtils";
   import FileSelector from "../../../Mail/Composer/Attachments/FileSelector.svelte";
   import Menu from "../../../Shared/Menu/Menu.svelte";
   import MenuItem from "../../../Shared/Menu/MenuItem.svelte";
@@ -104,20 +104,20 @@
   export let showObsolete: boolean;
 
   let showPassword = false;
-  let passphrase: string;
+  let passphrase = "";
   async function onFilePassword() {
     await onImportFile(passphrase);
     showPassword = false;
   }
 
   let fileSelector: FileSelector;
-  const acceptFileTypes = [ "application/pgp-secret-keys", ".asc", "application/pkcs8", "application/x-pem-file", ".key", "text/plain" ];
+  const acceptFileTypes = [ "application/pgp-secret-keys", ".asc", "application/pkcs8", "application/x-pem-file", ".key", "application/x-pkcs12", ".p12", ".pfx", "text/plain" ];
   async function onImportFile(passphrase: string) {
     let file = await fileSelector.selectFile();
     if (!file) {
       return;
     }
-    let fileContent = await file.text();
+    let fileContent = await readKeyFile(file, passphrase);
     let key = await importPrivateKey(fileContent, passphrase);
     let existing = identity.encryptionPrivateKeys.find(existing => existing.id == key.id);
     if (!existing) {
--- a/app/logic/Mail/Encryption/SMIME/PKCS12.ts
+++ b/app/logic/Mail/Encryption/SMIME/PKCS12.ts
@@ -21,10 +21,13 @@ export class PKCS12 {
    * Typically the certificate of the key, and the CAs that signed it. */
   certificates: Uint8Array[] = [];
   protected passphrase: string;
+  /** The passphrase in the form that the key derivation needs it */
+  protected passphraseBytes: Uint8Array;
 
   constructor(passphrase: string) {
     assert(typeof passphrase == "string", "Need the passphrase");
     this.passphrase = passphrase;
+    this.passphraseBytes = bmpString(passphrase);
   }
 
   /** Reads the contents of a .p12 file into `keys` and `certificates` */
@@ -66,8 +69,24 @@ export class PKCS12 {
    * This is also where a wrong passphrase shows up. */
   protected verifyMAC(macData: any, authSafe: Uint8Array) {
     let hash = sanitize.translate(macData.mac.digestAlgorithm.algorithm, kHashes);
+    if (this.macMatches(macData, authSafe, hash)) {
+      return;
+    }
+    // An empty passphrase and no passphrase at all are 2 different keys, and
+    // the file does not say which of them the app that wrote it used, so try
+    // the other one as well. This is the only place that can tell them apart.
+    if (!this.passphrase) {
+      this.passphraseBytes = kNoPassphrase;
+      if (this.macMatches(macData, authSafe, hash)) {
+        return;
+      }
+    }
+    throw new Error(gt`Wrong passphrase for this file`);
+  }
+
+  protected macMatches(macData: any, authSafe: Uint8Array, hash: Hash): boolean {
     let key = this.deriveKey(KeyPurpose.MAC, hash.outputLen, macData.macSalt, macData.iterations, hash);
-    assert(!indexedDB.cmp(hmac(hash, key, authSafe), macData.mac.digest), gt`Wrong passphrase for this file`);
+    return !indexedDB.cmp(hmac(hash, key, authSafe), macData.mac.digest);
   }
 
   /** @returns the `SafeContents` of one part of the file, decrypted if needed */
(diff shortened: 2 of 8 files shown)


benbucksch:
@NeilRashbrook Can you review, please?

app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
```diff
@@ -39,46 +40,30 @@ export class SMIMEPrivateKey extends SMIMEPublicKey implements PrivateKey {
...
-    if (!this.passphrase) {
-      let privateKey = PrivateKeyInfo.decodePEM(this.privateKeyArmored, { label: "PRIVATE KEY" });
-      if (privateKey.privateKeyAlgorithm.algorithm != "rsaEncryption") {
-        throw new Error("Unsupported private key algorithm");
+    return crypto.subtle.importKey("pkcs8", await this.decryptPKCS8(), { name: "RSASSA-PKCS1-v1_5", hash: "SHA-256" }, true, ["sign"]);
+  }
+
+  /** Decrypts our private key with our passphrase.
+   * @returns the key in PKCS#8 form */
+  protected async decryptPKCS8(): Promise<Uint8Array> {
```
NeilRashbrook: I didn't know that you could decrypt an encrypted private key simply using `crypto.subtle.decrypt` rather than `crypto.subtle.unwrapKey`. Furthermore, at one point my intention was to use `crypto.subtle` for signing and verification, but this never came to pass either. As such, this appears to mean that `decryptCryptoKey` is unnecessary. At the very least it could be removed, and then it might well be worth merging `decryptKey` with `decryptPKCS8` as it is the only caller and is only one line of code.

app/logic/Mail/Encryption/SMIME/PKCS12.ts
```diff
@@ -0,0 +1,215 @@
...
+  certificates: Uint8Array[] = [];
+  protected passphrase: string;
+
+  constructor(passphrase: string) {
+    this.passphrase = passphrase;
+  }
+
+  /** Reads the contents of a .p12 file into `keys` and `certificates` */
+  async read(fileContents: Uint8Array): Promise<void> {
+    let pfx = PFX.decode(fileContents, { berToDER: true });
```
NeilRashbrook, line 35: Nice, that `berToDER` change is paying off already!

app/logic/Mail/Encryption/SMIME/PKCS12.ts
```diff
@@ -0,0 +1,215 @@
...
+
+/** The passphrase as PKCS#12 uses it: UTF-16, big endian, with a final null.
+ * Characters outside the BMP go in as their 2 surrogates,
+ * which is what the apps that write these files do as well. */
+function bmpString(text: string): Uint8Array {
+  let bytes = new Uint8Array(text.length * 2 + 2);
+  for (let i = 0; i < text.length; i++) {
+    bytes[i * 2] = text.charCodeAt(i) >> 8;
+    bytes[i * 2 + 1] = text.charCodeAt(i) & 0xFF;
+  }
```
NeilRashbrook, line 214: I feel this would be clearer if it used a `DataView` to write the code units.

app/frontend/Settings/Mail/Account/EncryptionImport.svelte
```diff
@@ -111,13 +111,13 @@
...
   let fileSelector: FileSelector;
-  const acceptFileTypes = [ "application/pgp-secret-keys", ".asc", "application/pkcs8", "application/x-pem-file", ".key", "text/plain" ];
+  const acceptFileTypes = [ "application/pgp-secret-keys", ".asc", "application/pkcs8", "application/x-pem-file", ".key", "application/x-pkcs12", ".p12", ".pfx", "text/plain" ];
   async function onImportFile(passphrase: string) {
     let file = await fileSelector.selectFile();
     if (!file) {
       return;
     }
-    let fileContent = await file.text();
+    let fileContent = await readKeyFile(file, passphrase);
```
NeilRashbrook, line 120: Everything here assumes `passphrase` is a string. Except it never gets definitely initialised, so that if you try to import a `.p12` file without typing a password then the code crashes in `bmpString`. Just needs an initialiser on line 107 to fix this.

app/logic/Mail/Encryption/SMIME/PKCS12.ts
```diff
@@ -0,0 +1,215 @@
...
+   * first the private key, then the certificates. The key is encrypted
+   * again with the same passphrase, so that it is never stored in the clear.
+   *
+   * A file with several keys is not an S/MIME identity that we could
+   * represent, so only the first key is used. All certificates are kept,
+   * because they form the chain up to the CA.
+   */
+  async toPEM(): Promise<string> {
+    assert(this.keys.length, gt`This file contains no secret key`);
+    let parts = [await encryptPrivateKey(this.keys[0], this.passphrase)];
```
NeilRashbrook: This also assumes that `passphrase` is not empty; `SMIMEPrivateKey.decryptKey` assumes that the PEM will be the raw key PEM-encoded with the label `PRIVATE KEY` if the passphrase is empty.

NeilRashbrook (approved):
Basically fine, although there are a few nits that could improve the code slightly, e.g. should `pbes2.ts` be in upper case, like `PKCS12.ts`?

app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
benbucksch: Code has been rewritten

benbucksch: Can you please check again whether it's good now?

benbucksch:
@NeilRashbrook Can you please re-review? Esp. the removed custom decrypt/encrypt functions, and the "file without passphrase" case?

benbucksch:
Opus 5:

p12 files written without any passphrase now import.

  Sanity check: 25 encryption tests pass (14 in PKCS12.test.ts), tsc clean on PKCS12.ts. I
  also ran the negative check: with the fallback stubbed to if (false), the new test fails
  and the empty-string test still passes, so the two conventions are genuinely exercised
  independently rather than one masking the other.

  The change (PKCS12.ts, ~29 lines): the derivation input became state — passphraseBytes,
  initialised to bmpString(passphrase). verifyMAC() now tries that, and if it fails and the
  passphrase is empty, retries with kNoPassphrase (zero-length) and keeps whichever form
  verified, so the legacy RC2/3DES/RC4 bags derive consistently. macMatches() split out so
  the retry isn't duplicated code. PBES2 still takes this.passphrase directly — it derives
  from the UTF-8 password, empty either way.

  One deliberate divergence from OpenSSL: when a file has no MAC at all, OpenSSL assumes
  the null form. I kept the empty-string form there. Without a MAC there's no evidence
  which was used, and switching the default would break files that do use the empty-string
  form — a regression to buy nothing. The MAC is the only thing that can actually
  distinguish them, so that's the only place that retries.

  The fixture is independent of our code, which matters for a test like this. I found that
  OpenSSL's PKCS12KDF provider takes the password as raw bytes with no unicode conversion
  of its own, so openssl kdf -kdfopt hexpass:… is a true oracle. I validated it by
  rederiving the existing empty-password fixture's MAC (hexpass:0000 → exact match,
  47BA53FC…), then derived the zero-length variant (9E6C7CFB…, completely different), and
  patched that 32-byte digest into the file. openssl pkcs12 -passin pass: reads the result
  — so it's a file OpenSSL itself accepts, with a MAC computed entirely by OpenSSL's KDF
  and Python's HMAC.

app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
NeilRashbrook: Yes it's much better thanks! (Also the other fixes look good.)

PR #1399, opened 2026-08-24
Incoming invitation: Adds duplicate events

benbucksch:
EWS incoming invitations: when the user receives an email with an calendar invitation
and clicks [Confirm], he sometimes gets the event twice in his EWS calendar. In some
cases, it's even twice in the EWS *and* twice in the local calendar.

### DIFF
--- a/app/logic/Calendar/Invitation/IncomingInvitation.ts
+++ b/app/logic/Calendar/Invitation/IncomingInvitation.ts
@@ -36,9 +36,17 @@ export class IncomingInvitation {
   calEvent(): Event | null {
     // First find a standalone event or master.
     let event = this.calendar.events.find(event => event.calUID == this.event.calUID && !event.recurrenceStartTime);
+    if (!this.event.recurrenceStartTime) {
+      return event;
+    }
     // If this is an occurrence, find it now.
     // It may have been cancelled and excluded from the calendar already.
-    return this.event.recurrenceStartTime ? event?.getOccurrenceByDate(this.event.recurrenceStartTime, false) : event;
+    if (event) {
+      return event.getOccurrenceByDate(this.event.recurrenceStartTime, false);
+    }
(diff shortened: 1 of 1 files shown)


benbucksch:
Opus 5 wrote:

1. Two concurrent syncs (the "sometimes"). Confirming calls calendar.listEvents(), and
  the Exchange push notification for that very calendar item makes EWSAccount call
  calendar.listEvents() at the same moment. Both runs start from the same syncState, both
  get the same Create, both create an event and a DB row. CalDAVCalendar,
  ActiveSyncCalendar and EWSAddressbook all have a sync lock — EWSCalendar was the one that
  didn't.

2. The same item returned twice inside one sync. getEventByItemID() searches
  calendar.events, but syncFolder() collected newly fetched events in a local list and only
  moved them into calendar.events after the whole paging loop. So when Exchange returns
  the item as Create on one page and Update on the next — exactly what it does when it adds
  the calendar item for an invitation and then updates it on accept — the lookup misses
  and a second event is created. OWA has the identical structure. Person.saveLocally() puts
  itself into addressbook.persons; Event.saveLocally() doesn't, so each calendar has to do
  it, and CalDAV/ActiveSync do it immediately while EWS/OWA deferred it.

3. The UID check you suspected exists, and was broken. IncomingInvitation.calEvent() is
  the "is it already in the calendar" check. It skipped every event that has a
  recurrenceStartTime. When you're invited to a single occurrence of a series you don't
  have, we add the occurrence as a standalone event carrying that recurrenceStartTime — so
  calEvent() could never find it again. Consequences: InvitationButtons calls
  calEvent()?.save(), so the event was never written to the DB or the server at all; and
  myParticipation couldn't be read back, so the mail kept offering [Confirm] and each
  further click added another copy. This one is protocol-independent — local, CalDAV, JMAP,
  and the button state for EWS/OWA/ActiveSync.

@NeilRashbrook Can you please review? Look at the commits and commit messages.

Issue #1186, opened 2026-04-27
EWS: occurrence date not in recurrence


benbucksch:
Stack trace from error logs:
```
Error: Exchange 2016: occurrence date not in recurrence
  at assert (../../../../app/logic/util/util.ts:4:11)
  at Event$1.Event.getOccurrenceByDate (../../../../app/logic/Calendar/Event.ts:539:5)
  at Event$1.Event.makeExclusionLocally (../../../../app/logic/Calendar/Event.ts:549:10)
  at EWSEvent.fromXML (../../../../app/logic/Calendar/EWS/EWSEvent.ts:90:16)
  at EWSCalendar.getEvents (../../../../app/logic/Calendar/EWS/EWSCalendar.ts:237:17)
...
```

Is that a bug, or faulty data in the calendar?

NeilRashbrook:
Without seeing the XML, my best guess is a bug in handling of all-day events from other time zones. I'll see whether I can reproduce it.

NeilRashbrook:
This is what #1137 was supposed to fix - all-day events from other time zones are definitely broken before that change, but I wasn't able to reproduce the error after that change.

benbucksch:
The error appeared for the user in 0.9.26, which already has #1137.
He used Exchange 2016.

NeilRashbrook:
Well, in theory I could set up an isolated trial installation of Exchange 2016 to see whether I can reproduce the bug...

benbucksch:
We're not sure that the Exchange version is the cause. So, that would be a little excessive.

benbucksch:
@NeilRashbrook  See c1a5e755775bc0

NeilRashbrook:
Maybe we should put in a log so that we can tell whether this is still happening?

PR #1401, opened 2026-08-24
Calendar: EWS, OWA, ActiveSync: All-day recurrence exceptions were dropped

benbucksch:
Opus 5:

Exchange sends the times of an all-day event as the UTC time that would be midnight in the event's time zone, and setAllDayTimeToLocalMidnight() converts them to local midnight. The occurrence dates that the server sends
- the recurrence ID of a modified occurrence, and the deleted occurrences - need the same conversion, or they no longer match the recurrence that we generate from the converted start time.

The sync then logged "occurrence date not in recurrence" and dropped the modified occurrence, and deleted occurrences kept being shown. Happens only when the event has a different time zone than the machine.

### DIFF
--- a/app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
+++ b/app/logic/Calendar/ActiveSync/ActiveSyncCalendar.ts
@@ -146,7 +146,7 @@ export class ActiveSyncCalendar extends ExchangeCalendar implements ActiveSyncPi
           // Exceptions must be handled after the master event has been saved.
           for (let exception of ensureArray(item.ApplicationData.Exceptions?.Exception)) {
             if (exception.Deleted != "1") {
-              let exceptionTime = fromCompact(sanitize.nonemptystring(exception.ExceptionId || exception.ExceptionStartTime), event.allDay && this.account.protocolVersion == "16.1");
+              let exceptionTime = event.toLocalMidnight(fromCompact(sanitize.nonemptystring(exception.ExceptionId || exception.ExceptionStartTime), event.allDay && this.account.protocolVersion == "16.1"));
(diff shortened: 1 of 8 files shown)


benbucksch:
The cause

  "occurrence date not in recurrence" might come from Event.getOccurrenceByDate()
  (app/logic/Calendar/Event.ts:646), reached from EWSCalendar.getEvents() when the server
  sends a modified occurrence of a recurring series. It triggers for all-day series:

  Exchange sends all-day times as the UTC instant that is midnight in the event's own time 
  zone (2024-03-14T23:00:00Z for a Berlin event on 03-15).
  ExchangeEvent.setAllDayTimeToLocalMidnight() rewrites startTime/endTime to machine-local
  midnight and clears timezone. But the occurrence dates the server sends — RecurrenceId,
  DeletedOccurrences.Start, ActiveSync's ExceptionId — were never converted, so they no
  longer matched the recurrence we generate from the converted startTime. Verified
  numerically: on a UTC machine the rule yields 2024-03-17T00:00:00Z while the server's
  RecurrenceId is 2024-03-16T23:00:00Z. 

  That's the "rarely": it only bites when the event's time zone differs from the machine's.
  The same mismatch on the deletion path is what the #1186 guard in makeExclusionLocally()
  was silently swallowing — which means deleted occurrences of all-day series kept being
  displayed.

  The fix

  ExchangeEvent now keeps the server's zone in allDayTimezone (also carried into instances
  by copyFromRecurrenceMaster(), since ActiveSync Exception elements have no Timezone) and
  exposes toLocalMidnight(date), applied at every point where an occurrence date comes off
  the wire — EWS, OWA and ActiveSync, master and exception paths alike.

@NeilRashbrook Can you review, please?

NeilRashbrook (approved):
This `ExchangeEvent` class of yours has really paid off!

PR #1404, opened 2026-08-25
Outgoing invitations: Send the VTIMEZONE for Outlook

benbucksch:
We wrote only IANA timezones.
RFC 5545 3.6.5 requires the VTIMEZONE for every TZID.

Outlook does not know the IANA timezone names, but matches them against a list of well-known timezones, and gets only the base offset from it, because it reads the daylight saving time solely from the VTIMEZONE. It therefore shows every event during summer time 1 hour too early. Exchange ignores the name entirely and falls back to the timezone of the mailbox, so it converts the invitation wrongly before Outlook even sees it. Both bugs hide in winter, when base offset and real offset agree.

We derive the rules from the timezone database via `Intl`, for the year in which the event happens, and write only what Outlook and Exchange read: DTSTART, RRULE, TZOFFSETFROM and TZOFFSETTO, in local time.

### DIFF
--- a/app/logic/Calendar/ICal/VTimezone.ts
+++ b/app/logic/Calendar/ICal/VTimezone.ts
@@ -141,8 +141,9 @@ function utcOffsetChanges(timezone: string, year: number): Date[] {
  * @param before must have a different UTC offset than `after` */
 function findUTCOffsetChange(before: Date, after: Date, timezone: string): Date {
   let offsetBefore = utcOffset(before, timezone);
-  while (after.getTime() - before.getTime() > 1) {
-    let middle = new Date(Math.round((before.getTime() + after.getTime()) / 2));
+  // The timezone database changes the offset only at full minutes
+  while (after.getTime() - before.getTime() > k1MinuteMS) {
+    let middle = new Date(Math.round((before.getTime() + after.getTime()) / 2 / k1MinuteMS) * k1MinuteMS);
     if (utcOffset(middle, timezone) == offsetBefore) {
       before = middle;
     } else {
--- a/app/logic/Calendar/ICal/VTimezone.ts
+++ b/app/logic/Calendar/ICal/VTimezone.ts
@@ -183,9 +183,8 @@ function dayOfNthWeekday(year: number, month: number, nth: number, weekday: numb
     let firstWeekday = new Date(Date.UTC(year, month - 1, 1)).getUTCDay();
     return 1 + (weekday - firstWeekday + 7) % 7 + (nth - 1) * 7;
   }
-  let lastDay = new Date(Date.UTC(year, month, 0)).getUTCDate();
-  let lastWeekday = new Date(Date.UTC(year, month - 1, lastDay)).getUTCDay();
-  return lastDay - (lastWeekday - weekday + 7) % 7;
+  let lastDay = new Date(Date.UTC(year, month, 0));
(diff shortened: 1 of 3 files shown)


app/logic/Calendar/ICal/VTimezone.ts
```diff
@@ -0,0 +1,201 @@
...
+function utcOffset(time: Date, timezone: string): number {
+  // "lt" locale has date format YYYY-MM-DD hh:mm:ss,
+  // which we can easily read as UTC.
+  let local = new Date(time.toLocaleString("lt", { timeZone: timezone }).replace(" ", "T") + "Z");
+  return Math.round((local.getTime() - time.getTime()) / k1MinuteMS);
+}
+
+/** The year at `time` in `timezone`, which may differ from the year here */
+function yearIn(time: Date, timezone: string): number {
+  return new Date(time.getTime() + utcOffset(time, timezone) * k1MinuteMS).getUTCFullYear();
```
NeilRashbrook: The above code converts the given time into a string using the given timezone, subtracts the original time, divides by 60,000, multiplies by 60,000, adds the original time, turns back into a Date, then gets the year. I feel it would be more straightforward to do this:
```suggestion
  return Number(time.toLocaleDateString("lt", { timeZone: timezone, year: "numeric" }));
```

app/logic/Calendar/ICal/VTimezone.ts
```diff
@@ -0,0 +1,201 @@
...
+    startOffset = endOffset;
+  }
+  return changes;
+}
+
+/** The exact moment when the UTC offset of `timezone` changes
+ * @param before must have a different UTC offset than `after` */
+function findUTCOffsetChange(before: Date, after: Date, timezone: string): Date {
+  let offsetBefore = utcOffset(before, timezone);
+  while (after.getTime() - before.getTime() > 1) {
```
NeilRashbrook: Do we need millisecond accuracy?
```suggestion
  while (after.getTime() - before.getTime() > k1MinuteMS) {
```
(Maybe also round `middle` to the nearest minute?)

app/logic/Calendar/ICal/VTimezone.ts
```diff
@@ -0,0 +1,201 @@
...
+/** The day of the month of e.g. the 2nd Sunday (`nth` 2, `weekday` 0)
+ * or the last Sunday (`nth` -1), the reverse of `nthWeekdayOfMonth()`
+ * @param month 1 = January */
+function dayOfNthWeekday(year: number, month: number, nth: number, weekday: number): number {
+  if (nth > 0) {
+    let firstWeekday = new Date(Date.UTC(year, month - 1, 1)).getUTCDay();
+    return 1 + (weekday - firstWeekday + 7) % 7 + (nth - 1) * 7;
+  }
+  let lastDay = new Date(Date.UTC(year, month, 0)).getUTCDate();
+  let lastWeekday = new Date(Date.UTC(year, month - 1, lastDay)).getUTCDay();
```
NeilRashbrook: `Date.UTC(year, month - 1, lastDay)` is the same value as `Date.UTC(year, month, 0)`, so `new Date(Date.UTC(year, month, 0))` could go in its on local variable.

NeilRashbrook (commented):
(Comments from code inspection only.)

PR #1406, opened 2026-08-25
S/MIME: Expose signing certificate

NeilRashbrook:
With this patch, a completely new profile can view the certificate of an S/MIME signed message without having to import it from a .pem file into a contact first.

The signature icon starts off yellow but turns blue when clicked - I don't know if there's a better place to trigger the certificate verification.

I don't know if there's supposed to be UI to save the certificate to a contact (needed to be able to encrypt a reply) or what I might have to add for that.

### DIFF
--- a/app/logic/Mail/EMail.ts
+++ b/app/logic/Mail/EMail.ts
@@ -10,6 +10,7 @@ import { RawFilesAttachment } from "./Store/RawFilesAttachment";
 import { EMailProcessorList, ProcessingStartOn } from "./EMailProcessor";
 import type { ExtraData } from "./ExtraData";
 import type { SMLData } from "./SML/SMLData";
+import type { PublicKey } from "./Encryption/PublicKey";
 import { Event } from "../Calendar/Event";
 import { InvitationMessage, type iCalMethod } from "../Calendar/Invitation/InvitationStatus";
 import { FilterMoment } from "./FilterRules/FilterMoments";
@@ -108,6 +109,8 @@ export class EMail extends Message {
    * Format: `PublicKey.id` */
   @notifyChangedProperty
   signed: string | null = null;
+  /** As `signed` but giving you the actual PublicKey object */
+  signer: PublicKey | null = null;
   /** Contains the complete MIME message for sending.
    * Used for encrypted messages. */
   sendRawMIME: string | null = null;
--- a/app/logic/Mail/Encryption/KeyUtils.ts
+++ b/app/logic/Mail/Encryption/KeyUtils.ts
@@ -44,6 +44,9 @@ export async function getPublicKeyByKeyID(id: string | null, email?: EMail): Pro
       }
     }
   }
+  if (email?.signer?.id == id) {
+    return email.signer;
+  }
   if (email && email.system != EncryptionSystem.SMIME) {
     let key = await readAutoCryptKeys(email);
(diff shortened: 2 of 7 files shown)


benbucksch:
> The signature icon starts off yellow but turns blue when clicked - I don't know if there's a better place to trigger the certificate verification.

When you import the cert or process the message.

benbucksch:
> I don't know if there's supposed to be UI to save the certificate to a contact

When you click on a sender that is not in your addressbook (it shows a domain), you get a dropdown with "Add to <addressbook>". When you do that, we should save the cert as well.

Likewise, when the user replies, we should be able to encrypt the response. Esp. if the original email was encrypted.

> (needed to be able to encrypt a reply)

We might need to have an optional cert on a `PersonUID` and not just a `Person`.

app/logic/Mail/EMail.ts
```diff
@@ -108,6 +109,8 @@ export class EMail extends Message {
    * Format: `PublicKey.id` */
   @notifyChangedProperty
   signed: string | null = null;
+  /** As `signed` but giving you the actual PublicKey object */
+  signer: PublicKey | null = null;
```
benbucksch, line 113:
```suggestion
  signedKeyID: string | null = null;
  /** As `signed` but giving you the actual PublicKey object */
  signedKey: PublicKey | null = null;
```

benbucksch (approved):

app/logic/Mail/EMail.ts
benbucksch: Fixed by 7aef2525e

benbucksch:
> > I don't know if there's supposed to be UI to save the certificate to a contact
> 
> When you click on a sender that is not in your addressbook (it shows a domain), you get a dropdown with "Add to ". When you do that, we should save the cert as well.
> 
> Likewise, when the user replies, we should be able to encrypt the response. Esp. if the original email was encrypted.
> 
> > (needed to be able to encrypt a reply)
> 
> We might need to have an optional cert on a `PersonUID` and not just a `Person`.

Please address these in a new PR

benbucksch:
> > I don't know if there's supposed to be UI to save the certificate to a contact
> 
> When you click on a sender that is not in your addressbook (it shows a domain), you get a dropdown with "Add to ". When you do that, we should save the cert as well.
> 
> Likewise, when the user replies, we should be able to encrypt the response. Esp. if the original email was encrypted.
> 
> > (needed to be able to encrypt a reply)
> 
> We might need to have an optional cert on a `PersonUID` and not just a `Person`.

Fixed in d983f5ccac137a. @NeilRashbrook Can you please check? (Most of it is mechanical.)

NeilRashbrook:
> > The signature icon starts off yellow but turns blue when clicked - I don't know if there's a better place to trigger the certificate verification.
> 
> When you import the cert or process the message.

Create PR #1410.

> > > I don't know if there's supposed to be UI to save the certificate to a contact
> 
> Fixed in [d983f5c](https://github.com/mustang-im/mustang/commit/d983f5ccac137a7ac6c66e0d293ef738dce4d8d8). @NeilRashbrook Can you please check? (Most of it is mechanical.)

I can encrypt a reply to a newly received signed message. And I can add a new recipient with their certificate to the address book. But if the recipient already exists in the address book, I can't seem to add (or update?) the certificate.

PR #1408, opened 2026-08-26
CalDAV: Nextcloud: Save the event attachments as files on the server - fixes #1400

### LINKED ISSUE #1400: Calendar: Attachments: Not working on NextCloud calendar
### Reproduction 1

1. Go to Calendar
2. Create a new event in a Nextcloud calendar 
3. Add an attachment
4. Add a title
5. Click [Save]
6. Check the event on Nextcloud web app
7. You may need to click [More details] to see the attachment

### Actual Result 1

- The attachment appears in the attachment section with the filename
- When you click to open it it is empty and the link is `\null`

### Reproduction 2

1. Go to the Nextcloud web app
2. Go to Calendar
3. Create an event
4. Add an attachment
5. Click [Save]
6. Go to the Parula app
7. Click [Sync]
8. Open the event

### Actual Result 2

- The attachment is missing in Parula but present in Nextcloud
- The link in Nextcloud is valid with `\[id]`

### Expected Result

- The attachment is present in Parula and Nextcloud
- Nextcloud opens the attachment with a valid link

benbucksch:
Fixes #1400

Nextcloud and ownCloud cannot save the file in the event. They save it as a normal file of the user, reference it in the event only by URL, and show only such attachments in their web app. So, an event that we saved showed the file name with a link to `null` there, and the attachments that the user added in the web app were missing here, because we skipped `ATTACH` with a URI and read only the files that are inline in the event.

Upload the files into the file storage of the same server, with the same login, and read the attachments that are only referenced by URL. We fetch them via WebDAV, or, for the web page links `/f/<fileID>` that the Nextcloud web app writes, via its `direct` download API. Attachments on some foreign web server, e.g. a cloud drive link, open in the browser.

Email invitations still carry the files themselves, because the invitees have no access to our file storage.


benbucksch:
This is a huge amount of change, plus a DB schema change, just because Nextcloud cannot save attachments in events. I don't like it.

jermy-c:
Works, there's one difference:

- Saving on NextCloud web app the link saved is `https://[domain]/f/[id]` which opens into a preview on the web app.
- Saving on Parula the link saved is `https://[domain]/remote.php/dav/files/[user]/Calendar/[filename]` which downloads the file instead of previewing it on the web app.

PR #1409, opened 2026-08-26
EWS, OWA, ActS: Read and save the contact picture

benbucksch:
Exchange saves the picture as a file attachment of the contact,
marked with IsContactPhoto, so it needs its own calls,
after the contact itself exists on the server.


benbucksch:
Merged a different variant of this PR

* 26f41341a Contacts: ActiveSync: Read and save the contact picture
* b055d0dd5 Contacts: OWA: Read and save the contact picture
* 33395ef12 Contacts: EWS: Read and save the contact picture
* 25a40d718 OWA: Move the attachment requests out of the calendar

app/logic/Contacts/OWA/OWAPerson.ts
```diff
@@ -47,6 +59,36 @@ export class OWAPerson extends ExchangePerson {
     return this;
   }
 
+  /** The persona is an aggregate, e.g. of a GAL entry and our contact.
+   * The writable attribution is the contact in our mailbox.
+   * Fall back to the email address, which knows the contact as well,
+   * @see `OWAAddressbook.listGroups()` */
+  protected static itemIDFromJSON(json: any): string {
+    let contact = json.Attributions?.find(attribution => sanitize.boolean(attribution.IsWritable, false) && attribution.SourceId);
```
NeilRashbrook, line 67: When I look at the Exchange 2019 OWA trace, I see an `AttributionsArray`, which is in the format expected here. (Rather unfortunately, Exchange 2019 OWA can't edit a contact picture. It can view it, but the call returns an empty pixel if there's no picture.)

app/logic/Contacts/OWA/OWAPerson.ts
```diff
@@ -47,6 +59,36 @@ export class OWAPerson extends ExchangePerson {
...
+    return sanitize.nonemptystring(contact?.SourceId.Id ?? json.EmailAddress?.ItemId?.Id, "");
```
NeilRashbrook, line 68: I guess it's not worth checking all of the email addresses for the `ItemId`; they all seem to have the same one.

app/logic/Contacts/OWA/OWAPerson.ts
```diff
@@ -60,14 +102,58 @@ export class OWAPerson extends ExchangePerson {
 
   async saveToServer() {
     let fields = this.toFields();
-    if (Object.keys(fields).every(key => fields[key] == this.fields[key])) {
+    if (Object.keys(fields).some(key => fields[key] != this.fields[key])) {
+      let request = this.personaID ? new OWAUpdatePersonaRequest(this.personaID, this.fields, fields) : new OWACreatePersonaRequest(this.addressbook.folderID, this.fields, fields);
+      let response = await this.addressbook.callOWA(request);
+      this.name = sanitize.nonemptystring(response.DisplayName, "");
+      this.personaID = sanitize.nonemptystring(response.PersonaId.Id);
```
NeilRashbrook, line 109: When saving we should also save the new `ItemId` to avoid having to fetch it in `savePictureToServer` (both `CreatePersona` and `UpdatePersona` return the complete new/updated persona, including attribution array).

app/logic/Contacts/EWS/EWSPerson.ts
```diff
@@ -70,6 +73,33 @@ export class EWSPerson extends ExchangePerson {
...
+      // The contents come in a separate call, @see downloadPictureFromServer()
+      this.picture = null;
+    }
+  }
+
+  /** Exchange saves the picture as an attachment of the contact,
+   * so it needs a separate call. */
+  async downloadPictureFromServer() {
+    if (this.pictureAttachmentID && !this.picture) {
+      let response = await this.addressbook.account.callEWS({
```
NeilRashbrook, line 90: (Bah, I missed this on the event attachments PR) I'm not a fan of inlining the request into the server call but I guess it's not a problem.

app/logic/Contacts/EWS/ExchangePerson.ts
```diff
@@ -1,5 +1,36 @@
...
+
+  /**
+   * @param base64 The picture, as the server sent it
+   * @param mimeType As the server declared it
+   */
+  protected pictureFromBase64(base64: string, mimeType: string = kPictureMIMEType) {
+    if (!mimeType.startsWith("image/")) {
+      mimeType = kPictureMIMEType;
+    }
+    this.picture = sanitize.url(`data:${mimeType};base64,${base64}`, null, ["data"]);
```
NeilRashbrook, line 25: (Can't work out what the point of sanitising a URL we just constructed is.)

app/logic/Contacts/EWS/EWSAddressbook.ts
```diff
@@ -192,15 +194,11 @@ export class EWSAddressbook extends ExchangeAddressbook implements EWSSubscribab
...
-            await person.saveLocally();
-          } else {
-            person = new EWSPerson(this);
-            person.fromXML(result.Items.Contact);
-            await person.saveLocally();
-          }
+          let person = this.getPersonByItemID(sanitize.nonemptystring(result.Items.Contact.ItemId.Id)) ??
+            new EWSPerson(this);
+          person.fromXML(result.Items.Contact);
+          await person.downloadPictureFromServer();
```
NeilRashbrook, line 200: I like the way the new version batches the picture downloads. (I assume that pictures are small enough that the resulting response won't be unreasonably large.)

NeilRashbrook:
Filed PR #1413 for the bugs mentioned above.

Issue #1398, opened 2026-08-20
Contacts: Person: Bugs with person picture


jermy-c:
### Reproduction 1

1. Go to Contacts
2. Create a new person in a CardDAV or Nextcloud address book
3. Add picture
4. Add a name
5. Click save

### Actual Result

- Console error
```
Error: Saving the contact failed: HTTP 412 Precondition failed
    at assertHTTPResponseOK (netUtil.ts:129:9)
    at async CardDAVPerson.saveToServer (CardDAVPerson.ts:56:7)
    at async CardDAVPerson.save (Person.ts:60:5)
    at async Module.track_reactivity_loss (chunk-S4IO5ETD.js?v=52c32fe1:1873:15)
    at async save (NameBox.svelte:141:7)
    at async Module.track_reactivity_loss (chunk-S4IO5ETD.js?v=52c32fe1:1873:15)
    at async HTMLButtonElement.myOnClick (RoundButton.svelte:83:7)
```
- Picture is saved to the address even though there's an error
- The contact is saved by just adding a picture

### Reproduction 2

1. Go to Contacts
2. Edit an existing person without a picture
3. Add a picture
4. Click save

### Actual Result

- Console error
```
Error: Saving the contact failed: HTTP 415 Unsupported Media Type
    at assertHTTPResponseOK (netUtil.ts:129:9)
    at async CardDAVPerson.saveToServer (CardDAVPerson.ts:66:7)
    at async CardDAVPerson.save (Person.ts:60:5)
    at async Module.track_reactivity_loss (chunk-S4IO5ETD.js?v=52c32fe1:1873:15)
    at async onSelectPicture (SetPictureButton.svelte:31:3)
    at async Module.track_reactivity_loss (chunk-S4IO5ETD.js?v=52c32fe1:1873:15)
    at async HTMLButtonElement.myOnClick (Button.svelte:108:7)
```
- Picture is saved to the address even though there's an error

### Expected Result

- No error when adding picture or saving
- Don't save unless [Save] is clicked by the user in case there's any unwanted changes

### Related

- See commit 99eb986b7bd7f976c864432be4790542f1bd122f
- Tested with Nextcloud account
- For OWA the picture is not saved remotely and when synced from remote, it does not appear locally

jermy-c:
> The contact is saved by just adding a picture

We're saving the contact as soon as a picture is selected here.

[...]

PR #1412, opened 2026-08-27
OWA, ActS: Fix picture save

benbucksch:
@jermy-c reported:
                                                                                           
  I get the following error when saving a new contact with a picture:                      
                                                                                           
  Cannot find the contact  on the server Error: Cannot find the contact  on the server     
      at assert$3 (util.ts:4:11)                                                           
      at OWAPerson.savePictureToServer (OWAPerson.ts:103:5)                                
      at async OWAPerson.saveToServer (OWAPerson.ts:88:5)                                  
      at async OWAPerson.save (Person.ts:59:5)                                             
      at async save (NameBox.svelte:142:5)                                                 
      at async HTMLButtonElement.myOnClick (RoundButton.svelte:83:7)                       
                                                                                           
                                                                                          
  I get an error when saving a contact with a picture:                                     
                                                                                           
  [email address] ActiveSync: Malformed request ActiveSyncError:                           
  jeremy@beonex.onmicrosoft.com ActiveSync: Malformed request                              
      at ActiveSyncPerson.saveToServer (ActiveSyncPerson.ts:119:17)                        
      at async ActiveSyncPerson.save (Person.ts:59:5)                                      
      at async Module.track_reactivity_loss (chunk-JSQPOSIO.js?v=3555e86e:1873:15)         
      at async save (NameBox.svelte:141:7)                                                 
      at async Module.track_reactivity_loss (chunk-JSQPOSIO.js?v=3555e86e:1873:15)         
      at async HTMLButtonElement.myOnClick (RoundButton.svelte:83:7)    

Opus 5 wrote:

  Jeremy's ActiveSync error — root cause confirmed from the spec. [MS-ASCNTC] 2.2.2.58: the
  Picture value "MUST be limited to 48 KB of binary content that is encoded with base64
  encoding, or an image size of around 36 KB… If the value of the Picture element exceeds
  48 KB… the server MUST return a status error of 6." Sync status 6 is what our table
  labels "Malformed request". We sent the picked file at full size — a phone photo is 100×
  over. ContactBase.setPictureFromFile() now scales it to 240 px (96 px display, ×2 for
  HiDPI) and re-encodes as JPEG until the data URL fits in 48 kB, via a new
  scaleImageToDataURL() in logic/util/util.ts built the same way as SystemNotification.ts.
  It fills the canvas white first, since JPEG has no alpha and would render transparency
  black. This also helps EWS/OWA/CardDAV, which have their own size limits.
  
  Jeremy's OWA error — two bugs. The empty name in Cannot find the contact  on the server
  is itself the proof of the first: CreatePersona/UpdatePersona answer with the new
  PersonaId and nothing else, so this.name = sanitize.nonemptystring(response.DisplayName, 
  "") blanked the name he had just typed. It now keeps the existing name.
  
  The second is the assert: the picture hangs off the contact item, and a persona the 
  server created a moment ago names neither its contact (Attributions[].SourceId) nor its
  email address (EmailAddress.ItemId), so GetPersona hands back no ItemId.
  findItemIDOnServer() now falls back to the address book search — owaFindPersonsRequest()
  gained a QueryString parameter — matching on the PersonaId. Caveat worth your judgement:
  which of the two lookups actually works on Exchange Online I could not verify without a
  server. I inferred it from the report plus the fact that listPersons() merges the
  FindPeople result into the GetPersona persona, i.e. production has always had both
    sources at hand and now the single-contact path asks for both too. If Jeremy still hits
  it, the GetPersona JSON for a fresh persona would settle it.

  Fourth commit, separable if you disagree: Person.save()/Group.save() now saveLocally() in
  a finally. The contact was created on the server before the picture upload threw, and we
  discarded its ID — so Jeremy's every retry was creating another contact. Same ID-loss
  class as #1398.


NeilRashbrook:
Unsurprisingly I don't approve of the OWA changes.

The image scaling code looks reasonable, except that I think it should use the original file if the resulting data URL is small enough.

jermy-c:
The ActiveSync fix from this PR works.

benbucksch:
@jermy-c I've merged a heavily modified version of this PR to master, without OWA changes. Can you please check again?

jermy-c:
Just a note: ActiveSync change is commit 3613138a54d2e664a606e909df5c0e2c045e175b on master

jermy-c:
Confirming commit https://github.com/mustang-im/mustang/commit/3613138a54d2e664a606e909df5c0e2c045e175b on master works

benbucksch:
Thanks!

NeilRashbrook:
> I think it should use the original file if the resulting data URL is small enough.

Ignore that, Exchange doesn't support transparency, so we should always fill and resize the picture for consistency. (I wonder whether it's worth checking PNG format first though.)

PR #1417, opened 2026-08-31
Mail: Search on server

benbucksch:
* Implement server-side search for all protocols
  * IMAP
  * JMAP
  * EWS
  * OWA
* Combine the result from local DB server and server-side search, and de-dup

### DIFF
--- a/app/logic/Mail/EWS/ExchangeSearchEMail.ts
+++ b/app/logic/Mail/EWS/ExchangeSearchEMail.ts
@@ -1,0 +1,64 @@
+import { SearchEMail } from "../Store/SearchEMail";
+import type { ExchangeMailAccount } from "./ExchangeMailAccount";
+
+/** Lets the Exchange server search, using Advanced Query Syntax (AQS).
+ * <https://learn.microsoft.com/en-us/exchange/client-developer/exchange-web-services/how-to-perform-an-aqs-search-by-using-ews-in-exchange>
+ * Exchange restrictions cannot match the message body, so AQS is the only way
+ * to search the mail contents. EWS, OWA and Graph all send this query. */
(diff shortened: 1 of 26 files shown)


app/logic/Mail/EWS/ExchangeSearchEMail.ts
```diff
@@ -0,0 +1,64 @@
+import { SearchEMail } from "../Store/SearchEMail";
+import type { ExchangeMailAccount } from "./ExchangeMailAccount";
+
+/** Lets the Exchange server search, using Advanced Query Syntax (AQS).
+ * <https://learn.microsoft.com/en-us/exchange/client-developer/exchange-web-services/how-to-perform-an-aqs-search-by-using-ews-in-exchange>
+ * Exchange restrictions cannot match the message body, so AQS is the only way
+ * to search the mail contents. EWS, OWA and Graph all send this query. */
```
NeilRashbrook, line 7: I can't find a source for this. On the other hand, PR #906 works at least on Office 365 accounts. (I can't remember whether I tested this on Exchange 2019.)

NeilRashbrook (commented):
(Given that you get one response for each folder, you don't actually need to ask for the parent folder for each item, because all of the items in the same response will have the same parent folder.)

I haven't found any problems with this though code inspection.

app/logic/Mail/EWS/ExchangeSearchEMail.ts
benbucksch: Yes, that was a hallozination, which led down the wrong path.

> PR #906 works at least on Office 365 accounts

Thanks for the pointer. I'll incorporate that.
(The problem in that PR was the API.)

PR #1421, opened 2026-09-02
POP3: Implement - fixes #60

### LINKED ISSUE #60: Protocol: Mail: POP3
* Implement POP3 protocol to fetch mail
* Poll
* Save mails locally on disk, as original MIME source, in ZIP file
   * Needs: File API
   * Must store in persistance disk location, to not lose mails (e.g. no cache clear)

benbucksch:
Prompt for Fable 5.1:

Implement POP3. In the backend, make only a factory function to open a socket and pass
the parameters to it (hostname, port, tls on or off, tls options like allowing self-signed
or allowing expired etc.), and then use the socket entirely from app/logic/Mail/POP3/ .

Use the same code style as you find in IMAP, and put all functions (other than generic utils) in the appropriate class. Do not use a state machine code logic, but use async functions. Your primary state is whether you're logged in. The steps to get there are async functions in the login() function.

Given that we have the only storage of these emails, in addition to our normal email storage, add the MailDir content storage provider, just for the POP3 accounts.

First, write a clean implementation based on the RFCs. Second step, launch an Opus agent to search for servers that need special handling.
Launch another Opus agent to research whether there's a way to pull mails with POP3, and have the POP3 queue emptied, but not delete the messages in IMAP for the same mail account. The goal is: We want to leave messages in IMAP, but we also don't want to let the POP3 mail queue grow very large, because that slows down every mail check. Alternatively, check how you can make the mail check in POP3 fast even though the queue is long, e.g. by remembering until which mail number we have pulled the mail, and then check only from there (plus a leeway of maybe 10 messages, just in case). Obviously, make sure not to pull the same email multiple times.

When the user selects POP3 in mail setup (from autoconfig or manual config), show a warning why POP3 interferes with mobile email apps and that the user should use IMAP.

Once you're done, use an Opus agent to install some common open-source POP3 servers and test against them. Consider the limits of the environment you're in.


jermy-c:
All emails were marked as unread even if they were already read on the server for Outlook mail.

benbucksch:
> All emails were marked as unread even if they were already read on the server for Outlook mail.

That's normal. POP3 only pulls the raw mails and nothing else. It cannot pull any meta state: No read, flag, tag, nothing.

benbucksch:
@jermy-c Which settings did you try? Did the mails in your outlook inbox survive, or were all emails that you pulled with POP3 deleted on the server?

jermy-c:
> Which settings did you try?

I just used the default settings.

> Did the mails in your outlook inbox survive, or were all emails that you pulled with POP3 deleted on the server?

Yes, all mails survived and are still on the server.

benbucksch:
Merged manually

PR #1425, opened 2026-09-03
Accounts: Dependent accounts log in, before they start up

benbucksch:
"If the dependent account needs its own stream listener, it should do that in `login()`", as defined in
<https://github.com/mustang-im/mustang/pull/1292#issuecomment-4963453696>.

The shared mailboxes, calendars and address books of an Exchange account subscribed to their notification stream in `startup()`. The startup is the initial sync, and the caller may skip it and issue a single call instead, so the stream is then missing. It belongs to the connection, i.e. to `login()`.

Dependent accounts never had their `login()` called: the main account starts them up, and their own `login()` only forwards to the main account, which already logged in. So `startupDependentAccounts()` now calls both, in the order that the definition gives.

For the same reason, `loginAndStartup()` of a dependent account forwards to the main account only when that is not logged in yet. Otherwise, e.g. for the Login button of a shared mailbox whose account works, it logs in and starts up only that one account, instead of running the startup of every account again.

### DIFF
--- a/app/frontend/Settings/AccountSubAccounts.svelte
+++ b/app/frontend/Settings/AccountSubAccounts.svelte
@@ -110,7 +110,7 @@
     newAccount.mainAccount = account;
     getAppGlobalListForAccount(newAccount).add(newAccount);
     await newAccount.save();
-    await newAccount.syncOnStartup();
+    await newAccount.loginAndStartup(false);
   }
 </script>
 
--- a/app/logic/Abstract/Account.ts
+++ b/app/logic/Abstract/Account.ts
@@ -87,11 +87,15 @@ export class Account extends Observable {
   /**
    * Convenience method for accounts to use to start up dependent accounts.
    * This should be called after the account has finished its own startup.
+   * They log in first, e.g. for their own notification stream. Their `login()`
+   * returns right away, given that we are logged in already.
    */
   protected async startupDependentAccounts() {
     for (let dependent of this.dependentAccounts()) {
-      dependent.dependentStartupOnce.runOnce(() => dependent.syncOnStartup())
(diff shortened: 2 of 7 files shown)


NeilRashbrook:
I am confused as to when notifications are supposed to be started. Originally the behaviour was that any `login` would be directed up to the main account, after which it would run `startup`, since renamed to `syncOnStartup`, which would then sync all of the dependent accounts, plus hook up their notifications.

Current `master` has of course split up `login` and `syncOnStartup`, but `loginAndStartup` will again be directed up to the main account, which will sync all of the dependent accounts plus hook up their notifications. If however your UI element only calls `login`, no notifications are hooked up; calling `syncOnStartup` for a single dependent account will enable its notifications, or calling `syncOnStartup` on the main account will of course sync all of the dependent accounts and hook up their notifications.

Now with this patch, if you call `login` on the main account, again, no notifications are hooked up. If you call `login` on a subaccount (with or without a `login` on the main account), notifications are hooked up, and you can subsequently call `login` on other subaccounts to hook their notifications up. But if you want to hook up all of the notifications? Well that's still done via `syncOnStartup` for the main account, which is also currently the only way to hook up its notifications. (`loginAndStartup` will still work normally of course, because it always starts everything up.)

If I happen to have understood correctly, then to get everything to work, you need to:
- undo the change to `Account.startupDependentAccounts`
- make `OWAAccount.login` always call `super.login()` first, and then early return for dependent accounts
- make several changes to `EWSAccount.login`:
  - always call `super.login()` first
  - dependent accounts should subscribe to subaccount notifications and then early return
  - main account should subscribe to notifications instead of doing it in `syncOnStartup`
  - main account should log in to all dependent accounts to hook up their notifications

PR #1426, opened 2026-09-03
S/MIME: Interop fixes #1379

### LINKED ISSUE #1379: S/MIME: Ensure it works with real email apps
S/MIME needs to exchange messages with real email apps that are commonly used to send and receive S/MIME.

* Outlook Windows 2024 classic
  * #1378 
  * We can read Outlook Windows 2024 classic messages
    - [ ] Signed
    - [ ] Encrypted and signed
  * Outlook Windows 2024 classic can read our messages
    - [ ] Signed
    - [ ] Encrypted and signed
* New Outlook
  * We can read New Outlook messages
    - [ ] Signed
    - [ ] Encrypted and signed
  * New Outlook can read our messages
    - [ ] Signed
    - [ ] Encrypted and signed
* Outlook Android
  * We can read Outlook Android messages
    - [ ] Signed
    - [ ] Encrypted and signed
  * Android Outlook can read our messages
    - [ ] Signed
    - [ ] Encrypted and signed
* Apple Mail
  * We can read Apple Mail messages
    - [ ] Signed
    - [ ] Encrypted and signed
  * Apple Mail can read our messages
    - [ ] Signed
    - [ ] Encrypted and signed
* Thunderbird
  * We can read Thunderbird messages
    - [ ] Signed
    - [ ] Encrypted and signed
  * Thunderbird can read our messages
    - [ ] Signed
    - [ ] Encrypted and signed

benbucksch:
Fixes for #1379
Fable 5.1 orchestration, Opus 5 coding
    
### Commits
    
  - Send, 2 commits. The inner part keeps its Content-Transfer-Encoding, so single-part
    mail no longer arrives as raw quoted-printable in every client. A recipient without a
    key gives a message instead of a null dereference.
  - Read, 7 commits. 3DES and RC2 content decrypt using the ciphers already written for
    PKCS#12. Unsupported ciphers and key transports name themselves in a user-visible 
    error. A certificate-less key no longer aborts decryption for all keys. Recipients are
    matched by subjectKeyIdentifier. Signed content with LF line endings verifies. A p7m or
    p7s sent as application/octet-stream is read, as Microsoft specifies. A mislabelled
    part renders instead of leaking an ASN.1 parser error. RSA-OAEP key transport decrypts
    via WebCrypto.
  - ECDSA and PSS, 4 commits. ECDSA signatures on P-256, P-384 and P-521 verify, chains
    signed by EC CAs and by RSA-PSS CAs validate, and encrypting to an EC certificate gives
    a clear error instead of a crash.
  - Trust, 9 commits, your mid-job request. DN attribute types outside our OID map compare
    by value, which fixes signing and decryption for eIDAS-style certificates. A key you
    marked bad stays bad. The Signed state requires the certificate's address to equal
    From, for PGP too since the check lives in the shared rememberSigner. No fallback to
    the first certificate when the sid matches none. The contentType attribute is checked.
    signingTime is read, used for certificate validity at signing time, and compared with
    the Date header within an hour as PGP does. ESS signingCertificate and V2 are checked.
    Chains are checked for basicConstraints, pathLen, keyCertSign, rfc822 name constraints,
    and the leaf for keyUsage and extKeyUsage. A certificate whose key usage forbids
    encryption is never chosen for it.

### Verification
  
  Test count in the S/MIME directory rose from 47 to 97, the whole app suite from 757 to
  807 tests with only the three failures that already fail on master. svelte-check is at
  the same baseline. Every "must fail" fixture was cross-checked with openssl, and our
  signed and encrypted output was verified by the real NSS tools that Thunderbird uses.

### Behaviour changes to be aware of

  - Mail we signed with the old subject-in-sid bug now shows as unsigned, because that
    fallback is gone.
  - A signature whose Date header is more than about an hour from the signing time shows as
    unsigned, mirroring PGP and Thunderbird.
  - A From address not covered by the signer certificate shows as unsigned, also for PGP.
 
### Left out
    
  The SignerInfo sid in subjectKeyIdentifier form (Outlook 2010 RTM only) is rejected with
  a TODO. A multipart/signed nested inside a multipart/mixed wrapper is still never
  verified. Outlook Android reportedly shows raw source when the first signed part is a
  bare text/plain, which only plain-text send mode produces. Message signatures with
  RSA-PSS, ECDH key agreement and ECDSA signing remain unsupported. A latent PGP bug was
  found and not touched: the inner-versus-outer From check in the PGP reader runs before
  the MIME is replaced, so it can never fire. The own-identity and recipient-certificate
  gaps from the inventory earlier in this session (self-signed certificate creation, CSR
  without SAN, DER and .p7b import, dropped renewed certificates, GAL certificates not
  persisted) are untouched.
 
  Only a live test can settle what New Outlook, OWA and Outlook mobile put on the wire, how
  Outlook presents an untrusted signer, and whether current Apple Mail still emits 3DES.

  Summary: send-side encoding fix, 3DES/RC2/OAEP/SKI/octet-stream reading, ECDSA and PSS
  verification, and the From-binding plus certificate input checks you asked for.

### DIFF
--- a/app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
+++ b/app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
@@ -1,6 +1,6 @@
 import type { PrivateKey } from "../PrivateKey";
 import { SMIMEPublicKey, splitPEM } from "./SMIMEPublicKey";
-import { Any, PrivateKeyInfo, EncryptedPrivateKeyInfo, RSAPrivateKey, RSAPublicKey, CertificationRequestInfo, DigestInfo, CertificationRequest, Null } from "./SMIMEASN1";
+import { Any, PrivateKeyInfo, EncryptedPrivateKeyInfo, RSAPrivateKey, RSAPublicKey, SubjectPublicKeyInfo, CertificationRequestInfo, DigestInfo, CertificationRequest, Null } from "./SMIMEASN1";
 import { decryptPBES2, encryptPrivateKey } from "./PBES2";
 import { decrypt, padFF } from "./SMIMERSAES";
 import { sanitize } from "../../../../../lib/util/sanitizeDatatypes";
@@ -33,7 +33,13 @@ export class SMIMEPrivateKey extends SMIMEPublicKey implements PrivateKey {
     }
   }
 
-  async matches(key: RSAPublicKey): Promise<boolean> {
+  /** Whether the given certificate belongs to our private key,
+   * which is always an RSA key */
+  async matches(publicKey: SubjectPublicKeyInfo): Promise<boolean> {
+    if (publicKey.algorithmIdentifier.algorithm != "rsaEncryption") {
+      return false;
+    }
+    let key = RSAPublicKey.decode(publicKey.subjectPublicKey.data);
     let rawKey = await this.decryptKey();
     return key.n == rawKey.n && key.e == rawKey.e;
   }
--- a/app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
+++ b/app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
@@ -40,21 +40,24 @@ export class SMIMEPrivateKey extends SMIMEPublicKey implements PrivateKey {
 
   /** Decrypts our private key with our passphrase */
   async decryptKey(): Promise<RSAPrivateKey> {
-    let pkcs8: Uint8Array;
-    if (this.passphrase) {
-      let encryptedKey = EncryptedPrivateKeyInfo.decodePEM(this.privateKeyArmored, { label: "ENCRYPTED PRIVATE KEY" });
-      if (encryptedKey.encryptionAlgorithm.algorithm != "pkcs5PBES2") {
-        throw new Error("Unsupported private key encryption algorithm");
-      }
-      pkcs8 = await decryptPBES2(encryptedKey.encryptedData, encryptedKey.encryptionAlgorithm.parameters, this.passphrase);
-    } else {
-      pkcs8 = Any.decodePEM(this.privateKeyArmored, { label: "PRIVATE KEY" });
-    }
-    let privateKeyInfo = PrivateKeyInfo.decode(pkcs8);
+    let privateKeyInfo = PrivateKeyInfo.decode(await this.decryptKeyPKCS8());
     if (privateKeyInfo.privateKeyAlgorithm.algorithm != "rsaEncryption") {
       throw new Error("Unsupported private key algorithm");
     }
     return RSAPrivateKey.decode(privateKeyInfo.privateKey);
+  }
+
+  /** Decrypts our private key with our passphrase, in the PKCS#8 form
+   * that `crypto.subtle.importKey()` reads */
+  async decryptKeyPKCS8(): Promise<Uint8Array> {
+    if (!this.passphrase) {
+      return Any.decodePEM(this.privateKeyArmored, { label: "PRIVATE KEY" });
+    }
+    let encryptedKey = EncryptedPrivateKeyInfo.decodePEM(this.privateKeyArmored, { label: "ENCRYPTED PRIVATE KEY" });
+    if (encryptedKey.encryptionAlgorithm.algorithm != "pkcs5PBES2") {
+      throw new Error("Unsupported private key encryption algorithm");
+    }
+    return await decryptPBES2(encryptedKey.encryptedData, encryptedKey.encryptionAlgorithm.parameters, this.passphrase);
   }
 
   privateKeyAsFile(): File {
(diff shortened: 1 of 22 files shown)


app/logic/Mail/Encryption/SMIME/SMIMESend.ts
```diff
@@ -29,14 +29,19 @@ export class SMIMESend {
...
       let pos = mimeAsText.indexOf("\r\n\r\n");
       // Split on CRLF, but keep folded continuation lines (those starting with
       // whitespace) attached to their header.
       let headers = mimeAsText.slice(0, pos).split(/\r\n(?![ \t])/);
-      let contentTypeHeader = headers.find(header => /^Content-Type: /i.test(header)) ?? "Content-Type: text/plain";
+      let contentHeaders = headers.filter(header => /^Content-/i.test(header));
       let otherHeaders = headers.filter(header => !/^Content-/i.test(header));
-      mimeAsText = contentTypeHeader + mimeAsText.slice(pos);
+      if (!contentHeaders.some(header => /^Content-Type: /i.test(header))) {
+        contentHeaders.push("Content-Type: text/plain");
```
NeilRashbrook, line 42: Does it make sense to `unshift` this instead? (Also lower down.)

app/logic/Mail/Encryption/SMIME/SMIMEReadProcessor.ts
```diff
@@ -51,16 +52,32 @@ export class SMIMEReadProcessor extends EMailProcessor {
...
+    if (!sanitize.enum(cipher, ["aes128cbc", "aes192cbc", "aes256cbc", "desEDE3CBC", "rc2CBC"], null)) {
       return;
     }
-    let vector = OctetString.decode(envelopedData.content.encryptedContentInfo.contentEncryptionAlgorithm.parameters);
-    let encryptedContent = envelopedData.content.encryptedContentInfo.encryptedContent;
     let symmetricKey = await this.decryptSymmetricKey(email, envelopedData.content.recipientInfos);
-    let key = await crypto.subtle.importKey("raw", symmetricKey, "AES-CBC", false, ["decrypt"]);
-    let decryptedContent = await crypto.subtle.decrypt({ name: "AES-CBC", iv: vector }, key, encryptedContent);
+    let decryptedContent: Uint8Array;
+    if (cipher == "desEDE3CBC") {
```
NeilRashbrook, line 62: `switch (cipher) {`?

app/logic/Mail/Encryption/SMIME/SMIMEPrivateKey.ts
```diff
@@ -40,23 +40,26 @@ export class SMIMEPrivateKey extends SMIMEPublicKey implements PrivateKey {
...
 
+  /** Decrypts our private key with our passphrase, in the PKCS#8 form
+   * that `crypto.subtle.importKey()` reads */
+  async decryptKeyPKCS8(): Promise<Uint8Array> {
+    if (!this.passphrase) {
+      return Any.decodePEM(this.privateKeyArmored, { label: "PRIVATE KEY" });
+    }
+    let encryptedKey = EncryptedPrivateKeyInfo.decodePEM(this.privateKeyArmored, { label: "ENCRYPTED PRIVATE KEY" });
+    if (encryptedKey.encryptionAlgorithm.algorithm != "pkcs5PBES2") {
+      throw new Error("Unsupported private key encryption algorithm");
```
NeilRashbrook, line 58: I noticed in one of the other commits we were including the name/oid of the algorithm in the error message, but I don't know whether it makes sense to do that here.

app/logic/Mail/Encryption/SMIME/SMIMEVerify.ts
```diff
@@ -107,5 +107,7 @@ export async function verifySignedData(signedData: any, content: Uint8Array): Pr
 /** Compares two X.501 names, e.g. certificate issuer and subject */
 export function sameName(a: TBSCertificate["issuer"], b: TBSCertificate["issuer"]): boolean {
   return a.length == b.length &&
-    a.every((attr, i) => attr.type == b[i].type && attr.value.value == b[i].value.value);
+    // Attribute types that our OID map does not know, e.g. the eIDAS
+    // organizationIdentifier, decode to arrays, which `==` compares by identity
+    a.every((attr, i) => String(attr.type) == String(b[i].type) && attr.value.value == b[i].value.value);
```
NeilRashbrook, line 112: I think `indexedDB.cmp` would be better here; it will compare both strings and arrays, and returns a non-zero value if one argument is a string and the other is an array.

app/logic/Mail/Encryption/SMIME/SMIMEVerify.ts
```diff
@@ -71,8 +75,24 @@ export async function verifySignedData(signedData: any, content: Uint8Array): Pr
...
+    }
     signedContent = Attributes.encode(signerInfo.signedAttrs);
   }
+  // The certificate must have been valid when the message was signed. Mail
+  // that was signed before the certificate expired stays signed afterwards.
+  let { notBefore, notAfter } = cert.tbsCertificate.validity;
+  if (signedAt && (signedAt < notBefore.value || signedAt > notAfter.value)) {
+    console.log("the certificate was not valid when the message was signed");
+    return null;
+  }
```
NeilRashbrook, line 95: Is the UI happy with a certificate that's since expired? In particular what happens when:
- this message was correctly signed at the date it was sent, and we only have the included certificate
- we have a contact with a newer certificate for the same public key, which was not valid at the date the message was sent, and has also since expired
- we have a contact with a certificate for the same public key which is currently valid

NeilRashbrook (approved):
Seems mostly fine but feel free to implement any of these comments that you feel are worthy enough.

